PR #1, opened 2020-02-04
8237360: Add tests that test binding recipe generation directly

JornVernee:
Continuation of: _https://mail.openjdk.java.net/pipermail/panama-dev/2020-February/007417.html_
 
> Hi,
> 
> Please review the following patch that adds tests for the binding recipe 
> generators. These tests are more or less transcribed from the old 
> CallingSequenceBuilder tests on the foreign branch.
> 
> Bug: https://bugs.openjdk.java.net/browse/JDK-8237360
> Webrev: 
> http://cr.openjdk.java.net/~jvernee/panama/webrevs/8237360/webrev.00/
> 
> This also helped me find a couple of bugs in the Aarch64 CallArranger 
> that I introduced in an earlier patch, which I've fixed.
> 
> Thanks,
> Jorn

## Progress

## Issue
[JDK-8237360](https://bugs.openjdk.java.net/browse/JDK-8237360): Add tests that test binding recipe generation directly

## Approvers

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/Binding.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/Binding.java
@@ -24,6 +24,8 @@ package jdk.internal.foreign.abi;
 
 import jdk.incubator.foreign.MemoryLayout;
 
+import java.util.Objects;
+
 public abstract class Binding {
     static final int MOVE_TAG = 0;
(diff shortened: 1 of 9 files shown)


mcimadamore:
> 
> Yes, this is a good idea, the tag does not need to be a simple int
> (anymore).
> 
> I'd like to add that to the Binding cleanup patch instead though (this
> also changes some of the names), OK?

Sounds good.

PR #4, opened 2020-02-04
8238494: Odd behavior when using VarHandle::toMethodHandle on a memory access handle

mcimadamore:
Hi,
this change fixes a bug introduced by the recently pushed VarHandle adaptation support. More specifically, when calling `VarHandle::toMethodHandle` on a memory access handle, some spurious index out of bound exception is generated. This is caused by the fact that when `IndirectVarHandle::toMethodHandle` generates the required method handle, it binds it against `this`, rather than against the leaf direct var handle. Which means we're calling a method handle which expects a certain kind of a VarHandle (a memory access handle) with a different VarHandle which has no associated offset information. In fact, the memory access var handle helper classes will try to read the `offset` field from a class that doesn't have it - resulting in a spurious read.

I've updated the adapter test to do the final get in the various tests through a method handle obtained by the corresponding adapted var handle.

Cheers
Maurizio

## Progress

## Issue
[JDK-8238494](https://bugs.openjdk.java.net/browse/JDK-8238494): Odd behavior when using VarHandle::toMethodHandle on a memory access handle

## Approvers

### DIFF
--- a/src/java.base/share/classes/java/lang/invoke/IndirectVarHandle.java
+++ b/src/java.base/share/classes/java/lang/invoke/IndirectVarHandle.java
@@ -84,6 +84,6 @@ import java.util.function.BiFunction;
 
     @Override
     public MethodHandle toMethodHandle(AccessMode accessMode) {
-        return getMethodHandle(accessMode.ordinal()).bindTo(this);
+        return getMethodHandle(accessMode.ordinal()).bindTo(this.directTarget);
     }
 }
--- a/test/jdk/java/foreign/TestAdaptVarHandles.java
+++ b/test/jdk/java/foreign/TestAdaptVarHandles.java
@@ -82,7 +82,7 @@ public class TestAdaptVarHandles {
     }
 
     @Test
-    public void testFilterValue() {
+    public void testFilterValue() throws Throwable {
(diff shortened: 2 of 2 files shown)


mcimadamore:
> Is it possible to add a cast or assert somewhere checking that the bound VH is not an instance of IndirectVarhandle?

Not sure I understand what you mean. If you mean checking that the argument of a bindTo in VarHandle::toMethodHandle is a direct handle, I'm not sure how we can enforce that, as that's just a plain MH::bindTo call.

What could be perhaps more useful, I suppose, would be to tweak the VarHandle generation strategy so that the various methods just take a `VarHandle` and a cast is inserted in the helpers to make sure that we can't pass the 'wrong' handle of type A and treat it like a B. But I think we should do this for _all_ var handle code, even the array/bytebuffer ones - since they also use specific subclasses.

PaulSandoz:
Yes, that is what i was thinking.  Prior to adapting VHs it was more obviously correct by construction.
e.g.:
```
        @ForceInline
        static int get(VarHandle _handle, Object holder) {
            FieldInstanceReadOnly handle = (FieldInstanceReadOnly) _handle;
            return UNSAFE.getInt(Objects.requireNonNull(handle.receiverType.cast(holder)),
                                 handle.fieldOffset);
        }
```

PR #12, opened 2020-02-10
8237359: Clean up Binding

JornVernee:
Continuation of: https://mail.openjdk.java.net/pipermail/panama-dev/2020-February/007417.html

> Hi,
> 
> Please review the following patch that does a few cleanups of the 
> Binding class implementation:
> - Adds javadoc
> - Uses static factory methods instead of raw constructors. This allows 
> returning singletons for some of the operators for instance.
> - Constructing a COPY_BUFFER binding now requires a MemoryLayout, 
> instead of a separate size & alignment. This was easier given the use-cases.
> - Binding.AllocateBuffer renamed to Binding.Allocate (since we already 
> had Copy as well).
> - BoxAddress renamed to ConvertAddress, as a more general name to 
> describe the boxing & unboxing that can occur. Suggested by Maurizio.
> 
> Webrev: 
> http://cr.openjdk.java.net/~jvernee/panama/webrevs/cleanup_bindings/webrev.00/
> Bugs: https://bugs.openjdk.java.net/browse/JDK-8237359, 
> https://bugs.openjdk.java.net/browse/JDK-8238227, 
> https://bugs.openjdk.java.net/browse/JDK-8238235
> 
> This patch take care of several smaller issues all at once. It applies 
> on top of the JDK-8237360 patch.
> 
> Thanks,
> Jorn

## Progress

## Issues
[JDK-8237359](https://bugs.openjdk.java.net/browse/JDK-8237359): Use static factory methods to create Binding instances
[JDK-8238227](https://bugs.openjdk.java.net/browse/JDK-8238227): The COPY_BUFFER operator should take a MemoryLayout
[JDK-8238235](https://bugs.openjdk.java.net/browse/JDK-8238235): Add extensive documentation to Binding

## Approvers
 * Maurizio Cimadamore ([mcimadamore](@mcimadamore) - Committer) **Note!** Review applies to 40562472bef72dfc1058fd3aba2b594946190170

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/x64/sysv/CallArranger.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/x64/sysv/CallArranger.java
@@ -55,6 +55,7 @@ import java.util.Optional;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
+import static jdk.internal.foreign.abi.Binding.*;
 import static jdk.internal.foreign.abi.x64.X86_64Architecture.*;
 import static jdk.internal.foreign.abi.x64.sysv.SysVx64ABI.MAX_INTEGER_ARGUMENT_REGISTERS;
 import static jdk.internal.foreign.abi.x64.sysv.SysVx64ABI.MAX_VECTOR_ARGUMENT_REGISTERS;
@@ -119,7 +120,7 @@ public class CallArranger {
         if (!forUpcall) {
             //add extra binding for number of used vector registers (used for variadic calls)
             csb.addArgumentBindings(long.class, MemoryLayouts.SysV.C_LONG,
-                    List.of(new Binding.Move(rax, long.class)));
+                    List.of(move(rax, long.class)));
         }
 
         return new Bindings(csb.build(), returnInMemory, argCalc.storageCalculator.nVectorReg);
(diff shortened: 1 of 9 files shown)


JornVernee:
Comment from @mcimadamore on the old mail thread:
> > - Constructing a COPY_BUFFER binding now requires a MemoryLayout, instead of a separate size & alignment. This was easier given the use-cases.
> 
> Uhmmm - I'd agree with using a layoutm but patch says this:
> 
> ```
>  public AllocateBuffer(MemoryLayout layout) {
> +         private Allocate(long size, long alignment) {
>               super(ALLOC_BUFFER_TAG);
> -             this.size = layout.byteSize();
> -             this.alignment = layout.byteAlignment();
> +             this.size = size;
> +             this.alignment = alignment;
>           }
> ```
> 
> Seems like you went opposite way - e.g. from layout to size/alignment?

I like to keep my constuctors more 1-to-1 with the fields of the class, and handle any argument conversions in a static factory method, so that is what I've done here. The factory methods `allocate` and `copy` do take a `MemoryLayout`

JornVernee:
Comment from @mcimadamore on the old mail thread:

> I find the fact that Copy is also allocating a bit disturbing. I think copy should just mean that - e.g. copy from source address into target address. You can use Binding.allocate to allocate the target segment and Binding.BaseAddress to get the addresses. I think we should try to boil things down to primitive moves, so as to maximize reuse/composability.

I agree with the sentiment, but after looking into it, it complicates the IR further since we need to shuffle more than 2 values around on the operand stack. Something that would require either load/store operators and a local variable table, or a swap operator.

I'd like to investigate further and save that move for another time.

JornVernee:
I've also added the suggested builder (see Binding.Builder), but kept the static factories since they were being used in the tests to declare the expected binding recipe.

JornVernee:
/solves 8238227: The COPY_BUFFER operator should take a MemoryLayout

JornVernee:
/solves 8238235: Add extensive documentation to Binding

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/x64/sysv/CallArranger.java
```diff
@@ -119,7 +120,7 @@ public static Bindings getBindings(MethodType mt, FunctionDescriptor cDesc, bool
         if (!forUpcall) {
             //add extra binding for number of used vector registers (used for variadic calls)
             csb.addArgumentBindings(long.class, MemoryLayouts.SysV.C_LONG,
-                    List.of(new Binding.Move(rax, long.class)));
+                    Binding.builder().move(rax, long.class).build());
```
mcimadamore: can't you use the factory here?

JornVernee: Good catch. I started out doing everything with the builder, but if we have both, this should use the factory.

PR #15, opened 2020-02-11
8238226: Revisit FunctionDescriptor

JornVernee:
Continuation of: https://mail.openjdk.java.net/pipermail/panama-dev/2020-February/007418.html

> Hi,
> 
> Please review the following patch that revisits FunctionDescriptor: 1. 
> Adds appendArgumentLayouts, and changeReturnLayout combinator methods. 
> 2. Removes the variadic flag. This flag was not really being used in 
> practice, since we can not link variadic functions directly any ways. 
> Using the new appendArgumentLayouts method we can 'specialize' a base FD 
> with different argument layouts representing vararg arguments. I think 
> if a similar flag is needed in the future, we should consider adding 
> annotations to FunctionDescriptor, and making it an annotation instead, 
> but currently it doesn't seem like it's pulling it's weight by being in 
> the API.
> 
> Webrev: 
> http://cr.openjdk.java.net/~jvernee/panama/webrevs/cleanup_fd/webrev.00/
> Bugs: https://bugs.openjdk.java.net/browse/JDK-8238226, 
> https://bugs.openjdk.java.net/browse/JDK-8237580
> 
> Again, this patch applies on top of the Binding cleanup patch.
> 
> Thanks,
> Jorn

## Progress

## Issues
[JDK-8238226](https://bugs.openjdk.java.net/browse/JDK-8238226): Add combinators to FunctionDescriptor
[JDK-8237580](https://bugs.openjdk.java.net/browse/JDK-8237580): Add a method for transforming varargs FunctionDescriptors into fixed-arity ones

## Approvers

### DIFF
--- a/test/jdk/java/foreign/StdLibTest.java
+++ b/test/jdk/java/foreign/StdLibTest.java
@@ -182,30 +182,29 @@ public class StdLibTest extends NativeTestHelper {
 
                 strcat = abi.downcallHandle(lookup.lookup("strcat"),
                         MethodType.methodType(MemoryAddress.class, MemoryAddress.class, MemoryAddress.class),
-                        FunctionDescriptor.of(C_POINTER, false, C_POINTER, C_POINTER));
+                        FunctionDescriptor.of(C_POINTER, C_POINTER, C_POINTER));
 
                 strcmp = abi.downcallHandle(lookup.lookup("strcmp"),
                         MethodType.methodType(int.class, MemoryAddress.class, MemoryAddress.class),
-                        FunctionDescriptor.of(C_INT, false, C_POINTER, C_POINTER));
+                        FunctionDescriptor.of(C_INT, C_POINTER, C_POINTER));
 
                 puts = abi.downcallHandle(lookup.lookup("puts"),
                         MethodType.methodType(int.class, MemoryAddress.class),
-                        FunctionDescriptor.of(C_INT, false, C_POINTER));
+                        FunctionDescriptor.of(C_INT, C_POINTER));
 
                 strlen = abi.downcallHandle(lookup.lookup("strlen"),
                         MethodType.methodType(int.class, MemoryAddress.class),
-                        FunctionDescriptor.of(C_INT, false, C_POINTER));
+                        FunctionDescriptor.of(C_INT, C_POINTER));
 
                 gmtime = abi.downcallHandle(lookup.lookup("gmtime"),
                         MethodType.methodType(MemoryAddress.class, MemoryAddress.class),
-                        FunctionDescriptor.of(C_POINTER, false, C_POINTER));
+                        FunctionDescriptor.of(C_POINTER, C_POINTER));
 
-                qsortComparFunction = FunctionDescriptor.of(C_INT, false,
-                        C_POINTER, C_POINTER);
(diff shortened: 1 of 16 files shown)


JornVernee:
/solves 8237580: Add a method for transforming varargs FunctionDescriptors into fixed-arity ones

JornVernee:
Here I've uploaded the first revision that removes the variadic information from FunctionDescriptor, according to the last message on the old mail thread:

> ...
> Let's just remove variadic-ness info for now. As I said, jextract is fully able to recover that info from the jextract declaration API. Before, we needed that info _in the layouts/descriptors_ because they were the only thing that came out of jextract - and they were what drove the binder. Now the story is more nuanced and jextract will be able to generate the right MH w/o needed any special hook in function descriptor.
> 
> Maurizio

test/jdk/java/foreign/StdLibTest.java
```diff
@@ -354,7 +353,7 @@ private MethodHandle specializedPrintf(List<PrintfArg> args) {
                 argLayouts.add(arg.layout);
             }
             MethodHandle mh = abi.downcallHandle(printfAddr, mt,
-                    FunctionDescriptor.of(C_INT, false, argLayouts.toArray(new MemoryLayout[0])));
+                    FunctionDescriptor.of(C_INT, argLayouts.toArray(new MemoryLayout[0])));
```
mcimadamore: It feels like we should be able to simplify this routine with the new FunctionDescriptor::append ?

JornVernee: Good catch! - updated

JornVernee:
I'll create another patch for foreign-jextract to take into account these changes as well.

test/jdk/java/foreign/StdLibTest.java
```diff
@@ -347,14 +349,9 @@ private MethodHandle specializedPrintf(List<PrintfArg> args) {
...
             }
-            //function
-            List<MemoryLayout> argLayouts = new ArrayList<>();
-            argLayouts.add(C_POINTER); //format
-            for (PrintfArg arg : args) {
-                argLayouts.add(arg.layout);
-            }
-            MethodHandle mh = abi.downcallHandle(printfAddr, mt,
-                    FunctionDescriptor.of(C_INT, false, argLayouts.toArray(new MemoryLayout[0])));
+            FunctionDescriptor printfSpec = StdLibHelper.printfBase.appendArgumentLayouts(
```
mcimadamore: I would have dealt with expanding the function descriptor in the same loop where we expand the method type - but what you have is also fine.

JornVernee: Yeah, I like that better as well. I'll change it.

I guess I got a little tunnel visioned since the comment pointed at the line with `FunctionDescriptor.of(C_INT, argLayouts.toArray(new MemoryLayout[0])))` on GitHub, and doesn't show the for loop where the MethodType is expanded in the first place (neither does the email it looks like). From what I understand, multi-line review comments are in the works, so in the future it should be possible to attach a comment to a larger block of code.

slowhog:
Looks good to me. I think foreign-jextract branch need to change to adapt this after merge, is that merge still happen automatically with git workflow?

PR #16, opened 2020-02-11
8238712: ConcurrentModificationException thrown when forward reference to record type in function prototype

slowhog:
CME get thrown when nested call of createTree introduce a new declaration into the map, this is too restrictive as demonstrated by the test case that it is likely to happen when forward reference of a struct, which will introduce a new declaration while processing the function declaration.

The fix allows nested call to add new declaration and only throw CME if replace actually happens.

Also included a minor fix that should abort generate function or could cause NPE.

## Progress

## Issue
[JDK-8238712](https://bugs.openjdk.java.net/browse/JDK-8238712): ConcurrentModificationException thrown when forward reference to record type in function prototype

## Approvers

### DIFF
--- a/src/jdk.incubator.jextract/share/classes/jdk/incubator/jextract/tool/StaticWrapperSourceFactory.java
+++ b/src/jdk.incubator.jextract/share/classes/jdk/incubator/jextract/tool/StaticWrapperSourceFactory.java
@@ -25,16 +25,14 @@
 
 package jdk.incubator.jextract.tool;
 
-import jdk.incubator.jextract.Declaration;
-import jdk.incubator.jextract.Type;
+import java.lang.invoke.MethodType;
+import java.util.List;
 import jdk.incubator.foreign.FunctionDescriptor;
 import jdk.incubator.foreign.MemoryAddress;
 import jdk.incubator.foreign.MemoryLayout;
 import jdk.incubator.foreign.MemorySegment;
-
-import java.lang.invoke.MethodType;
-import java.util.List;
-import java.util.stream.Collectors;
+import jdk.incubator.jextract.Declaration;
+import jdk.incubator.jextract.Type;
 
 public class StaticWrapperSourceFactory extends HandleSourceFactory {
     public StaticWrapperSourceFactory(String clsName, String pkgName, List<String> libraryNames) {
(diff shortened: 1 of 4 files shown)


src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/TreeMaker.java
```diff
@@ -54,7 +54,18 @@ public TreeMaker() {}
     TypeMaker typeMaker = new TypeMaker(this);
 
     private <T extends Declaration> T checkCache(Cursor c, Class<T> clazz, Supplier<Declaration> factory) {
-        return clazz.cast(treeCache.computeIfAbsent(c, cx->factory.get()));
+        // The supplier function may lead to adding some other type, which will cause CME using computeIfAbsent.
+        // This implementation relax the constraint a bit only check for same key
```
mcimadamore: Nice workaround - basically only throw if we see that we're about to add something that was already there,

test/jdk/java/jextract/SmokeTest.java
```diff
@@ -52,6 +58,28 @@ public void testParser() {
         checkConstant(d, "ZERO", intType, 0L);
     }
 
+    @Test
+    public void test8238712() {
```
mcimadamore: At some point (e.g. separate changeset) it would be nice to split the tests from the framework here. E.g. SmokeTest has accumulated a lot of useful stuff which allow us to write API tests easily. But, currently, we are putting the tests together with the framework. We should make the framework a superclass, and then have a bunch of separate test subclasses - one per header.

src/jdk.incubator.jextract/share/classes/jdk/incubator/jextract/tool/StaticWrapperSourceFactory.java
```diff
@@ -52,6 +50,7 @@ public Void visitFunction(Declaration.Function funcTree, Declaration parent) {
         FunctionDescriptor descriptor = Type.descriptorFor(funcTree.type()).orElse(null);
         if (descriptor == null) {
             //abort
+            return null;
```
mcimadamore: good catch!

PR #19, opened 2020-02-13
8238192: Reimplement MemoryAddress memory access handles on top of var handle combinators

mcimadamore:
This is a continuation of the RFR here:
https://mail.openjdk.java.net/pipermail/panama-dev/2020-January/007366.html

## Progress

## Issue
[JDK-8238192](https://bugs.openjdk.java.net/browse/JDK-8238192): Reimplement MemoryAddress memory access handles on top of var handle combinators

## Approvers

### DIFF
--- a/make/gensrc/GensrcVarHandles.gmk
+++ b/make/gensrc/GensrcVarHandles.gmk
@@ -167,6 +167,8 @@ define GenerateVarHandleMemoryAddress
 
   $1_Type := $2
 
+  $1_FILENAME := $(VARHANDLES_GENSRC_DIR)/VarHandleMemoryAddressAs$$($1_Type)s.java
+
   ifeq ($$($1_Type), Byte)
     $1_type := byte
     $1_BoxType := $$($1_Type)
@@ -246,21 +248,6 @@ define GenerateVarHandleMemoryAddress
     $1_ARGS += -KfloatingPoint
   endif
 
-  ifeq ($$($1_Type), MemoryAddress)
-    $1_type := Object
-    $1_BoxType := Object
-
-    $1_rawType := long
-    $1_RawType := Long
-    $1_RawBoxType := Long
-
-    $1_ARGS += -KMemoryAddressProxy
-
-    $1_FILENAME := $(VARHANDLES_GENSRC_DIR)/VarHandleMemoryAddressAs$$($1_Type)es.java
-  else
-    $1_FILENAME := $(VARHANDLES_GENSRC_DIR)/VarHandleMemoryAddressAs$$($1_Type)s.java
-  endif
-
   $$($1_FILENAME): $(VARHANDLES_SRC_DIR)/X-VarHandleMemoryAddressView.java.template $(BUILD_TOOLS_JDK)
 	$$(call MakeDir, $$(@D))
 	$(RM) $$@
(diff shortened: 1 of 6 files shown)


JornVernee:
It's unfortunate that AddressVarHandleGenerator still needs to be aware of the MemoryAddressProxy as a carrier. It would be nice if this wasn't the case, and it only knew about `long`, and we could handle the adaptation in places where we call JLI.memoryAddressViewVarHandle. The downside with the current setup is that AddressVarHandleGenerator silently changes MemoryAddressProxy to long, and expects code that is elsewhere (Utils::fixUpVarHandle) to apply the adaptation. These Seem prone to going out of sync...

For changing where the adaptation happens, the only problematic places seem to be MemoryHandles::withOffset and MemoryHandles::withStride. Since they use JLI.memoryAddressCarrier to find the carrier type. I also notice they re-wrap the direct VarHandle target, so any adaptation is discarded... I think we should require a direct VarHandle as a base for these methods, i.e. modify MethodHandleImpl::checkMemAccessHandle to throw if `handle.isDirect()` returns false, since we don't seem to be able to apply to re-apply the adaptations any ways (at least now). That would also mean that VarHandles with a MemoryAddress carrier would be disallowed for withOffset and withStride, and maybe that is  okay.

But, that should also allow use to remove knowledge about MemoryAddressProxy as a carrier from AddressVarHandleGenerator and instead handle the adaptation in MemoryHandles::varHandle and LayoutPath::dereferenceHandle instead, by requesting a VarHandle with 'long' as a carrier, and then applying the adaptation. Then all the adaptation happens in one place.

mcimadamore:
> It's unfortunate that AddressVarHandleGenerator still needs to be aware of the MemoryAddressProxy as a carrier. It would be nice if this wasn't the case, and it only knew about `long`, and we could handle the adaptation in places where we call JLI.memoryAddressViewVarHandle. The downside with the current setup is that AddressVarHandleGenerator silently changes MemoryAddressProxy to long, and expects code that is elsewhere (Utils::fixUpVarHandle) to apply the adaptation. These Seem prone to going out of sync...


Well, I tried doing it the other way and ran into a roadblock with the MemoryHandles methods - as yu note below.


> 
> For changing where the adaptation happens, the only problematic places seem to be MemoryHandles::withOffset and MemoryHandles::withStride. Since they use JLI.memoryAddressCarrier to find the carrier type. I also notice they re-wrap the direct VarHandle target, so any adaptation is discarded... I think we should require a direct VarHandle as a base for these methods, i.e. modify MethodHandleImpl::checkMemAccessHandle to throw if `handle.isDirect()` returns false, since we don't seem to be able to apply to re-apply the adaptations any ways (at least now). That would also mean that VarHandles with a MemoryAddress carrier would be disallowed for withOffset and withStride, and maybe that is okay.

Except none of the memory access handles are really direct - since we have to adapt them all to fixup the MemoryAddressProxy mismatch. Also, not super sure about the restriction on MemoryAddress adaptations...

> 
> But, that should also allow use to remove knowledge about MemoryAddressProxy as a carrier from AddressVarHandleGenerator and instead handle the adaptation in MemoryHandles::varHandle and LayoutPath::dereferenceHandle instead, by requesting a VarHandle with 'long' as a carrier, and then applying the adaptation. Then all the adaptation happens in one place.

While I see where you want to go (I wanted to go there too), I don't see a straightforward path to get there. I think that, as we discussed at some point, the real issue here is that the adapters in MemoryHandles should really be dropped in favor of "real" VH adapters - but we can't go down there (at the moment) because to do those kind of adaptations you would need to use MemoryAddress::addOffset, which is slow (because of outstanding VM/C2 issues).

So, while I get the discomfort with the approach in this patch, I don't think we can do much better if we carry the MemoryHandles adapters around. And, as you note, this patch has issues if clients start mixing regular VH adapters with MemoryHandles adapters. I don't know how to solve that, given that we are using adaptation internally as a result of JDK-8237349.

mcimadamore:
Thinking more - I see only two real solutions:

* we drop adapters from MemoryHandles completely
* we re-implement the adapters on top of "official" VH adaptation support - that is:

a) withOffset takes a VarHandle whose coordinates are (MemoryAddress, ...) and returns another VarHandle with coordinates (MemoryAddress, ...) where the memory address has been offset-ed using the provided arg

b) withStride takes a VarHandle whose coordinates are (MemoryAddress, ...) and returns another VarHandle with coordinates (MemoryAddress, long, ...) that is, where an extra index is added to the coordinate - this will be multiplied with the provided stride, and the result will be added to the offset of the provided MemoryAddress

Then we can drop most of the machinery to expose special info on memory access handle. As mentioned before, the memory access handles created using adapters with the approach outlined above will probably not be as efficient as the ones you would get from MemoryLayout::varHandle (since those are flatter, and have no addOffset operations).

PR #21, opened 2020-02-16
8239128: Macro defines constant pointers can crash VM via jextract API

slowhog:
The fundamental issue of the crash is that clang cursor/type may no longer valid after reparse another macro, and with pointer type lazily resolve the pointee type, if the clang cursor/type no longer valid, access invalid memory cause the crash.

The other issue is the a macro pointer to a record type like struct/union can cause NPE, and that cause the macro to be ignored on generation without a warning.

The fix remove the laziness from Type API perspective, but leave that as an implementation detail, so implementation of Pointer type need to figure out how to do that safely internally.

Record type in macro are reduces to void, as currently the Declaration implementation is depending on clang as well, we cannot guarantee that works after reparse.

---------
### Progress

### Issue
 * [JDK-8239128](https://bugs.openjdk.java.net/browse/JDK-8239128): 8239128: Macro defines constant pointers can crash VM via jextract API

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/21/head:pull/21`
`$ git checkout pull/21`


src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/TypeImpl.java
```diff
@@ -137,17 +138,28 @@ public Type type() {
...
         private final Supplier<Type> pointeeFactory;
+        private Type pointee;
 
         public PointerImpl(Supplier<Type> pointeeFactory) {
             super(Kind.POINTER, Optional.empty());
             this.pointeeFactory = pointeeFactory;
+            this.pointee = null;
+        }
+
+        public PointerImpl(Type pointee) {
```
mcimadamore: Can't we implement this constructor by creating a factory which always return the same pointee (e.g. the constructor parameter) and drop the pointee field?

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/TypeMaker.java
```diff
@@ -27,23 +27,47 @@
...
+    private final List<TypeImpl.PointerImpl> pointers = new ArrayList<>();
 
     public TypeMaker(TreeMaker treeMaker) {
         this.treeMaker = treeMaker;
     }
 
+    /**
+     * Resolve all type references. This method should be called before discard clang cursors/types
+     */
+    void resolveTypeReferences() {
```
mcimadamore: This approach seems flexible. It relies on keeping track of which pointers are created and then have the client call, at a later point 'resolveTypeReferences' to make sure that we fully convert all pointee clang types into clang-agnostic types.

I have few observations here:
* this mechanism is, general and should allow for all pointers to remove dependencies from clang-originated types, which should allow us to be able to close the clang index after we parse
* I'm a bit worried that the resolve logic is a bit too simplistic. In principle, it could be possible for resolve() to add new pointer types into `pointers`, so, resolve should use some kind of fixed point logic - e.g. keep going until we're sure _all_ types have been resolved, no matter how deep.

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/MacroParserImpl.java
```diff
@@ -81,16 +81,24 @@ public MacroParserImpl(TranslationUnit tu, Collection<String> args) {
...
     }
 
     MacroResult reparse(String snippet) {
-        return reparser.reparse(snippet)
+        MacroResult rv = reparser.reparse(snippet)
                 .filter(c -> c.kind() == CursorKind.VarDecl &&
                         c.spelling().contains("jextract$"))
                 .map(c -> compute(c))
                 .findAny().get();
+        typeMaker.resolveTypeReferences();
```
mcimadamore: Should this call go here? Probably - if we drop the toplevel unit after subsequent reparse. But, I think we should also put a call on this at the end of parsing? Let's say I have  something like:

```
struct A {
   struct B *b;
}

struct B {
  struct A *a;
}
```

No macros here - how does this work? I believe you will add pointers in the todo list (e.g. when we parse `struct A` we will add `B*`). Who resolves that? And, if we did a resolve, wouldn't that end up adding another pointer (`A*`) to the list? Who would resolve that? And if we re-run resolve on the new pointer, how do we prevent the logic from re-adding the already resolved pointer type `B*` ?

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/TypeImpl.java
slowhog: No, note that the design is to throw away factory after used(resolved), because  there is no guarantee that factory can still work in this case. Which is why we call type() in TypeMaker.resolveTypeReferences()

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/MacroParserImpl.java
slowhog: resolveTypeReferences suppose to be called after parse or reparse, so we no longer depends on clang type.

Theoretically, we also should call resolveTypeReference after parse for non-macro case, but as I noted in the code, we haven't completely get rid of dependency on cursor for declaration API, so we now still have to operate under an valid clang TU for non-macro case, thus there is no need to call that just yet. Because when type() is called on the pointer, the resolve happens automatically.

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/TypeMaker.java
slowhog: Good point, in case of pointer to pointer, current way of create pointer type will add another pointer and that could lead to issues. I'll add a test case to that.

However, we are setting for the resolveTypeReferences after we finished parsing a header, thus all types suppose to be known, and there is no more reason not to create pointer with pointee type directly.

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/TypeImpl.java
mcimadamore: Yeah - I think what I'm saying here is that you are using a side-effect to subtly alter the way pointer types are represented. Before the type() call there's a factory, after the type() call, the factory is gone and you have a pointee. So you have two mutable things, and some side-effects.

Another way to approach this problem would be to add an indirection. Let's say that you call `makeType` and you find a pointer type; what we could do is to setup a PointerTypeImpl, whose 'pointee' supplied does something like this:

```
new TypeImpl.PointerImpl(() -> getPointee(pointerId++));
```

That is, as we turn clang types into high-level types, we also build a side-map from pointers to pointees. Initially this map will contain no entries, but after you resolve all the pending pointers, you will start adding entries for them in the table. Which means that, when the client calls PointerTypeImpl::type(), the lookup succeeds and the client gets the type it wants (w/o looking up into clang specific info).

I think an approach like this can be made to work for all the situations in which we have to 'delay' pointee info computation.

Actually, if all clang types get registered into the TypeMaker cache (as they are now), we don't even to do anything special - the pointee supplier can simply do a lookup on the cache using the identity of the clang type. Note that, while, after we get rid of the index, or after reparsing we can no longer use a clang type with the clang API, the Type object we have is still valid and can be used to to e.g. Map lookups.

So, I think there is a simpler solution here, where we:

* retain the old `PointerImpl` implementation
* when we create a pointer type, we setup a supplier lambda - but instead of having the lambda call `makeType` as before, we make the lambda to a direct lookup on the TypeMaker cache using the clang type object as a key
* we also register the pointer to the 'todo' list, as you are doing now
* when `resolveTypeReferences` is called, we go through the todo list, and call makeType on all the pointees in there - thus populating the cache

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/TypeMaker.java
slowhog: I had something like this before, but as you can see, it's only used by Pointer, thus I simply consolidated them.

We cannot use j.i.clang.Type for identity, as the equal method could actually call into native code. The fact that same type not necessary by same instance of CXType struct actually bit me when I use clang.Type as key for unresolved map instead of plain list.

mcimadamore: I mean - `IdentityHashMap`? I know that equals goes into native, I'm saying let's just use the object identity.

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/Parser.java
```diff
@@ -103,7 +103,9 @@ public Parser(JextractTask.ConstantParser constantParser) {
                 }
             });
 
-        return treeMaker.createHeader(tuCursor, decls);
+        Declaration.Scoped rv = treeMaker.createHeader(tuCursor, decls);
+        treeMaker.freeze();
```
mcimadamore: Does this mean that we can close the index after this?

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/TypeMaker.java
```diff
@@ -41,7 +43,39 @@
 
     TreeMaker treeMaker;
     private final Map<jdk.internal.clang.Type, Type> typeCache = new HashMap<>();
-    private final List<TypeImpl.PointerImpl> pointers = new ArrayList<>();
+    private List<ClangTypeReference> unresolved = new ArrayList<>();
+
+    private class ClangTypeReference implements Supplier<Type> {
```
mcimadamore: Nice - I like it; simpler than what I proposed, but equally effective.

src/jdk.incubator.jextract/share/classes/jdk/incubator/jextract/Type.java
```diff
@@ -415,15 +413,6 @@
...
     static Type.Delegated pointer(Type pointee) {
-        return new TypeImpl.PointerImpl(() -> pointee);
-    }
-
-    /**
-     * Creates a new pointer type with given pointee type.
-     * @param pointee factory to (lazily) build the pointee type.
-     * @return a new pointer type with given pointee type (lazily built from factory).
-     */
-    static Type.Delegated pointer(Supplier<Type> pointee) {
```
mcimadamore: This shouldn't be removed! It's the only way for a client to create mutually referring pointer types.

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/TypeMaker.java
```diff
@@ -119,12 +182,16 @@ Type makeType(jdk.internal.clang.Type t) {
             }
             case Enum:
             case Record: {
-                return Type.declared((Declaration.Scoped)treeMaker.createTree(t.getDeclarationCursor()));
+                if (treeMaker == null) {
```
mcimadamore: Do we get a crash if we go here inside macro eval? Why the special casing - I mean I get that we probably don't care, but trying to understand.

test/jdk/java/jextract/JextractApiTestBase.java
```diff
@@ -0,0 +1,141 @@
+/*
```
mcimadamore: Thx for splitting this out!

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/Parser.java
slowhog: Once we implement the freeze to also remove the Declaration/Cursor dependency, we can close the clang Index.

src/jdk.incubator.jextract/share/classes/jdk/incubator/jextract/Type.java
slowhog: This change is what I meant at beginning to make this an implementation detail.

I have trouble to figure out how to use this API without an API providing something like the TypeReference, which I can implement as a Type that is a transparent delegate.

Imagine Type.declare(Declareation.struct(pos, "foo", Type.pointer(<what can I do to provide this struct type>)).

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/TypeMaker.java
slowhog: We get a NPE since treeMaker is null. If we simply plug-in a TreeMaker, then we get crash as TU in Macro is not usable.

I am thinking we could provide a inquiry-only TreeMaker of the precompile TU for macro case, then we can get the correct type.

slowhog:
Access cursor via CursorPosition can cause crash after close the Index. However, CursorPosition is only available internally, this is relative safe.

Alternatively, we can implement freeze on CursorPosition but that means we keep reference of all CursorPosition instances.

src/jdk.incubator.jextract/share/classes/jdk/incubator/jextract/Type.java
mcimadamore: The problem is always with cases like:

```
struct A {
  struct B *b;
};

struct B {
  struct A *a;
};
```

If you have a supplier-accepting constructor, you can imagine an approach where you store the pointers in different fields, and then you set up the lambdas to retrieve the field values - that will work in such pathological cases.

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/TreeMaker.java
```diff
@@ -125,24 +125,31 @@ Position toPos(Cursor cursor) {
 
     static class CursorPosition implements Position {
         private final Cursor cursor;
+        private final Path path;
```
mcimadamore: Very nice - this way we no longer depend on cursors :+1:

PR #23, opened 2020-02-18
8239345: need an enum for standard C types and a way to get a memory layout for those C types

sundararajana:
---------
### Progress

### Issue
 * [JDK-8239345](https://bugs.openjdk.java.net/browse/JDK-8239345): need an enum for standard C types and a way to get a memory layout for those C types

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/23/head:pull/23`
`$ git checkout pull/23`


src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryLayouts.java
```diff
@@ -144,139 +144,6 @@ private MemoryLayouts() {
      */
     public static final ValueLayout JAVA_DOUBLE = MemoryLayout.ofValueBits(64, ByteOrder.nativeOrder());
 
-    /**
-     * The {@code _Bool} native type.
-     */
-    public static final ValueLayout C_BOOL;
```
JornVernee: Can't we keep all these constants as well?

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryLayouts.java
```diff
@@ -144,139 +144,6 @@ private MemoryLayouts() {
...
-    public static final ValueLayout C_DOUBLE;
-
-    /**
-     * The {@code T*} native type.
-     */
-    public static final ValueLayout C_POINTER;
-
-    static {
-        SystemABI abi = SystemABI.getInstance();
-        switch (abi.name()) {
```
JornVernee: We could use `SystemABI::layoutFor` here to do the initialization instead though.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/aarch64/AArch64ABI.java
```diff
@@ -72,5 +75,41 @@ public String name() {
         return SystemABI.ABI_AARCH64;
     }
 
-
+    @Override
+    public Optional<MemoryLayout> layoutFor(NativeType type) {
+        switch (type) {
```
JornVernee: Could use a switch expression here. Same for the other impls of SystemABI::layoutFor

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryLayouts.java
sundararajana: I think the idea is to remove all C constants from MemoryLayouts.java (only keep Java layouts etc)

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryLayouts.java
sundararajana: Please see above. Per discussion, the idea is to keep C-independent layout constants in MemoryLayouts.java.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/SystemABI.java
```diff
@@ -110,4 +110,96 @@ static SystemABI getInstance() {
         }
         throw new UnsupportedOperationException("Unsupported os or arch: " + os + ", " + arch);
     }
+
+    public enum NativeType {
```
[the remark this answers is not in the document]
sundararajana: Agree. SystemABI.NativeType can be changed to SystemABI.Type

sundararajana:
Will send fresh PR with changes suggested.

PR #25, opened 2020-02-21
8238316: jextract emits a C_BOOL when source says char

sundararajana:
Using SystemABI.Type of MemoryLayouts to generate layout constants

---------
### Progress

### Issue
 * [JDK-8238316](https://bugs.openjdk.java.net/browse/JDK-8238316): jextract emits a C_BOOL when source says char

### Reviewers
 * Jorn Vernee ([jvernee](@JornVernee) - Committer) ⚠️ Review applies to ca463ab8eebc978955a78f99e78156316b0f07f4

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/25/head:pull/25`
`$ git checkout pull/25`

### DIFF
--- a/src/jdk.incubator.jextract/share/classes/jdk/incubator/jextract/tool/JavaSourceBuilder.java
+++ b/src/jdk.incubator.jextract/share/classes/jdk/incubator/jextract/tool/JavaSourceBuilder.java
@@ -30,8 +30,10 @@ import jdk.incubator.foreign.GroupLayout;
 import jdk.incubator.foreign.MemoryAddress;
 import jdk.incubator.foreign.MemoryLayout;
 import jdk.incubator.foreign.MemoryLayouts;
+import jdk.incubator.foreign.MemoryLayouts.SysV;
 import jdk.incubator.foreign.MemorySegment;
 import jdk.incubator.foreign.SequenceLayout;
+import jdk.incubator.foreign.SystemABI;
 import jdk.incubator.foreign.ValueLayout;
 
 import java.lang.invoke.MethodType;
@@ -48,6 +50,7 @@ import java.util.stream.Stream;
  * method is called to get overall generated source string.
  */
 class JavaSourceBuilder {
+    private static final String ABI = SystemABI.getInstance().name();
     // buffer
     protected StringBuffer sb;
     // current line alignment (number of 4-spaces)
@@ -86,6 +89,9 @@ class JavaSourceBuilder {
         sb.append("import java.lang.invoke.VarHandle;\n");
         sb.append("import jdk.incubator.foreign.*;\n");
         sb.append("import jdk.incubator.foreign.MemoryLayout.PathElement;\n");
+        sb.append("import static ");
+        sb.append(HandleSourceFactory.C_LANG_CONSTANTS_HOLDER);
+        sb.append(".*;\n");
     }
 
     protected void classBegin(String name) {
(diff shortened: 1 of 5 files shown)


src/jdk.incubator.jextract/share/classes/jdk/incubator/jextract/tool/JavaSourceBuilder.java
```diff
@@ -30,8 +30,10 @@
 import jdk.incubator.foreign.MemoryAddress;
 import jdk.incubator.foreign.MemoryLayout;
 import jdk.incubator.foreign.MemoryLayouts;
+import jdk.incubator.foreign.MemoryLayouts.SysV;
```
JornVernee: This import seems to be unused.
```suggestion
```

src/jdk.incubator.jextract/share/classes/jdk/incubator/jextract/tool/JavaSourceBuilder.java
```diff
@@ -125,24 +131,32 @@ protected void addLayout(String elementName, MemoryLayout layout) {
...
+            if (type == SystemABI.Type.LONG_DOUBLE) {
+                if (ABI != SystemABI.ABI_SYSV) {
+                    throw new RuntimeException("long double is supported only for SysV ABI");
+                } else {
+                    sb.append("C_LONGDOUBLE");
                 }
-            }
-            if (!found) {
-                throw new AssertionError("Should not get here: " + l);
+            } else {
```
JornVernee: `long double` is a standard type, so it should be fine to support on other platforms as well. e.g. Windows: https://docs.microsoft.com/en-us/cpp/cpp/fundamental-types-cpp?view=vs-2019

Looks like the constant is just missing from MemoryLayouts.WinABI. Should be 8 bytes there.
```java
public static final ValueLayout C_LONGDOUBLE = MemoryLayout.ofValueBits(64, ByteOrder.LITTLE_ENDIAN)
        .withAnnotation(AbstractLayout.NATIVE_TYPE, SystemABI.Type.LONG_DOUBLE);
```

Looks like on arm it is 16 bytes, so you can copy the impl from SysV.

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/LayoutUtils.java
```diff
@@ -61,21 +61,26 @@ public static String getName(Cursor cursor) {
 
     public static MemoryLayout getLayout(Type t) {
         switch(t.kind()) {
-            case Char_S:
-            case Char_U:
             case UChar:
+            case Char_U:
+                return C_UCHAR;
```
JornVernee: Could use

```java
            case UChar, Char_U:
                return C_UCHAR;
```

Here and elsewhere in the same switch.

Or maybe it's nice to switch the whole switch to a switch-expression using `->`, since then you can remove all the `return`s?

test/jdk/tools/jextract/testStruct/LibStructTest.java
```diff
@@ -40,4 +42,51 @@ public void testMakePoint() {
...
+                 case "ll":
+                     assertEquals(type, Type.LONG_LONG);
+                     break;
+                 case "ull":
+                     assertEquals(type, Type.UNSIGNED_LONG_LONG);
+                     break;
+             }
+             fieldCount++;
+         }
+         assertEquals(fieldCount, 10);
```
JornVernee: You could shorten this a bit using `MemoryLayout::select`:

```java
private static void checkFieldABIType(GroupLayout group, String fieldName, SystemABI.Type expected) {
    assertEquals(group.select(groupElement(fieldName)).abiType().orElseThrow(), expected);
}

@Test
public void testFieldTypes() {
    GroupLayout g = (GroupLayout)AllTypes$LAYOUT;
    checkFieldABIType(g, "sc",  Type.SIGNED_CHAR);
    checkFieldABIType(g, "uc",  Type.UNSIGNED_CHAR);
    checkFieldABIType(g, "s",   Type.SHORT);
    checkFieldABIType(g, "us",  Type.UNSIGNED_SHORT);
    checkFieldABIType(g, "i",   Type.INT);
    checkFieldABIType(g, "ui",  Type.UNSIGNED_INT);
    checkFieldABIType(g, "l",   Type.LONG);
    checkFieldABIType(g, "ul",  Type.UNSIGNED_LONG);
    checkFieldABIType(g, "ll",  Type.LONG_LONG);
    checkFieldABIType(g, "ull", Type.UNSIGNED_LONG_LONG);
}
```

src/jdk.incubator.jextract/share/classes/jdk/incubator/jextract/tool/JavaSourceBuilder.java
sundararajana: Nice suggestion. Will add long double for other ABIs

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/LayoutUtils.java
sundararajana: Agree on combining cases. But there is control flow (throw with couple of more statements). I'd prefer minimal change for this fix.

test/jdk/tools/jextract/testStruct/LibStructTest.java
sundararajana: Thanks. Will incorporate this (along with additional field test for "long double")

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/LayoutUtils.java
```diff
@@ -61,21 +61,24 @@ public static String getName(Cursor cursor) {
 
     public static MemoryLayout getLayout(Type t) {
```
mcimadamore: What about COMPLEX_LONGDOUBLE and friends here?

PR #27, opened 2020-02-24
Use Windows and Aarch64 LongDouble constants in LayoutUtils

JornVernee:
Hi,

Please review this small fix that lets LayoutUtils use the new LongDouble constants instead of Double on Windows and Aarch64 to initialize the global `C_LONGDOUBLE` value. Currently the use of Double results in an incorrect ABI attribute, causing a test failure.

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/27/head:pull/27`
`$ git checkout pull/27`


slowhog:
Looks reasonable to me

JornVernee:
> 
> 
> Looks good - I guess I'm not sure of what is the failure you were seeing? Shouldn't DOUBLE and LONGDOUBLE be the same on Win?

As far as both being 64 bits yes, but now that we have the ABI type attribute this is different for DOUBLE and LONGDOUBLE. 

The particular check that is failing: https://github.com/openjdk/panama-foreign/blob/foreign-jextract/test/jdk/tools/jextract/testStruct/LibStructTest.java#L66

Since the actual Type that is found is `Type.DOUBLE`

PR #28, opened 2020-02-24
8239491: Missing CursorKind and TypeKind constants for LLVM 9

JornVernee:
Adds some missing LLVM9 CursorKind and TypeKind enum constants. Whenever these kinds are encountered during parsing and exception is throw, even if the value is otherwise ignored.

- Also renames ParamDecl to ParmDecl to keep parity with [`Index.h`](https://github.com/llvm/llvm-project/blob/master/clang/include/clang-c/Index.h#L1765)
- Also removes the `FirstExtraDecl` and `LastExtraDecl` constants, since these are just markers used in C, and we don't need them in Java (actually, they might cause problems by overwriting the actual values in the lookup map).

---------
### Progress

### Issue
 * [JDK-8239491](https://bugs.openjdk.java.net/browse/JDK-8239491): Missing CursorKind and TypeKind constants for LLVM 9

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/28/head:pull/28`
`$ git checkout pull/28`


JornVernee:
@mcimadamore ParmDecl was working, but I renamed it from ParamDecl, since it's also called ParmDecl in the clang header.

mcimadamore:
> @mcimadamore ParmDecl was working, but I renamed it from ParamDecl, since it's also called ParmDecl in the clang header.

argh - got confused when looking at the diffs; that explains it, thanks

PR #31, opened 2020-02-25
8237585: Dismantle ForeignUnsafe (foreign-abi part)

JornVernee:
Removes ForeignUnsafe, moves `getUnsafeBase` and `getUnsafeAccess` to `MemoryAddress`, and moves `ofNativeUnchecked` to `MemorySegment`.

I've added a warning about the safety of these methods to the javadoc.

Use of the methods is guarded by a check against a system property, `jdk.incubator.foreign.permitUnsafeInterop` for the first 2, and `jdk.incubator.foreign.permitUncheckedSegments` for the latter. This needs to be set to `true` when running an application, or otherwise the use of these methods results in an `IllegalAccessError` being thrown.

I've updated the test accordingly, and also remove the FASTPATH option from the SysV ABI implementation, since it was not actually being used, but still causing the tests to be run multiple times.

---------
### Progress

### Issue
 * [JDK-8237585](https://bugs.openjdk.java.net/browse/JDK-8237585): Dismantle ForeignUnsafe

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/31/head:pull/31`
`$ git checkout pull/31`


src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryAddress.java
```diff
@@ -142,4 +143,50 @@ static MemoryAddress ofLong(long value) {
...
     }
+
+    // The following methods can be used in conjunction with the java.foreign API.
+
+    /**
+     * Obtain the base object (if any) associated with this address. This can be used in conjunction with
+     * {@link #getUnsafeOffset(MemoryAddress)} in order to obtain a base/offset addressing coordinate pair
+     * to be used with methods like {@link sun.misc.Unsafe#getInt(Object, long)} and the likes.
+     * <p>
+     * This method is <em>unsafe</em>. It's use can result in putting the VM in a corrupt state when used incorrectly,
```
mcimadamore: Typo: `It's` -> `Its` (same in others)

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryAddress.java
```diff
@@ -142,4 +143,50 @@ static MemoryAddress ofLong(long value) {
...
+     * and is provided solely to cover use-cases that can not otherwise be addressed safely. When used incorrectly, there
+     * are no guarantees made about the behaviour of the program. Particularly, incorrect use is not guaranteed to
```
mcimadamore: I'd replace `Particularly, incorrect use is not guaranteed to result in a VM crash, but might instead silently cause memory to be corrupted.` with `More specifically, incorrect uses of this method might result in a JVM crash or, worse, might silently result in memory corruption.`

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -28,6 +28,7 @@
 
 import java.nio.ByteBuffer;
 
+import jdk.internal.foreign.MemorySegmentImpl;
```
sundararajana: Is this import used in this file?

JornVernee: It is not. Thanks for catching!

JornVernee:
Thanks for the comments.

I've uploaded another revision that incorporates the suggested changes.

JornVernee:
I've uploaded another revision. This removes the `getUnsafeBase` method and renames `getUnsafeOffset` to `asLong`. I've left to method `static` to hint that it does something special (the access check). It will also throw an exception when passed a heap address.

PR #33, opened 2020-02-25
8239809: Need API to access attributes

slowhog:
Add API to allow associate attributes with Declaration.

An attribute can have multiple values, thus a list of ConstantDesc.

The test case is based on [JDK-8223105](https://bugs.openjdk.java.net/browse/JDK-8223105), Windows test is simply parsing at this point, need to add validation as the header is implemented differently. 

We may want to add a more generic test case with some standard attributes.

---------
### Progress

### Issue
 * [JDK-8239809](https://bugs.openjdk.java.net/browse/JDK-8239809): Need API to access attributes

### Reviewers
 * Maurizio Cimadamore ([mcimadamore](@mcimadamore) - Committer) ⚠️ Review applies to 49b308e475d996474f13a5726cfba08ec6d6a6e0

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/33/head:pull/33`
`$ git checkout pull/33`

### DIFF
--- a/src/jdk.incubator.jextract/share/classes/jdk/incubator/jextract/Declaration.java
+++ b/src/jdk.incubator.jextract/share/classes/jdk/incubator/jextract/Declaration.java
@@ -26,8 +26,9 @@
 
 package jdk.incubator.jextract;
 
-import java.lang.constant.ConstantDesc;
+import java.lang.constant.Constable;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
@@ -56,11 +57,24 @@ public interface Declaration {
     String name();
 
     /**
+     * Get a declaration with specified attributes
+     * @param attributes The map of attribute names and their values
+     * @return the Declaration with attributes
+     */
+    Declaration withAttributes(Map<String, List<Constable>> attributes);
+
+    /**
+     * Get a declaration without current attributes
+     * @return the Declatation without any attributes
+     */
+    Declaration stripAttributes();
+
+    /**
      * The values of the specified attribute.
      * @param name The attribute to retrieve
      * @return The list of values associate with this attribute
      */
-    Optional<List<ConstantDesc>> getAttribute(String name);
+    Optional<List<Constable>> getAttribute(String name);
 
     /**
      * The attributes associated with this declaration
--- a/src/jdk.incubator.jextract/share/classes/jdk/incubator/jextract/Declaration.java
+++ b/src/jdk.incubator.jextract/share/classes/jdk/incubator/jextract/Declaration.java
@@ -28,7 +28,6 @@ package jdk.incubator.jextract;
 
 import java.lang.constant.Constable;
 import java.util.List;
-import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
@@ -57,11 +56,15 @@ public interface Declaration {
     String name();
 
     /**
-     * Get a declaration with specified attributes
-     * @param attributes The map of attribute names and their values
+     * Get a declaration with specified attribute.
+     * Set the values to the specified attribute while other attributes remains unchanged. If the specified attribute
+     * already exist, the new values are replacing the old ones. By not specifying any value, the attribute will become
+     * empty as {@link #getAttribute(String) getAttribute(name).isEmpty()} will return true.
+     * @param name The attribute name
+     * @param values More attribute values
      * @return the Declaration with attributes
      */
-    Declaration withAttributes(Map<String, List<Constable>> attributes);
+    Declaration withAttribute(String name, Constable... values);
 
     /**
      * Get a declaration without current attributes
--- a/src/jdk.incubator.jextract/share/classes/jdk/incubator/jextract/Declaration.java
+++ b/src/jdk.incubator.jextract/share/classes/jdk/incubator/jextract/Declaration.java
@@ -26,13 +26,15 @@
 
 package jdk.incubator.jextract;
 
-import jdk.incubator.foreign.MemoryLayout;
-import jdk.internal.jextract.impl.DeclarationImpl;
-
+import java.lang.constant.ConstantDesc;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import jdk.incubator.foreign.MemoryLayout;
+import jdk.internal.jextract.impl.DeclarationImpl;
 
 /**
  * Instances of this class are used to model declaration elements in the foreign language.
@@ -53,6 +55,9 @@ public interface Declaration {
      * @return The name associated with this declaration.
      */
     String name();
+
+    Optional<List<ConstantDesc>> getAttribute(String name);
+    Set<String> availableAttributes();
 
     /**
      * Entry point for visiting declaration instances.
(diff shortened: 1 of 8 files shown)


src/jdk.incubator.jextract/share/classes/jdk/incubator/jextract/Declaration.java
```diff
@@ -54,6 +56,19 @@
...
+     * @param name The attribute to retrieve
+     * @return The list of values associate with this attribute
+     */
+    Optional<List<ConstantDesc>> getAttribute(String name);
+
+    /**
+     * The attributes associated with this declaration
+     * @return The attributes associated with this declaration
+     */
+    Set<String> availableAttributes();
```
mcimadamore: Change this to attributeNames()

src/jdk.incubator.jextract/share/classes/jdk/incubator/jextract/Declaration.java
```diff
@@ -277,8 +292,8 @@
      * @param type the global variable declaration type.
      * @return a new global variable declaration with given name and type.
      */
-    static Declaration.Variable globalVariable(Position pos, String name, Type type) {
-        return new DeclarationImpl.VariableImpl(type, Declaration.Variable.Kind.GLOBAL, name, pos);
+    static Declaration.Variable globalVariable(Position pos, String name, Type type, Map<String, List<ConstantDesc>> attrs) {
```
mcimadamore: Uhmmm. I think I'd prefer not to have a Map accepting factory, but to have a `Declaration::withAttribute(String, ConstantDesc)` method. This makes the API a lot tidier (and is similar to what we did for layouts).

slowhog: I removed changes to factory methods, they are from parsing internally, so this is fine.

I agree withAttributes is preferred, but not certain about take single attribute/value. While this maybe the use case most of time, there are situations I need to do multiple modifications or copy and that leads to unnecessary rounds of object creation.

We can leave that out for now, since attributes are populated via parsing, and we don't have a use case where user should be adding attributes.

slowhog:
> 
> This also reminds me that we never sorted out lack of equals/hashcode in Declaration/Type implementation classes. We should address that (and make these method take into account attributes too). But it's probably preferrable to do this in a separate pull request.

I actually had implemented a matcher for Declaration/Type using Visitor, which is probably more flexible.

src/jdk.incubator.jextract/share/classes/jdk/incubator/jextract/Declaration.java
```diff
@@ -54,6 +55,19 @@
...
+     * @param name The attribute to retrieve
+     * @return The list of values associate with this attribute
+     */
+    Optional<List<ConstantDesc>> getAttribute(String name);
+
+    /**
+     * The attributes associated with this declaration
+     * @return The attributes associated with this declaration
+     */
+    Set<String> attributeNames();
```
mcimadamore: I find it odd that the API has ways to query attributes, but no way to set them. I really think that withAttribute/dropAttributes should be a pretty minimal way to get there.

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/DeclarationImpl.java
```diff
@@ -57,25 +62,41 @@ public Position pos() {
...
+    @Override
+    public Set<String> attributeNames() { return Collections.unmodifiableSet(attributes.keySet()); }
+
     public static class VariableImpl extends DeclarationImpl implements Declaration.Variable {
 
         final Variable.Kind kind;
         final Type type;
         final Optional<MemoryLayout> layout;
 
+        private VariableImpl(Type type, Optional<MemoryLayout> layout, Variable.Kind kind, String name, Position pos, Map<String, List<ConstantDesc>> attrs) {
```
mcimadamore: Rather than changing these factories to add attribute variants, consider either using `withAttribute` and drop the changes, or add an internal method in DeclarationImp which clones a declaration and adds annotations (even using a single Map).

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/TreeMaker.java
```diff
@@ -80,23 +81,29 @@ public void freeze() {
...
+                .collect(Collectors.groupingBy(
+                        attr -> attr.kind().name(),
+                        Collectors.mapping(Cursor::spelling, Collectors.toList())
+                ));
     }
 
     public Declaration createTree(Cursor c) {
-        switch (Objects.requireNonNull(c).kind()) {
+        Objects.requireNonNull(c);
+        Map<String, List<ConstantDesc>> attrs = collectAttributes(c);
```
mcimadamore: Rather than collecting attributes upfront, and then having to pass down the attributes to all the various createXYZ functions, wouldn't it be easier to take the parsed declaration and adding the attributes after the fact? That way the createXYZ could remain unchanged and the attribute changes would be more orthogonal. This could be implemented on top of either a public or an impl-only withAttribute functionality.

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/DeclarationImpl.java
```diff
@@ -141,7 +166,7 @@ public ScopedImpl(Kind kind, List<Declaration> declarations, String name, Positi
         }
 
         ScopedImpl(Kind kind, Optional<MemoryLayout> optLayout, List<Declaration> declarations, String name, Position pos) {
```
mcimadamore: No addition of an annotation-accepting ScopeImpl constructor?

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/TreeMaker.java
slowhog: That's certainly possible, but the change won't be that straightforward. We will need to change the checkCache mechanism as what need to be in cache should be with attributes after parsing.

Thus I don't think the change will be smaller, and the constructor with map are still needed. I'll give it a go and see how we feel about that.

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/DeclarationImpl.java
```diff
@@ -26,22 +26,27 @@
...
 import jdk.incubator.jextract.Type;
 
-import java.util.List;
-import java.util.Optional;
-
 public abstract class DeclarationImpl implements Declaration {
 
     private final String name;
     private final Position pos;
+    private final Map<String, List<ConstantDesc>> attributes;
```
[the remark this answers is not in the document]
slowhog: I gave it some thoughts, and I tend to agree with you.

At beginning, I am thinking we need something universal that can turn into the value as needed by API user, thus ConstantDesc seems to make sense, as we don't need the value necessary to be Constable.

I don't expect to see jextract API to be involved in any direction of that operation, so it really up to the user of attribute, thus ConstantDesc seems to be less restrictive. 

However, require the value to be "Constable" make sense as well and that, as you said, live value is easier to work with. After all, API user may end up generating code/class and care more about turn the value into constants.

slowhog:
The drawback with the API change is that no specialized subclass override, we can add those but seems too much. Cast if really needed should not be a big issue. If there is an elegant way to do that, I would like to know.

mcimadamore:
> The drawback with the API change is that no specialized subclass override, we can add those but seems too much. Cast if really needed should not be a big issue. If there is an elegant way to do that, I would like to know.

We had a similar issue in MemoryLayout API. You can see what we did there; in principle you can add an intermediate, non public, abstract class which uses generic types - e.g.

```
/* package */ abstract class AbstractDeclaration<X extends AbstractDeclaration<X>> implements Declaration {
    public X withAttribute(String, Constable)
}

/* package */ class ScopeDeclarationImpl extends AbstractDeclaration<ScopeDeclarationImpl> { ... }
```

In practice though, javadoc doesn't work well with this (but fixes are in the works) - so we had to re-define all the methods in the layout subclasses to make javadoc happy.

PR #34, opened 2020-02-27
8240173: Confusing overflow error when trying to dereference a nothing segment

mcimadamore:
This simple patch attempts to generate a more explicit error message when trying to dereference an address which is based on the Nothing segment. Note that the first problem here is caused by the fact that the Nothing segment is considered "small" - which means most of the real address values will fall outside its range. While we could simply fix that (which will improve the error message - from reporting an overflow during offset computation to report an access outside the bounds of the segment), I thought it was better to mark the Nothing segment as non-accessible, and generate an explicit error message.

---------
### Progress

### Issue
 * [JDK-8240173](https://bugs.openjdk.java.net/browse/JDK-8240173): Confusing overflow error when trying to dereference a nothing segment

### Reviewers
 * Paul Sandoz ([psandoz](@PaulSandoz) - Committer) ⚠️ Review applies to d3ea2ea0f0375d89cfbfc4ada87da4709e298944

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/34/head:pull/34`
`$ git checkout pull/34`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemorySegmentImpl.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemorySegmentImpl.java
@@ -60,14 +60,19 @@ public final class MemorySegmentImpl implements MemorySegment, MemorySegmentProx
 
     final static int READ_ONLY = 1;
     final static int SMALL = READ_ONLY << 1;
+    final static int NO_ACCESS = SMALL << 1;
     final static long NONCE = new Random().nextLong();
 
     public static MemorySegmentImpl NOTHING =
-            new MemorySegmentImpl(0, null, 0, 0, null, MemoryScope.GLOBAL);
+            new MemorySegmentImpl(0, null, 0, NO_ACCESS, null, MemoryScope.GLOBAL);
 
-    public MemorySegmentImpl(long min, Object base, long length, int mask, Thread owner, MemoryScope scope) {
+    public MemorySegmentImpl(long min, Object base, long length, Thread owner, MemoryScope scope) {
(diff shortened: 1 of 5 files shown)


slowhog:
Looks good. I think making NO_ACCESS explicit rather than depending on size checking is the right thing and glad to see the unused mask argument gone before we actually need it.

slowhog:
Looks good. I think making NO_ACCESS explicit rather than depending on size checking is the right thing and glad to see the unused mask argument gone before we actually need it.

mcimadamore:
> This started to make me wonder whether this ?nothing? segment is really an ?untrusted? segment, which could even have a bound. It does represent something, but that something cannot be fully trusted and as such its contents cannot from be accessed within Java code.

There is an overlapping between Untrusted and Nothing. In the past we considered having distinction between trusted (or managed) and not trusted. Having the notion of the Nothing segment allowed us to speak about non-trusted addresses in a way which didn't need any other special concepts (and to reuse the 'rebase' operation to trust an untrusted address). In other words, there is a certain appeal in not adding a new kind of segment (untrusted) and simply piggy back on the concepts we already have. At the time, the only way to make a segment inaccessible was to give it zero length - but I'm planning a revamp of the access modes, to expose, instead of a single asReadOnly method a set of orthogonal access modes READ | WRITE | CLOSE - at which an untrusted segment will simply be a segment whose READ | WRITE (and maybe CLOSE) bits are unset.

PaulSandoz:
Re: access modes.  That makes sense.  I was not suggesting that a new segment type be added.  The characteristic of such a segment can be derived from its properties as you indicate.

PR #35, opened 2020-02-27
8240181: jextract does not generate constants for anonymous enum fields

sundararajana:
anonymous enums were filtered by TreeMaker.

---------
### Progress

### Issue
 * [JDK-8240181](https://bugs.openjdk.java.net/browse/JDK-8240181): jextract does not generate constants for anonymous enum fields

### Reviewers
 * Maurizio Cimadamore ([mcimadamore](@mcimadamore) - Committer) ⚠️ Review applies to 4d78fe58d5d7713a356263a508f60bff61fa583e

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/35/head:pull/35`
`$ git checkout pull/35`

### DIFF
--- a/test/jdk/tools/jextract/JextractToolProviderTest.java
+++ b/test/jdk/tools/jextract/JextractToolProviderTest.java
@@ -37,6 +37,7 @@ import static org.testng.Assert.assertTrue;
  * @test
  * @modules jdk.incubator.jextract
  * @build JextractToolRunner
+ * @bug 8240181
  * @run testng/othervm -Duser.language=en --add-modules jdk.incubator.jextract JextractToolProviderTest
  */
 public class JextractToolProviderTest extends JextractToolRunner {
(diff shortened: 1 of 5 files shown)


test/jdk/tools/jextract/JextractToolProviderTest.java
```diff
@@ -37,6 +37,7 @@
  * @test
  * @modules jdk.incubator.jextract
  * @build JextractToolRunner
+ * @bug 8240181
  * @run testng/othervm -Duser.language=en --add-modules jdk.incubator.jextract JextractToolProviderTest
  */
 public class JextractToolProviderTest extends JextractToolRunner {
```
mcimadamore: One comment on testing - note that we also have a separate framework to just test the API (w/o jextract). After all, this is not, strictly speaking a jextract bug, but a bug in the declaration API parsing, so perhaps a more direct test would be preferrable?

sundararajana: Agree. This test is ported from older jextract tests (doesn't mean that we've to use the same  test holder class). I'll fix this.

sundararajana: On extract API testing. I'll address that under another bug.

slowhog:
Hmm, I sneak in the fix in the Attribtue API, but I think yours is probably better as it won't depend on clang Cursor.

PR #36, opened 2020-02-28
8239490: jextract throws "No matching declaration found for bitfield" because wrong bitfield content is found

mcimadamore:
This patch fixes an issue wit bitfields handling; the issue is that we are using a map to keep track of which layouts are associated with bitfield contents. Unfortunately, most of the times, bitfield contents is associated with "shared" layouts, such as int layouts - so we end up overwriting the map, associating different contents to the same underlying layout and this causes downstream issues in the code that needs to access this information.

A more robust solution is to use layout annotations to do this - these are not exposed in a public API, but we already have some reflective hacks in order to use them from within jextract. I've added one more hack to add a single annotation to a layout. Longer term, we plan to make this part of the layout API public, so that no more hacks will be required (as we have other use cases for public metadata on layouts).

I've also added a test to verify that bitfields are correctly parsed. While writing the test, I've realized that our testing framework, while decent, can probably be streamlined more, I'll start thinking about ways on how to do that as a separate change, so that we can write these kind of tests more quickly (which hopefully will lead us to write more of them).

---------
### Progress

### Issue
 * [JDK-8239490](https://bugs.openjdk.java.net/browse/JDK-8239490): jextract throws "No matching declaration found for bitfield" because wrong bitfield content is found

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/36/head:pull/36`
`$ git checkout pull/36`


make/autoconf/lib-clang.m4
```diff
@@ -102,14 +102,14 @@ AC_DEFUN_ONCE([LIB_SETUP_LIBCLANG],
...
 
-    BASIC_FIXUP_PATH(CLANG_INCLUDE_PATH)
-    BASIC_FIXUP_PATH(CLANG_LIB_PATH)
-    BASIC_FIXUP_PATH(CLANG_INCLUDE_AUX_PATH)
+    UTIL_FIXUP_PATH(CLANG_INCLUDE_PATH)
+    UTIL_FIXUP_PATH(CLANG_LIB_PATH)
+    UTIL_FIXUP_PATH(CLANG_INCLUDE_AUX_PATH)
 
     if test "x$CLANG_INCLUDE_PATH" != "x"; then
         LIBCLANG_CPPFLAGS="-I$CLANG_INCLUDE_PATH"
```
mcimadamore: These changes are necessary, as the build seems to be broken after the latest merge from jdk/jdk - these are orthogonal to the fix tackled in this issue

PR #37, opened 2020-03-02
8240300: jextract produces non compilable code with repeated declarations

sundararajana:
Filtering repeated global variables and functions in code generator (added necessary hashCode, equals for decls, types)

---------
### Progress

### Issue
 * [JDK-8240300](https://bugs.openjdk.java.net/browse/JDK-8240300): jextract produces non compilable code with repeated declarations

### Reviewers
 * Maurizio Cimadamore ([mcimadamore](@mcimadamore) - Committer) ⚠️ Review applies to f1720cf0f8c23ab2abcf671b8000082493cab707
 * Henry Jen ([henryjen](@slowhog) - Committer)

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/37/head:pull/37`
`$ git checkout pull/37`

### DIFF
--- a/src/jdk.incubator.jextract/share/classes/jdk/incubator/jextract/tool/StaticWrapperSourceFactory.java
+++ b/src/jdk.incubator.jextract/share/classes/jdk/incubator/jextract/tool/StaticWrapperSourceFactory.java
@@ -26,7 +26,9 @@
 package jdk.incubator.jextract.tool;
 
 import java.lang.invoke.MethodType;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 import jdk.incubator.foreign.FunctionDescriptor;
 import jdk.incubator.foreign.MemoryAddress;
 import jdk.incubator.foreign.MemoryLayout;
@@ -35,6 +37,9 @@ import jdk.incubator.jextract.Declaration;
 import jdk.incubator.jextract.Type;
 
 public class StaticWrapperSourceFactory extends HandleSourceFactory {
+    private final Set<Declaration.Variable> variables = new HashSet<>();
+    private final Set<Declaration.Function> functions = new HashSet<>();
+
     public StaticWrapperSourceFactory(String clsName, String pkgName, List<String> libraryNames) {
         super(clsName, pkgName, libraryNames);
     }
@@ -46,6 +51,9 @@ public class StaticWrapperSourceFactory extends HandleSourceFactory {
 
     @Override
     public Void visitFunction(Declaration.Function funcTree, Declaration parent) {
+        if (! functions.add(funcTree)) {
+            return null;
+        }
         MethodType mtype = typeTranslator.getMethodType(funcTree.type());
         FunctionDescriptor descriptor = Type.descriptorFor(funcTree.type()).orElse(null);
         if (descriptor == null) {
@@ -91,6 +99,10 @@ public class StaticWrapperSourceFactory extends HandleSourceFactory {
 
     @Override
     public Void visitVariable(Declaration.Variable tree, Declaration parent) {
+        if (parent == null && !(variables.add(tree))) {
+            return null;
+        }
+
         String fieldName = tree.name();
         String symbol = tree.name();
         assert !symbol.isEmpty();
--- a/src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/DeclarationImpl.java
+++ b/src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/DeclarationImpl.java
@@ -94,6 +94,11 @@ public abstract class DeclarationImpl implements Declaration {
         return name().equals(decl.name());
     }
 
+    @Override
+    public int hashCode() {
+        return Objects.hash(name);
+    }
+
     public static final class VariableImpl extends DeclarationImpl implements Declaration.Variable {
 
         final Variable.Kind kind;
@@ -157,7 +162,7 @@ public abstract class DeclarationImpl implements Declaration {
 
         @Override
         public int hashCode() {
-            return Objects.hash(kind, type);
+            return Objects.hash(super.hashCode(), kind, type);
         }
     }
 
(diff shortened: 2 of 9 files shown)


src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/DeclarationImpl.java
```diff
@@ -176,6 +191,20 @@ public Function withAttributes(Map<String, List<Constable>> attrs) {
         public Function stripAttributes() {
             return new FunctionImpl(type, params, name(), pos(), null);
         }
+
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) return true;
+            if (o == null || getClass() != o.getClass()) return false;
+            FunctionImpl function = (FunctionImpl) o;
+            return params.equals(function.params) &&
```
slowhog: Type would have covered both parameter type and return type. Compare params depends on how Variable implements equals and may cause same type declaration with different argument name to fail(not the case here).

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/TypeImpl.java
```diff
@@ -205,6 +260,21 @@ public boolean varargs() {
...
         }
+
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) return true;
+            if (o == null || getClass() != o.getClass()) return false;
+            FunctionImpl function = (FunctionImpl) o;
+            return varargs == function.varargs &&
+                    argtypes.equals(function.argtypes) &&
+                    restype.equals(function.restype);
```
slowhog: Function with typedef and raw type will not be equal?

typedef int* intptr;
void intConsumer(int *ptr);
void intConsumer(intptr);

mcimadamore: Ok, this is maybe something worth considering - should equals for type redefinition be extended/relaxed to cover the following the cases?
```
A == typedef A B 
typedef A B == A
```
I don't think the code in this patch is *wrong* - and I can see pros and cons (and use cases) for both approaches.

I guess one thing we should evaluate is whether clients which use equality e.g. for purposes similar to how equality is used in this patch (e.g. come up with a cache of types which have same properties) will benefit more from a 'smart' equality or not.

Now, in our case, since basic jextract normalizes all type references to follow typedefs, it just happens that we really care about deep type equality where typedefs have been replaced. But smarter code generators which preserve the 'shape' of the original code (e.g. which spin synthetic interface for typedeffed types), might not consider `typedef A B` and `A` as the same thing.

So, again, like in the case for declaration, I believe that, perhaps the safest move is to do what Sundar is doing here - and build every other kind of smarter comparison on top (and jextract would need such a smarter comparison too).

src/jdk.incubator.jextract/share/classes/jdk/incubator/jextract/tool/StaticWrapperSourceFactory.java
```diff
@@ -91,6 +99,10 @@ public Void visitFunction(Declaration.Function funcTree, Declaration parent) {
 
     @Override
     public Void visitVariable(Declaration.Variable tree, Declaration parent) {
+        if (parent == null && !(variables.add(tree))) {
```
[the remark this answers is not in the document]
sundararajana: Yes. Struct fields will result in the same method being called (please check parent check in the same method). We just want to check globals for repeated decls. Note that struct field re-declaration will be prevented by compiler anyway (cannot have two fields of same name in the same struct)

sundararajana:
> Looks good - thanks for adding equals/hashcode to types and declaration. I guess we should add them to the top interfaces as well (so that we can document that these things can be compared with equals()) ?

Yes. I agree. I'll add equals and hashCode to Declaration and Type as suggested.

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/DeclarationImpl.java
sundararajana: Agree. Function parameters should not be compared. Only type and (function) name should be. Will fix this and expand test to cover this case (void func(int x) and void func(int y) as duplicate declarations)

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/DeclarationImpl.java
```diff
@@ -176,6 +191,20 @@ public Function withAttributes(Map<String, List<Constable>> attrs) {
         public Function stripAttributes() {
             return new FunctionImpl(type, params, name(), pos(), null);
         }
+
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) return true;
+            if (o == null || getClass() != o.getClass()) return false;
```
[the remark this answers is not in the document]
sundararajana: Going to add equals and hashCode to Declaration and Type as suggested by Maurizio. Will fix equality of type with instanced check.

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/TypeImpl.java
sundararajana: Yes.  Not sure if we can handle it without underlying clang Type objects around. I'll see if simple cases can be handled at least.

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/TypeImpl.java
```diff
@@ -129,10 +129,10 @@ public int hashCode() {
...
         public boolean equals(Object o) {
             if (this == o) return true;
-            if (o == null || getClass() != o.getClass()) return false;
-            DelegatedBase that = (DelegatedBase) o;
-            return kind == that.kind &&
-                    name.equals(that.name);
+            if (!(o instanceof Type.Delegated)) return false;
+            Type.Delegated that = (Type.Delegated) o;
+            return kind == that.kind() &&
+                    name.equals(that.name());
```
slowhog: Delegate is tricky, I notice that we didn't check for type equals.

Just wondering what do we want for equals on multiple-delegated type in different order? Like a unsigned atomic int and a atomic unsigned int?

PR #40, opened 2020-03-05
Improve jextract error reporting

JornVernee:
Hi,

I spent the last couple of days tracking down a crash I was seeing sometimes when running jextract. In the process I ended up doing several improvements to jextract's error checking and reporting, that I think are useful to keep in the long run.

The changes include:
- Using `clang_parseTranslationUnit2` to parse, since that returns an error code, instead of a null translation unit. This also means we can still process diagnostic for the returned translation unit (now returned in a buffer).
- Checking the return codes for `clang_parseTranslationUnit2`, `clang_saveTranslationUnit`, and `clang_reparseTranslationUnit` (using a proper enum) and throwing an exception in case they are erroneous.
- Catching BadMacroException instead of Throwable in the macro parser. We only care about the former, and catching Throwable could lead to too many exceptions being eaten. Technically, if there is an unknown exception, we could just skip the macro, but this isn't always the right strategy. For example, when re-parsing fails the used translation unit becomes invalid, so we really need to do something else than just skipping the macro in that case (to be addressed in another patch https://bugs.openjdk.java.net/browse/JDK-8240614). Imo it's better to just crash than to skip when we encounter an unknwon exception, so that we're more likely to find issues.
- Enabling libclang crash recovery by default. By default the crash recovery was being disabled. This leads to the process exiting with OS exception code. If we enable this however, it will instead try to return an error code from the libclang call, which we can then check for and handle the failure however we want.

Thanks,
Jorn

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/40/head:pull/40`
`$ git checkout pull/40`

### DIFF
--- a/src/jdk.incubator.jextract/share/classes/jdk/internal/clang/LibClang.java
+++ b/src/jdk.incubator.jextract/share/classes/jdk/internal/clang/LibClang.java
@@ -36,8 +36,8 @@ public class LibClang {
     public static Index createIndex(boolean local) {
         Index index = new Index(Index_h.clang_createIndex(local ? 1 : 0, 0));
         Index_h.clang_toggleCrashRecovery(CRASH_RECOVERY ? 1 : 0);
-        if (DEBUG && !CRASH_RECOVERY) {
-            System.err.println("LibClang crash recovery disabled");
+        if (DEBUG) {
+            System.err.println("LibClang crash recovery " + (CRASH_RECOVERY ? "enabled" : "disabled"));
         }
         return index;
     }
--- a/src/jdk.incubator.jextract/share/classes/jdk/internal/clang/LibClang.java
+++ b/src/jdk.incubator.jextract/share/classes/jdk/internal/clang/LibClang.java
@@ -31,12 +31,12 @@ import jdk.internal.clang.libclang.Index_h;
 
 public class LibClang {
     private static final boolean DEBUG = Boolean.getBoolean("libclang.debug");
-    private static final boolean NO_CRASH_RECOVERY = Boolean.getBoolean("libclang.disable_crash_recovery");
+    private static final boolean CRASH_RECOVERY = Boolean.getBoolean("libclang.crash_recovery");
 
     public static Index createIndex(boolean local) {
         Index index = new Index(Index_h.clang_createIndex(local ? 1 : 0, 0));
-        Index_h.clang_toggleCrashRecovery(NO_CRASH_RECOVERY ? 0 : 1);
-        if (DEBUG && NO_CRASH_RECOVERY) {
+        Index_h.clang_toggleCrashRecovery(CRASH_RECOVERY ? 1 : 0);
+        if (DEBUG && !CRASH_RECOVERY) {
             System.err.println("LibClang crash recovery disabled");
         }
         return index;
--- a/src/jdk.incubator.jextract/share/classes/jdk/internal/clang/SaveError.java
+++ b/src/jdk.incubator.jextract/share/classes/jdk/internal/clang/SaveError.java
@@ -25,31 +25,37 @@
  */
 package jdk.internal.clang;
 
-import jdk.incubator.foreign.MemoryAddress;
-import jdk.incubator.foreign.MemorySegment;
-import jdk.internal.clang.libclang.Index_h;
+import java.util.Arrays;
+import java.util.Map;
+import java.util.NoSuchElementException;
+import java.util.function.Function;
 
-public class LibClang {
-    private static final boolean DEBUG = Boolean.getBoolean("libclang.debug");
-    private static final boolean NO_CRASH_RECOVERY = Boolean.getBoolean("libclang.disable_crash_recovery");
+import static java.util.stream.Collectors.toMap;
+import static jdk.internal.clang.libclang.Index_h.CXSaveError_InvalidTU;
+import static jdk.internal.clang.libclang.Index_h.CXSaveError_None;
+import static jdk.internal.clang.libclang.Index_h.CXSaveError_TranslationErrors;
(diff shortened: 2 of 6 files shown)


src/jdk.incubator.jextract/share/classes/jdk/internal/clang/SaveError.java
```diff
@@ -0,0 +1,61 @@
...
+
+    public int code() {
+        return code;
+    }
+
+    private static final Map<Integer, SaveError> lookup = Arrays.stream(values())
+            .collect(toMap(SaveError::code, Function.identity()));
+
+    public static SaveError valueOf(int code) {
+        return lookup.computeIfAbsent(code, k -> { throw new NoSuchElementException("No SaveError with code: " + k); });
```
sundararajana: Do we need this computeIfAbsent with lambda? All we do is to throw exception always if key is not found.

src/jdk.incubator.jextract/share/classes/jdk/internal/clang/TranslationUnit.java
```diff
@@ -62,9 +62,8 @@ public Cursor getCursor() {
 
     public final void save(Path path) throws TranslationUnitSaveException {
         try (MemorySegment pathStr = Utils.toNativeString(path.toAbsolutePath().toString())) {
-            int res = Index_h.clang_saveTranslationUnit(tu,
-                    pathStr.baseAddress(), 0);
-            if (res != 0) {
+            SaveError res = SaveError.valueOf(Index_h.clang_saveTranslationUnit(tu, pathStr.baseAddress(), 0));
+            if (res != SaveError.None) {
                 throw new TranslationUnitSaveException(path);
```
sundararajana: Would it be useful to store the SaveError kind in TranslationUnitSaveException?

src/jdk.incubator.jextract/share/classes/jdk/internal/clang/LibClang.java
```diff
@@ -31,11 +31,12 @@
 
 public class LibClang {
     private static final boolean DEBUG = Boolean.getBoolean("libclang.debug");
+    private static final boolean NO_CRASH_RECOVERY = Boolean.getBoolean("libclang.disable_crash_recovery");
 
     public static Index createIndex(boolean local) {
         Index index = new Index(Index_h.clang_createIndex(local ? 1 : 0, 0));
-        Index_h.clang_toggleCrashRecovery(0);
-        if (DEBUG) {
+        Index_h.clang_toggleCrashRecovery(NO_CRASH_RECOVERY ? 0 : 1);
```
sundararajana: I vaguely recall a specific reason for this default. Maurizio may recall about this

src/jdk.incubator.jextract/share/classes/jdk/internal/clang/TranslationUnit.java
JornVernee: Yes, good idea.

src/jdk.incubator.jextract/share/classes/jdk/internal/clang/SaveError.java
JornVernee: `Map` doesn't have a `getOrThrow`, but `computeIfAbsent` effectively works the same way. The alternative would be `get` plus null check.

JornVernee:
Wrt the crash recovery; It seems to work reliably on Windows, so I think we should at least turn it on there, since the alternative is having a jextract that crashes quite frequently on Windows (maybe 1 in 5 runs, depending on the header contents).

sundararajana:
> Looks good, my only concern is the disable clang crash recovery default was disabled but now default to enable.
> 
> The past experience is that crash recovery cause JVM crash from time to time, so unless we are certain that situation changes and should be default to enable, perhaps we should just default to disable as before.

I agree with Henry. It's better to restore the default but make it configurable via system property. Having different defaults on different platforms would be confusing to the user. Because it is configurable, we can try the config for debugging on any platform.

JornVernee:
I've uploaded another revision where I've change the default of crash recovery back to disabled, but kept the property.

src/jdk.incubator.jextract/share/classes/jdk/internal/clang/LibClang.java
```diff
@@ -31,12 +31,12 @@
...
     private static final boolean DEBUG = Boolean.getBoolean("libclang.debug");
-    private static final boolean NO_CRASH_RECOVERY = Boolean.getBoolean("libclang.disable_crash_recovery");
+    private static final boolean CRASH_RECOVERY = Boolean.getBoolean("libclang.crash_recovery");
 
     public static Index createIndex(boolean local) {
         Index index = new Index(Index_h.clang_createIndex(local ? 1 : 0, 0));
-        Index_h.clang_toggleCrashRecovery(NO_CRASH_RECOVERY ? 0 : 1);
-        if (DEBUG && NO_CRASH_RECOVERY) {
+        Index_h.clang_toggleCrashRecovery(CRASH_RECOVERY ? 1 : 0);
+        if (DEBUG && !CRASH_RECOVERY) {
```
sundararajana: Crash recovery is disabled by default, right? Shouldn't the debug for enabling? (because that is the explicit config behaviour - or else this message will be printed always)

JornVernee: This was the way it was before when crash recovery was always being disabled. It would always print the message when running in debug mode.

I guess now we can print whether the crash recovery is enabled or disabled instead.

sundararajana: Looks good. Go ahead!

PR #45, opened 2020-03-11
8240874: Add finer-grained access control for memory segments

mcimadamore:
This patch adds a finer-grained access control for memory segments; instead of the ad-hoc MemorySegment::asReadOnly, we now support a more complete MemorySegment::withAccessModes(), where access modes is an int mask.

Supported access modes are READ, WRITE, CLOSE and ACQUIRE, which gives much more freedom to clients when it comes to create restricted views (I've added a new section on that in the MemorySegment toplevel javadoc).

I also did some minor changes, mostly to get rid of confinement restrictions where such restrictions were not needed (e.g. MemorySegment::asSlice).

### DIFF
--- a/test/jdk/java/foreign/TestSegments.java
+++ b/test/jdk/java/foreign/TestSegments.java
@@ -38,6 +38,7 @@ import java.nio.ByteOrder;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Consumer;
 import java.util.function.LongFunction;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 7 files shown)


test/jdk/java/foreign/TestSegments.java
```diff
@@ -219,4 +237,51 @@ static Object defaultValue(Class<?> c) {
             return null;
         }
     }
+
+    @DataProvider(name = "accessModes")
+    public Object[][] accessModes() {
+        int nActions = AccessActions.values().length;
+        Object[][] results = new Object[nActions * nActions][];
```
JornVernee: I think this should be:
```suggestion
        Object[][] results = new Object[1 << nActions][];
```
Since the number of combinations is 2^4. Though for 4 access modes 4 * 4 and 1 << 4 are both 16, it becomes a problem if we add or remove access modes.

mcimadamore: whoops - good catch

mcimadamore:
> Suggest adding a `hasAccessModes` which does the obvious thing but can make code more readable.

I'll take a stab and see what happens.

PR #47, opened 2020-03-12
Merge openjdk/panama-foreign:foreign-memaccess

mcimadamore:
Merge with foreign-memaccess

---------
### Progress

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/47/head:pull/47`
`$ git checkout pull/47`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemorySegmentImpl.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemorySegmentImpl.java
@@ -35,6 +35,8 @@ import jdk.internal.misc.Unsafe;
 import jdk.internal.vm.annotation.ForceInline;
 
 import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Objects;
 import java.util.Random;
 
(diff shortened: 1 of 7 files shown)


test/jdk/java/foreign/TestByteBuffer.java
```diff
@@ -390,7 +390,8 @@ public void testBufferOnClosedScope() {
         try (MemorySegment segment = MemorySegment.allocateNative(bytes)) {
             leaked = segment;
         }
-        leaked.asByteBuffer();
+        ByteBuffer byteBuffer = leaked.asByteBuffer(); // ok
+        byteBuffer.get(); // should throw
     }
```
slowhog: I am a little confused with this change. What can we do with the returned ByteBuffer? Should we check the liveness  in asByteBuffer()?

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemorySegmentImpl.java
```diff
@@ -58,11 +60,16 @@
     final Thread owner;
     final MemoryScope scope;
 
-    final static int READ_ONLY = 1;
-    final static int SMALL = READ_ONLY << 1;
+    final static int SMALL = ACQUIRE << 1;
```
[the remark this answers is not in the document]
mcimadamore: I noted this as I was merging too, and I think we need to split the mask in two parts - public and private flags, and mask the private flags when users ask for them.

test/jdk/java/foreign/TestByteBuffer.java
mcimadamore: This change is just a merge change from the fix of JDK-8240874. The goal of that change is to determine a failure caused by liveness. Before this change the code was (redundantly) treating asByteBuffer as an operation which could only be performed on live segments. I've rectified that, but that means this test no longer threw the expected exception, so I added another call which was triggering the failure (get).

PR #53, opened 2020-03-13
8241017: Enhance AllocationScope to support "unbounded" mode

mcimadamore:
This patch adds support for allocation scopes whose size is not known statically; as such these new unbounded allocation scope can be used in more dynamic use cases where e.g. the amount of memory to be allocated depends on the result of other native calls.

Internally, the bounded version works as before, and is the more optimized. The unbounded version is similar in spirit to the old Panama/foreign scope - where new segments are allocated depending on needs.

I've enhanced the existing test covering allocation scopes to also cover the unbounded variants, as well as the bounded.

---------
### Progress

### Issue
 * [JDK-8241017](https://bugs.openjdk.java.net/browse/JDK-8241017): Enhance AllocationScope to support "unbounded" mode

### Reviewers
 * Jorn Vernee ([jvernee](@JornVernee) - Committer) ⚠️ Review applies to fbf5ef776be35e38adb98e710aade0dd1e3b268b

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/53/head:pull/53`
`$ git checkout pull/53`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/AllocationScope.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/AllocationScope.java
@@ -38,7 +38,7 @@ import java.util.OptionalLong;
  * of the allocation scope is known statically. If an application knows before-hand how much memory it needs to allocate the values it needs,
  * using a <em>bounded</em> allocation scope will typically provide better performances than independently allocating the memory
  * for each value (e.g. using {@link MemorySegment#allocateNative(long)}), or using an <em>unbounded</em> allocation scope.
- * For this reason, using a bound allocation scope is recommended in cases where programs might need to emulate native stack allocation.
+ * For this reason, using a bounded allocation scope is recommended in cases where programs might need to emulate native stack allocation.
  */
 public abstract class AllocationScope implements AutoCloseable {
 
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/AllocationScope.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/AllocationScope.java
@@ -43,7 +43,7 @@ import java.util.OptionalLong;
 public abstract class AllocationScope implements AutoCloseable {
 
     /**
-     * If this allocation scope is bound, returns the size, in bytes, of this allocation scope.
+     * If this allocation scope is bounded, returns the size, in bytes, of this allocation scope.
      * @return the size, in bytes, of this allocation scope (if available).
      */
     public abstract OptionalLong byteSize();
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/AllocationScope.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/AllocationScope.java
@@ -26,44 +26,33 @@
 
 package jdk.incubator.foreign;
 
-import jdk.internal.foreign.MemoryAddressImpl;
-import jdk.internal.foreign.MemorySegmentImpl;
-import jdk.internal.foreign.Utils;
+import jdk.internal.foreign.BoundedAllocationScope;
+import jdk.internal.foreign.UnboundedAllocationScope;
 
 import java.lang.invoke.VarHandle;
+import java.util.OptionalLong;
 
 /**
  * This class provides a scope of given size, within which several allocations can be performed. An allocation scope can be backed
- * either by heap, or off-heap memory (see {@link AllocationScope#heapScope(int)} and {@link AllocationScope#nativeScope(long)},
- * respectively). If an application knows before-hand how much memory it needs to allocate the values it needs,
- * using an allocation scope will typically provide better performances than independently allocating the memory for each value
- * (e.g. using {@link MemorySegment#allocateNative(long)}). For this reason, using an allocation scope is recommended
- * in cases where programs might need to emulate native stack allocation.
+ * either by heap, or off-heap memory. Allocation scopes can be either <em>bounded</em> or <em>unbounded</em>, depending on whether the size
+ * of the allocation scope is known statically. If an application knows before-hand how much memory it needs to allocate the values it needs,
+ * using a <em>bound</em> allocation scope will typically provide better performances than independently allocating the memory
+ * for each value (e.g. using {@link MemorySegment#allocateNative(long)}), or using an <em>unbounded</em> allocation scope.
+ * For this reason, using a bound allocation scope is recommended in cases where programs might need to emulate native stack allocation.
  */
-public class AllocationScope implements AutoCloseable {
-    private final MemorySegment segment, acquiredSegment;
-    private long sp = 0L;
+public abstract class AllocationScope implements AutoCloseable {
 
     /**
-     * Returns the size, in bytes, of this allocation scope.
-     * @return the size, in bytes, of this allocation scope.
+     * If this allocation scope is bound, returns the size, in bytes, of this allocation scope.
+     * @return the size, in bytes, of this allocation scope (if available).
      */
-    public long byteSize() {
-        return segment.byteSize();
-    }
+    public abstract OptionalLong byteSize();
 
     /**
      * Returns the number of allocated bytes in this allocation scope.
      * @return the number of allocated bytes in this allocation scope.
      */
-    public long allocatedBytes() {
-        return sp;
-    }
-
-    private AllocationScope(MemorySegment segment) {
-        this.segment = segment;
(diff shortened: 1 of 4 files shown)


src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/AllocationScope.java
```diff
@@ -26,44 +26,33 @@
...
 /**
  * This class provides a scope of given size, within which several allocations can be performed. An allocation scope can be backed
- * either by heap, or off-heap memory (see {@link AllocationScope#heapScope(int)} and {@link AllocationScope#nativeScope(long)},
- * respectively). If an application knows before-hand how much memory it needs to allocate the values it needs,
- * using an allocation scope will typically provide better performances than independently allocating the memory for each value
- * (e.g. using {@link MemorySegment#allocateNative(long)}). For this reason, using an allocation scope is recommended
- * in cases where programs might need to emulate native stack allocation.
+ * either by heap, or off-heap memory. Allocation scopes can be either <em>bounded</em> or <em>unbounded</em>, depending on whether the size
+ * of the allocation scope is known statically. If an application knows before-hand how much memory it needs to allocate the values it needs,
+ * using a <em>bound</em> allocation scope will typically provide better performances than independently allocating the memory
```
JornVernee: Typo
```suggestion
 * using a <em>bounded</em> allocation scope will typically provide better performances than independently allocating the memory
```

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/AllocationScope.java
```diff
@@ -26,44 +26,33 @@
...
+ * for each value (e.g. using {@link MemorySegment#allocateNative(long)}), or using an <em>unbounded</em> allocation scope.
+ * For this reason, using a bound allocation scope is recommended in cases where programs might need to emulate native stack allocation.
```
JornVernee: Same typo
```suggestion
 * For this reason, using a bounded allocation scope is recommended in cases where programs might need to emulate native stack allocation.
```

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/AllocationScope.java
```diff
@@ -26,44 +26,33 @@
...
  */
-public class AllocationScope implements AutoCloseable {
-    private final MemorySegment segment, acquiredSegment;
-    private long sp = 0L;
+public abstract class AllocationScope implements AutoCloseable {
 
     /**
-     * Returns the size, in bytes, of this allocation scope.
-     * @return the size, in bytes, of this allocation scope.
+     * If this allocation scope is bound, returns the size, in bytes, of this allocation scope.
```
JornVernee:
```suggestion
     * If this allocation scope is bounded, returns the size, in bytes, of this allocation scope.
```

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/UnboundedAllocationScope.java
```diff
@@ -0,0 +1,87 @@
...
+    @Override
+    public MemoryAddress allocate(long bytesSize, long bytesAlignment) {
+        for (int i = 0; i < 2; i++) {
+            long min = ((MemoryAddressImpl) segment.baseAddress()).unsafeGetOffset();
+            long start = Utils.alignUp(min + sp, bytesAlignment) - min;
+            try {
+                MemorySegment slice = segment.asSlice(start, bytesSize)
+                        .withAccessModes(MemorySegment.READ | MemorySegment.WRITE | MemorySegment.ACQUIRE);
+                sp = start + bytesSize;
+                size += Utils.alignUp(bytesSize, bytesAlignment);
```
JornVernee: Aren't `sp` and `size` always the same? Are 2 separate fields really needed?

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/UnboundedAllocationScope.java
```diff
@@ -0,0 +1,87 @@
...
+                return slice.baseAddress();
+            } catch (IndexOutOfBoundsException ex) {
+                sp = 0L;
+                usedSegments.add(segment);
+                segment = segmentFactory.apply(BLOCK_SIZE);
```
slowhog: So we have a hard limit of allocation size no larger then 4K, is the intentional?

I would like to see what we have for NativeScope  in the foreign-branch for unbounded, when allocation larger than the block, we can just allocate it and give back that as a single-used MS.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/UnboundedAllocationScope.java
mcimadamore: The idea is that size keeps track of total allocation size, sp is just a pointer into current segment and is reset every time - unless I made a mistake

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/UnboundedAllocationScope.java
mcimadamore: I considered allocating standalone segments of any size - this is easy to do in the implementation, but there's an issue: we do not have any way to guarantee alignment for heap-allocated segments, so we can't really preserve the behavior that we had before.

I'd prefer to go ahead with this change as is - but we should keep thinking about how we want to support (if at all) alignment for heap allocated allocation scopes. One option could be to drop heap allocation scope altogether - since alignment doesn't really make sense there. Perhaps one day we will have a way to create aligned arrays in the heap, in which case we might be able to add that functionality back.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/UnboundedAllocationScope.java
JornVernee: Oh I see. Yeah, then it makes sense to have 2 fields.

mcimadamore:
> I have some disagreements on the implementation. The expected behaviour,
> in my mind anyway, is that the returned MemoryAddress is *ALWAYS* just
> an offset of a larger, always same MemorySegment,

The goal of AllocationScope is to provide some extra aid to those clients who need to allocate multiple segments which feature the same temporal bounds - e.g. so that they can be closed all at once. It gives you back an address to the allocated memory. How it's implemented inside, it's well, an implementation detail, and I don't really see why a client should assume certain behavior out of the addresses that it gets out of AllocationScope. Do you expect that all addresses given back to you via malloc are consecutive? I hope not!

As for resizing a segment dynamically, I've said many times that the design centre for memory segment API is that to have immutable segments, so that bound checks can be better optimized. If that's the constraint, then, in order to have unbounded allocation, you have to allocate multiple segments and keep them in a list. Maybe we can do some tuning on the threshold and make the impl smarter at reusing the memory, but this is mostly beyond the point - as this is not supposed to be an highly efficient allocator - this is just supposed to be an help for clients building native application - one thing that we found was pretty common for clients to do when we started porting more and more code using the new bindings.

mcimadamore:
> If this is the intended purpose then it should, IMO, be changed to
> "BlockScope". "UnboundedAllocationScope" implies same-segment, dynamic
> allocation.

This thread is essentially a request for adding something that you asked already few weeks ago - adding support for realloc. I've explained then why it's not a great idea, and how it doesn't fit into where we wanna go (other people chimed in too, saying that you might be reading too much into what realloc does).

Anyway, this being Java, I find it funny that we have an interface, and we're arguing over whether the implementation is desirable or not. The interface allows you to allocate multiple chunks of memory, making _no guarantees_ as to whether the chunks will be allocated. What you are asking is _not_ just an implementation change. You want the interface to make stronger guarantees about the fact that all the memory that will ever be returned by the scope will be _contiguous_. That's not going to happen.

As for the name, an allocation scope just provides a 'scope' within which all allocation happens, so that you can close everything at once. In the bounded case, you can come up with a block size ahead of time, hence the single segment. In the unbounded case, we're doing what the old Panama/foreign scope was doing; which I think is fine enough for simple use cases (which is what this interface is for).

If you want to support a realloc-based strategy on top of memory segment you would have to invalidate the existing segments on each realloc call (because you don't know whether realloc has decided to just realloc, or to alloc + copy + free, in which case old segments are just invalid).

Using stdlib is possible - in fact we have a test for it - maybe you should look there:

https://github.com/openjdk/panama-foreign/blob/686b4fafad1c198273b218f6df39deba5268d7da/test/jdk/java/foreign/StdLibTest.java

To get to the standard library you need LibraryLookup::ofDefault.

PR #54, opened 2020-03-18
JDK-8241154: Clarify the role of MemorySegments

mcimadamore:
This patch proposes a restacking of the memory access API. Currently, memory segments are playing a dual role: they are both view of a resource, and they impersonate the resource itself. This creates confusion when thinking about operationssuch as close() and acquire().

The idea put forward by this patch is to put all segments on equal footing; e.g. remove the distinction between normal segments and *acquired* segments. Now *all* segments are, in a sense, acquired from some memory *source*. A memory source, in other words, model the actual memory that the segment is a view of.

Memory sources are unconfined, which makes them ideal to support operation such as registration with cleaners (to allow for automatic cleanup, where needed). Moreover, since we can support many kinds of memory sources, this patch also adds a MappedMemorySource which is specific to mapped segments; such memory source contains methods for syncing contents of memory against the mapped file (e.g. force()). 

This split between memory segment and memory source allows us to keep the memory segment API sane, while at the same time providing us room to expand the API in the future to add more memory sources. And it makes the API cleaner too, as we can put methods where they belong (e.g. see difference between MemorySegment::isAlive vs. MemorySource::isReleased).

A javadoc for this refactoring is available here:

http://cr.openjdk.java.net/~mcimadamore/panama/8241154_javadoc/javadoc/jdk/incubator/foreign/package-summary.html

Many thanks to Brian, John, Jorn, Paul and Stuart for the feedback which led to this iteration.

---------
### Progress

### Issue
 * [JDK-8241154](https://bugs.openjdk.java.net/browse/JDK-8241154): Clarify the role of MemorySegments

### Reviewers
 * Paul Sandoz ([psandoz](@PaulSandoz) - Committer) ⚠️ Review applies to b2ddc2f38a092f74b76d64ed90ed683116f8aa53
 * Jorn Vernee ([jvernee](@JornVernee) - Committer) ⚠️ Review applies to b2ddc2f38a092f74b76d64ed90ed683116f8aa53

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/54/head:pull/54`
`$ git checkout pull/54`


src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MappedMemorySource.java
```diff
@@ -0,0 +1,81 @@
...
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+package jdk.incubator.foreign;
+
+/**
+ * A mapped memory source describes an underlying memory region which is memory-mapped from a given path.
+ * As such, it provides additional operations to allow for syncing the contents of the mapped memory source with
+ * the ones of the underlying file backing the memory mapping (see {@link #force()} and {@link #force(MemorySegment)}).
```
PaulSandoz: s/with the ones of/with that of

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MappedMemorySource.java
```diff
@@ -0,0 +1,81 @@
...
+ */
+public interface MappedMemorySource extends MemorySource {
+    /**
+     * Forces any changes made to this memory source to be written to the
+     * storage device containing the path associated with this memory source.
+     *
+     * <p> If the path associated with this memory source resides on a local storage
+     * device then when this method returns it is guaranteed that all changes
+     * made to this memory source since it was created, or since this method was last
+     * invoked, will have been written to that device.
```
PaulSandoz: Something to consider later: we might need to document a happens-before relation.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MappedMemorySource.java
```diff
@@ -0,0 +1,81 @@
...
+     *
+     * <p> If the path associated to this memory source does not reside on a local device,
+     * then no such guarantee is made.
+     *
+     * <p> If this the map mode associated with this memory source is not ({@link
```
PaulSandoz: s/if this the/if the ?

How does one obtain the map mode?

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MappedMemorySource.java
```diff
@@ -0,0 +1,81 @@
...
+     *
+     * <p> If the path associated with this memory source resides on a local storage
+     * device then when this method returns it is guaranteed that all changes
+     * made to this memory source since it was created, or since this method was last
+     * invoked, will have been written to that device.
+     *
+     * <p> If the path associated to this memory source does not reside on a local device,
+     * then no such guarantee is made.
+     *
+     * <p> If this the map modes associated with this memory source is not ({@link
```
PaulSandoz: s/if this the/if the ?

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MappedMemorySource.java
```diff
@@ -0,0 +1,81 @@
...
+     * java.nio.channels.FileChannel.MapMode#READ_WRITE}) then invoking this method may have no effect.
+     * This method may or may not have an effect for implementation-specific mapping modes.</p>
+     *
+     * @param segment the memory segment whose contents are to be written back to storage.
+     *
+     * @throws IllegalStateException if the memory region backing this memory source has already been released (see
```
PaulSandoz: And also if the segment is not derived from this memory source?

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySource.java
```diff
@@ -0,0 +1,71 @@
...
+ */
+public interface MemorySource {
+    /**
+     * Has this memory source been released?
+     * @return {@code true}, if this memory source been released.
+     */
+    boolean isReleased();
+
+    /**
+     * Register this memory source against a {@link java.lang.ref.Cleaner}; this means that when all memory segments
```
PaulSandoz: s/Register/Registers

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -52,39 +52,32 @@
...
- * by native memory.
+ * It is possible to obtain a memory segment backed by a memory-mapped file using the factory method
+ * {@link MemorySegment#mapFromPath(Path, long, FileChannel.MapMode)}. Such memory segments are backed by a
+ * <em>mapped</em> memory source (see {@link MappedMemorySource}).
  * <p>
- * Finally, it is also possible to obtain a memory segment backed by a memory-mapped file using the factory method
- * {@link MemorySegment#mapFromPath(Path, long, FileChannel.MapMode)}. Such memory segments are called <em>mapped memory segments</em>.
+ * Finally, it is also possible to obtain a memory segment backed by an existing Java byte buffer (see {@link ByteBuffer}),
+ * using the factory method {@link MemorySegment#ofByteBuffer(ByteBuffer)}. Memory segments obtained in this can be
+ * backed by a native memory source, an heap memory source, or a mapped memory source, depending on the characteristics
```
PaulSandoz: s/an heap/a heap

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -152,12 +142,11 @@
     MemoryAddress baseAddress();
 
     /**
-     * Obtains an <a href="#thread-confinement">acquired</a> memory segment which can be used to access memory associated
-     * with this segment from the current thread. As a side-effect, this segment cannot be closed until the acquired
-     * view has been closed too (see {@link #close()}).
-     * @return an <a href="#thread-confinement">acquired</a> memory segment which can be used to access memory associated
+     * Obtains a new memory segment backed by the same memory source as this segment which can be used to access memory associated
```
PaulSandoz: Consider later perhaps: now we have `MemorySource` we could name this method `share` since we have the ability to talk about what is shared.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MappedMemorySource.java
mcimadamore: Obtaining map modes is hard, although I agree it would be desirable. Unfortunately the MappedByteBuffer class throws them away soon after creation, and only keeps an 'isSync' flag around (which is just for NV memory).

In other words, the javadoc of this method mimics that of MappedByteBuffer - similarly, MappedByteBuffer has no way to tell which map mode was used at creation, and the javadoc vaguely refers to the map modes being used at FileChannel::map time. I'm trying to do the same here.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MappedMemorySource.java
mcimadamore: right

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
mcimadamore: Not 100% sold on `share`. I don't like `acquire` that much, but I think `acquire` suggest some act performed by the current thread on the underlying memory source (at least that's how my brain parses it). For some reason, `share` seems to imply that some state change happens in the segment on which the method is called (e.g. we go from confined to shared mode, which is not really what happens) ?

mcimadamore:
Uploaded new version which addresses review comments and adds some extra API points to obtain raw address/base object associated with memory source (this makes ForeignUnsafe redundant). Javadoc available here:

http://cr.openjdk.java.net/~mcimadamore/panama/8241154_javadoc_v2/javadoc/jdk/incubator/foreign/package-summary.html

mcimadamore:
Just realized that for this to work properly I need to add reachabilty fences in all the var handle code (at least, possibly AB invokers need to do the same). Should I do it as part of this efort, or should I split and deal with registerCleaner later?

PaulSandoz:
Drat, those fences are a pain. Might be easier to separate out (esp. for review purposes, i am already loosing track of what is going on!).

mcimadamore:
Closing this for now - we'll bring some of this stuff back in upcoming patches.

PR #55, opened 2020-03-18
8241148: need a way to create MemorySegment with contents from java String and a way to read a C char* as java String

sundararajana:
Adding utility methods to ForeignUnsafe class.

---------
### Progress

### Issue
 * [JDK-8241148](https://bugs.openjdk.java.net/browse/JDK-8241148): need a way to create MemorySegment with contents from java String and a way to read a C char* as java String

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/55/head:pull/55`
`$ git checkout pull/55`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/unsafe/ForeignUnsafe.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/unsafe/ForeignUnsafe.java
@@ -27,6 +27,7 @@
 package jdk.incubator.foreign.unsafe;
 
 import java.lang.invoke.VarHandle;
+import java.nio.charset.Charset;
 import jdk.incubator.foreign.MemoryAddress;
 import jdk.incubator.foreign.MemoryLayout;
 import jdk.incubator.foreign.MemorySegment;
@@ -95,24 +96,34 @@ public final class ForeignUnsafe {
      * @param str the Java String
      * @return a new native memory segment
      */
-    public static MemorySegment makeNativeString(String str) {
-        return makeNativeString(str, str.length() + 1);
+    public static MemorySegment toCString(String str) {
+        return toCString(str.getBytes());
     }
 
-    private static MemorySegment makeNativeString(String str, int length) {
-        MemoryLayout strLayout = MemoryLayout.ofSequence(length, C_CHAR);
+    /**
+     * Returns a new native memory segment holding contents of the given Java String
+     * @param str The Java String
+     * @param charset The Charset to be used to encode the String
+     * @return a new native memory segment
+     */
+    public static MemorySegment toCString(String str, Charset charset) {
+        return toCString(str.getBytes(charset));
+    }
+
+    private static MemorySegment toCString(byte[] bytes) {
+        MemoryLayout strLayout = MemoryLayout.ofSequence(bytes.length + 1, C_CHAR);
         MemorySegment segment = MemorySegment.allocateNative(strLayout);
         MemoryAddress addr = segment.baseAddress();
-        for (int i = 0 ; i < str.length() ; i++) {
-            byteArrHandle.set(addr, i, (byte)str.charAt(i));
+        for (int i = 0 ; i < bytes.length; i++) {
+            byteArrHandle.set(addr, i, bytes[i]);
         }
-        byteArrHandle.set(addr, (long)str.length(), (byte)0);
+        byteArrHandle.set(addr, (long)bytes.length, (byte)0);
         return segment;
     }
 
     /**
      * Returns a Java String from the contents of the given C '\0' terminated string
-     * @param addr the address of the C string
+     * @param addr The address of the C string
      * @return a Java String
      */
     public static String toJavaString(MemoryAddress addr) {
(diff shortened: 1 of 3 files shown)


test/jdk/java/foreign/Test8241148.java
```diff
@@ -0,0 +1,85 @@
...
+        try (var seg = ForeignUnsafe.makeNativeString("java")) {
+            assertEquals((int) strlen.invoke(seg.baseAddress()), 4);
+        }
+        try (var pathSeg = ForeignUnsafe.makeNativeString("PATH")) {
+            var path = (MemoryAddress) getenv.invoke(pathSeg.baseAddress());
+            if (!path.equals(NULL)) {
+                int len = (int) strlen.invoke(path);
+                var pathStr = ForeignUnsafe.toJavaString(path);
+                assertEquals(pathStr.length(), len);
+                System.out.println("PATH = " + pathStr);
```
mcimadamore: Should we compare it with System.getEnv?

https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/System.html#getenv()

Perhaps we could strengthen the test that way by ensuring that the naive getenv returns same values for all the keys in the System.getEnv Map.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/unsafe/ForeignUnsafe.java
```diff
@@ -81,4 +83,49 @@ public static long getUnsafeOffset(MemoryAddress address) {
...
+        return segment;
+    }
+
+    /**
+     * Returns a Java String from the contents of the given C '\0' terminated string
+     * @param addr the address of the C string
+     * @return a Java String
+     */
+    public static String toJavaString(MemoryAddress addr) {
+        StringBuilder buf = new StringBuilder();
```
slowhog: There are use cases to take a C string and make it a byte array in nio, and we have to retain the encoding of byte array, some pointed out to me that toJavaString(addr).getBytes() may not be proper if encoding is not properly specified(not my experience though).

Thus I have couple functions like following,
https://github.com/slowhog/panama-foreign/blob/eb79b8c14cfcdeae6a869d6f434929f44ab73dd5/src/java.base/unix/classes/sun/nio/FFIUtils.java#L176

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/unsafe/ForeignUnsafe.java
```diff
@@ -81,4 +83,49 @@ public static long getUnsafeOffset(MemoryAddress address) {
...
+    /**
+     * Returns a new native memory segment holding contents of the given Java String
+     * @param str the Java String
+     * @return a new native memory segment
+     */
+    public static MemorySegment makeNativeString(String str) {
+        return makeNativeString(str, str.length() + 1);
+    }
+
+    private static MemorySegment makeNativeString(String str, int length) {
```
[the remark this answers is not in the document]
slowhog: We also need a function to make a byte array into native string.

Although I think makeNativeString(new String(byte[])) should be good enough, but I think makeNativeString(string.getBytes()) is more accurate on keeping byte array encoding.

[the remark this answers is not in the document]
sundararajana: Current feature is only about conversion to/from C and Java Strings. Array conversions - not just byte[] - could be another feature.

test/jdk/java/foreign/Test8241148.java
sundararajana: Will fix this.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/unsafe/ForeignUnsafe.java
sundararajana: Will add Charset accepting variant as well.

mcimadamore:
> However, if it were move to MemoryAddress and it's assumed the address
> is safe, the method could just basically be a copy/paste of the old
> Pointer.toString from old Panama[1], with a variant that accepts a Charset.

I think the points you raise are valid ones.

We could indeed treat these conversions as 'safe' and expect the user to do the unsafe conversion.

That said, let's not focus too much on the fact that ForeignUnsafe is outside the API, as we have another change we're working on which will bring ForeignUnsafe inside the official API (guarded by some flag), with probably a different name.

So I think the choice here is whether we want to "pollute" the memory access API with functionalities which pertain to native interop or not. I think the status quo is that all the native interop stuff should go into a separate class - and keep the memory API simple (for users who want to use it w/o need to do native interop). But the distinction is blurry, and we can re-evaluate as needed.

test/jdk/java/foreign/Test8241148.java
mcimadamore: I guess what I had in mind was more a for loop where we iterate over all the System.getEnv entries and we validate that the strings we get are the same as those we get from the native function call. I'm saying this because maybe some env properties have weirder strings attached to them which might run into encoding issues and the likes - also we'd remove dependency on PATH in case it's not defined on a given system.

sundararajana:
> That said, let's not focus too much on the fact that ForeignUnsafe is outside the API, as we have another change we're working on which will bring ForeignUnsafe inside the official API (guarded by some flag), with probably a different name.

Hi Ty, 
As Maurizio mentioned, we have a plan to bring ForeignUnsafe inside the official API. So for the current change, let's keep these string utils in ForeignUnsafe.

sundararajana:
> Added comment on test

I'll fix the test to iterate through all System env. vars.

PR #56, opened 2020-03-19
8237585: Dismantle ForeignUnsafe

JornVernee:
Hi,

This is a redo of the earlier patch to dismantle ForeignUnsafe. It replaces ForeignUnsafe with the Foreign interface, and the InternalForeign implementation of that interface.

This PR removes the need to pass --add-exports to make ForeignUnsafe accessible, and replaces it with a runtime check on the `jdk.incubator.foreign.Foreign` system property, which is checked when calling `Foreign::getInstance`. The `Foreign` object returned from that method acts as an access token for accessing native interop functionality. The SystemABI::getInstance factory has also been moved to Foreign, along with the methods that were previously found in ForeignUnsafe. All have now been made non-static.

I've updated the tests and also removed the SysV ABI `jdk.internal.foreign.NativeInvoker.FASTPATH` which was no longer being used.

Jorn

---------
### Progress

### Issue
 * [JDK-8237585](https://bugs.openjdk.java.net/browse/JDK-8237585): Dismantle ForeignUnsafe

### Reviewers
 * Maurizio Cimadamore ([mcimadamore](@mcimadamore) - Committer) ⚠️ Review applies to f8d1c809156b9343e5b6ecccf42802d5973822b5

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/56/head:pull/56`
`$ git checkout pull/56`


JornVernee:
@mcimadamore Yes, thanks for mentioning that.

This is probably not the final solution, but it's an improvement over the status quo.

JornVernee:
> 
> 
> Looks like we have some unused imports?

Thanks for catching this. I had just integrated the PR, so I'll create another one to clean up the imports and some other general cleanups.

PR #59, opened 2020-03-19
Miscellaneous cleanups

JornVernee:
Hi,

This patch applies some general cleanups suggested by IntelliJ, such as:
- Removing unused imports
- Fixing typos
- Applying `final` when possible
- Using Optional::orElseThrow instead of Optional::get where it's uncertain if the optional contains a value
- Using StringBuilder::append instead of `+` in some places (to avoid indy concat).
- Changing indexed to for-each loop
- Fixing a link in javadoc
- Removing a spurious semicolon
- Removing an unused method

Jorn

---------
### Progress

### Reviewers
 * Henry Jen ([henryjen](@slowhog) - Committer)

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/59/head:pull/59`
`$ git checkout pull/59`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/Foreign.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/Foreign.java
@@ -27,11 +27,6 @@
 package jdk.incubator.foreign;
 
 import jdk.internal.foreign.InternalForeign;
-import jdk.internal.foreign.MemoryAddressImpl;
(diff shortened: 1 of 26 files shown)


mcimadamore:
> While the changes look generally good, I think there's a problem with this PR as it seems to apply to both foreign-memaccess and foregn-abi classes - I think you need to split it into two?

I see this has already been integrated - we should try to backport the changes to foreign-memaccess :-(

PR #64, opened 2020-03-24
8241504: Expose MemoryLayout annotations/attributes in the public API

JornVernee:
Hi,

This PR exposed MemoryLayout attributes in the public API.

A summary of the changes:
- Renamed annotation -> attribute everywhere to avoid confusion with Java language annotations
- Made sure that annotations were being serialized as part of a MemoryLayout's ConstantDesc, as well as the alignment which was previously left out. (The latter also meant changing PaddingLayout::hasNaturalAlignment to always return `true`, so serializing alignment is skipped for padding).
- Removed special casing for the abiType attribute. This can now instead be handled through the generic attribute mechanism.

Thanks,
Jorn

---------
### Progress

### Issue
 * ⚠️ Temporary failure when trying to retrieve information on issue `8241504`.

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/64/head:pull/64`
`$ git checkout pull/64`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryLayout.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryLayout.java
@@ -305,7 +305,7 @@ public interface MemoryLayout extends Constable {
     }
 
     /**
-     * Returns a new MemoryLayout with the given addtional attribute
+     * Returns a new MemoryLayout with the given additional attribute
      *
      * @param name the name of the attribute
      * @param value the value of the attribute
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryLayout.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryLayout.java
@@ -41,6 +41,7 @@ import java.util.OptionalLong;
 import java.util.Set;
 import java.util.function.Function;
 import java.util.function.UnaryOperator;
+import java.util.stream.Stream;
 
 /**
  * A memory layout can be used to describe the contents of a memory segment in a <em>language neutral</em> fashion.
@@ -290,6 +291,30 @@ public interface MemoryLayout extends Constable {
      * @throws IllegalArgumentException if {@code bitAlignment} is not a power of two, or if it's less than than 8.
      */
     MemoryLayout withBitAlignment(long bitAlignment);
+
+    /**
+     * Returns the attribute with the given name if it exists, or an empty optional
+     *
+     * @param name the name of the attribute
+     * @return the optional attribute
+     */
+    Optional<Constable> attribute(String name);
+
+    /**
+     * Returns a new MemoryLayout with the given addtional attribute
+     *
+     * @param name the name of the attribute
+     * @param value the value of the attribute
+     * @return the new MemoryLayout
+     */
+    MemoryLayout withAttribute(String name, Constable value);
+
+    /**
+     * Returns a stream of the names of the attributes of this layout
+     *
+     * @return the stream of names
+     */
+    Stream<String> attributes();
 
     /**
      * Computes the offset, in bits, of the layout selected by a given layout path, where the path is considered rooted in this
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryLayout.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryLayout.java
@@ -228,13 +228,6 @@ public interface MemoryLayout extends Constable {
     Optional<String> name();
 
     /**
-     * Return the ABI <em>type</em> (if any) associated with this layout.
-     *
-     * @return the layout ABI <em>type</em> (if any).
-     */
-    Optional<SystemABI.Type> abiType();
-
-    /**
      * Creates a new layout which features the desired layout <em>name</em>.
      *
      * @param name the layout name.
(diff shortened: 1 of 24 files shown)


test/jdk/java/foreign/TestLayoutConstants.java
```diff
@@ -106,6 +106,9 @@ public void testDescribeResolveFunction(MemoryLayout layout, boolean isVoid) {
                         MemoryLayout.ofStruct(
                                 MemoryLayouts.PAD_8,
                                 MemoryLayouts.BITS_32_BE)) },
+                { MemoryLayouts.BITS_32_LE.withName("myInt") },
+                { MemoryLayouts.BITS_32_LE.withBitAlignment(8) },
+                { MemoryLayouts.BITS_32_LE.withAttribute("xyz", "xyz") },
```
mcimadamore: Maybe change the test to have name != value, so that we also detect bugs where we serialize the name for both.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/aarch64/CallArranger.java
```diff
@@ -200,13 +201,13 @@ static boolean isHomogeneousFloatAggregate(MemoryLayout type) {
         if (!(baseType instanceof ValueLayout))
             return false;
 
-        var optAbiType = baseType.abiType();
+        var optAbiType = baseType.attribute(SystemABI.NATIVE_TYPE, SystemABI.Type.class);
```
mcimadamore: Since all ABI layouts are going to have an annotation of this kind, would it make sense to add a static helper function in the SystemABI.Type enum to get the ABI type from a layout e.g.

```
Type fromLayout(Layout layout) throws IllegalArgumentException //if not an ABI layout
```

test/jdk/java/foreign/TestLayoutConstants.java
JornVernee: Good idea!

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/aarch64/CallArranger.java
JornVernee: Good idea, will add this as well.

JornVernee:
I've also added some very basic tests of the new methods, mostly to sort of document the expectations w.r.t. how NAME is just another attribute, plus how using the attribute(String,Class) overload filters the optional with an instanceof check as opposed to simply casting (and potentially failing that cast with CCE).

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/SystemABI.java
```diff
@@ -174,7 +179,20 @@ default void freeUpcallStub(MemoryAddress address) {
...
-        POINTER
+        POINTER;
+
+        /**
+         * Retrieve the ABI type attached to the given layout,
+         * or throw an {@code IllegalArgumentException} if there is none
+         *
+         * @param ml the layout to retrieve the ABI type of
+         * @return the retrieved ABI type
+         * @throws IllegalArgumentException if the given layout does not have an ABI type annotation
```
mcimadamore:
```suggestion
         * @throws IllegalArgumentException if the given layout does not have an ABI type attribute
```

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryLayout.java
```diff
@@ -291,6 +285,41 @@ default long byteAlignment() {
...
+    Optional<Constable> attribute(String name);
+
+    /**
+     * Returns the attribute with the given name and the given type if it exists, or an empty optional
+     *
+     * @param name the name of the attribute
+     * @param type the type to filter by
+     * @return the optional attribute
+     */
+    default <T extends Constable> Optional<T> attribute(String name, Class<T> type) {
```
mcimadamore: Not sure I like this method. While it works assuming all constable implementation are non-generic, it will stop working correctly as soon as you have a generic class implementing a constable. Maybe save the sugar for later?

JornVernee: I see what you mean. It's not that easy to get a `Class<T>` when `T` is a generic type. I'll remove this helper for now. (adding SystemABI.Type::fromLayout removed most use-cases we currently have as well).

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/x64/windows/CallArranger.java
```diff
@@ -183,7 +183,10 @@ private static TypeClass classifyValueType(ValueLayout type) {
         } else if(clazz == ArgumentClassImpl.POINTER) {
             return TypeClass.POINTER;
         } else if (clazz == ArgumentClassImpl.SSE) {
-            if (type.attribute(VARARGS_ATTRIBUTE_NAME, String.class).map(Boolean::parseBoolean).orElse(false)) {
+            if (type.attribute(VARARGS_ATTRIBUTE_NAME)
+                    .filter(String.class::isInstance)
```
mcimadamore: Not convinced by the `filter` here? You get an Optional<Constable>, you do a cast to String and then parse as boolean - why the filtering? If there's an annotation with that name, but different type, I'd prefer to crash than to end up doing something wrong. Same applies elsewhere.

PR #65, opened 2020-03-25
8237573: Add a basic call overhead benchmark

JornVernee:
Hi,

Per my earlier email [1], this PR adds some basic call overhead benchmarks.

2 benchmarks are added:
- CallOverhead, which measures both an empty and an identity function call.
- PointsDistance, which extends the existing points benchmarks with one that measures the distance between 2 points doing the calculation is C (using a native call).

The latter also shows the difference between passing a pointer to a struct, vs passing a struct by value, PointsDistance.panama_MemoryAddress and PointsDistance.panama_MemorySegment respectively.

This acts as a basis to test the coming performance improvements in this area.

Thanks,
Jorn

[1] : https://mail.openjdk.java.net/pipermail/panama-dev/2020-March/008007.html

---------
### Progress

### Issue
 * [JDK-8237573](https://bugs.openjdk.java.net/browse/JDK-8237573): Add a basic call overhead benchmark

### Reviewers
 * Maurizio Cimadamore ([mcimadamore](@mcimadamore) - Committer) ⚠️ Review applies to b8b73345ba2540c608df5274c1d39632dbbe9bcb

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/65/head:pull/65`
`$ git checkout pull/65`


mcimadamore:
I'm trying to run this on Linux - building of the benchmark fails with this:

```
=== Output from failing command(s) repeated here ===
* For target support_test_micro_native_support_libPoint_libPoint.o:
../test/micro/org/openjdk/bench/jdk/incubator/foreign/points/support/libPoint.c: In function 'distance':
../test/micro/org/openjdk/bench/jdk/incubator/foreign/points/support/libPoint.c:34:17: error: implicit declaration of function 'abs' [-Werror=implicit-function-declaration]
     int xDist = abs(p1.x - p2.x);
                 ^~~
```
I'll try to take a look what's up and propose a suggestion.

test/micro/org/openjdk/bench/jdk/incubator/foreign/points/support/libPoint.c
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+#include <math.h>
```
mcimadamore:
```suggestion
#include <stdlib.h>
#include <math.h>
```

mcimadamore: I needed this change to make it compile on Linux (I actually need both includes).

test/micro/org/openjdk/bench/jdk/incubator/foreign/points/support/PanamaPoint.java
```diff
@@ -43,8 +43,8 @@
 public class PanamaPoint implements AutoCloseable {
 
     public static final MemoryLayout LAYOUT = MemoryLayout.ofStruct(
-        MemoryLayouts.JAVA_INT.withOrder(ByteOrder.nativeOrder()).withName("x"),
-        MemoryLayouts.JAVA_INT.withOrder(ByteOrder.nativeOrder()).withName("y")
+        MemoryLayouts.C_INT.withOrder(ByteOrder.nativeOrder()).withName("x"),
```
mcimadamore: Do we need the 'withOrder' call?

JornVernee: No, that's not needed. Must have been a left over from experimenting :)

PR #69, opened 2020-03-26
8241504: Expose MemoryLayout annotations/attributes in the public API (foreign-jextract part)

JornVernee:
Hi,

These are the jextract changes needed after exposing layout attributes in the public API. This also fixes the pending merge conflict, which was closely related to the needed changes.

Jorn

---------
### Progress

### Issue
 * [JDK-8241504](https://bugs.openjdk.java.net/browse/JDK-8241504): Expose MemoryLayout annotations/attributes in the public API

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/69/head:pull/69`
`$ git checkout pull/69`

### DIFF
--- a/test/jdk/java/foreign/NativeTestHelper.java
+++ b/test/jdk/java/foreign/NativeTestHelper.java
@@ -38,11 +38,7 @@ public class NativeTestHelper {
     public static final SystemABI ABI = Foreign.getInstance().getSystemABI();
 
     public static boolean isIntegral(MemoryLayout layout) {
-        var optAbiType = layout.abiType();
-        if (!optAbiType.isPresent()) {
-            return false;
-        }
-        return switch(optAbiType.get()) {
+        return switch(SystemABI.Type.fromLayout(layout)) {
             case BOOL, UNSIGNED_CHAR, SIGNED_CHAR, CHAR, SHORT, UNSIGNED_SHORT,
                 INT, UNSIGNED_INT, LONG, UNSIGNED_LONG, LONG_LONG, UNSIGNED_LONG_LONG -> true;
             default -> false;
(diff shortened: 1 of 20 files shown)


JornVernee:
So, this includes a merge for: https://github.com/openjdk/panama-foreign/pull/68 and some other jextract fixes. Both are needed to get foreign-jextract building & passing tests again. Just the other changes can be seen when looking at the diff for the last commit: https://github.com/openjdk/panama-foreign/pull/69/commits/a1074729fb3190614f2e59947dfbed492e968ad3

test/jdk/java/foreign/NativeTestHelper.java
```diff
@@ -38,19 +38,15 @@
     public static final SystemABI ABI = Foreign.getInstance().getSystemABI();
 
     public static boolean isIntegral(MemoryLayout layout) {
-        var optAbiType = layout.abiType();
-        if (!optAbiType.isPresent()) {
-            return false;
-        }
-        return switch(optAbiType.get()) {
+        return switch(SystemABI.Type.fromLayout(layout)) {
```
slowhog: Need to catch the exception?

JornVernee: No, I don't think so. The intent is to have a hard failure if the ABI type is missing. The old code was failing more silently by returning `false`.

JornVernee: Similar motivation here: https://github.com/openjdk/panama-foreign/pull/64#discussion_r398488164

slowhog: OK, I didn't look at use cases for this method, just thinking of compatibility of existing code. If the cases are expecting native ABI type to be available all the time, then yes, I agree.

PR #70, opened 2020-03-27
8241712: AArch64: TestDowncall fails with IllegalArgumentException: No ABI attribute present

nick-arm:
Tests cases where the structure has an initial floating point element followed by padding fail with this exception. The padding element doesn't have the NATIVE_TYPE attribute.

```
  test TestDowncall.testDowncall("f0_V_S_FD", VOID, [STRUCT], [FLOAT, DOUBLE]): failure
  java.lang.IllegalArgumentException: No ABI attribute present
          at jdk.incubator.foreign/jdk.incubator.foreign.SystemABI$Type.lambda$fromLayout$0(SystemABI.java:195)
          at java.base/java.util.Optional.orElseThrow(Optional.java:401)
          at jdk.incubator.foreign/jdk.incubator.foreign.SystemABI$Type.fromLayout(SystemABI.java:195)
          at jdk.incubator.foreign/jdk.internal.foreign.abi.aarch64.CallArranger.isHomogeneousFloatAggregate(CallArranger.java:207)
          at jdk.incubator.foreign/jdk.internal.foreign.abi.aarch64.CallArranger.classifyStructType(CallArranger.java:220)
```

---------
### Progress

### Issue
 * [JDK-8241712](https://bugs.openjdk.java.net/browse/JDK-8241712): AArch64: TestDowncall fails with IllegalArgumentException: No ABI attribute present

### Reviewers
 * Jorn Vernee ([jvernee](@JornVernee) - Committer) ⚠️ Review applies to aee157e793ae7ef0a52093be4a3990c010c3ab44

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/70/head:pull/70`
`$ git checkout pull/70`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/aarch64/CallArranger.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/aarch64/CallArranger.java
@@ -203,9 +203,11 @@ public class CallArranger {
            return false;
 
         for (MemoryLayout elem : groupLayout.memberLayouts()) {
+            if (!(elem instanceof ValueLayout))
+                return false;
+
             ArgumentClassImpl argClass = AArch64ABI.argumentClassFor(SystemABI.Type.fromLayout(elem));
-            if (!(elem instanceof ValueLayout) ||
-                    elem.bitSize() != baseType.bitSize() ||
+            if (elem.bitSize() != baseType.bitSize() ||
                     elem.bitAlignment() != baseType.bitAlignment() ||
                     baseArgClass != argClass) {
                 return false;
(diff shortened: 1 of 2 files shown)


mcimadamore:
> Looks good to me. Thanks for fixing!

I wonder if we have same problem on Linux...

JornVernee:
> 
> 
> > Looks good to me. Thanks for fixing!
> 
> I wonder if we have same problem on Linux...

The tests are currently passing, and looking at the code there is one case where trying to look up the ABI type of padding could cause an exception, but an exception would be thrown for padding any ways:

```
private static List<ArgumentClassImpl> classifyValueType(ValueLayout type) {
    ArrayList<ArgumentClassImpl> classes = new ArrayList<>();
    ArgumentClassImpl clazz = SysVx64ABI.argumentClassFor(SystemABI.Type.fromLayout(type));
    if (clazz == null) {
        //padding not allowed here
        throw new IllegalStateException("Unexpected value layout: could not determine ABI class");
    }
...
```
So, I think we're good.

nick-arm:
I realised the was a mistake with the data types in the callarranger test and it didn't trigger the bug. I just pushed an extra commit to fix that. Not sure if the PR needs to be approved again?

PR #71, opened 2020-03-27
JDK-8241772: MemorySegment should offer a spliterator

mcimadamore:
This patch rethinks how shared memory segment work; up to now, our story for sharing a segment across multiple threads has been to support a ref-counting strategy (see MemorySegment::acquire). While this mechanism is very general, and can be used to support different idioms, it also significantly increases the complexity surface of the memory access API; also, the constrained nature of the acquire() mechanism is, in some occasions, still not general enough to support all the use cases - the cases which cannot workaround the confinement guarantees.

Because of this, I believe it is a better course of action to restrict the memory segment sharing capabilities to those use cases that are well-behaved enough: parallel processing of a segment via divide-and-conquer. This can be achieved rather succintly by having a segment offering a spliterator, which can then be used to create e.g. a parallel stream (see example in the test), or to create a fork/join recursive action (again, see test for some examples).

Of course we will need some kind of escape hatch for when a "true" shared segment is required - this could be provided by either offering a way to register a segment against a Cleaner, and giving up the deterministic deallocation guarantees in that case; the need for such GC-managed segment has already arisen when dealing with native interop use cases (e.g. a binding might need to create long-lived constants which are basically never closed).

Alternatively, the escape hatch could be an unsafe segment which is not thread confined, and can be closed from any thread (this is a solution that, I'm sure, some of the people I've chatted with will appreciate).

Either way, it is important to realize that the long-lived, shared segment is pretty far from the design center of the memory access API - whose goal is, after all, to improve over some of the things that direct buffers aren't so great at. If you are after a big, long-lived slab of shared memory, chances are that the ByteBuffer is already good enough at handling those.

Where the pressure comes from, is, we believe, in those allocation-intensive scenarios, where the cleaner mechanism (which is the only choice for direct buffers) is simply not enough to guarantee a reasonable throughput. These are also the cases that are crucial to achieve an efficient native interop support.

So, let's focus on what this API does best - which is managing confined memory segments - and let's build on that to cover the divide and conquer use cases (through the spliterator machinery).

Sidebar: for those of you worried that the lack of acquire() will make native interop harder, we are planning to _remove_ confinement restrictions from addresses obtained from native libraries - meaning that if you get a MemoryAddress from a native library, you will be able to share it with any thread you want, and such threads will be able to dereference the address w/o any need to workaround the confinement guarantees.

Moving forward, there are other things we need to do after this patch is done - as mentioned, providing a GC-backed escape hatch would probably be a good addition; but we also have to add more ways to share a segment with other threads in serial-confinement mode (e.g. one thread at a time); the main operation we're looking at is an 'handoff' operation which replaces the owner thread with a different one (useful in producer/consumer use cases) and, possibly, also a detach/attach pair of operations which can be used to temporarily remove ownership from a segment, and then have it picked up by a second thread - this is is effectively similar to handoff, but where the two threads don't know each other.

In addition to that, I'm planning to add a bunch of helper methods to SequenceLayout to help with reshaping and flattening sequence layouts; this operation is very useful when a client needs/want to define a splitearator which works on multiple elements at a time (see the ParallelSum benchmark for an example). I'll be filing a PR for this separately.

Cheers
Maurizio

---------
### Progress

### Issue
 * [JDK-8241772](https://bugs.openjdk.java.net/browse/JDK-8241772): MemorySegment should offer a spliterator

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/71/head:pull/71`
`$ git checkout pull/71`


mcimadamore:
Javadoc here
http://cr.openjdk.java.net/~mcimadamore/panama/8241772_javadoc/javadoc/jdk/incubator/foreign/package-summary.html

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -94,13 +94,18 @@
...
  * <p>
- * If a memory segment S owned by thread A needs to be used by thread B, B needs to explicitly <em>acquire</em> S,
- * which will create a so called <em>acquired</em> memory segment owned by B (see {@link #acquire()}) backed by the same resources
- * as S. A memory segment can be acquired multiple times by one or more threads; in that case, a memory segment S cannot
- * be closed until all the acquired memory segments derived from S have been closed. Furthermore, closing an acquired
- * memory segment does <em>not</em> trigger any deallocation action. It is therefore important that clients remember to
- * explicitly close the original segment from which the acquired memory segments have been obtained in order to truly
- * ensure that off-heap resources associated with the memory segment are released.
+ * In some cases, it might be useful for multiple threads to process the contents of the same memory segment concurrently
+ * (e.g. in the case of parallel processing); while memory segments provide string confinement guarantees, it is possible
```
PaulSandoz: s/string/strong

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -152,14 +157,27 @@
     MemoryAddress baseAddress();
 
     /**
-     * Obtains an <a href="#thread-confinement">acquired</a> memory segment which can be used to access memory associated
-     * with this segment from the current thread. As a side-effect, this segment cannot be closed until the acquired
-     * view has been closed too (see {@link #close()}).
-     * @return an <a href="#thread-confinement">acquired</a> memory segment which can be used to access memory associated
-     * with this segment from the current thread.
-     * @throws IllegalStateException if this segment has been closed.
+     * Returns a spliterator for this memory segment. The returned spliterator must feature the
```
PaulSandoz: Keeping with existing terminology:
"
The returned spliterator reports {@link Spliterator#SIZED}, {@link Spliterator#SUBSIZED}, {@link Spliterator#IMMUTABLE}, and {@link Spliterator#NONNULL}.
"

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -152,14 +157,27 @@
...
+     * {@link Spliterator#SIZED}, {@link Spliterator#SUBSIZED}, {@link Spliterator#IMMUTABLE} and {@link Spliterator#NONNULL}
+     * characteristics.
+     * <p>
+     * The returned spliterator should split the segment according to the specified sequence layout; that is,
```
PaulSandoz: "
The returned spliterator splits this segment ...
"

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -152,14 +157,27 @@
...
+     * if the supplied layout is a sequence layout whose element count is {@code N}, then calling {@link Spliterator#trySplit()}
+     * will result in a spliterator serving approximatively {@code N/2} elements (depending on whether N is even or not).
+     * As such, splitting is possible as long as {@code N >= 2}.
+     * <p>
+     * The returned spliterator effectively allows to slice a segment into disjoint sub-segments, which can then
+     * be processed in parallel by multiple threads (if the access mode {@link #ACQUIRE} is set).
+     * While closing the segment (see {@link #close()}) during pending concurrent execution will generally
+     * fail with an exception, it is possible to close a segment when a spliterator has been obtained but not thread
```
PaulSandoz: "but no thread"

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemorySegmentImpl.java
```diff
@@ -285,4 +308,88 @@ private int id() {
...
+        SegmentSplitter(long elementSize, long elemCount, MemorySegmentImpl segment) {
+            this.segment = segment;
+            this.elementSize = elementSize;
+            this.elemCount = elemCount;
+        }
+
+        @Override
+        public SegmentSplitter trySplit() {
+            if (currentIndex == 0 && elemCount > 1) {
+                MemorySegmentImpl old = segment;
```
PaulSandoz: s/old/parent

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemorySegmentImpl.java
```diff
@@ -285,4 +308,88 @@ private int id() {
...
+                long rem = elemCount % 2;
+                elemCount  = elemCount / 2;
+                long lobound = elemCount * elementSize;
+                long hibound = lobound + (rem * elementSize);
+                segment = old.asSliceNoCheck(0, lobound);
+                return new SegmentSplitter(elementSize, elemCount + rem, old.asSliceNoCheck(lobound, hibound));
```
PaulSandoz: We should return the left hand side, so it's a strict prefix.  We have an expected encounter order of the elements so we can also add ORDERED to the set of reported characteristics.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemorySegmentImpl.java
```diff
@@ -285,4 +308,88 @@ private int id() {
...
+            } else {
+                return false;
+            }
+        }
+
+        @Override
+        public void forEachRemaining(Consumer<? super MemorySegment> action) {
+            Objects.requireNonNull(action);
+            if (currentIndex < elemCount) {
+                MemorySegmentImpl acquired = segment.acquire();
```
PaulSandoz: I would ensure the spliterator state is consumed before performing the bulk loop (take local copies before updating and then looping).  Thereby if the consumer throws an exception it does not leave the spliterator in a partial state i.e. it is guaranteed to be fully traversed if subsequently operated on.  In the current implementation i believe an NPE will be thrown.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemorySegmentImpl.java
```diff
@@ -285,4 +308,88 @@ private int id() {
...
+        }
+
+        @Override
+        public long estimateSize() {
+            return elemCount;
+        }
+
+        @Override
+        public int characteristics() {
+            return NONNULL | SUBSIZED | SIZED | IMMUTABLE;
```
PaulSandoz: Add ORDERED

PR #75, opened 2020-03-30
JDK-8241866:Add API to reshape sequence layouts

mcimadamore:
https://docs.scipy.org/doc/numpy/reference/generated/numpy.reshape.html

As in Python, clients can use _at most_ one `-1` dimension value, and that dimension value will be inferred by remaining dimensions.

The changes also add the `flatten` method, which does the opposite - e.g. take a sequence layout, whose element layout could be other nested sequence layouts and flatten everything into an outermost sequence layout (where element count is adjusted as needed).

These routines are fairly general and helpful and I found myself in need for something similar when writing code doing parallel processing of memory segments, but there are many other possible use cases.

---------
### Progress

### Reviewers
 * Paul Sandoz ([psandoz](@PaulSandoz) - Committer) ⚠️ Review applies to 494bdd4e2fe2f4cada6ef7427479f8d96965c5fa

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/75/head:pull/75`
`$ git checkout pull/75`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/SequenceLayout.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/SequenceLayout.java
@@ -202,19 +202,25 @@ public final class SequenceLayout extends AbstractLayout {
      * }</pre>
      * @return a new sequence layout with the same size as this layout (but, possibly, with different
      * element count), whose element layout is not a sequence layout.
-     * @throws UnsupportedOperationException if this sequence layout does not have an element count.
+     * @throws UnsupportedOperationException if this sequence layout, or one of the nested sequence layouts being
+     * flattened, does not have an element count.
      */
     public SequenceLayout flatten() {
         if (!elementCount().isPresent()) {
-            throw new UnsupportedOperationException("Cannot flatten a sequence layout whose element count is unspecified");
+            throw badUnboundSequenceLayout();
         }
         long count = elementCount().getAsLong();
         MemoryLayout elemLayout = elementLayout();
         while (elemLayout instanceof SequenceLayout) {
-            count = count * ((SequenceLayout)elemLayout).elementCount().getAsLong();
-            elemLayout = ((SequenceLayout)elemLayout).elementLayout();
+            SequenceLayout elemSeq = (SequenceLayout)elemLayout;
+            count = count * elemSeq.elementCount().orElseThrow(this::badUnboundSequenceLayout);
+            elemLayout = elemSeq.elementLayout();
         }
         return MemoryLayout.ofSequence(count, elemLayout);
+    }
+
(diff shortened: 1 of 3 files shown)


src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/SequenceLayout.java
```diff
@@ -107,6 +108,115 @@ public SequenceLayout withElementCount(long elementCount) {
...
+     * @throws UnsupportedOperationException if this sequence layout does not have an element count.
+     */
+    public SequenceLayout flatten() {
+        if (!elementCount().isPresent()) {
+            throw new UnsupportedOperationException("Cannot flatten a sequence layout whose element count is unspecified");
+        }
+        long count = elementCount().getAsLong();
+        MemoryLayout elemLayout = elementLayout();
+        while (elemLayout instanceof SequenceLayout) {
+            count = count * ((SequenceLayout)elemLayout).elementCount().getAsLong();
```
slowhog: Should we throw same UOE here if no elementCount() or just let the NoSuchElementException pass through?

test/jdk/java/foreign/TestReshape.java
```diff
@@ -0,0 +1,194 @@
+/*
+ * Copyright (c) 2020, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
```
JornVernee: Wrong copyright header

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/SequenceLayout.java
```diff
@@ -107,6 +108,121 @@ public SequenceLayout withElementCount(long elementCount) {
...
+     * If one of the provided element count is the special value {@code -1}, then the element
+     * count in that position will be inferred from the remaining element counts and the
+     * element count of the flattened projection of this layout. For instance, a layout equivalent to
+     * the above {@code reshapeSeq} can also be computed in the following ways:
+     * <pre>{@code
+    var reshapeSeqImplicit1 = seq.reshape(-1, 6);
+    var reshapeSeqImplicit2 = seq.reshape(2, -1);
+     * }</pre>
+     * @param elementCounts an array of element counts, of which at most one can be {@code -1}.
+     * @return a new sequence layout where element element layouts in the flattened projection of this
```
PaulSandoz: s/element element/element

PR #77, opened 2020-03-31
Backport misc changes from foreign-abi

JornVernee:
Hi,

This patch backports several changes that went into the foreign-abi branch, but which are in retrospect useful to have on the memaccess branch, and as part of the memory access API as well.

These changes include:
- Replacing ForeignUnsafe with Foreign
- Adding the `ofNativeUnchecked` method to create unchecked MemorySegments, MemoryAddress::ofLong to forge MemoryAddresses from longs, as well as the Nothing segment and rebase support. Mainly to support going from a `long` to `MemoryAddress`, and also to `MemorySegment` (MA -> long is already covered)
- The memory layout attributes exposed in the public API (could be interesting to investigate attaching carrier types directly to layouts).

The MemoryAddress carrier support for VarHandles is left out for now, since the current thinking is that it will be dropped and replaced with VarHandle combinators to convert between long and MA instead.

Thanks,
Jorn

---------
### Progress

### Reviewers
 * Paul Sandoz ([psandoz](@PaulSandoz) - Committer) ⚠️ Review applies to c6e250880660bf801475fcda610734968293116e

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/77/head:pull/77`
`$ git checkout pull/77`


src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryAddress.java
```diff
@@ -116,4 +126,22 @@ static void copy(MemoryAddress src, MemoryAddress dst, long bytes) {
         MemoryAddressImpl.copy((MemoryAddressImpl)src, (MemoryAddressImpl)dst, bytes);
     }
 
+    /**
+     * A native memory address instance modelling the {@code NULL} pointer. This address is backed by a memory segment
+     * which can be neither closed, nor dereferenced.
+     */
+    MemoryAddress NULL = MemorySegmentImpl.NOTHING.baseAddress();
```
mcimadamore: I'd leave this out for now? This seems very C oriented.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryAddressImpl.java
```diff
@@ -119,4 +128,22 @@ public boolean equals(Object that) {
...
+
+    public static long addressof(MemoryAddress address) {
+        MemoryAddressImpl addressImpl = (MemoryAddressImpl)address;
+        if (addressImpl.unsafeGetBase() != null) {
+            throw new IllegalStateException("Heap address!");
+        }
+        return addressImpl.unsafeGetOffset();
+    }
+
+    public static MemoryAddress ofLongUnchecked(long value) {
```
mcimadamore: I believe these two are probably unused?

JornVernee: You're right, thanks for catching!

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryAddress.java
JornVernee: OK - My initial thought was that NULL is pretty universal, and that having a singleton NULL might be handy. But I guess there's no real need to use such a value.

PR #84, opened 2020-04-02
JDK-8242011: Add support for memory address combinator

mcimadamore:
Following recent changes (see https://github.com/openjdk/panama-foreign/pull/77)which backport some of the goodies in foreign-abi into foreign-memaccess,
this patch brings support for a VarHandle combinator which turns a regular memory access var handle into a var handle which gets/sets a MemoryAddress (e.g. instead of a long).

This patch also addresses the general problem of the co-existence between combinators in MemoryHandles and the general var handle combinators in MethodHandles.

This coexistence has always been tricky, because the combinators in MemoryHandles like to create a new 'flattened' memory access var handle, which provides best possible performances, and also performs certain alignment checks upfront.

However, if a memory access handle is adapted (e.g. the carrier type is changed) this simplistic approach no longer works, as, by reconstructing the memory access var handle from scratch we will also lose all the adaptations.

The solution is to detect as to wheter the target handle is a "direct" memory access var handle or not (special provisions are made for the ubiquitous adaptation added on all memory access var handle creation as a workaround for JDK-8237349). If that's the case, and, if the stride or the offset matches the alignment constraint, then we go the fast/flattened path and we can adapt by reconstructing the handle from scratch. If any of these two conditions are not met (there's complex adaption that would be dropped on the floor, or the offset/stride parameter doesn't match with alignment constraint), then the slow path is taken - the target var handle is kept around and is adapted using the standard combinator API. This leads to a less performant VarHandle (because, unfortunately, calling addOffset() currently breaks C2 optimizations), but also guarantees that alignment constraints will be checked in full (this is because the memory access var handle implementation always checks the alignment of the base address passed to it - only for the offset part is the alignment check skipped - on the basis that this has already been verified by construction).

The result is that we can lift a lot of the restrictions surrounding the combinators in MemoryHandles; such combinators can now work on _any_ var handle (provided the var handle has a first coordinate type of type MemoryAddress). Also, I've lifted the alignment restrictions, since these will either be enforced dynamically (by taking the slow path), or they won't be enforced because we have already statically proven that the constraints are satisfied (fast path).

I've also took the chance to rename some of the classes surrounding memory access var handles to use the "memory access var handle" terminology, which is the one that stuck (currently, some classes use the word "address var hande" which is ambiguous).

---------
### Progress

### Issue
 * [JDK-8242011](https://bugs.openjdk.java.net/browse/JDK-8242011): Add support for memory address combinator

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/84/head:pull/84`
`$ git checkout pull/84`

### DIFF
--- a/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
+++ b/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
@@ -5256,9 +5256,8 @@ System.out.println((int) f0.invokeExact("x", "y")); // 2
         }
 
         List<Class<?>> newCoordinates = new ArrayList<>(targetCoordinates);
-        for (Class<?> arg : filter.type().parameterList()) {
-            newCoordinates.add(pos, arg);
-        }
+        newCoordinates.remove(pos);
+        newCoordinates.addAll(pos, filter.type().parameterList());
 
         return new IndirectVarHandle(target, target.varType(), newCoordinates.toArray(new Class<?>[0]),
                 (mode, modeHandle) -> MethodHandles.collectArguments(modeHandle, 1 + pos, filter));
--- a/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
+++ b/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
@@ -5041,7 +5041,6 @@ System.out.println((int) f0.invokeExact("x", "y")); // 2
     public static VarHandle filterCoordinates(VarHandle target, int pos, MethodHandle... filters) {
         Objects.nonNull(target);
         Objects.nonNull(filters);
-        if (filters.length == 0) return target;
 
         List<Class<?>> targetCoordinates = target.coordinateTypes();
         if (pos < 0 || pos >= targetCoordinates.size()) {
@@ -5049,6 +5048,8 @@ System.out.println((int) f0.invokeExact("x", "y")); // 2
         } else if (pos + filters.length > targetCoordinates.size()) {
             throw new IllegalArgumentException("Too many filters");
         }
+
+        if (filters.length == 0) return target;
 
         List<Class<?>> newCoordinates = new ArrayList<>(targetCoordinates);
         for (int i = 0 ; i < filters.length ; i++) {
@@ -5095,7 +5096,6 @@ System.out.println((int) f0.invokeExact("x", "y")); // 2
     public static VarHandle insertCoordinates(VarHandle target, int pos, Object... values) {
         Objects.nonNull(target);
         Objects.nonNull(values);
-        if (values.length == 0) return target;
 
         List<Class<?>> targetCoordinates = target.coordinateTypes();
         if (pos < 0 || pos >= targetCoordinates.size()) {
@@ -5103,6 +5103,8 @@ System.out.println((int) f0.invokeExact("x", "y")); // 2
         } else if (pos + values.length > targetCoordinates.size()) {
             throw new IllegalArgumentException("Too many values");
         }
+
+        if (values.length == 0) return target;
 
         List<Class<?>> newCoordinates = new ArrayList<>(targetCoordinates);
         for (int i = 0 ; i < values.length ; i++) {
@@ -5261,6 +5263,44 @@ System.out.println((int) f0.invokeExact("x", "y")); // 2
 
         return new IndirectVarHandle(target, target.varType(), newCoordinates.toArray(new Class<?>[0]),
                 (mode, modeHandle) -> MethodHandles.collectArguments(modeHandle, 1 + pos, filter));
+    }
+
+    /**
+     * Returns a var handle which will discard some dummy coordinates before delegating to the
+     * target var handle. As a consequence, the resulting var handle will feature more
+     * coordinate types than the target var handle.
+     * <p>
+     * The {@code pos} argument may range between zero and <i>N</i>, where <i>N</i> is the arity of the
+     * target var handle's coordinate types. If {@code pos} is zero, the dummy coordinates will precede
+     * the target's real arguments; if {@code pos} is <i>N</i> they will come after.
+     * <p>
+     * The resulting var handle will feature the same access modes (see {@link java.lang.invoke.VarHandle.AccessMode} and
+     * atomic access guarantees as those featured by the target var handle.
+     *
+     * @param target the var handle to invoke after the dummy coordinates are dropped
+     * @param pos position of first coordinate to drop (zero for the leftmost)
+     * @param valueTypes the type(s) of the coordinate(s) to drop
+     * @return an adapter var handle which drops some dummy coordinates,
+     *         before calling the target var handle
+     * @throws NullPointerException if either {@code target}, {@code valueTypes} are {@code == null}.
+     * @throws IllegalArgumentException if {@code pos} is not between 0 and the target var handle coordinate arity, inclusive.
+     */
+    public static VarHandle dropCoordinates(VarHandle target, int pos, Class<?>... valueTypes) {
+        Objects.nonNull(target);
+        Objects.nonNull(valueTypes);
+
+        List<Class<?>> targetCoordinates = target.coordinateTypes();
+        if (pos < 0 || pos > targetCoordinates.size()) {
+            throw newIllegalArgumentException("Invalid position " + pos + " for coordinate types", targetCoordinates);
+        }
+
+        if (valueTypes.length == 0) return target;
+
+        List<Class<?>> newCoordinates = new ArrayList<>(targetCoordinates);
+        newCoordinates.addAll(pos, List.of(valueTypes));
+
+        return new IndirectVarHandle(target, target.varType(), newCoordinates.toArray(new Class<?>[0]),
+                (mode, modeHandle) -> MethodHandles.dropArguments(modeHandle, 1 + pos, valueTypes));
     }
 
     private static void noCheckedExceptions(MethodHandle handle) {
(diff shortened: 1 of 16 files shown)


src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -5256,9 +5256,8 @@ public static VarHandle collectCoordinates(VarHandle target, int pos, MethodHand
         }
 
         List<Class<?>> newCoordinates = new ArrayList<>(targetCoordinates);
```
mcimadamore, line 5258: This was a bug I discovered in the collectCoordinates implementation - the old coordinate was not removed - leading to WrongMethodTypeException.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryAddress.java
```diff
@@ -126,14 +126,22 @@ static void copy(MemoryAddress src, MemoryAddress dst, long bytes) {
         MemoryAddressImpl.copy((MemoryAddressImpl)src, (MemoryAddressImpl)dst, bytes);
     }
 
+    /**
+     * A native memory address instance modelling the {@code NULL} pointer. This address is backed by a memory segment
+     * which can be neither closed, nor dereferenced.
+     */
+    MemoryAddress NULL = MemorySegmentImpl.NOTHING.baseAddress();
```
mcimadamore, line 133: In the end, this was added back :-)

src/java.base/share/classes/jdk/internal/access/JavaLangInvokeAccess.java
```diff
@@ -113,8 +113,14 @@
...
      */
-    VarHandle memoryAddressViewVarHandle(Class<?> carrier, long alignmentMask,
-                                         ByteOrder order, long offset, long[] strides);
+    VarHandle memoryAccessVarHandle(Class<?> carrier, long alignmentMask,
+                                    ByteOrder order, long offset, long[] strides);
+
+    /**
+     * Is {@code handle} a memory access varhandle?
+     * Used by {@code jdk.incubator.foreign.MemoryHandles}.
+     */
```
JornVernee, line 122: Does this need `@param` and `@return` tags?

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryAddress.java
JornVernee: :)

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryHandles.java
```diff
@@ -174,85 +199,128 @@ public static VarHandle varHandle(Class<?> carrier, long alignmentBytes, ByteOrd
...
+     * That is, when the returned var handle receives a memory address coordinate pointing at a memory location at
+     * offset <em>O</em>, a new memory address coordinate pointing at a memory location at offset <em>(S * X) + O</em>
+     * is created, and then passed to the target var handle,
+     * where <em>S</em> is a constant <em>stride</em>, whereas <em>X</em> is a dynamic value that will be
+     * provided as an additional access coordinate (of type {@code long}).
      *
-     * @apiNote the resulting var handle features certain <a href="#memaccess-mode">access mode restrictions</a>,
-     * which are common to all memory access var handles.
+     * The returned var handle will feature the same type as the target var handle; an additional access coordinate
+     * of type {@code long} will be <em>prepended</em> to the access coordinate types of the target var handle.
```
JornVernee, line 253: Doesn't the MemoryAddress also counts as an access coordinate?

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryHandles.java
```diff
@@ -174,85 +199,128 @@ public static VarHandle varHandle(Class<?> carrier, long alignmentBytes, ByteOrd
...
      * @param bytesStride the stride, in bytes, by which to multiply the coordinate value. Must be greater than zero.
-     * @return the new memory access var handle.
-     * @throws IllegalArgumentException when the target var handle is not a memory access var handle,
-     * or if {@code bytesStride <= 0}, or otherwise incompatible with the alignment constraint.
+     * @return the adapted var handle.
+     * @throws IllegalArgumentException if {@code bytesStride <= 0}, or if the first access coordinate type
+     * is not of type {@link MemoryAddress}.
      */
     public static VarHandle withStride(VarHandle target, long bytesStride) {
         if (bytesStride == 0) {
```
JornVernee, line 262: Pre-existing, but it seems that this should check for `byteStride <= 0` instead?

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryHandles.java
```diff
@@ -265,4 +333,20 @@ private static long carrierSize(Class<?> carrier) {
         long bitsAlignment = Math.max(8, Wrapper.forPrimitiveType(carrier).bitWidth());
         return Utils.bitsToBytesOrThrow(bitsAlignment, IllegalStateException::new);
     }
+
+    private static MemoryAddress longToAddress(long value) {
+        return MemoryAddress.ofLong(value);
+    }
+
+    private static long addressToLong(MemoryAddress value) {
+        return ((MemoryAddressImpl)value).unsafeGetOffset();
```
JornVernee, line 342: Maybe this should use MemoryAddressImpl::addressof, since that also checks for heap addresses. I don't think we want to allow serializing heap addresses?

test/jdk/java/foreign/TestAddressHandle.java
```diff
@@ -0,0 +1,111 @@
...
+    public void testBadAdaptFloat() {
+        VarHandle floatHandle = MemoryHandles.varHandle(float.class, ByteOrder.nativeOrder());
+        MemoryHandles.asAddressVarHandle(floatHandle);
+    }
+
+    @Test(expectedExceptions = IllegalArgumentException.class)
+    public void testBadAdaptDouble() {
+        VarHandle doubleHandle = MemoryHandles.varHandle(double.class, ByteOrder.nativeOrder());
+        MemoryHandles.asAddressVarHandle(doubleHandle);
+    }
```
JornVernee, line 80: The adapter code also checks for `!carrier.isPrimitive()` that's also a case that could be tested.

test/jdk/java/foreign/TestAddressHandle.java
```diff
@@ -0,0 +1,111 @@
...
+
+                // char
+                { MemoryHandles.asAddressVarHandle(MemoryHandles.varHandle(char.class, ByteOrder.nativeOrder())) },
+                { MemoryHandles.asAddressVarHandle(MemoryHandles.withOffset(MemoryHandles.varHandle(char.class, ByteOrder.nativeOrder()), 0)) },
+                { MemoryHandles.asAddressVarHandle(MemoryLayouts.JAVA_CHAR.varHandle(char.class)) },
+
+                // byte
+                { MemoryHandles.asAddressVarHandle(MemoryHandles.varHandle(byte.class, ByteOrder.nativeOrder())) },
+                { MemoryHandles.asAddressVarHandle(MemoryHandles.withOffset(MemoryHandles.varHandle(byte.class, ByteOrder.nativeOrder()), 0)) },
+                { MemoryHandles.asAddressVarHandle(MemoryLayouts.JAVA_BYTE.varHandle(byte.class)) }
```
JornVernee, line 108: `boolean` would also be an interesting case to test. explicitCastArguments does have handling to convert between boolean and long as well. Maybe that's a carrier you also want to explicitly disallow in the adapter?

src/java.base/share/classes/jdk/internal/access/JavaLangInvokeAccess.java
mcimadamore: I went with the style of the other comments surrounding this. This is internal API, seems like the main need (as arised during RFR) is to indicate where a routine is used :-)

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryHandles.java
mcimadamore: good point, will fix

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryHandles.java
mcimadamore: also good point, will fix and add a test

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryHandles.java
mcimadamore: yeah - that's what it should do

test/jdk/java/foreign/TestAddressHandle.java
mcimadamore: I'll add something

test/jdk/java/foreign/TestAddressHandle.java
mcimadamore: Good point, I'll go for disallow, since there's only two pointers you can express with that :-)

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryHandles.java
mcimadamore: After offline discussion I've removed these restrictions. Now all offset and stride values are legal - but if you use 0 some very minimal adaptation (or none at all) will take place. "Be liberal in what you accept".

PR #107, opened 2020-04-08
Add JMH test for loop over sliced segment

mcimadamore:
This simple patch adds a simple benchmark that does a loop on a segment slice. I found this test useful when investigating escape analysis issues of derived segments, so I'm adding it here. With the current implementation (MemorySegmentImpl) the difference between the new `segment_loop` and `segment_loop_slice` are negligible, so this can be used to spot regressions in that logic.

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/107/head:pull/107`
`$ git checkout pull/107`

### DIFF
--- a/test/micro/org/openjdk/bench/jdk/incubator/foreign/LoopOverNonConstant.java
+++ b/test/micro/org/openjdk/bench/jdk/incubator/foreign/LoopOverNonConstant.java
@@ -137,9 +137,9 @@ public class LoopOverNonConstant {
     }
 
     @Benchmark
-    public int segment_loop_slice() {
+    public int segment_loop_readonly() {
         int sum = 0;
-        MemoryAddress base = segment.asSlice(0, segment.byteSize()).baseAddress();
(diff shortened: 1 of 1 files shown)


test/micro/org/openjdk/bench/jdk/incubator/foreign/LoopOverNonConstant.java
```diff
@@ -126,6 +126,26 @@ public int segment_loop() {
...
+        int sum = 0;
+        MemoryAddress base = segment.asSlice(0, segment.byteSize()).baseAddress();
+        for (int i = 0; i < ELEM_SIZE; i++) {
+            sum += (int) VH_int.get(base, (long) i);
+        }
+        return sum;
+    }
+
+    @Benchmark
+    public int segment_loop_slice() {
```
JornVernee, line 140: Looks like the method is accidentally defined twice.

PR #109, opened 2020-04-10
JDK-8242495: Restructure implementation of memory segments

mcimadamore:
Currently memory segments use a monomorphic implementation. This implementation doesn't differentiate between heap segments and native segments and, as a result can lead to type profile pollution.

Moreover, for native segment, the accessor to the unsafe base object is redundant (e.g. base is always null), but we can't easily shortcircuit that if we use a monomorphic implementation.

Finally, it would be desirable, at some point, to add public sub-interfaces of MemorySegment, such as MappedMemorySegment which could provide a force() method. Unfortunately it would be cumbersome to do that with an underlying monomorphic implementation, unless we make force() total on all segments (this work will likely be addressed in a separate, follow-up PR).

This patch introduces a new abstract class which contain most of the memory segment implementation (it is essentially the same as the old MemorySegmentImpl).

Three subclasses are also added:

* NativeMemorySegmentImpl - to model native segments (base() returns null)
* HeapMemorySegmentImpl - to model heap segments (base() returns some array)
* MappedMemorySegmentImpl - to model memory mapped segments (this extends from NativeMemorySegment)

Each subclass overrides a dup() method, which must be used by the abstract subclass to implement view methods in a generic way. Each subclass also defines what base() and min() look like.

Since now we have different segment implementation classes, I also decided to move all the factory methods from Utils to the segment impl class they belong, which leads to cleaner code.

Note that, for heap segment there is a subtle trick: instead of storing the `base` object (an array) we store a *lambda* which retrieves the base object. This allows us to retain a monomorphic heap segment implementation, but, thanks to lambdas, we get sharply typed accessors to the base object, which helps out C2 when determining whether memory barriers should be added or not. This approach strikes a good balance between performance and code readability. Should we find cases where this approach is not enough, we could always fall back to create separate subclasses for each array type.

I've added benchmarks for both heap and mapped segments (they borrow heavily from the current LoopOverNonConstant), so that at least we can check performances in all cases. Numbers look good and similar across the whole spectrum of segments. As a comparison, before this patch, heap segments were roughly 10x slower than off-heap segments.

This patch also tidies up the byte buffer support; that is, it make sure that when going from a buffer to a segment (and back) all the correct information is preserved; while, before this patch, this was the case for native vs. heap segments, information about mapped segments was lost (as a "normal" direct buffer was created out of a mapped segment). This was also true when creating a segment from a mapped buffer - where a regular native segment was created. While semantics of access operations was not impacted by the aforementioned problems, calling force() on a buffer derived from a mapped segment might not behave as expected; this patch fixes that. To make sure that information is correctly preserved, I have enhanced the byte buffer test.

---------
### Progress

### Issue
 * [JDK-8242495](https://bugs.openjdk.java.net/browse/JDK-8242495): Restructure implementation of memory segments

### Reviewers
 * Paul Sandoz ([psandoz](@PaulSandoz) - Committer) ⚠️ Review applies to 505dab16e6005ae9ddd0851f6cec02aa124a1217
 * Jorn Vernee ([jvernee](@JornVernee) - Committer) ⚠️ Review applies to 505dab16e6005ae9ddd0851f6cec02aa124a1217

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/109/head:pull/109`
`$ git checkout pull/109`


mcimadamore:
> There is a trade-off being monomorphic in type or monomorphic in unsafe access. One problem with `Buffer` was all the different types would cause profile pollution if many were intermixed in the same code (e.g. on- and off-heap etc).
> 
> There might be a similar issue here, although given the primary target is off-heap there is less risk and i believe the intermixing of native and mapped should not be an issue. Still, it might be worth checking how intermixing on- and off-heap behaves.

I believe Vlad did some tests in this area and found similar issues.

test/micro/org/openjdk/bench/jdk/incubator/foreign/ParallelSum.java
```diff
@@ -186,31 +186,50 @@ protected Integer compute() {
         }
     }
 
-    static class SumSegment extends RecursiveTask<Integer> {
+    static class SumSegment extends CountedCompleter<Integer> {
```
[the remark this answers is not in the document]
PaulSandoz: Oh, I missed that, yes that's better.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/HeapMemorySegmentImpl.java
```diff
@@ -0,0 +1,127 @@
...
+public class HeapMemorySegmentImpl<H> extends AbstractMemorySegmentImpl {
+
+    private static final Unsafe UNSAFE = Unsafe.getUnsafe();
+    private static final int BYTE_ARR_BASE = UNSAFE.arrayBaseOffset(byte[].class);
+
+    final long offset;
+    final Supplier<H> baseProvider;
+
+    @ForceInline
+    HeapMemorySegmentImpl(long offset, Supplier<H> baseProvider, long length, int mask, Thread owner, MemoryScope scope) {
```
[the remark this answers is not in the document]
mcimadamore: Just tried this - makes benchmarks revert to old behavior - roughly 10x slower. Worth a try, but I think using reflection probably makes thing less transparent to the JIT.

test/jdk/java/foreign/TestByteBuffer.java
```diff
@@ -569,4 +599,21 @@ static Object defaultValue(Class<?> c) {
...
+    public static Object[][] bufferSources() {
+        Predicate<MemorySegment> heapTest = segment -> segment instanceof HeapMemorySegment;
+        Predicate<MemorySegment> nativeTest = segment -> segment instanceof NativeMemorySegment;
+        Predicate<MemorySegment> mappedTest = segment -> segment instanceof MappedMemorySegment;
+        try (FileChannel channel = FileChannel.open(tempPath)) {
+            return new Object[][]{
+                    { ByteBuffer.wrap(new byte[256]), heapTest },
+                    { ByteBuffer.allocate(256), heapTest },
+                    { ByteBuffer.allocateDirect(256), nativeTest },
+                    { channel.map(FileChannel.MapMode.READ_ONLY, 0L, 256), mappedTest }
```
slowhog, line 616: Isn't the file channel closed after return? I am suppose this is OK as we don't access it and probably intentional here.

What I am curious about, is what suppose to happen to the MS after this channel is closed? Since this is from MappedByteBuffer, I suppose it inherits the unspecified behavior.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/HeapMemorySegmentImpl.java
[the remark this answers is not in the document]
PaulSandoz: Interesting, I thought it might be the explicit casts introduced in the MH machinery of lambdas, and since everything is final and stable (?) the casts usually help the JIT, if so then there is clearly more going on with the MH machinery.

test/jdk/java/foreign/TestByteBuffer.java
mcimadamore: My understanding is that lifecycle of a mapped buffer is not tied to that of the channel where it came from. The API for `FileChannel::map` says:

```
 A mapping, once established, is not dependent upon the file channel that was used to create it. Closing the channel, in particular, has no effect upon the validity of the mapping. 
```

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/HeapMemorySegmentImpl.java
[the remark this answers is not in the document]
mcimadamore: that's true, although I'd expect that with MH the Class constant is stored in some separate spinned class - e.g. it is a true constant, not a final field load?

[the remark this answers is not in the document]
JornVernee: Maybe the speedup comes from each lambda having a separate type (a class is spun for each), so C2 can apply type profiling & guarded inlining, and within the inlined section a sharper array type can be assumed.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/InternalForeign.java
```diff
@@ -49,18 +49,18 @@ public static InternalForeign getInstancePrivileged() {
 
     @Override
     public MemoryAddress withSize(MemoryAddress base, long byteSize) throws IllegalAccessError {
-        return Utils.makeNativeSegmentUnchecked(base.toRawLongValue(), byteSize, null, false)
+        return NativeMemorySegmentImpl.makeNativeSegmentUnchecked(base.toRawLongValue(), byteSize, null, false)
```
JornVernee, line 52: Pre-existing; should this check that `base` has the Nothing segment as segment? (otherwise access modes are dropped for instance).

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
```diff
@@ -406,4 +377,52 @@ public int characteristics() {
...
+        Object base = nioAccess.getBufferBase(bb);
+        UnmapperProxy unmapper = nioAccess.unmapper(bb);
+
+        int pos = bb.position();
+        int limit = bb.limit();
+
+        MemoryScope bufferScope = new MemoryScope(bb, null);
+        int size = limit - pos;
+        if (base != null) {
+            return new HeapMemorySegmentImpl<>(bbAddress + pos, () -> (byte[])base, size, defaultAccessModes(size), Thread.currentThread(), bufferScope);
```
JornVernee, line 399: Seems to be pre-existing, but using the default access modes here would allow dropping access modes from a segment by going to ByteBuffer and then back to MemorySegment. I think at least the read-only property of the BB should be respected here, but maybe we can just retrieve the originally captured MemorySegmentProxy and use the access modes from that?

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/NativeMemorySegmentImpl.java
```diff
@@ -0,0 +1,113 @@
+/*
+ *  Copyright (c) 2019, Oracle and/or its affiliates. All rights reserved.
```
JornVernee, line 2: Update copyright year?
```suggestion
 *  Copyright (c) 2020, Oracle and/or its affiliates. All rights reserved.
```

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/HeapMemorySegmentImpl.java
```diff
@@ -0,0 +1,127 @@
+/*
+ *  Copyright (c) 2019, Oracle and/or its affiliates. All rights reserved.
```
JornVernee, line 2: Update copyright year. 
```suggestion
 *  Copyright (c) 2020, Oracle and/or its affiliates. All rights reserved.
```

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
```diff
@@ -44,68 +44,62 @@
...
 
-    private MemorySegmentImpl() {
-        this.length = 0L;
-        this.mask = 0;
-        this.min = 0L;
-        this.base = null;
-        this.owner = null;
-        this.scope = MemoryScope.GLOBAL;
-    }
+    static JavaNioAccess nioAccess = SharedSecrets.getJavaNioAccess();
```
JornVernee, line 66: Could be `final`
```suggestion
    final static JavaNioAccess nioAccess = SharedSecrets.getJavaNioAccess();
```

PR #111, opened 2020-04-15
Memory access implementation rewrite - post cleanup

mcimadamore:
* Fix access modes not preserved in AbstractMemorySegmentImpl.ofBuffer
* Add check to ensure that Foreign::withSize and Foreign::asMallocSegment are called with unchecked address

---------
### Progress

### Reviewers
 * Jorn Vernee ([jvernee](@JornVernee) - Committer) ⚠️ Review applies to c7e953b299b2e65c80b9fce3f1d9130809d96505

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/111/head:pull/111`
`$ git checkout pull/111`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
@@ -399,15 +399,17 @@ public abstract class AbstractMemorySegmentImpl implements MemorySegment, Memory
         MemorySegment bufferSegment = (MemorySegment)nioAccess.bufferSegment(bb);
         int modes = bufferSegment == null ?
                 defaultAccessModes(size) : bufferSegment.accessModes();
+        Thread owner = bufferSegment == null ?
+                Thread.currentThread() : bufferSegment.ownerThread();
         if (bb.isReadOnly()) {
             modes &= ~WRITE;
         }
         if (base != null) {
-            return new HeapMemorySegmentImpl<>(bbAddress + pos, () -> (byte[])base, size, modes, Thread.currentThread(), bufferScope);
+            return new HeapMemorySegmentImpl<>(bbAddress + pos, () -> (byte[])base, size, modes, owner, bufferScope);
         } else if (unmapper == null) {
-            return new NativeMemorySegmentImpl(bbAddress + pos, size, modes, Thread.currentThread(), bufferScope);
+            return new NativeMemorySegmentImpl(bbAddress + pos, size, modes, owner, bufferScope);
         } else {
-            return new MappedMemorySegmentImpl(bbAddress + pos, unmapper, size, modes, Thread.currentThread(), bufferScope);
+            return new MappedMemorySegmentImpl(bbAddress + pos, unmapper, size, modes, owner, bufferScope);
         }
     }
 
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
@@ -395,12 +395,19 @@ public abstract class AbstractMemorySegmentImpl implements MemorySegment, Memory
 
         MemoryScope bufferScope = new MemoryScope(bb, null);
         int size = limit - pos;
+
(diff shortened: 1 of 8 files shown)


test/jdk/java/foreign/TestByteBuffer.java
```diff
@@ -608,12 +610,18 @@ static Object defaultValue(Class<?> c) {
...
                     { ByteBuffer.allocate(256), heapTest },
                     { ByteBuffer.allocateDirect(256), nativeTest },
-                    { channel.map(FileChannel.MapMode.READ_ONLY, 0L, 256), mappedTest }
+                    { channel.map(FileChannel.MapMode.READ_WRITE, 0L, 256), mappedTest },
+
+                    { ByteBuffer.wrap(new byte[256]).asReadOnlyBuffer(), heapTest },
+                    { ByteBuffer.allocate(256).asReadOnlyBuffer(), heapTest },
+                    { ByteBuffer.allocateDirect(256).asReadOnlyBuffer(), nativeTest },
+                    { channel.map(FileChannel.MapMode.READ_WRITE, 0L, 256).asReadOnlyBuffer(),
+                            nativeTest /* this seems to be an existing bug in the BB implementation */ }
```
mcimadamore, line 624: Note: this test uncovered an issue with mapped buffers: any projections of a mapped buffer (slices, read only) will create a direct buffer which will have lost the original file descriptor. As such, the behavior of the projection will not be one of a mapped buffer (e.g. calling force() will do nothing). As such, our implementation will simply return a native segment in such cases. I'll discuss this issue with somebody from core-libs.

mcimadamore: Apparently, a very old issue :-)
https://bugs.openjdk.java.net/browse/JDK-4833719

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
```diff
@@ -395,12 +395,19 @@ public static AbstractMemorySegmentImpl ofBuffer(ByteBuffer bb) {
...
+                defaultAccessModes(size) : bufferSegment.accessModes();
+        if (bb.isReadOnly()) {
+            modes &= ~WRITE;
+        }
         if (base != null) {
-            return new HeapMemorySegmentImpl<>(bbAddress + pos, () -> (byte[])base, size, defaultAccessModes(size), Thread.currentThread(), bufferScope);
+            return new HeapMemorySegmentImpl<>(bbAddress + pos, () -> (byte[])base, size, modes, Thread.currentThread(), bufferScope);
         } else if (unmapper == null) {
-            return new NativeMemorySegmentImpl(bbAddress + pos, size, defaultAccessModes(size), Thread.currentThread(), bufferScope);
+            return new NativeMemorySegmentImpl(bbAddress + pos, size, modes, Thread.currentThread(), bufferScope);
```
PaulSandoz, line 408: I think this and prior code can result in an integrity issue with segment -> buffer -> segment. One can close the original segment while retaining access on another thread to the segment obtained from the buffer.

Consider the following example:
```
        CountDownLatch a = new CountDownLatch(1);
        CountDownLatch b = new CountDownLatch(1);
        CompletableFuture<?> r;
        try (MemorySegment s1 = MemorySegment.allocateNative(intArrayLayout)) {
            r = CompletableFuture.runAsync(() -> {
                try {
                    ByteBuffer bb = s1.asByteBuffer();

                    MemorySegment s2 = MemorySegment.ofByteBuffer(bb);
                    a.countDown();

                    try {
                        b.await();
                    } catch (InterruptedException e) {
                    }

                    MemoryAddress base = s2.baseAddress();
                    intElemHandle.set(base, 1L, -42);
                } catch (RuntimeException e) {
                    e.printStackTrace();
                    throw e;
                }
            });

            a.await();
            MemoryAddress base = s1.baseAddress();
            intElemHandle.set(base, 1L, 42);
        }

        b.countDown();
        r.get();
```

JornVernee: Yeah. Instead of always using Thread.currentThread(), I think we can use the owner thread of `bufferSegment` if it's not null. That makes the above example fail.

mcimadamore: Good point, I will fix this. It's interesting how all these issues were latent in the code and having a more direct implementation for the various segments made them pop out more :-)

PR #113, opened 2020-04-16
Preserve memory scope for buffer segments

mcimadamore:
Not 100% sure about this, but wanted to share this PR for evaluation. Basically, I've realized that we are also not preserving the memory scope of the original segment when we do a segment -> buffer -> segment transition.

This means that, e.g. code like this:

```
MemorySegment segment = MemorySegment.allocateNative(16);
segment = MemorySegment.ofByteBuffer(segment.asByteBuffer());
segment.close(); // no cleanup of native memory
```

will not really 'close' the original segment. Similarly, if the original segment is being worked upon by some other thread (through a spliterator) it is possible for the morphed segment to lose this info, and to allow a close - if the scope is created afresh (as is now).

This patch fixes this, but I'm torn as to whether it makes things better; preserving access modes and thread ownership is mostly about not allowing clients to use the byte buffer escape hatch as a way to defy the segment restrictions. I don't think preserving scope falls into the same bucket - e.g. this is a separate question as to whether a segment S2 obtained through S1 -> BB -> S2 is just a *view* of S1 (with same temporal bounds), or is just an independent segment, with new temporal bounds. I think both answers are legitimate, and it's mostly a question of making up our mind.

My take is that inferring scope like this can be seen a bit too magic - and you are never sure of what happens when you do a MemorySegment::close, so perhaps what we have w/o this patch is preferrable. What do you think?

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/113/head:pull/113`
`$ git checkout pull/113`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
@@ -392,15 +392,21 @@ public abstract class AbstractMemorySegmentImpl implements MemorySegment, Memory
 
         int pos = bb.position();
         int limit = bb.limit();
-
-        MemoryScope bufferScope = new MemoryScope(bb, null);
         int size = limit - pos;
 
-        MemorySegment bufferSegment = (MemorySegment)nioAccess.bufferSegment(bb);
-        int modes = bufferSegment == null ?
-                defaultAccessModes(size) : bufferSegment.accessModes();
-        Thread owner = bufferSegment == null ?
-                Thread.currentThread() : bufferSegment.ownerThread();
+        AbstractMemorySegmentImpl bufferSegment = (AbstractMemorySegmentImpl)nioAccess.bufferSegment(bb);
+        final MemoryScope bufferScope;
+        int modes;
+        final Thread owner;
+        if (bufferSegment != null) {
+            bufferScope = bufferSegment.scope;
+            modes = bufferSegment.mask;
+            owner = bufferSegment.owner;
+        } else {
+            bufferScope = new MemoryScope(bb, null);
+            modes = defaultAccessModes(size);
+            owner = Thread.currentThread();
+        }
         if (bb.isReadOnly()) {
             modes &= ~WRITE;
         }
--- a/test/jdk/java/foreign/TestSharedAccess.java
+++ b/test/jdk/java/foreign/TestSharedAccess.java
@@ -138,6 +138,23 @@ public class TestSharedAccess {
         } //should fail here!
     }
 
+    @Test(expectedExceptions=IllegalStateException.class)
+    public void testBadCloseWithPendingAcquireBuffer() throws InterruptedException {
+        MemorySegment segment = MemorySegment.allocateNative(16);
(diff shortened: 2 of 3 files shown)


src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
```diff
@@ -392,15 +392,21 @@ public static AbstractMemorySegmentImpl ofBuffer(ByteBuffer bb) {
...
-
-        MemoryScope bufferScope = new MemoryScope(bb, null);
         int size = limit - pos;
 
-        MemorySegment bufferSegment = (MemorySegment)nioAccess.bufferSegment(bb);
-        int modes = bufferSegment == null ?
-                defaultAccessModes(size) : bufferSegment.accessModes();
-        Thread owner = bufferSegment == null ?
-                Thread.currentThread() : bufferSegment.ownerThread();
+        AbstractMemorySegmentImpl bufferSegment = (AbstractMemorySegmentImpl)nioAccess.bufferSegment(bb);
```
JornVernee, line 397: Wondering if we can't just return `bufferSegment` here? It should have the write type right? I think the only thing we'd need to do is drop the WRITE access mode if `bb` is readOnly.

JornVernee:
Wrt your concerns. I think we need to preserve the scope to prevent dead accesses. Consider this example:

```java
        MemorySegment s1 = MemorySegment.allocateNative(intArrayLayout);
        MemorySegment s2 = MemorySegment.ofByteBuffer(s1.asByteBuffer());

        s1.close(); // memory freed

        intElemHandle.set(s2.baseAddress(), (long) 0, 10); // Dead access!
```

It fails with this patch, but wrongly succeeds without it.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
mcimadamore: uhmmm - not quite - the buffer could have been sliced?

> Wrt your concerns. I think we need to preserve the scope to prevent dead accesses. Consider this example:
> 
> ```
>         MemorySegment s1 = MemorySegment.allocateNative(intArrayLayout);
>         MemorySegment s2 = MemorySegment.ofByteBuffer(s1.asByteBuffer());
> 
>         s1.close(); // memory freed
> 
>         intElemHandle.set(s2.baseAddress(), (long) 0, 10); // Dead access!
> ```
> 
> It fails with this patch, but wrongly succeeds without it.

Very good point. Aliasing is an issue w.r.t. closing.

test/jdk/java/foreign/TestSharedAccess.java
```diff
@@ -138,6 +138,23 @@ public void testBadCloseWithPendingAcquire() throws InterruptedException {
         } //should fail here!
     }
 
+    @Test(expectedExceptions=IllegalStateException.class)
+    public void testBadCloseWithPendingAcquireBuffer() throws InterruptedException {
```
PaulSandoz, line 142: I think this test can be improved to avoid the 5s execution time, plus a dangling thread hanging around for 500s.  With some latches we can ensure the main thread calls `segment.close()` while another other thread is within the `spliterator.forEachRemaining` (and further ensure the runnable task completes before the test completes).  I'll send a pull request.

PR #117, opened 2020-04-17
JDK-8243073: Add support for serial confinement

mcimadamore:
This patch adds support for serial confinement - e.g. a handoff operation to transfer ownership from thread A to thread B.

The implementation is based on some iterations we have done in the past. I've added a test, and I've also rewritten some of the existing tests not to rely on waits, but to use latches instead, as observed by Paul.

---------
### Progress

### Issue
 * [JDK-8243073](https://bugs.openjdk.java.net/browse/JDK-8243073): Add support for serial confinement

### Reviewers
 * Jorn Vernee ([jvernee](@JornVernee) - Committer) ⚠️ Review applies to 78dbc41f8ffe0ec2ac61638b2e35875408300e09

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/117/head:pull/117`
`$ git checkout pull/117`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
@@ -100,6 +100,10 @@ import java.util.function.Consumer;
  * the segment using a memory access var handle. Any attempt to perform such operations from a thread other than the
  * owner thread will result in a runtime failure.
  * <p>
+ * Memory segments support <em>serial thread confinement</em>; that is, ownership of a memory segment can change (see
+ * {@link #withOwnerThread(Thread)}). This allows, for instance, for two threads {@code A} and {@code B} to share
+ * a segment in a controlled, cooperative and race-free fashion.
+ * <p>
  * In some cases, it might be useful for multiple threads to process the contents of the same memory segment concurrently
  * (e.g. in the case of parallel processing); while memory segments provide strong confinement guarantees, it is possible
  * to obtain a {@link Spliterator} from a segment, which can be used to slice the segment and allow multiple thread to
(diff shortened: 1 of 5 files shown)


test/jdk/java/foreign/TestSharedAccess.java
```diff
@@ -53,6 +55,32 @@
...
 
+    @Test
+    public void testConfined() throws Throwable {
+        Thread owner = Thread.currentThread();
+        MemorySegment s = MemorySegment.allocateNative(4);
+        AtomicReference<MemorySegment> confined = new AtomicReference<>(s);
+        setInt(s.baseAddress(), 42);
+        assertEquals(getInt(s.baseAddress()), 42);
+        List<Thread> threads = new ArrayList<>();
+        for (int i = 0 ; i < 1000 ; i++) {
```
PaulSandoz, line 66: That's a lot of threads created, 2 would be sufficient to prove the case of the ping-pong hand offs between the main thread and the others.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -199,9 +199,9 @@
      * but with different owner thread. As a side-effect, this segment will be marked as <em>not alive</em>,
      * and subsequent operations on this segment will result in runtime errors.
      * <p>
-     * It is important that the segment returned by this method is published by clients in a safe fashion
-     * (e.g. through a {@code volatile} field, or some other form of synchronization). Failures to do so might
-     * result in the new owner thread reading stale values from the memory region backing this segment.
+     * Write accesses to the segment’s content happens before hand-over from the current owner thread to the
```
PaulSandoz, line 202: Minor tweak.  Suggest adding link to the first happens-before e.g.:

(Unsure what the base path is for the link.)
```
<a href="<...>/java/util/concurrent/package-summary.html#MemoryVisibility"><i>happens-before</i></a>
```

and subsequent declarations using just emphasis.

PR #122, opened 2020-04-21
JDK-8243284: Remove Foreign class

mcimadamore:
The Foreign class is currently used for 3 methods:

* withSize, which takes an unchecked address and gives it a size
* asUnconfined, which takes an existing segment and turns it into an unconfined segment
* asMallocSegment, which takes an unchecked address and converts that into a segment which can be closed, and where closed is mapped into "free"

This patch replaces all these with this method:

static MemorySegment ofNativeUnsafe(MemoryAddress addr, long bytesSize, Thread owner, AutoCloseable cleanup, Object attachment)

Which is then moved directly into MemorySegment - so that we can garbage-collect the Foreign class.

I've also made some changes on the JDK property - new name is simply "jdk.foreign" and I now made sure that it cannot be bypassed with a simple static initializer.

---------
### Progress

### Issue
 * [JDK-8243284](https://bugs.openjdk.java.net/browse/JDK-8243284): Remove Foreign class

### Reviewers
 * Athijegannathan Sundararajan ([sundar](@sundararajana) - Committer) ⚠️ Review applies to 10b7929a7df5d121907c94f1017aa00ea0083f03
 * Jorn Vernee ([jvernee](@JornVernee) - Committer) ⚠️ Review applies to 10b7929a7df5d121907c94f1017aa00ea0083f03
 * Henry Jen ([henryjen](@slowhog) - Committer) ⚠️ Review applies to 286c8d07683991cdfb604070ac07efe6493ecdf7

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/122/head:pull/122`
`$ git checkout pull/122`


src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -473,6 +475,56 @@ static MemorySegment allocateNative(long bytesSize, long alignmentBytes) {
...
+     *                returned segment. If {@code cleanup == null}, no cleanup action is executed.
+     * @param attachment an object that must be kept alive by the returned segment; this can be useful when
+     *                   the returned segment depends on memory which could be released if a certain object
+     *                   is determined to be unreacheable. In most cases this will be set to {@code null}.
+     * @return a new native memory segment with given base address, size, owner, cleanup action and object attachment.
+     * @throws IllegalArgumentException if {@code bytesSize <= 0}.
+     * @throws UnsupportedOperationException if {@code addr} is associated with an heap segment.
+     * @throws IllegalAccessError if the runtime property {@code foreign.unsafe} is not set to either
+     * @throws NullPointerException if {@code addr == null}.
+     * {@code permit}, {@code warn} or {@code debug} (the default value is set to {@code deny}).
```
JornVernee, line 517: This last line should be moved up it looks like

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/Utils.java
```diff
@@ -85,4 +77,27 @@ public static VarHandle fixUpVarHandle(VarHandle handle) {
...
+            }
+            case "permit" -> {}
+            default -> throwIllegalAccessError(foreignAccess);
+        }
+    }
+
+    private static void throwIllegalAccessError(String value) {
+        throw new IllegalAccessError("Can not access jdk.incubator.foreign.Foreign." +
+                " System property 'jdk.incubator.foreign.Foreign' is set to '" + value + "'");
+    }
```
JornVernee, line 102: The message string here should be updated to talk about the new property/method (they still talk about `Foreign`). Might also be nice to pass a string for the source of the check to checkRestrictedAccess so that it can be used in the exception message.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/Utils.java
```diff
@@ -85,4 +77,27 @@ public static VarHandle fixUpVarHandle(VarHandle handle) {
...
+
+    public static void checkRestrictedAccess() {
+        switch (foreignAccess) {
+            case "deny" -> throwIllegalAccessError(foreignAccess);
+            case "warn" -> System.err.println("WARNING: Accessing jdk.incubator.foreign.Foreign.");
+            case "debug" -> {
+                StringBuilder sb = new StringBuilder("DEBUG: Accessing jdk.incubator.foreign.Foreign.");
+                StackWalker.getInstance().walk(s -> {
+                    s
+                            .forEach(f -> sb.append(System.lineSeparator()).append("\tat ").append(f));
```
slowhog, line 89: Is this format intentional? Also it seems to me this can be replaced with

StackWalker().getInstance().forEach()?

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -473,6 +475,56 @@ static MemorySegment allocateNative(long bytesSize, long alignmentBytes) {
...
+     * cleanup operation; this can be useful in cases where the user would like to take an existing native segment
+     * and tweak some of the properties of the original segment - as demonstrated below:
+     * <blockquote><pre>{@code
+    MemorySegment segment = MemorySegment.allocateNative(100);
+    MemorySegment unconfinedSegment = MemorySegment.ofNativeUnsafe(
+                                              segment.baseAddress(),
+                                              segment.byteSize(),
+                                              null,
+                                              segment,
+                                              null);
```
slowhog, line 494: In such use case, close() called on either segment will affect the other as well, correct?
I am assuming in use case like this, the new segment close is preferred to the old segment and developer most likely should not keep reference to the old segment?

mcimadamore: Yes - calling close will also affect the other segment. We have three options here:
* forget about passing a segment - just accept a Runnable; but this will make it harder e.g. to take an existing mapped segment and make it "unconfined" it. Also, still doesn't remove the problem, since you can have a native memory segment for an address X, and you can create another unchecked memory segment at same address which calls Unsafe::freeMemory when you close - which will again invalidate the segment
* make it so that when you pass a segment as an argument, the input segment is also killed - so that it becomes "not alive"; this can be bypassed in the same way described above
* leave it as is

There's no great solution here; I think I like either the current approach, or the first one, where we just take a `Runnable` and leave it up to the user to define what should happen when you close.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/Utils.java
mcimadamore: I didn't change this code, I just moved it around; I'll make another pass.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -512,16 +497,16 @@ static MemorySegment allocateNative(long bytesSize, long alignmentBytes) {
...
+     * @throws NullPointerException if {@code addr == null}.
      */
-    static MemorySegment ofNativeUnsafe(MemoryAddress addr, long bytesSize, Thread owner, AutoCloseable cleanup, Object attachment) {
+    static MemorySegment ofNativeRestricted(MemoryAddress addr, long bytesSize, Thread owner, Runnable cleanup, Object attachment) {
         Objects.requireNonNull(addr);
         if (bytesSize <= 0) {
             throw new IllegalArgumentException("Invalid size : " + bytesSize);
         }
-        Utils.checkRestrictedAccess();
+        Utils.checkRestrictedAccess("MemorySegment.ofNativeUnsafe");
```
JornVernee, line 509:
```suggestion
        Utils.checkRestrictedAccess("MemorySegment.ofNativeRestricted");
```

mcimadamore:
> 
> That's a lot of arguments. Could simplified overrides be provided too?
> For example:
> 
> static MemorySegment ofNativeUnsafe(MemoryAddress addr, long bytesSize,
> AutoCloseable cleanup)

The point of this API is to provide a very general way to construct any kind of native segments. I understand there are many arguments - but then you have a tradeoff between having multiple unsafe API points vs. just one, and I'm not sure that usability is the main concern when stepping onto the unsafe path. This is mostly to allow devs to link their own custom allocators with memory segments, so I don't think the extra arguments should be a big issue (but lack of generality could be).

As for providing pre-baked Runnable cleanup actions, I thought about that and I also thought that we should keep in mind what will happen once we have the ABI work - having a call to "free" won't require writing any JNI code, so by putting these pieces together it should be possible to implement pretty much any behavior.

In terms of usability, I think having a 'resize' method on (unchecked) MemoryAddress would probably be the most appealing choice. For now I think there is a certain appeal in keeping the number of unsafe entry points as low as possible, so that the user can reason about them.

PR #133, opened 2020-04-28
8243978: jextract should generate separate classes with static util methods for each struct, union

sundararajana:
JavaSourceBuilder refactored. HeaderBuilder, StructBuilder are added.

---------
### Progress

### Issue
 * [JDK-8243978](https://bugs.openjdk.java.net/browse/JDK-8243978): jextract should generate separate classes with static util methods for each struct, union

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/133/head:pull/133`
`$ git checkout pull/133`

### DIFF
--- a/src/jdk.incubator.jextract/share/classes/jdk/incubator/jextract/tool/HeaderBuilder.java
+++ b/src/jdk.incubator.jextract/share/classes/jdk/incubator/jextract/tool/HeaderBuilder.java
@@ -1,0 +1,133 @@
+/*
+ * Copyright (c) 2020, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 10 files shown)


src/jdk.incubator.jextract/share/classes/jdk/incubator/jextract/tool/HeaderBuilder.java
```diff
@@ -0,0 +1,133 @@
...
+ */
+class HeaderBuilder extends JavaSourceBuilder {
+    HeaderBuilder(String className, String pkgName, ConstantHelper constantHelper) {
+        super(className, pkgName, constantHelper);
+    }
+
+    public List<JavaFileObject> build() {
+        String res = sb.toString();
+        this.sb.delete(0, res.length());
+        List<JavaFileObject> outputs = new ArrayList<>(constantHelper.getClasses());
```
JornVernee, line 46: Now that the same ConstantHelper instance is shared by mutliple generators, I think it makes sense to move this call to `OutputFactory::generate`, to show that it's output is not just an implementation detail of `HeaderBuilder`.

PR #137, opened 2020-04-30
8244244: Better representing typedef in jextract API

slowhog:
Currently jextract API only present typedef declaration for record types, and using Declaration.Scoped to do so.

This makes downstream tools like jextract impossible to generate more comprehensive type. For example, if the tool would like to create a carrier type for size_t with similar name, because we pass through that, this is impossible.

This patch try to present typedef as Declaration.Variable with Kind TYPE. This will give us more transparency on dealing with typedef, and enable factoring an Declaration.Variable.TYPE from a Type.Delegated.TYPEDEF. 

This would also be useful when a tool collecting type dependencies and encounter anonymous type or types without declaration, this gave the tool a chance to inject an implicit typedef to associate a type with a proper name.

There is one thing worth further discussion is that if we would like to consider generate array with typedef, currently the array element type won't be a Type.Delegated.TYPEDEF, but canonical type. The test case try to explain what should be expected for most use cases, if there is any possible way to use typedef, I'll try to add them.

---------
### Progress

### Issue
 * [JDK-8244244](https://bugs.openjdk.java.net/browse/JDK-8244244): Better representing typedef in jextract API

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/137/head:pull/137`
`$ git checkout pull/137`


slowhog:
> I disagree (pretty strongly) with the direction proposed in this patch. Making a `typedef` a variable is a workaround. Under no circumstances you can get a value out of a typedef - a typedef does not model a variable, and trying to bend the API to do so is, IMHO, wrong.
> 

While I agree with your statement, the Declaration.Variable API doesn't present a value(), only Constant has a value().

In a sense, type() is the value if a Type declaration.

> To me the big issue is that a typedef declaration, as it stands, is only good to wrap a declaration. But the solution I think, would be to have a more explicit `TypeDef` declaration tree, which has a name and a type (where the type can be a `Declared`, if need be).

As current Declaration.Variable stands, it's perfect, with name(), type() and kind(). We can either create a different interface, or rename to something else if that helps.

mcimadamore:
> As current Declaration.Variable stands, it's perfect, with name(), type() and kind(). We can either create a different interface, or rename to something else if that helps.

Variable is perfect in the sense that it has the right structure. But it's not at all what a typdef is! Imagine I write a visitor for the declaration API: if I see vistVariable and I'd expect to see a global, or a struct field, I would now have to defend _against_ typedefs?

From a logical point of view a typedef has nothing to do with a variable declaration.

mcimadamore:
> Sure. We should add a new Visitor method as well?

I think so.

slowhog:
> There are many tests which still depend on Variable.kind.TYPE - how can they pass, given that the code never seem to create such vars?

I didn't realizes TestAnonymousDecl.java was in there. That is something from another branch and have dependency on other part not included in this patch.

I must added by accident. Only TestTypedef.java is suppose to be added.

Strangely, I thought that should be caught by make run-test-jdk_jextract?

PR #138, opened 2020-04-30
8244172: jextract should generate upcall stub creating helper in the functional interface class

sundararajana:
* moved the upcall stub creating utility method as static method of functional interface
* added free upcall stub utility to RuntimeHelper
* added a simple test for function pointer callback

---------
### Progress

### Issue
 * [JDK-8244172](https://bugs.openjdk.java.net/browse/JDK-8244172): jextract should generate upcall stub creating helper in the functional interface class

### Reviewers
 * Maurizio Cimadamore ([mcimadamore](@mcimadamore) - Committer) ⚠️ Review applies to eedbc34214d2e757e2a53a2bd3a73cf4e0720174

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/138/head:pull/138`
`$ git checkout pull/138`

### DIFF
--- a/src/jdk.incubator.jextract/share/classes/jdk/incubator/jextract/tool/HeaderBuilder.java
+++ b/src/jdk.incubator.jextract/share/classes/jdk/incubator/jextract/tool/HeaderBuilder.java
@@ -111,7 +111,7 @@ class HeaderBuilder extends JavaSourceBuilder {
 
     private void addFunctionalFactory(String className, MethodType mtype, FunctionDescriptor fDesc) {
         indent();
-        sb.append(PUB_MODS + "MemoryAddress $make(" + className + " fi) {\n");
+        sb.append(PUB_MODS + "MemoryAddress allocate(" + className + " fi) {\n");
         incrAlign();
         indent();
         sb.append("return RuntimeHelper.upcallStub(" + className + ".class, fi, " + functionGetCallString(className, fDesc) + ", " +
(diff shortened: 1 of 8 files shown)


src/jdk.incubator.jextract/share/classes/jdk/incubator/jextract/tool/HeaderBuilder.java
```diff
@@ -121,4 +108,16 @@ public void addStaticFunctionWrapper(String javaName, String nativeName, MethodT
         sb.append("}\n");
         decrAlign();
     }
+
+    private void addFunctionalFactory(String className, MethodType mtype, FunctionDescriptor fDesc) {
+        indent();
+        sb.append(PUB_MODS + "MemoryAddress $make(" + className + " fi) {\n");
```
mcimadamore, line 114: I'd call this "allocate" for similarity with other constructs.

src/jdk.incubator.jextract/share/classes/jdk/incubator/jextract/tool/resources/RuntimeHelper.java.template
```diff
@@ -77,6 +77,10 @@ public class RuntimeHelper {
         }
     }
 
+    public static void freeUpcallStub(MemoryAddress addr) {
+        ABI.freeUpcallStub(addr);
```
mcimadamore, line 81: Perhaps we could register the returned MemoryAddress with a cleaner and then let the GC pick it up and call the `freeUpcallStub` ? Note that either way we're unsafe - e.g. a client might call this `freeUpcallStub` method while some native code is accessing it (which causes a crash). Or the memory address might become unrecheable while some native code is accessing the upcall. So, either way, some discipline is required.

What I like more about the cleaner approach is that, like with libraries, code is typically longer-lived than allocations, so, IMHO it makes sense to leave that one to the GC, so that the user code doesn't have to worry about deallocating that.

PR #141, opened 2020-05-01
JDK-8244270: reorganize ABI-dependent layout constants (second attempt)

mcimadamore:
This is another attempt to reorganize ABI-dependent layout constants. The constant reorg is identical to the one proposed in:

 https://git.openjdk.java.net/panama-foreign/pull/97

But this time I've also integrated the changes into the jextract branch (as illustrated in this PR).

The main realization which led to this work was that we don't need the ABI type enum: we already have the jextract primitive type enum. So, we could just define constants for the common ABI types which are useful for ABI classification in SystemABI - and then we could start from here in jextract, and augment these layouts by attaching an extra attribute with jextract's Primitve/Kind enum.

I've simplified the API a bit by moving the layout inside the enum itself (so that the layout can be accessed directly from the enum constant) - and also simplified the primitive type factory.

Everything worked out as expected - note how much duplicated layout creation we managed to get rid of.

One problem I had was that the loader we use inside jextract test was not delegating correctly (was using `null` as parent loader) so I had to fix that.

I'm not 100% sure that this jextract/type extra annotation makes sense - at the end of the day it is now only used inside tests (as it makes layout comparison easier - since layouts might have additional attributes such as name and such).

So I'm open to suggestion on whether we want to keep it or not - I believe that tests which look for specific layouts should instead probably check for specific types instead. Anyway, we can also start from this cleanup, and take care of these issues later; the important thing is that, with this rewrite we are finally able to cleanup the ABI constants the way we wanted.

---------
### Progress

### Issue
 * [JDK-8244270](https://bugs.openjdk.java.net/browse/JDK-8244270): reorganize ABI-dependent layout constants (second attempt)

### Reviewers
 * Jorn Vernee ([jvernee](@JornVernee) - Committer) ⚠️ Review applies to 83fd4be90c51ef88c2a9e1d8d8ef3eb871d71e72

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/141/head:pull/141`
`$ git checkout pull/141`


slowhog:
Overall I think this is a nice move, we want constants in SystemABI with proper MemoryLayout as we needed for access ABI specific type.

I don't like the Utils.pick where we need to pass in all ABI candidates, we have picked an implementation class at that point, we should be able to get those just from the implementation class. I would think instead of get rid of the enum, we may actually hide the types internally and have each ABI class provide layout. However, this can be clean up later, the important thing is that the constants are in place.

The other thing I don't think we should be doing is to make CLASS_ATTRIBUTE like "abi/sysv/class" part of the layout. 

SystemABI constants should be a delegation to the implementation class, and strip the abi-class attribute. Trusting this attribute from external is dangerous, a tool can compose a bad layout to do some harm?

Bottom line, type is reasonable to expose to public, ABI implementation detail such as classification is not.

slowhog:
> The other thing I don't think we should be doing is to make CLASS_ATTRIBUTE like "abi/sysv/class" part of the layout.
> 
> SystemABI constants should be a delegation to the implementation class, and strip the abi-class attribute. Trusting this attribute from external is dangerous, a tool can compose a bad layout to do some harm?
> 
> Bottom line, type is reasonable to expose to public, ABI implementation detail such as classification is not.

Actually expose type or classification have the same effect, as classification is derived from type, the attribute value is an instance of a hidden class type so it's somewhat protected.

mcimadamore:
> Bottom line, type is reasonable to expose to public, ABI implementation detail such as classification is not.

There's a tradeoff here. The ABI obviously needs more info than those attached to a layout to do the classification; otherwise the layout of a `float` and that of an `int` would be the same on most platform. One might argue - but why not just using the carrier types attached to the `MethodType` (also passed to the ABI) to do the classification? Because that's brittle - it works for primitives, it doesn't work for structs passed by value, which, on certain ABI (e.g. SysV) require a fair amount of recursive classification.

So the choices are between augmenting layouts (with attributes) and coming up with a _new_ abstraction which is essentially a layout + some classification info (and maybe a Java carrier). In the name of minimality of the API surface we opted for the first option. Every time we considered moving away from layouts, the "bang for bucks" factor seemed relatively low.

PR #142, opened 2020-05-03
Alternative scalable MemoryScope

plevart:
This is an alternative MemoryScope which is more scalable when used in a scenario where child scope is frequently acquired and closed concurrently from multiple threads (for example in parallel Stream.findAny())

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/142/head:pull/142`
`$ git checkout pull/142`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
@@ -184,12 +184,7 @@ public abstract class AbstractMemorySegmentImpl implements MemorySegment, Memory
         if (owner == newOwner) {
             throw new IllegalArgumentException("Segment already owned by thread: " + newOwner);
         } else {
-            try {
-                return dup(0L, length, mask, newOwner, scope.dup());
-            } finally {
-                //flush read/writes to memory before returning the new segment
-                VarHandle.fullFence();
-            }
+            return dup(0L, length, mask, newOwner, scope.dup());
         }
     }
 
@@ -203,7 +198,7 @@ public abstract class AbstractMemorySegmentImpl implements MemorySegment, Memory
     }
 
     private final void closeNoCheck() {
-        scope.close(true);
+        scope.close();
     }
 
     final AbstractMemorySegmentImpl acquire() {
@@ -421,7 +416,7 @@ public abstract class AbstractMemorySegmentImpl implements MemorySegment, Memory
             modes = bufferSegment.mask;
             owner = bufferSegment.owner;
         } else {
-            bufferScope = new MemoryScope(bb, null);
+            bufferScope = MemoryScope.create(bb, null);
             modes = defaultAccessModes(size);
             owner = Thread.currentThread();
         }
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
@@ -184,7 +184,12 @@ public abstract class AbstractMemorySegmentImpl implements MemorySegment, Memory
         if (owner == newOwner) {
             throw new IllegalArgumentException("Segment already owned by thread: " + newOwner);
         } else {
-            return dup(0L, length, mask, newOwner, scope.dup());
+            try {
+                return dup(0L, length, mask, newOwner, scope.dup());
+            } finally {
+                //flush read/writes to segment memory before returning the new segment
+                VarHandle.fullFence();
+            }
         }
     }
 
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
@@ -133,20 +133,14 @@ abstract class MemoryScope {
     }
 
     private static final class Root extends MemoryScope {
-        private final LongAdder acquires;
-        private final LongAdder releases;
+        private final LongAdder acquires = new LongAdder();
+        private final LongAdder releases = new LongAdder();
         private final Object ref;
         private final Runnable cleanupAction;
 
-        private Root(LongAdder acquires, LongAdder releases, Object ref, Runnable cleanupAction) {
-            this.acquires = acquires;
-            this.releases = releases;
+        private Root(Object ref, Runnable cleanupAction) {
             this.ref = ref;
             this.cleanupAction = cleanupAction;
-        }
-
-        private Root(Object ref, Runnable cleanupAction) {
-            this(new LongAdder(), new LongAdder(), ref, cleanupAction);
         }
 
         @Override
@@ -154,14 +148,13 @@ abstract class MemoryScope {
             // increment acquires 1st
             acquires.increment();
             // check state 2nd
-            int state = (int) STATE.getVolatile(this);
-            while (state > STATE_OPEN) {
+            int state;
+            while ((state  = (int) STATE.getVolatile(this)) > STATE_OPEN) {
                 if (state == STATE_CLOSED) {
                     releases.increment();
                     throw new IllegalStateException("This scope is already closed");
                 }
                 Thread.onSpinWait();
-                state = (int) STATE.getVolatile(this);
             }
             return new Child();
         }
@@ -180,10 +173,8 @@ abstract class MemoryScope {
             if (state == STATE_CLOSED) {
                 throw new IllegalStateException("This scope is already closed");
             }
-            // pre-allocate duped scope so we don't get OOME later and be left with this scope closed;
-            // we only return this instance if releases.sum() == acquires.sum(), so both LongAdder(s)
-            // can be reused...
-            var duped = close ? null : new Root(acquires, releases, ref, cleanupAction);
+            // pre-allocate duped scope so we don't get OOME later and be left with this scope closed
+            var duped = close ? null : new Root(ref, cleanupAction);
             // modify state to STATE_CLOSING 1st
             STATE.setVolatile(this, STATE_CLOSING);
             // check for absence of active acquired children 2nd
@@ -201,7 +192,6 @@ abstract class MemoryScope {
                 }
                 return null;
             } else {
-                // assert releases.sum() == acquires.sum() && state == STATE_CLOSED;
                 return duped;
             }
         }
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
@@ -28,99 +28,215 @@ package jdk.internal.foreign;
 
 import java.lang.invoke.MethodHandles;
 import java.lang.invoke.VarHandle;
+import java.util.concurrent.atomic.LongAdder;
 
 /**
  * This class manages the temporal bounds associated with a memory segment. A scope has a liveness bit, which is updated
- * when the scope is closed (this operation is triggered by {@link AbstractMemorySegmentImpl#close()}). Furthermore, a scope is
- * associated with an <em>atomic</em> counter which can be incremented (upon calling the {@link #acquire()} method),
- * and is decremented (when a previously acquired segment is later closed).
+ * when the scope is closed (this operation is triggered by {@link AbstractMemorySegmentImpl#close()}). Furthermore,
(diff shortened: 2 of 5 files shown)


src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
```diff
@@ -28,99 +28,215 @@
...
-                throw new IllegalStateException("Segment is not alive");
-            } else if (value == MAX_ACQUIRE) {
-                //overflow
-                throw new IllegalStateException("Segment acquire limit exceeded");
-            }
-        } while (!COUNT_HANDLE.compareAndSet(this, value, value + 1));
-        return new MemoryScope(ref, this::release);
-    }
+    private static final class Root extends MemoryScope {
+        private final LongAdder acquires;
```
PaulSandoz, line 136: Did you consider collapsing into just one `LongAdder` using `increment` (for acquiring), `decrement` (for releasing), and `sum` (checking a zero sum for closing)?  Perhaps it was discussed already, i lost track of all the discussion.

PaulSandoz:
I think this should work under the well-defined `Spliterator` use case but might be problematic if made public sometime later on.  For example:

```
MemorySegment s = ...
var a = s.acquire();
a.close();
var d = s.dup();
a.acquire();  // transiently affecting state of d since the adders are shared
```

There also might be ways to poke a `Spliterator` to induce such behaviour e.g. the `Consumer` passed to `tryAdvance` creating a new `Spliterator` from the given segment, and operating on that after the `tryAdvance` completes.  Clearly that's not what someone should be doing but it's starting complex enough that it's hard to reason about the possible effects and whether they are harmless or not, at least for me :-)

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
```diff
@@ -28,99 +28,215 @@
...
-            if (value <= UNACQUIRED) {
-                //cannot get here - we can't close segment twice
-                throw new IllegalStateException();
+        @Override
+        MemoryScope acquire() {
+            // increment acquires 1st
+            acquires.increment();
+            // check state 2nd
+            int state = (int) STATE.getVolatile(this);
+            while (state > STATE_OPEN) {
```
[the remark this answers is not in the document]
plevart: Sure, that's possible. Thanks.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
plevart: That would not work. LongAdder.sum() is not an atomic operation. Consider the following scenario:
- precondition: adder.sum() == 0
- thread A1 successfully acquires child segment: adder.sum() == 1
- thread C calls close() on root segment
    * sets state = CLOSING
    * calls adder.sum() which starts summing LongAdder cells and just misses adder.increment() of thread A2...
- thread A2 tries to acquire child segment and so it 1st does adder.increment()
- thread A2 reads state == CLOSING and therefore undoes previous increment with adder.decrement() and fails, but...
- thread C that just missed A2's adder.increment() does see A2's adder.decrement() because it is performed on a different LongAdder cell which C reads after it is incremented, so what C gets as a result of adder.sum() is 0, it completes closing the root segment by setting state = CLOSED and executing cleanup, but thread A1 still has access to the child segment and BOOM!!!

Having two LongAdders (acquires, releases) and doing the sum() per-partes sequentially: 1st releases.sum(), then acquires.sum() we guarantee that we never miss an acquiring thread's acquires.increment() but see the thread's releases.increment(), so above scenario is not possible.

plevart:
So, in your scenario:
```
MemorySegment s = ...
var a = s.acquire();
a.close();
var d = s.dup();
a.acquire();  // transiently affecting state of d since the adders are shared
```
...the a.acquire() will fail, because s.state == CLOSED, but the shared acquires.sum() will briefly be greater than releases.sum() and therefore could affect d.close() in a way that it would cause spurious failure of d.close(). This is of course undesirable (like it was undesirable to have spurious failures of acquire() in previous versions of this MemoryScope until I added a two-phase close), so it would be best to not share the acquires/releases LongAdders with a duped MemoryScope. I'll fix this by creating new LongAdder instances for duped scope. Thanks.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
```diff
@@ -184,12 +184,7 @@ public MemorySegment withOwnerThread(Thread newOwner) {
...
             throw new IllegalArgumentException("Segment already owned by thread: " + newOwner);
         } else {
-            try {
-                return dup(0L, length, mask, newOwner, scope.dup());
-            } finally {
-                //flush read/writes to memory before returning the new segment
-                VarHandle.fullFence();
-            }
+            return dup(0L, length, mask, newOwner, scope.dup());
         }
```
plevart, line 188: Maybe you're asking why did I remove this fullFence() call? I don't think it is needed. I checked all implementations of MemorySegment and they all contain just final fields. It is therefore safe to publish such instances via data races and no explicit fences are needed.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
mcimadamore: I agree with Peter that using a single long would misses update and you could see 'zero' when in reality something has happened but `sum` didn't get the memo. The entire approach, as far as I understand, relies on subtle happens-before relationship which arise precisely because of the use of the two counters (and the fact that one counter is always updated before the other).

plevart:
I agree that the main problem here is the constant acquire/close-ing of memory segment in Spliterator.tryAdvance(). It is interesting to observe a slight performance improvement even in serial stream (according to above results) while the concurrent contention in current MemoryScope is obvious in parallel stream. The above results are of course "distorted" by the fact that there are lots of small elements in the testing stream (int == 4 bytes) so performance figures are dominated by the overhead. But there may be valid compositions of .findAny() or such that use bigger elements so that the overhead plays a minor role. Still, current MemoryScope could present a problem even in such compositions where the alternative MemoryScope would be perfectly acceptable.
Anyway I was also thinking of "improving" the Spliterator API with two empty default methods that would present real improvement in such scenarios:

```
public interface Spliterator<T> {

    /**
     * Before a sequence of {@link #tryAdvance} calls from a thread, this method may
     * be called in the same thread to establish a context that may help in improving the
     * efficiency of subsequent {@code tryAdvance} calls. If this method is called and it returns
     * normally, then it is guaranteed that the sequence of {@code tryAdvance} calls will be
     * followed by a call to {@link #endTryAdvanceSequence} method to end such sequence.
     * The end of sequence of {@code tryAdvance} calls may be announced before all elements of
     * the Spliterator have been exhausted.
     *
     * @implNote the default implementation of this method does nothing
     * @since 15
     */
    default void beginTryAdvanceSequence() {
    }

    /**
     * If {@link #beginTryAdvanceSequence} method is called and the call returns normally before
     * a sequence of {@link #tryAdvance} calls, then this method is guaranteed to be called to end such
     * sequence and give this Spliterator a chance to clean-up what was established in
     * {@code beginTryAdvanceSequence}.
     *
     * @implNote the default implementation of this method does nothing
     * @since 15
     */
    default void endTryAdvanceSequence() {
    }
```

Would this be an acceptable "addtition" to Spliterator API?

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
plevart: ...right and the fact that the sum() of both counters is performed in the opposite order as they are updated.

mcimadamore:
> Would this be an acceptable "addtition" to Spliterator API?

I'll leave this to Paul. IMHO doing this for an incubating API seems a bit too much (although I can buy that there can be other benefits in doing this). Another option would be in having this be a JDK-internal spliterator API, so that our own spliterators could implement that and get better stream performances (after all, I don't expect a lot of users to come up with their own spliterators). If stream sees that one of those 'special' stream is being used, it can use the more efficient implementation.

plevart:
This would be universally useful. I once already missed such methods in a Spliterator I was trying to design, backed by database.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
mcimadamore: I think we had a discussion on this; we knew publishing segments was safe (all fields final), but there were some perplexities that the fences added at the end of the constructor were enough to warrant flushing of writes to the off-heap memory backing the segment. This effect is not covered (afaik) by the JMM.

plevart: Ah, I see. Will revert that back then.

mcimadamore: > But why is AbstractMemorySegmentImpl.withOwnerThread() doing this differently than AbstractMemorySegmentImpl.acquire() ? acquire() too is handing access to a different thread and in acquire() there is no explicit fence.

This is a valid point. I guess from the past, acquire() meant "racy" - whereas for handoff patterns (e.g. withOwnerAccess) we wanted to be extra careful that only one thread at a time has access.

Note that, currently, while all acquired segments slices can be operated concurrently in a race-free fashion, unfortunately there is still the main segment that can be written/read concurrently by the original thread. So adding fences in this situation just kicks the can down the road.

If we had a way so that the original segment became inaccessible while splitted with a spliterator, then I'd agree with you that a fence would be ideal (since this is another case of handoff).

But it is very hard to invalidate the original segment; few options I've considered:

* make the 'valid state' check more complex - e.g. in your implementation we could check whether the two counters yield same value before allowing access to the original thread - but I think this would be slow, and I think it could still allow races (this is a case of check & act - where we have no guarantees that, by the time memory is read/written, the condition we checked still holds true)

* as we do in other places of the API - just kill the original segment after you split (e.g. make it not alive). This will make it inaccessible, which is consistent with what we want. But now there's a problem: how do you get back the original segment after you're done with the spliterator? Joining slices is messy. And add synchronization to allow for spliterators to be 'closed' atomically, will just add more contention.

So, it seems like the cost for fixing this particular issue is higher than the actual problem; there's an access mode called ACQUIRE - if a segment owner is not OK with having races with different threads - that mode can be disabled.

plevart: Yeah, I was just thinking of the situation where a segment is passed to parallel stream and the threads of parallel stream operate on slices (acquired children) in read-only way, since this is the source of information and should not be mutated. Now if the segment was created and written to in one thread before handed to parallel stream then workers from the FJPool may not see the content of their slices (acquired children) fully initialized since normal synchronization like dispatch of control to other thread presumably doesn't have the same effect on the off-heap memory (on all platforms)?

mcimadamore: > Yeah, I was just thinking of the situation where a segment is passed to parallel stream and the threads of parallel stream operate on slices (acquired children) in read-only way, since this is the source of information and should not be mutated. Now if the segment was created and written to in one thread before handed to parallel stream then workers from the FJPool may not see the content of their slices (acquired children) fully initialized since normal synchronization like dispatch of control to other thread presumably doesn't have the same effect on the off-heap memory (on all platforms)?

I'm fine with adding the fence in principle - if it adds too much overhead (which I didn't try), I think we should leave it out though, since the benefits are limited.

PaulSandoz:
Thanks Peter for the explanation.  I now see why two monotonically increasing counters work (with careful ordering the non-atomic sums).  My sense is we should let this soak in the panama repo.

With regards to the proposed changes to `Spliterator`, i fear this will complicate the already complicated set of interactions that are possible, and require clients to take into account what is in effect a new characteristic.  Instead i am wondering if we can add a new traversal method that can short-circuit, for example:

```
    // traverses until no elements or the action returns false, which ever is first
    // if action returns false and if there are elements remaining then traversing may continue
    // with further calls to this method, or tryAdvance, or forEachRemaining
    default void forEachSomeRemaining(Predicate<? super T> action) {
        boolean[] _continue = new boolean[1]; // ugly side return
        boolean continue;
        do {
            boolean continue = tryAdvance(e -> { _continue[0] = action.test(e); });
        } while (continue && _continue[0]);
    }

```

PR #157, opened 2020-05-11
Add MemoryLayout::byteOffset

ChrisHegarty:
Hi,

As part of feedback on the Foreign Memory API (when experimenting with
its usage internally in the JDK), a small number of potential usability
enhancements could be made to the API. This is the first such.

This change:
 1) renames MemoryLayout::offset to MemoryLayout::bitOffset, and
 2) adds MemoryLayout::byteOffset

This allows for easier interoperation with MemoryAddress, which deals
with offsets and lengths in terms of bytes (rather than bits), e.g.
   addr.addOffset(layout.byteOffset(groupElement("foo")))

This brings a nice symmetry to the API; there is a trio of bit-wise
methods: bitAlignment, bitOffset, bitSize, and their matching byte-wise
counterparts; byteAlignment, byteOffset, byteSize.

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/157/head:pull/157`
`$ git checkout pull/157`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryLayout.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryLayout.java
@@ -97,7 +97,7 @@ import java.util.stream.Stream;
  * Layout paths are typically expressed as a sequence of one or more {@link PathElement} instances.
  * <p>
  * Layout paths are for example useful in order to obtain offsets of arbitrarily nested layouts inside another layout
- * (see {@link MemoryLayout#offset(PathElement...)}), to quickly obtain a memory access handle corresponding to the selected
+ * (see {@link MemoryLayout#bitOffset(PathElement...)}), to quickly obtain a memory access handle corresponding to the selected
  * layout (see {@link MemoryLayout#varHandle(Class, PathElement...)}), to select an arbitrarily nested layout inside
  * another layout (see {@link MemoryLayout#select(PathElement...)}, or to transform a nested layout element inside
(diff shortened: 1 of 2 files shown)


JornVernee:
This seems like it should go to the `foreign-memaccess` branch, but is currently targeting the `foreign-jextract` branch. If you click the "Edit" button on the right next to the title of the PR, you should be able to change the target branch from the drop-down just below the PR title.

mcimadamore:
> This reminds me something I meant to bring up awhile ago: it would be
> nice to have a counterpart method "ofPaddingBytes" and pre-defined
> layouts like "PAD_BYTE_1" to the existing Bit based versions.

Right now all factories take bits (not bytes) as input, as that's more "primitive" and scales more to things like packed layouts. That said, I see your point. My feeling is that layout construction typically happens once, and then information is derived from the layout. When you construct you don't care much whether to use bits or bytes (other than for stylistic preferences) - but when you access layout information, you might be interested in getting either bits or bytes out, depending on which code you might need to interact with. So, in my view, fixing methods like "offset" takes the priority here. We'll keep evaluating against use cases and see how annoying that is. Thanks for bringing this up.

ChrisHegarty:
> This seems like it should go to the `foreign-memaccess` branch, but is currently targeting the `foreign-jextract` branch. If you click the "Edit" button on the right next to the title of the PR, you should be able to change the target branch from the drop-down just below the PR title.

Thanks Jorn, the intent was to target the `foreign-memaccess` branch. PR target branch updated.

PR #158, opened 2020-05-11
8244720: Check MethodType and FunctionDescritpor used when linking

JornVernee:
Hi,

This patch adds exhaustive checking to the MethodType and FunctionDescriptor used to link down calls and upcalls.

These checks define and enforce a set of acceptable carrier types, as well as which carrier type & memory layout combinations are acceptable. The set of accepted carrier types is:

1. The primitives: byte, short, char, int, long, float and double (excluding void and boolean)
2. MemoryAddress
3. MemorySegment

For (1), it is also checked that the used MemoryLayout is a ValueLayout, and that the size of the carrier matches the size of the layout. For (2) the expected layout must also be a ValueLayout, and again the size is checked. For (3) it is only checked that the layout is a GroupLayout, (since we don't have access to the size of the segment when linking).

This makes it easier to reason about the set of MethodType and FunctionDescriptor combinations that can be used during linking, as well as helping to catch any errors made with mismatching carrier types and memory layouts.

The additional checks turned up 2 cases of carrier type to memory layout mismatch in StdLibTest, which I've fixed.

Thanks,
Jorn

---------
### Progress

### Issue
 * [JDK-8244720](https://bugs.openjdk.java.net/browse/JDK-8244720): Check MethodType and FunctionDescritpor used when linking

### Reviewers
 * Maurizio Cimadamore ([mcimadamore](@mcimadamore) - Committer) ⚠️ Review applies to 6e55e6ccb8851174bb563c950e4cf92ccc815355

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/158/head:pull/158`
`$ git checkout pull/158`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/SharedUtils.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/SharedUtils.java
@@ -25,20 +25,18 @@
 package jdk.internal.foreign.abi;
 
 import jdk.incubator.foreign.FunctionDescriptor;
+import jdk.incubator.foreign.GroupLayout;
 import jdk.incubator.foreign.MemoryAddress;
+import jdk.incubator.foreign.MemoryLayout;
 import jdk.incubator.foreign.MemorySegment;
+import jdk.incubator.foreign.SequenceLayout;
 import jdk.incubator.foreign.SystemABI;
+import jdk.incubator.foreign.ValueLayout;
 import jdk.internal.foreign.MemoryAddressImpl;
 import jdk.internal.foreign.Utils;
-
-import jdk.incubator.foreign.GroupLayout;
-import jdk.incubator.foreign.MemoryLayout;
-import jdk.incubator.foreign.SequenceLayout;
-import jdk.incubator.foreign.ValueLayout;
 import jdk.internal.foreign.abi.aarch64.AArch64ABI;
 import jdk.internal.foreign.abi.x64.sysv.SysVx64ABI;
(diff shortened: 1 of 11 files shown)


src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/SharedUtils.java
```diff
@@ -186,11 +188,49 @@ private static MemoryAddress bufferCopy(MemoryAddress dest, MemorySegment buffer
...
+        return carrier == byte.class
+            || carrier == short.class
+            || carrier == char.class
+            || carrier == int.class
+            || carrier == long.class
+            || carrier == float.class
+            || carrier == double.class;
+    }
+
+    private static void checkCompatibleType(Class<?> carrier, MemoryLayout layout, long addressSize) {
```
mcimadamore, line 201: I recalled that we do this check in the foreign memory access API (see `LayoutPath.java`):

```
if (!carrier.isPrimitive() || carrier == void.class || carrier == boolean.class // illegal carrier?
                || Wrapper.forPrimitiveType(carrier).bitWidth() != layout.bitSize()) { // carrier has the right size?
            throw new IllegalArgumentException("Invalid carrier: " + carrier + ", for layout " + layout);
        }
```
Can we perhaps unify a little? It seems like your check should subsume the existing check (at least for ValueLayouts).

JornVernee: I've moved some of the common code into jdk.internal.foreign.Utils

PR #160, opened 2020-05-11
Fix concurrent MemoryScope.close() or MemoryScope.dup() race

plevart:
In case a MemorySegment is not confined to an "owner" thread, it is possible to call close() or withOwnerThread(newThread) on it from multiple threads concurrently and such calls are not screened for thread confinement but just forwarded to MemoryScope.close() and .dup(). It is therefore vital that close() and dup() are made atomic so that cleanupAction is executed just once.

---------
### Progress

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/160/head:pull/160`
`$ git checkout pull/160`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
@@ -170,13 +170,23 @@ abstract class MemoryScope {
         }
 
         private MemoryScope closeOrDup(boolean close) {
-            if (state == STATE_CLOSED) {
-                throw new IllegalStateException("This scope is already closed");
-            }
             // pre-allocate duped scope so we don't get OOME later and be left with this scope closed
             var duped = close ? null : new Root(ref, cleanupAction);
-            // modify state to STATE_CLOSING 1st
-            STATE.setVolatile(this, STATE_CLOSING);
+            // try to modify state from STATE_OPEN -> STATE_CLOSING 1st (atomically)
+            int state;
+            if ((state = (int) STATE.compareAndExchange(this, STATE_OPEN, STATE_CLOSING)) != STATE_OPEN) {
+                while (state == STATE_CLOSING) {
+                    Thread.onSpinWait();
+                    state = (int) STATE.getVolatile(this);
+                }
+                if (state == STATE_CLOSED) {
+                    throw new IllegalStateException("This scope is already closed");
+                } else {
+                    // assert state == STATE_OPEN;
+                    throw new IllegalStateException("Cannot close this scope as it has active acquired children");
+                }
+            }
+            // if we get this far, we are the winner of a closing race (only relevant in non-confined segments)
             // check for absence of active acquired children 2nd
             // IMPORTANT: 1st sum releases, then sum acquires !!!
             if (releases.sum() != acquires.sum()) {


plevart:
The ParallelSum benchmark doesn't show and regressions:
```
before fix:

Benchmark                                         Mode  Cnt     Score     Error  Units
ParallelSum.segment_parallel                      avgt   30    42.298 ±   0.139  ms/op
ParallelSum.segment_parallel_bulk                 avgt   30    38.268 ±   0.318  ms/op
ParallelSum.segment_serial                        avgt   30   109.087 ±   0.618  ms/op
ParallelSum.segment_stream_findany_parallel       avgt   30  1721.486 ± 306.951  ms/op
ParallelSum.segment_stream_findany_parallel_bulk  avgt   30    32.263 ±   1.572  ms/op
ParallelSum.segment_stream_findany_serial         avgt   30  5090.839 ±  19.335  ms/op
ParallelSum.segment_stream_findany_serial_bulk    avgt   30   105.765 ±   1.527  ms/op
ParallelSum.segment_stream_parallel               avgt   30    38.995 ±   0.236  ms/op
ParallelSum.segment_stream_parallel_bulk          avgt   30    38.146 ±   0.225  ms/op
ParallelSum.unsafe_parallel                       avgt   30     6.476 ±   0.226  ms/op
ParallelSum.unsafe_serial                         avgt   30   107.892 ±   0.882  ms/op

after fix:

Benchmark                                         Mode  Cnt     Score     Error  Units
ParallelSum.segment_parallel                      avgt   30    42.195 ±   0.094  ms/op
ParallelSum.segment_parallel_bulk                 avgt   30    38.216 ±   0.269  ms/op
ParallelSum.segment_serial                        avgt   30    98.562 ±   9.357  ms/op
ParallelSum.segment_stream_findany_parallel       avgt   30  1532.052 ± 325.223  ms/op
ParallelSum.segment_stream_findany_parallel_bulk  avgt   30    32.167 ±   1.664  ms/op
ParallelSum.segment_stream_findany_serial         avgt   30  5443.200 ± 370.581  ms/op
ParallelSum.segment_stream_findany_serial_bulk    avgt   30   104.543 ±   0.924  ms/op
ParallelSum.segment_stream_parallel               avgt   30    39.010 ±   0.281  ms/op
ParallelSum.segment_stream_parallel_bulk          avgt   30    38.161 ±   0.314  ms/op
ParallelSum.unsafe_parallel                       avgt   30     6.386 ±   0.250  ms/op
ParallelSum.unsafe_serial                         avgt   30    99.642 ±   8.711  ms/op
```

plevart:
On 5/12/20 3:55 PM, Maurizio Cimadamore wrote:
>
> *@mcimadamore* commented on this pull request.
>
> I've been playing around a bit and discussed this offline with 
> @JornVernee <https://github.com/JornVernee> too. A realization I had 
> some time ago is that what we're trying to do here is a read/write 
> lock, where multiple acquire can occur at the same time, but where 
> close of the root should be exclusive. I think all the machinery 
> around CLOSING is essentially due to the code trying to mimic that 
> pattern.
>
> I've put together this code:
> https://gist.github.com/mcimadamore/54fecd888bcdc8fcf8aefaa51d4cbcbd
>
> Which, I think, simplifies over the status quo, and also adds back the 
> atomicity seeked in the proposed patch. Also, since now there some 
> guarantee that no acquire can take place while a close is also taking 
> place (and pending acquires will be invalidated - using the optimistic 
> read logic), then we can just use a single long adder instead of two.
>
> The state is also simplified, since we just need a boolean flag.
>
> This seems to provide same numbers as what we have now, but IMHO the 
> code is bit easier to follow, as there are less moving parts. What do 
> you think?
>

I think this is correct implementation. And perhaps easier to follow 
too. Nice use of StampedLock which hides the complexity of read-write 
locking under the easy to use API. And the scalable counter state is 
reduced by half without reducing scalability too, which is also an 
improvement - a particular thread typically stabilizes to always update 
the same cell in the LongAdder so there should be little to no 
contention even with one counter.

I think this should replace current version. I'm all for it. Good work.


Regards, Peter

mcimadamore:
> I think this should replace current version. I'm all for it. Good work.

Cool! Then I suggest you maybe close this PR and I will submit another with the new impl, if that's ok?

plevart:
Closing this PR in favor of Maurizio's implementation.

PR #161, opened 2020-05-13
Add MemorySegment::fill

ChrisHegarty:
Hi,

As part of feedback on the Foreign Memory API (when experimenting with
its usage internally in the JDK), a small number of potential usability
enhancements could be made to the API. This is the second such.

  MemorySegment::fill

Which fills a value into the given memory segment. Fill can be useful to
initialize or reset the memory of a segment, similar(ish) to memset.

There are obviously different ways to provide such functionality, e.g.
accepting a fill value with a bit width greater than 8 bits, but on
balance this single method will likely satisfy the majority of simple
use-cases. Other more advanced initialization scenarios would likely be 
served better from an interaction with the Vector API, or some such.

---------
### Progress

### Reviewers
 * Jim Laskey ([jlaskey](@JimLaskey) - Committer) ⚠️ Review applies to e66ab857c91a48aec9985b367fce031c88b12d23
 * Jorn Vernee ([jvernee](@JornVernee) - Committer) ⚠️ Review applies to e66ab857c91a48aec9985b367fce031c88b12d23
 * Paul Sandoz ([psandoz](@PaulSandoz) - Committer) ⚠️ Review applies to af894de2e57b94424da9119e85d0f3db77d3f146

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/161/head:pull/161`
`$ git checkout pull/161`

### DIFF
--- a/test/jdk/java/foreign/TestSegments.java
+++ b/test/jdk/java/foreign/TestSegments.java
@@ -32,10 +32,10 @@ import jdk.incubator.foreign.MemoryLayouts;
 import jdk.incubator.foreign.MemorySegment;
 import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
-
 import java.lang.invoke.VarHandle;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
+import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 import java.util.ArrayList;
 import java.util.List;
@@ -43,8 +43,9 @@ import java.util.Spliterator;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.LongFunction;
+import java.util.function.Supplier;
 import java.util.stream.Stream;
-
+import static jdk.incubator.foreign.MemorySegment.WRITE;
 import static org.testng.Assert.*;
 
 public class TestSegments {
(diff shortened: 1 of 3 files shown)


test/jdk/java/foreign/TestSegments.java
```diff
@@ -161,6 +162,102 @@ public void testAccessModes(int accessModes) {
...
+        try (MemorySegment segment = MemorySegment.ofArray(new byte[2]).asSlice(0, 0)) {
+            MemorySegment.fill(segment, (byte) 0xFF);
+        }
+        try (MemorySegment segment = MemorySegment.ofByteBuffer(ByteBuffer.allocateDirect(0))) {
+            MemorySegment.fill(segment, (byte) 0xFF);
+        }
+    }
+
+    @Test(expectedExceptions = NullPointerException.class)
+    public void testFillWithNull() {
```
JimLaskey, line 257: Not "fill with null" but "fill null".

mcimadamore:
> I think we need to consider grouping bulk operations in the same place.

I suggested the location to Chris. Since the method operates on segment it felt right to add it on MemorySegment. If we add it on MemoryAddress it will look a tad odd IMHO, since it works on segments.

We could tweak fill to work on MemoryAddress, but that means that the user has to pass an address and a length (which is what a segment is).

Or we could make fill an instance method on segment.

I'm not sure there's one answer that's better than the others here - but I don't have any strong feelings one way or another, so I'm open to suggestions.

ChrisHegarty:
I agree with the logic of why this method fits better in MemorySegment, rather than MemoryAddress.

Fill seems more like a "utility", rather than an operation on the segment itself, so I marginally favour a static method.

PaulSandoz:
I think `copy` is a little odd where it is located, for the same reasons argued if `fill` was located on `MemoryAddress`, since both really operate on segments.

Using `MemoryAddress` affords some flexibility in terms of offset selection but that can also be achieved by slicing segments.

As separate issues perhaps consider moving or duplicating `copy` on `MemorySegment` and exposing a `fill` accepting a source segment?

mcimadamore:
Thought a bit more about this.

There's an additional angle to the problem of segment vs. address which we didn't discuss. While all segments have a *base address*, the reverse is not true: if an address is *unchecked* (e.g. created from `MemoryAddress::ofLong`) it has no backing segment and can therefore not be dereferenced.

Because of this, one could argue that the signature of the `MemoryAddress::copy` is overly general, as it allows both *checked* and *unchecked* addresses, while only clearly making sense for the former.

A way out would be again, to express `copy` in terms of segment - and, as John points out, the starting point could be to look at `System::arraycopy` - e.g. assuming that `MemorySegment` ~= array:

```
static void copy(MemorySegment src, long srcOffset, MemorySegment dst, long dstOffset, long length)
```

This is all and well; now we can only pass segments to the routine, which rules out the *unchecked* address issue. But still feels a bit redundant - after all, we can easily express offsets in terms of segment slices (which was pointed out by Paul at some point). So we can actually do this:

```
static void copy(MemorySegment src, MemorySegment dst)
```

where `dst` is assumed to be bigger (or equal) than `src`, where both starting offsets are assumed to be zero and where the length of the copy is assumed to be the length of the `src` segment.
Since now the signature is expressed in terms of segment, it makes sense to move the method under `MemorySegment` too.

This is nice, but I think that once we view this method as operating on segments, I think it makes sense to turn them into proper instance methods, and just do:

```
void copyTo(MemorySegment dst)
```
Can we do the same with `fill`? I think so:

```
void fill(byte value)
```
Very simple (instance) method.

And, looking a bit ahead, we will want to have a functionality equivalent to `Arrays.mismatch` - which we can also express as an instance method:

```
int mismatch(MemorySegment that)
```

So my recommendation would be for:

1. moving all methods under `MemorySegment`
2. express all signatures in terms of `MemorySegment`
3. make them instance methods (rather than static)

This should lead to a more self-consistent, and discoverable API. Thoughts?

ChrisHegarty:
@mcimadamore I really like this suggestion (instance methods on MemorySegment). Especially given unchecked memory addresses, your proposal to move `copy` to `MemorySegment`, will result in the API point being more obviuosly correct by design ( less potential to use incorrectly ). And as you mention, we probably want to get this right now, since `mismatch` is next on my todo list ;-) ( And mismatch will not work on unchecked addresses )

This design assumes slices are relatively cheap, given that more slicing will be needed to interact with API points, but that should probably be ok.

PaulSandoz:
Shifting over `copy` to an instance method on `MemorySegment` so it can be with its friend `fill`, is i think the right move. Other friends can join later.

For arrays we never really had any choice (static vs. instance), although i still hold out hope someday we can make arrays implement a richer array'ish interface.

rose00:
As you might guess from my leading comments, I very much like the idea of
putting side-effecting fill and copy methods onto MS, not statics as in ju.Arrays.
Using instance methods instead of static methods helps the reader make a
mental model of "where it all happens" — in the receiver of the method.

The problem with static methods is all the arguments "look alike", and so
there's a perennial problem of deciding which is the source and which is
the sink.  This is not an issue with `fill` since you know the scalar is the
source, but it's a big problem with `copy` (cf. `System.arraycopy`).

If a MS has a utility method for sending data somewhere else, it shouldn't
be called `copy` but rather `copyRange` (as in ju.Arrays) or `copyOut`
or `copyTo`.  I have a very mild preference for `copyFrom` for the normal
polarity of a copy, which copies into the receiver, but that would tend to
drag in `fillFrom` or `fillWith`, which looks like overkill.

My $0.02.

ChrisHegarty:
The instance method ( for `fill` ) feels much more intuitive ( as can be seem from the test updates ). The moving and renaming of copy will be done in a subsequent PR.

PR #163, opened 2020-05-13
Simplify MemoryScope implementation to use StampedLock

mcimadamore:
I've been playing around a bit and discussed this offline with @JornVernee too. A realization I had some time ago is that what we're trying to do here is a read/write lock, where multiple acquire can occur at the same time, but where close of the root should be exclusive. I think all the machinery around `CLOSING` is essentially due to the code trying to mimic that pattern.

This patch simplifies over the status quo, by using a `StampedLock` instead of managing races manually through a `CLOSING` state. The tweaked implementation also adds back the atomicity w.r.t. close() vs. dup() which was seeked in a previous attempt (see https://git.openjdk.java.net/panama-foreign/pull/160). Also, since now there some guarantee that no acquire can take place while a close is also taking place (and pending acquires will be invalidated - using the optimistic read logic), then we can just use a single long adder instead of two.

The state is also simplified, since we just need a boolean flag.

I ran the `ParallelSum` benchmark and could not spot any obvious regression compared to the previous code.

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/163/head:pull/163`
`$ git checkout pull/163`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
@@ -29,6 +29,7 @@ package jdk.internal.foreign;
 import java.lang.invoke.MethodHandles;
 import java.lang.invoke.VarHandle;
 import java.util.concurrent.atomic.LongAdder;
+import java.util.concurrent.locks.StampedLock;
 
 /**
  * This class manages the temporal bounds associated with a memory segment. A scope has a liveness bit, which is updated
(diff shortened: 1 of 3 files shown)


src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
```diff
@@ -127,36 +124,50 @@ final boolean isAliveThreadSafe() {
...
-                if (state == STATE_CLOSED) {
-                    releases.increment();
-                    throw new IllegalStateException("This scope is already closed");
+            //try to optimistically acquire the lock
+            long stamp = lock.tryOptimisticRead();
+            try {
+                for (; ; stamp = lock.readLock()) {
+                    if (stamp == 0L)
+                        continue;
+                    checkAliveConfined(); // plain read is enough here (either successful optimistic read, or ful read lock)
```
PaulSandoz, line 154: s/ful/full

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
```diff
@@ -127,36 +124,50 @@ final boolean isAliveThreadSafe() {
...
+
+                    // increment acquires
+                    acquired.increment();
+                    // did a call to close() occurred since we acquired the lock?
```
PaulSandoz, line 158: s/occurred/occur

PR #167, opened 2020-05-15
Move "owner" field and thread-confinement checks to MemoryScope

plevart:
Now MemoryScope is simplified, I re-based this change and am opening this PR on top. Currently MemorySegment is encapsulating thread-confinement logic and state (owner field) while MemoryScope is encapsulating temporal-confinement logic and state. But the interplay between the two must be very carefully caried out (for example, close() or dup() on child scopes may only be called in owner thread). By moving the thread-confinement logic and state to MemoryScope, I think we get better encapsulation as all MemoryScope methods become "safe" - some can still be called in owner thread only, but failing to do so throws IllegalSateException instead of exhibiting undefined behavior.

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/167/head:pull/167`
`$ git checkout pull/167`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
@@ -92,7 +92,7 @@ abstract class MemoryScope {
     }
 
     private final Thread owner;
-    boolean closed; // = false
+    private boolean closed; // = false
     private static final VarHandle CLOSED;
 
     static {
@@ -188,20 +188,18 @@ abstract class MemoryScope {
             if (owner != Thread.currentThread()) {
                 throw new IllegalStateException("Attempted access outside owning thread");
             }
-            checkAliveConfined();
+            checkAliveConfined(this);
         }
     }
 
     /**
      * Checks that this scope is still alive.
-     * This method is a MemoryScope internal API and is package-private only
-     * as an implementation detail. Not for direct consumption from other classes.
      *
      * @throws IllegalStateException if this scope is already closed
      */
     @ForceInline
-    final void checkAliveConfined() {
-        if (closed) {
+    private static void checkAliveConfined(MemoryScope scope) {
+        if (scope.closed) {
             throw new IllegalStateException("This scope is already closed");
         }
     }
@@ -226,7 +224,7 @@ abstract class MemoryScope {
                 for (; ; stamp = lock.readLock()) {
                     if (stamp == 0L)
                         continue;
-                    checkAliveConfined(); // plain read is enough here (either successful optimistic read, or full read lock)
+                    checkAliveConfined(this); // plain read is enough here (either successful optimistic read, or full read lock)
 
                     // increment acquires
                     acquired.increment();
@@ -267,13 +265,13 @@ abstract class MemoryScope {
             // enter critical section - no acquires are possible past this point
             long stamp = lock.writeLock();
             try {
-                checkAliveConfined(); // plain read is enough here (full write lock)
+                checkAliveConfined(this); // plain read is enough here (full write lock)
                 // check for absence of active acquired children
                 if (acquired.sum() > 0) {
                     throw new IllegalStateException("Cannot close this scope as it has active acquired children");
                 }
                 // now that we made sure there's no active acquired children, we can mark scope as closed
-                closed = true; // plain write is enough here (full write lock)
+                CLOSED.set(this, true); // plain write is enough here (full write lock)
             } finally {
                 // leave critical section
                 lock.unlockWrite(stamp);
@@ -303,7 +301,7 @@ abstract class MemoryScope {
             @Override
             void close() {
                 checkValidState(); // child scope is always checked
-                closed = true;
+                CLOSED.set(this, true);
                 // following acts as a volatile write after plain write above so
                 // plain write gets flushed too (which is important for isAliveThreadSafe())
                 Root.this.acquired.decrement();
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
@@ -265,7 +265,7 @@ abstract class MemoryScope {
             // enter critical section - no acquires are possible past this point
             long stamp = lock.writeLock();
             try {
-                checkAliveConfined(this); // plain read is enough here (full write lock)
+                checkValidState(); // plain read is enough here (full write lock)
                 // check for absence of active acquired children
                 if (acquired.sum() > 0) {
                     throw new IllegalStateException("Cannot close this scope as it has active acquired children");
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
@@ -26,38 +26,73 @@
 
 package jdk.internal.foreign;
 
+import jdk.internal.vm.annotation.ForceInline;
+
 import java.lang.invoke.MethodHandles;
 import java.lang.invoke.VarHandle;
+import java.util.Objects;
 import java.util.concurrent.atomic.LongAdder;
 import java.util.concurrent.locks.StampedLock;
 
 /**
- * This class manages the temporal bounds associated with a memory segment. A scope has a liveness bit, which is updated
- * when the scope is closed (this operation is triggered by {@link AbstractMemorySegmentImpl#close()}). Furthermore,
- * a scope is either root scope ({@link #create(Object, Runnable) created} when memory segment is allocated) or child scope
- * ({@link #acquire() acquired} from root scope). When a child scope is acquired from another child scope, it is actually
- * acquired from the root scope. There is only a single level of children. All child scopes are peers.
- * A child scope can be {@link #close() closed} at any time, but root scope can only be closed after all its children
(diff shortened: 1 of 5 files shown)


src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
```diff
@@ -26,38 +26,73 @@
...
+     * that all methods may be called in any thread and that {@link #checkValidState()}
+     * does not check for temporal bounds.
+     *
+     * @param ref           an optional reference to an instance that needs to be kept reachable
+     * @param cleanupAction an optional cleanup action to be executed when returned scope is closed
+     * @param owner         the desired owner thread. If {@code owner == null},
+     *                      the returned scope is <em>not</em> thread-confined and not checked.
+     * @return a root MemoryScope
+     */
+    static MemoryScope createUnchecked(Object ref, Runnable cleanupAction, Thread owner) {
```
mcimadamore, line 90: For some (merely stylistic) reason, I'd prefer to see `owner` coming first in the parameter list

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
```diff
@@ -26,38 +26,73 @@
...
+        return new Root(owner, ref, cleanupAction);
     }
 
-    boolean closed = false;
+    private final Thread owner;
+    boolean closed; // = false
```
mcimadamore, line 95: Is there any real advantage in not having the initialization here? I mean, I get we probably save one putfield - but I would say that if the code is hot, C2 is probably able to detect that the store is useless? I kind of prefer the code to be self-evident rather then to use comments, where possible, of course. I've seen in the past cases where seemingly innocuous redundant initializer subtly changed happens-before order; but this shouldn't be the case here?

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
```diff
@@ -85,67 +123,104 @@ private MemoryScope() {
...
     /**
-     * Duplicates this scope and {@link #close() closes} it. If this is a root scope,
-     * new root scope is returned. If this is a child scope, new child scope is returned.
-     * This method may only be called in "owner" thread.
-     * The returned instance may be published unsafely to and used in any thread, but methods that explicitly state that
-     * they may only be called in "owner" thread, must strictly be called in single thread that has been selected to be the
-     * "owner" thread.
+     * Duplicates this scope with given new "owner" thread and {@link #close() closes} it.
+     * If this is a root scope, new root scope is returned, this root scope is closed but
+     * eventual cleanup action is not executed yet - it is inherited by duped scope.
```
mcimadamore, line 139:
```suggestion
     * If this is a root scope, a new root scope is returned; this root scope is closed, but
     * without executing the cleanup action, which is instead transferred to the duped scope.
```

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
```diff
@@ -85,67 +123,104 @@ private MemoryScope() {
...
+     * If this is a child scope, new child scope is returned.
+     * This method may only be called in "owner" thread of this scope unless the
```
mcimadamore, line 141:
```suggestion
     * If this is a child scope, a new child scope is returned.
     * This method may only be called in the "owner" thread of this scope unless the
```

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
```diff
@@ -85,67 +123,104 @@ private MemoryScope() {
...
     /**
      * Checks that this scope is still alive.
-     * This method may only be called in "owner" thread.
+     * This method is a MemoryScope internal API and is package-private only
+     * as an implementation detail. Not for direct consumption from other classes.
      *
      * @throws IllegalStateException if this scope is already closed
      */
+    @ForceInline
     final void checkAliveConfined() {
```
mcimadamore, line 203: Consider making this private

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
```diff
@@ -171,18 +246,19 @@ MemoryScope acquire() {
         }
 
         @Override
-        MemoryScope dup() { // always called in owner thread
-            return closeOrDup(false);
+        MemoryScope dup(Thread owner) {
+            Objects.requireNonNull(owner, "owner");
```
mcimadamore, line 250: Uhm - not sure about this. Let's say that you have an unchecked segment - you start off unconfined - but then, after some processing, you want to kill the segment and obtained a confined view. Seems a legitimate use case? Although I agree that, in that case, we'd need at least some sort of atomic handshake on the closed bit, to ensure the update either fails or succeeds across multiple racy threads.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
```diff
@@ -171,18 +246,19 @@ MemoryScope acquire() {
...
 
         @Override
-        void close() { // always called in owner thread
-            closeOrDup(true);
+        void close() {
+            closeOrDup(null);
         }
 
-        private MemoryScope closeOrDup(boolean close) {
+        private MemoryScope closeOrDup(Thread newOwner) {
```
mcimadamore, line 259: I think I don't like this way of sharing the code. Since we're in cleanup mode, I think the code would be more readable by having two separate methods `close` and `dup` which both uses the same underlying logic for flipping the liveness bit (using the lock).

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
mcimadamore: > So maybe just rename the parameter to newOwner ?

Yes please

mcimadamore:
> So maybe just remove the comment?

No, leave it there - I'm fine with the explanation.

mcimadamore:
> ((MemoryScope)this).checkAliveConfined();
> 
> Do you prefer such call gymnastics? If this class gets used elsewhere
> then maybe it will be required...

Ugh - I missed subclass usages. Perhaps a private static routine in the toplevel class. But it's also ok if you leave it like that.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
```diff
@@ -246,19 +246,24 @@ MemoryScope acquire() {
...
+            // pre-allocate duped scope so we don't get OOME later and be left with this scope closed
+            var duped = new Root(newOwner, ref, cleanupAction);
+            justClose();
+            return duped;
         }
 
         @Override
         void close() {
-            closeOrDup(null);
+            justClose();
```
mcimadamore, line 259: This is exactly what I had in mind - I think the code is cleaner. Perhaps rename `justClose` to `doClose` (and make `doClose` private) - but that's minor - I'll leave that up to you.

plevart: I think justClose() tells more that doClose() and private it is.

plevart:
I hid checkAliveConfined (by making it private static) and also the closed field which has the same poblem of being access from subclass if it is private. But there's no such problem if it is accessed via VarHandle. Would that be OK?

plevart:
I ran the tests and the TestSegments.testOpOutsideConfinement(close) fails with this change. The fix follows.

mcimadamore:
> I ran the tests and the TestSegments.testOpOutsideConfinement(close) fails with this change. The fix follows.

In the latest iteration I don't see any change that could possibly affect the test (compared to the previous iteration I mean). Are you also working on a followup fix for the test?

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
```diff
@@ -85,80 +123,115 @@ private MemoryScope() {
...
+     *                               in checked scope
+     */
+    @ForceInline
+    final void checkValidState() {
+        if (owner != null) {
+            if (owner != Thread.currentThread()) {
+                throw new IllegalStateException("Attempted access outside owning thread");
+            }
+            checkAliveConfined(this);
+        }
```
JornVernee, line 192: This changes the behavior we had before in AbstractMemorySegmentImpl::checkValidState. We should still do the liveness check even if there is no owner.

mcimadamore: > This changes the behavior we had before in AbstractMemorySegmentImpl::checkValidState. We should still do the liveness check even if there is no owner.

True - I suspect Peter would argue that such check doesn't make much sense in a racy context - but still, there's a change in behavior here.

JornVernee: IMO It's not necessarily racy, it's just that the API doesn't guarantee that there is no race, that's up to the user to do. e.g. by making sure that the thread doing the access can see an up to date liveness flag by using other synchronization.

plevart: Ok, then another change follows...

plevart:
Tests still pass after this change.

PR #169, opened 2020-05-15
Move MemoryAddress::copy

mcimadamore:
As discussed, this patch moves `MemoryAddress::copy` as an *instance* method of `MemorySegment`, namely `MemorySegment::copyFrom`.
I chose the `copyFrom` suggestion from John, since it allows us more freedom, in the future, to add overloads for different sources for the copy operation.

I've rewrorked the documentation a bit to speak about segment offsets rather than addresses.

I believe all use cases touched by this change actually lead to simpler code, which is good.

I've also added a benchmark for some of the bulk operations such as fill and copyFrom - the numbers I've got look good:

```
Benchmark             Mode  Cnt       Score       Error  Units
BulkOps.segment_copy  avgt   30  441675.750 ? 12794.881  ns/op
BulkOps.segment_fill  avgt   30  122465.908 ?  1760.711  ns/op
BulkOps.unsafe_copy   avgt   30  430153.694 ?  9055.253  ns/op
BulkOps.unsafe_fill   avgt   30  120413.079 ?  3799.484  ns/op
```

Cheers

---------
### Progress

### Reviewers
 * Chris Hegarty ([chegar](@ChrisHegarty) - no project role) ⚠️ Review applies to 3776b9e917df989ae03014b1ee5cb664f4ef317f
 * Jorn Vernee ([jvernee](@JornVernee) - Committer) ⚠️ Review applies to 3776b9e917df989ae03014b1ee5cb664f4ef317f

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/169/head:pull/169`
`$ git checkout pull/169`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
@@ -31,6 +31,7 @@ import java.nio.ByteBuffer;
 import jdk.internal.foreign.AbstractMemorySegmentImpl;
 import jdk.internal.foreign.HeapMemorySegmentImpl;
 import jdk.internal.foreign.MappedMemorySegmentImpl;
+import jdk.internal.foreign.MemoryAddressImpl;
 import jdk.internal.foreign.NativeMemorySegmentImpl;
 import jdk.internal.foreign.Utils;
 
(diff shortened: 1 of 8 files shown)


test/micro/org/openjdk/bench/jdk/incubator/foreign/BulkOps.java
```diff
@@ -0,0 +1,94 @@
+/*
+ *  Copyright (c) 2020, Oracle and/or its affiliates. All rights reserved.
+ *  DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ *  This code is free software; you can redistribute it and/or modify it
+ *  under the terms of the GNU General Public License version 2 only, as
+ *  published by the Free Software Foundation.  Oracle designates this
+ *  particular file as subject to the "Classpath" exception as provided
+ *  by Oracle in the LICENSE file that accompanied this code.
```
JornVernee, line 9: Wrong copyright
```suggestion
 *  published by the Free Software Foundation.
```

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -308,6 +309,29 @@
      */
     MemorySegment fill(byte value);
 
+    /**
+     * Perform bulk copy from given source segment to this segment. More specifically, the bytes at
```
PaulSandoz, line 313: "Performs a bulk copy..."

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -308,6 +309,29 @@
...
+     * offset {@code 0} through {@code src.byteSize() - 1} in the source segment are copied into this segment
+     * at offset {@code 0} through {@code src.byteSize() - 1}.
+     * If the source segment overlaps with this segment, then the copying is performed as if the bytes at
+     * offset {@code 0} through {@code src.byteSize() - 1} in the source segment were first copied into a
+     * temporary segment with size {@code bytes}, and then the contents of the temporary segment were copied into
+     * this segment at offset {@code 0} through {@code src.byteSize() - 1}.
+     * <p>
+     * The result of a bulk copy is unspecified if, in the uncommon case, the source segment does not overlap with
+     * this segment, but it instead refers to an overlapping regions of the same backing storage using different addresses.
```
PaulSandoz, line 322: "The result of a bulk copy is unspecified if, in the uncommon case, the source segment and this segment do not overlap, but refer to overlapping regions of the same backing storage using different addresses"

test/micro/org/openjdk/bench/jdk/incubator/foreign/BulkOps.java
```diff
@@ -0,0 +1,92 @@
...
+    static {
+        for (int i = 0 ; i < bytes.length ; i++) {
+            bytes[i] = i;
+        }
+    }
+
+    @Benchmark
+    @OutputTimeUnit(TimeUnit.NANOSECONDS)
+    public void unsafe_fill() {
+        unsafe.setMemory(unsafe_addr, ALLOC_SIZE, (byte)0);
```
PaulSandoz, line 72: Out of an abundance of caution you might wanna choose a non-zero value just in case the JIT knows it's zeroing already zeroed memory (i suspect it does not).

PR #171, opened 2020-05-15
Move MemoryAddress::copy (jextract version)

mcimadamore:
This patch addresses the remaining issues with jextrct templating mentioning the old MemoryAddress:copy method.

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/171/head:pull/171`
`$ git checkout pull/171`

### DIFF
--- a/src/jdk.incubator.jextract/share/classes/jdk/incubator/jextract/tool/resources/C-X.java.template
+++ b/src/jdk.incubator.jextract/share/classes/jdk/incubator/jextract/tool/resources/C-X.java.template
@@ -23,11 +23,6 @@ public class C-X {
     private static final VarHandle handle = LAYOUT.varHandle(CARRIER);
     private static final VarHandle arrayHandle = arrayHandle(LAYOUT, CARRIER);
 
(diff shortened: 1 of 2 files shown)


src/jdk.incubator.jextract/share/classes/jdk/incubator/jextract/tool/resources/C-X.java.template
```diff
@@ -25,7 +25,7 @@ public class C-X {
 
     private static void copy(MemoryAddress addr, ${CARRIER}[] arr) {
         var heapSegment = MemorySegment.ofArray(arr);
-        MemoryAddress.copy(heapSegment.baseAddress(), addr, arr.length * sizeof());
+        addr.segment().copyFrom(heapSegment);
```
JornVernee, line 28: The behavior is not necessarily the same if `addr` is not the base address of the segment. I think the right move here is to change the parameter from MemoryAddress to MemorySegment (same for Cstring).

PR #173, opened 2020-05-16
Add unsigned adapter handles

ChrisHegarty:
Hi,

As part of feedback on the Foreign Memory API (when experimenting with its usage internally in the JDK), a small number of potential usability enhancements could be made to the API. This is the third such.

  MemoryHandles::asUnsigned

When dealing with _unsigned_ native data types it is often convenient to model them as
wider _signed_ Java primitives (with appropriate magnitude checks). For example, it is convenient to model an _unsigned short_ as a Java _int_ to avoid dealing with negative values, which would be the case if modeled as a Java short. We do this all the time in the JDK implementation, and it even bleeds into the APIs sometimes.

To model this, I found myself reaching for `MemoryHandles::filterValue` to do the narrowing and widening, so that my layout class could expose the set of memory handles with appropriate carrier types that the higher-level java code was expecting. This worked fine, but Maurizio pointed out that this is something that the API should probably expose at a slightly higher level. Writing these widening and narrowing adapters using _filterValue_ is not a lot of code, but easy to make mistakes and a place for bugs to hide.

A single static adapter factory, `MemoryHandles::asUnsigned(VarHandle target, Class<?> adaptedType)`, (thanks Maurizio for the suggestion) would be sufficient to provide such functionality. It adapts a target var handle by narrowing incoming values and widening outgoing values, to and from the given type, respectively. When calling _set_ on the resulting var handle, the incoming value (of type adaptedType) is converted by a narrowing primitive conversion and then passed to the target var handle. Conversely, when calling _get_ on the resulting var handle, the returned value obtained from the target var handle is converted by an unsigned widening conversion before being returned to the caller.

We don't necessarily need to, or can, support all combinations, but there seems to be a sweet spot where Java longs and ints can be used to model _unsigned char_, _unsigned short_, and _unsigned int_, [1] which covers the majority of the use-cases. This also seems to align nicely with the primitive wrapper unsigned widening methods, e.g. `Byte::toUnsignedInt`, `Byte::toUnsignedLong`, etc.

I started this discussion as a PR so that hopefully the code changes can help convey the idea and inform readers.

Comments welcome.

[1] https://cr.openjdk.java.net/~chegar/foreign/asUnsigned.pdf

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/173/head:pull/173`
`$ git checkout pull/173`


JornVernee:
I don't think the source code generation is really needed. It seems like it should be enough to have a set of TO/FROM MethodHandles for each adaptation, and then have the if/else in `asUnsigned` pick the right pair to pass to `filterValue`.

Also, the whole implementation could be put in the MemoryHandles class. We do this for `asAddressVarHandle` as well. No need to put things in j.l.i IMO.

ChrisHegarty:
Thanks @JornVernee and @mcimadamore for your comments.

The templating was cute when I started, but as rows in the table were eliminated it no longer carries its own weight.

This update removes the templating. I decided to keep the implementation in its own class for now, so that we can see if we really want to merge it into MemoryHandles. Most of the implementation is not really sharable with other parts of MemoryHandles.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/UnsignedAdapters.java
```diff
@@ -0,0 +1,159 @@
...
+            throw new InternalError("expected byte carrier, but got: " + target.varType());
+        return MemoryHandles.filterValue(target, LONG_TO_BYTE, BYTE_TO_UNSIGNED_LONG);
+    }
+
+    // long to short
+    private static VarHandle longToUnsignedShort(VarHandle target) {
+        if (target.varType() != short.class)
+            throw new InternalError("expected byte carrier, but got: " + target.varType());
+        return MemoryHandles.filterValue(target, LONG_TO_SHORT, SHORT_TO_UNSIGNED_LONG);
+    }
```
mcimadamore, line 124: I'm not super sure of the added value of these - the check is essentially a no op, since you already have a switch where, based on the var handle carrier, you decide which method to call. So perhaps we can just inline the right MH filter inside the switch above?

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/UnsignedAdapters.java
```diff
@@ -0,0 +1,159 @@
...
+
+    private static void checkNarrowable(Class<?> type) {
+        if (!(type == int.class || type == long.class)) {
+            throw newIllegalArgumentException("illegal adapter type", type.getSimpleName());
+        }
+    }
+
+    private static void checkTargetWiderThanCarrier(Class<?> carrier, Class<?> target) {
+        if (Wrapper.forPrimitiveType(target).bitWidth() <= Wrapper.forPrimitiveType(carrier).bitWidth()) {
```
mcimadamore, line 146: Note that MemoryHandles has already a routine to retrieve carrier size - maybe there's some reuse possible there

ChrisHegarty:
I am going to review and update the set of values in the test's data providers.

PR #175, opened 2020-05-18
Add benchmark to measure performance of VH adapters

mcimadamore:
This patch adds a straightforward benchmark to measure performances of var handle adapters.

The benchmark is set up to read values from an int array, both with a var handle(obtained from `MethodHandles`), an associated `MethodHandle` (derived from the first var handle) and with a segment-based, memory access var handle.

Then the same test is repeated, but with an extra adaptation step inserted in the middle - rather than reading `int` values directly, the adapter turns `ints`
into instances of `IntBox` and the loop code doing the sum converts them back into ints.

Numbers are extremeluy solid on my machine:

```
TestAdaptVarHandles.mh_box_loop       avgt   30  0.306 ? 0.009  ms/op
TestAdaptVarHandles.mh_loop           avgt   30  0.297 ? 0.009  ms/op
TestAdaptVarHandles.segment_box_loop  avgt   30  0.308 ? 0.009  ms/op
TestAdaptVarHandles.segment_loop      avgt   30  0.307 ? 0.008  ms/op
TestAdaptVarHandles.vh_box_loop       avgt   30  0.296 ? 0.005  ms/op
TestAdaptVarHandles.vh_loop           avgt   30  0.291 ? 0.003  ms/op
```

I thought it would have been nice to add this to our benchmark suites since we do not have anything that tests VH adaptation directly.

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/175/head:pull/175`
`$ git checkout pull/175`

### DIFF
--- a/test/micro/org/openjdk/bench/jdk/incubator/foreign/TestAdaptVarHandles.java
+++ b/test/micro/org/openjdk/bench/jdk/incubator/foreign/TestAdaptVarHandles.java
@@ -1,0 +1,161 @@
+/*
+ * Copyright (c) 2020 Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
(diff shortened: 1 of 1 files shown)


JornVernee:
With the adapter creating IntBox, this benchmark is also testing whether C2 can eliminate the allocation of the box.

If the goal is to measure an adapted VarHandle against a non-adapted baseline VarHandle, I think it makes sense to have the adapters be the identity function, since then the _only_ difference is whether an adapter is present or not.

WDYT?

mcimadamore:
I guess my (non stated) goal was to benchmark non trivial adaptation where users could create their own carriers on top of the foreign memory access API. I wanted to make sure that the VH machinery didn't add extra cost when doing so.

PR #179, opened 2020-05-20
8245459: Add support for complex filter value var handle adaptation

mcimadamore:
It is sometimes necessary to apply a carrier transform on a VarHandle that is expressed in more than just a pair of unary function (e.g. from `A->B` and `B->A`). Sometimes the transform needs to act on some bit of external state.

It would be nice if `MemoryHandles::filterValue` would be enhanced to accept all functions of the kind:

`(C... A) ->B`
`(C... B) -> A`

So that the resulting VarHandle will apply carrier transform from A to B and will also add extra coordinate types (C...) which can be used by the transform filters. 

To enhance `MemoryHandles::filterValue` this way regular MH adapters are not enough; while we can implement setters, by combining the filters with `MethodHandles::collectArguments` there is no way to implement the getters - since that would require a method handle transform which adapts a return value *and* adds some extra parameters to the adapter handle (which are then forwarded to the filter).

This patch adds the missing method handle primitive to do that; for now it's a package private method, namely `MethodHandles::collectReturnValue`. If people think that this could be useful, we can evaluate whether to open this up for method handle combinator API as well.

This required to add a new kind of lambda form using the `LambdaFormEditor`, which was a bit tricky to get right (we currently have no transforms which operates on both arguments *and* return values, which makes it tricky).

The implementation of `MemoryHandles::filterValue` is also a bit more complex, since if there are additional coordinates added by collecting arguments/return values, such additional coordinates might be added multiple times, so we need some steps to 'join' the extra coordinates (with a call to `MethodHandle::permuteArguments`) so that extra coordinate show up only once in the final adapter.

If the filter functions used for adapting are just unary functions, then the implementation falls back to the simpler behavior we had before, and the permutation step is skipped (in such cases, `collectArguments` == `filterArguments` and `collectReturnValue` == `filterReturnValue`).

---------
### Progress

### Issue
 * [JDK-8245459](https://bugs.openjdk.java.net/browse/JDK-8245459): Add support for complex filter value var handle adaptation

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/179/head:pull/179`
`$ git checkout pull/179`

### DIFF
--- a/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
+++ b/src/java.base/share/classes/java/lang/invoke/MethodHandles.java
@@ -5475,6 +5475,40 @@ System.out.println((int) f0.invokeExact("x", "y")); // 2
     }
 
     /**
+     * Filter the return value of a target method handle with a filter function. The filter function is
+     * applied to the return value of the original handle; if the filter specifies more than one parameters,
+     * then any remaining parameter is appended to the adapter handle. In other words, the adaptation works
+     * as follows:
+     * <blockquote><pre>{@code
+     * T target(A...)
+     * V filter(B... , T)
+     * V adapter(A... a, B... b) {
+     *     T t = target(a...);
+     *     return filter(b..., t);
+     * }</pre></blockquote>
+     * <p>
+     * If the filter handle is a unary function, then this method behaves like {@link #filterReturnValue(MethodHandle, MethodHandle)}.
+     *
+     * @param target the target method handle
+     * @param filter the filter method handle
+     * @return the adapter method handle
+     */
+    /* package */ static MethodHandle collectReturnValue(MethodHandle target, MethodHandle filter) {
+        MethodType targetType = target.type();
+        MethodType filterType = filter.type();
+        BoundMethodHandle result = target.rebind();
+        LambdaForm lform = result.editor().collectReturnValueForm(filterType.basicType());
+        MethodType newType = targetType.changeReturnType(filterType.returnType());
+        if (filterType.parameterList().size() > 1) {
+            for (int i = 0 ; i < filterType.parameterList().size() - 1 ; i++) {
+                newType = newType.appendParameterTypes(filterType.parameterType(i));
+            }
+        }
+        result = result.copyWithExtendL(newType, lform, filter);
+        return result;
+    }
+
+    /**
      * Adapts a target method handle by pre-processing
      * some of its arguments, and then calling the target with
      * the result of the pre-processing, inserted into the original
--- a/src/java.base/share/classes/java/lang/invoke/LambdaFormEditor.java
+++ b/src/java.base/share/classes/java/lang/invoke/LambdaFormEditor.java
@@ -922,6 +922,48 @@ class LambdaFormEditor {
         return putInCache(key, form);
     }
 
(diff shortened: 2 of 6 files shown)


src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -5474,6 +5474,40 @@ private static void filterReturnValueChecks(MethodType targetType, MethodType fi
             throw newIllegalArgumentException("target and filter types do not match", targetType, filterType);
     }
 
+    /**
+     * Filter the return value of a target method handle with a filter function. The filter function is
```
PaulSandoz, line 5478: Suggested improvement to JavaDoc:
```
Adapts a target method handle by post-processing
its return value and additional arguments (if any) with a filter 
(another method handle).
The result of the filter is returned from the adapter.

If the target returns a value, the filter must accept that value as
as the trailing argument, with the additional arguments proceeding that.
If the target returns void, the filter must accept only the additional arguments.

The return type of the filter
replaces the return type of the target
in the resulting adapted method handle.

The trailing argument type of the filter (if any) must be identical to the
return type of the target.

The additional argument types of the filter (if any) are appended (in order) 
to the argument types of the target method handle and become the argument
types of the resulting adapted method handle.

Here is pseudocode for the resulting adapter. In the code,
{@code T}/{@code t} represent the result type and value of the
{@code target}; {@code V}, the result type of the {@code filter};
{@code A}/{@code a}, the types and values of the parameters and arguments
of the {@code target} as well as the resulting adapter.
{@code B}/{@code b}, the types and values of the additional parameters and arguments
of the {@code filter} as well as appended to the resulting adapter.
...
```

src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -5474,6 +5474,40 @@ private static void filterReturnValueChecks(MethodType targetType, MethodType fi
...
+     *     return filter(b..., t);
+     * }</pre></blockquote>
+     * <p>
+     * If the filter handle is a unary function, then this method behaves like {@link #filterReturnValue(MethodHandle, MethodHandle)}.
+     *
+     * @param target the target method handle
+     * @param filter the filter method handle
+     * @return the adapter method handle
+     */
+    /* package */ static MethodHandle collectReturnValue(MethodHandle target, MethodHandle filter) {
```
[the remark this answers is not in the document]
mcimadamore: yeah - ideally I'd need these checks - in reality the checks are performed on the VH side for now, and this is not exposed publicly, but I can add all required checks.

As for void, yes, for consistency with filterReturnValue we should eventually support that too - but it wasn't the goal of this patch to provide a full blown MH adapter; I have implemented what was necessary for the VH adaptation. I can add the other parts (but would prefer to deal with that separately).

src/java.base/share/classes/java/lang/invoke/MethodHandles.java
mcimadamore: Thanks - I'll see how much work it is to add support of the void case - hopefully it won't be too bad.

mcimadamore:
Pushed as is for now - will consider completing the work on the MH side as a separate PR. Thanks!

PR #180, opened 2020-05-20
Add MemorySegment::mismatch

ChrisHegarty:
Hi,

As part of feedback on the Foreign Memory API (when experimenting with its usage internally in the JDK), a small number of potential usability enhancements could be made to the API. This is the fourth such, and last on my current todo list.

MemorySegment::mismatch

The mismatch semantic is very useful for building equality and comparison logic on top of segments. I found that I needed such when modeling and comparing native socket address in the JDK implementation. It is possible to write your own, but requires a non-trivial amount of not-trivial code - to do it right! I also think that we can provide a more efficient implementation building on top of the JDK's internal mismatch support.

I still need to do some perf testing and add a micro-benchmake ( Maurizio suggested possibly amending TestBulkOps ). There is also the question about possibly improving the JDK's internal implementation to work on long sizes (which could be done separately). For now, I just want to share the idea, along with the proposed specification and initial implementation.

Comments welcome.

---------
### Progress

### Reviewers
 * Paul Sandoz ([psandoz](@PaulSandoz) - Committer) ⚠️ Review applies to caf136fde6fb6a97cbb47ea6fd4c47208280d81f

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/180/head:pull/180`
`$ git checkout pull/180`

### DIFF
--- a/src/java.base/share/classes/jdk/internal/util/ArraysSupport.java
+++ b/src/java.base/share/classes/jdk/internal/util/ArraysSupport.java
@@ -171,12 +171,7 @@ public class ArraysSupport {
         long remaining = length;
         int i ;
         while (remaining > 7) {
-            int size;
-            if (remaining > Integer.MAX_VALUE) {
-                size = Integer.MAX_VALUE;
-            } else {
-                size = (int) remaining;
-            }
+            int size = (int) Math.min(Integer.MAX_VALUE, remaining);
             i = vectorizedMismatch(
                     a, aOffset + off,
                     b, bOffset + off,
--- a/src/java.base/share/classes/jdk/internal/util/ArraysSupport.java
+++ b/src/java.base/share/classes/jdk/internal/util/ArraysSupport.java
@@ -160,6 +160,37 @@ public class ArraysSupport {
         }
     }
 
+    /**
+     * Mismatch over long lengths.
+     */
+    public static long vectorizedMismatchLarge(Object a, long aOffset,
+                                               Object b, long bOffset,
+                                               long length,
+                                               int log2ArrayIndexScale) {
+        long off = 0;
+        long remaining = length;
+        int i ;
+        while (remaining > 7) {
+            int size;
+            if (remaining > Integer.MAX_VALUE) {
+                size = Integer.MAX_VALUE;
+            } else {
+                size = (int) remaining;
+            }
+            i = vectorizedMismatch(
+                    a, aOffset + off,
+                    b, bOffset + off,
+                    size, log2ArrayIndexScale);
+            if (i >= 0)
+                return off + i;
+
+            i = size - ~i;
+            off += i;
+            remaining -= i;
+        }
+        return ~off;
+    }
+
     // Booleans
     // Each boolean element takes up one byte
 
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
@@ -344,7 +344,8 @@ public interface MemorySegment extends AutoCloseable {
      * between the two segments at that offset within the respective segments.
      * If one segment is a proper prefix of the other then the returned offset is
      * the smaller of the segment sizes, and it follows that the offset is only
-     * valid for the larger segment. Otherwise, there is no mismatch.
+     * valid for the larger segment. Otherwise, there is no mismatch and {@code
+     * -1} is returned.
      *
      * @param other the segment to be tested for a mismatch with this segment
      * @return the relative offset, in bytes, of the first mismatch between this
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
@@ -333,6 +333,31 @@ public interface MemorySegment extends AutoCloseable {
     void copyFrom(MemorySegment src);
 
     /**
+     * Finds and returns the offset, in bytes, of the first mismatch between
+     * this segment and a given other segment. The offset is relative to the
+     * {@link #baseAddress() base address} of each segment and will be in the
+     * range of 0 (inclusive) up to the {@link #byteSize() size} (in bytes) of
+     * the smaller memory segment (exclusive).
+     * <p>
+     * If the two segments share a common prefix then the returned offset is
+     * the length of the common prefix and it follows that there is a mismatch
+     * between the two segments at that offset within the respective segments.
+     * If one segment is a proper prefix of the other then the returned offset is
+     * the smaller of the segment sizes, and it follows that the offset is only
+     * valid for the larger segment. Otherwise, there is no mismatch.
+     *
+     * @param other the segment to be tested for a mismatch with this segment
+     * @return the relative offset, in bytes, of the first mismatch between this
+     * and the given other segment, otherwise -1 if no mismatch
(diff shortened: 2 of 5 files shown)


src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -332,6 +332,31 @@
...
+     * {@link #baseAddress() base address} of each segment and will be in the
+     * range of 0 (inclusive) up to the {@link #byteSize() size} (in bytes) of
+     * the smaller memory segment (exclusive).
+     * <p>
+     * If the two segments share a common prefix then the returned offset is
+     * the length of the common prefix and it follows that there is a mismatch
+     * between the two segments at that offset within the respective segments.
+     * If one segment is a proper prefix of the other then the returned offset is
+     * the smaller of the segment sizes, and it follows that the offset is only
+     * valid for the larger segment. Otherwise, there is no mismatch.
```
mcimadamore, line 347: there is no mismatch and the returned value is (I think you say `-1` below, perhaps would be good to state that here also).

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
```diff
@@ -131,6 +132,53 @@ public void copyFrom(MemorySegment src) {
...
     }
 
+    @Override
+    public long mismatch(MemorySegment other) {
+        AbstractMemorySegmentImpl that = (AbstractMemorySegmentImpl)other;
+        final long thisSize = this.byteSize();
+        final long thatSize = that.byteSize();
+        final long minSize = Math.min(thisSize, thatSize);
+
+        this.checkRange(0, minSize, false);
```
mcimadamore, line 142: This is of course correct but also a bit redundant (there's a similar issue, although to a smaller extent in `copyFrom`). Let's see how performance numbers are - basically this code can be simplified by omitting the bound check on both segments (since you already picked a size that must work on both) - so that we just need confinement + read access check.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
```diff
@@ -131,6 +132,53 @@ public void copyFrom(MemorySegment src) {
...
+                return off + i;
+            }
+            i = size - ~i;
+
+            off += i;
+            remaining -= i;
+        }
+
+        for (; off < minSize; off++) {
+            if (UNSAFE.getByte(this.base(), this.min() + off) != UNSAFE.getByte(that.base(), that.min() + off)) {
```
mcimadamore, line 175: This code could be simplified/rewritten to use `MemoryAddress' and VH, instead of unsafe access with object/offset addressing. E.g. you could maintain a `MemoryAddress` local variable in the loop instead of the `offset` and keep increasing that address on each iteration of `ArraySupport::vectorizedMismatch`. Then, when you get out of the loop, the address already points at the base of the region to compare, and a simple for loop with an indexed VH should do the job.

test/jdk/java/foreign/TestMismatch.java
```diff
@@ -0,0 +1,236 @@
...
+                assertEquals(s1.mismatch(s2), expectedMismatchOffset);
+                assertEquals(s2.mismatch(s1), expectedMismatchOffset);
+            } else {
+                assert s1.byteSize() < s2.byteSize();
+                var off = Math.min(s1.byteSize(), expectedMismatchOffset);
+                assertEquals(s1.mismatch(s2), off);  // proper prefix
+                assertEquals(s2.mismatch(s1), off);
+            }
+        }
+    }
```
mcimadamore, line 99: How important is it that these tests operate on slices? Looking at the test code, it could have worked equally well if the input parameters were just two sizes, and then you did an explicit allocation (or maybe also receive a segment factory from the provider, so that you can test different segment kinds).

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
ChrisHegarty: Correct. Added, Otherwise, there is no mismatch "and {@code  -1} is returned".

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
ChrisHegarty: Yeah, I did think this too at the time, but then forgot. How about I just add a private method `checkStateAndAccess`.

mcimadamore: For now I think we can keep it as is, as it's cleaner and reuses existing abstractions. If this becomes a performance problem going forward we can revisit.

test/jdk/java/foreign/TestMismatch.java
ChrisHegarty: Originally I had a version of the test that did compare specific segments, but it didn't scale well to different sizes and kinds ( we need to test both above and below the 8 byte threshold ). I removed a number of slice sizes, which greatly reduces the combinations. This may be enough, or I can certainly revisit the test's structure.

src/java.base/share/classes/jdk/internal/util/ArraysSupport.java
```diff
@@ -160,6 +160,37 @@ public static int vectorizedMismatch(Object a, long aOffset,
...
+    public static long vectorizedMismatchLarge(Object a, long aOffset,
+                                               Object b, long bOffset,
+                                               long length,
+                                               int log2ArrayIndexScale) {
+        long off = 0;
+        long remaining = length;
+        int i ;
+        while (remaining > 7) {
+            int size;
+            if (remaining > Integer.MAX_VALUE) {
```
PaulSandoz, line 175: Simplify as the following?
```
int size = (int) Math.min(Integer.MAX_VALUE, remaining);
```

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
```diff
@@ -131,6 +134,42 @@ public void copyFrom(MemorySegment src) {
...
+            i = length - ~i;
+        }
+        MemoryAddress thisAddress = this.baseAddress();
+        MemoryAddress thatAddress = that.baseAddress();
+        for (; i < length; i++) {
+            if ((byte) BYTE_HANDLE.get(thisAddress, i) != (byte) BYTE_HANDLE.get(thatAddress, i)) {
+                return i;
+            }
+        }
+        return thisSize != thatSize ? length : -1;
```
mcimadamore, line 170: I guess I'm a bit confused - shouldn't this method return (as per javadoc), `-1` if there's no mismatch?  In this code path we found no mismatch, and yet, if sizes are different we return `length`, suggesting there's a mismatch. I now realize that *mismatch* is taken quite literally - e.g. no mismatch really means the two things are identical in contents and size --- which is, I realize, what `Arrays::mismatch` also does.

IMHO the javadoc of the various `mismatch` routines could use more clarity around what a *mismatch* is. But maybe that's something for another day.

ChrisHegarty:
The proposed `MemorySegment::mismatch` spec wording is aligned with similar in the buffer classes and elsewhere. If it needs clarification then we should probably do so consistently. As @mcimadamore said, "that's something for another day" ;-)

PR #184, opened 2020-05-27
8245988: Add a special VaList carrier

JornVernee:
Hi,

This patch adds a special VaList carrier to CSupport that can be used by C linker implementations to pass `va_list` arguments to upcalls and downcalls.

Currently Windows VaList and SysV VaLists are implemented, but AArach64 not yet.

The API for the Reader and Builder might be a bit strange; the limitation to int, long, and double comes from the standard argument promotions that are done in C when calling a variadic function. Since we do not know how to promote any arbitrary value, at least for now the API restricts itself to accepting only types that would not be promoted in C, leaving any needed conversion up to the user.

For reviewers; I moved CallArranger.TypeClass + classification code to the top level, since it is now shared with the VaList implementations, so that's where the CallArranger diff comes from.

Thanks,
Jorn

---------
### Progress

### Issue
 * [JDK-8245988](https://bugs.openjdk.java.net/browse/JDK-8245988): Add a special VaList carrier

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/184/head:pull/184`
`$ git checkout pull/184`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CSupport.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CSupport.java
@@ -55,31 +55,137 @@ public class CSupport {
     }
 
     /**
+     * An interface that models a C {@code va_list}.
+     *
      * Per the C specification (see C standard 6.5.2.2 Function calls - item 6),
      * arguments to variadic calls are erased by way of 'default argument promotions',
      * which erases integral types by way of integer promotion (see C standard 6.3.1.1 - item 2),
      * and which erases all {@code float} arguments to {@code double}.
      *
-     * As such, this reader interface only supports reading {@code int}, {@code double},
-     * and any other type that fits into a {@code long} (when given it's layout).
+     * As such, this interface only supports reading {@code int}, {@code double},
+     * and any other type that fits into a {@code long}.
      */
-    public interface VaList extends AutoCloseable /* permits */ {
+    public interface VaList extends AutoCloseable {
+
+        /**
+         * Reads a value into an {@code int}
+         *
+         * @param layout the layout of the value
+         * @return the value read as an {@code int}
+         */
         int readInt(MemoryLayout layout);
+
+        /**
+         * Reads a value into a {@code long}
+         *
+         * @param layout the layout of the value
+         * @return the value read as an {@code long}
+         */
         long readLong(MemoryLayout layout);
+
+        /**
+         * Reads a value into a {@code double}
+         *
+         * @param layout the layout of the value
+         * @return the value read as an {@code double}
+         */
         double readDouble(MemoryLayout layout);
+
+        /**
+         * Reads a value into a {@code MemoryAddress}
+         *
+         * @param layout the layout of the value
+         * @return the value read as an {@code MemoryAddress}
+         */
         MemoryAddress readPointer(MemoryLayout layout);
+
+        /**
+         * Reads a value into a {@code MemorySegment}
+         *
+         * @param layout the layout of the value
+         * @return the value read as an {@code MemorySegment}
+         */
         MemorySegment readStructOrUnion(MemoryLayout layout);
+
+        /**
+         * Skips a number of va arguments with the given memory layouts.
+         *
+         * @param layouts the layout of the value
+         */
         void skip(MemoryLayout...layouts);
 
+        /**
+         * A predicate used to check if this va list is alive,
+         * or in other words; if {@code close()} has been called on this
+         * va list.
+         *
+         * @return true if this va list is still alive.
+         * @see #close()
+         */
         boolean isAlive();
+
+        /**
+         * Closes this va list, releasing any resources it was using.
+         *
+         * @see #isAlive()
+         */
         void close();
+
+        /**
+         * Copies this va list.
+         *
+         * @return a copy of this va list.
+         */
         VaList copy();
 
+        /**
+         * A builder interface used to construct a va list.
+         */
         interface Builder {
+
+            /**
+             * Adds a native value represented as an {@code int} to the va list.
+             *
+             * @param layout the native layout of the value.
+             * @param value the value, represented as an {@code int}.
+             * @return this builder.
+             */
             Builder intArg(MemoryLayout layout, int value);
+
+            /**
+             * Adds a native value represented as a {@code long} to the va list.
+             *
+             * @param layout the native layout of the value.
+             * @param value the value, represented as a {@code long}.
+             * @return this builder.
+             */
             Builder longArg(MemoryLayout layout, long value);
+
+            /**
+             * Adds a native value represented as a {@code double} to the va list.
+             *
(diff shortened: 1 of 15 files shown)


JornVernee:
Reader is an attempt to allow users to re-read the same VaList. It also gives an oppertunity to declare how many arguments the VaList is expected to hold (something we don't know when constructing the VaList instance). I think moving the reader methods to VaList makes sense as well. 

I'll look into implementing a `copy()` method as well.

As discussed offline, I'll also look into adding methods to convert back and forth between a MemoryAddress from a VaList. As well as adding a benchmark that compares VaList with a manually specialized varargs down call (using `...`).

JornVernee:
Updated to address review comments:
- Merged Reader and VaList.
- Added methods to convert between a VaList and a MemoryAddress directly.
- Let SysVVaList use the underlying va_list's memory instead of have separate fields (this makes copying and converting to MemoryAddress easier/possible).
- Added javadoc to VaList and VaList.Builder (forgot to do this before).
- Added a copy() method for copying a VaList, plus some accompanying tests.
- Adding missing tests for the skip method, and fixed a bug I found with that.
- Added a benchmark that compares passing a VaList with using a specilalized varargs invocation that creates a va_list on the native side instead.

wrt the benchmark; the VaList case is currently 15x slower on Windows :( I have to do more investigation to figure out exactly why, but my guess is that this is mostly due to the need for native heap allocations when we create the VaList in Java, while the native implementation can allocate on the stack. Off course, call intrinsification will probably also have a big effect here.

Still, VaList provides easier access to native functions accepting a va_list, and for upcalls, makes passing a variable number of arguments possible in the first place.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CSupport.java
```diff
@@ -49,6 +50,163 @@ public static ForeignLinker getSystemLinker() {
...
+     */
+    public interface VaList extends AutoCloseable {
+
+        /**
+         * Reads a value into an {@code int}
+         *
+         * @param layout the layout of the value
+         * @return the value read as an {@code int}
+         */
+        int readInt(MemoryLayout layout);
```
mcimadamore, line 76: Here some confusion might arise as to what `int` means. `int` means Java `int`.
As discussed offline, it would be better to use some hint e.g. `from` to indicate that we are converting *from* a Java value to a native one.
Also, `read` for some reason is reminiscent of I/O.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CSupport.java
```diff
@@ -49,6 +50,163 @@ public static ForeignLinker getSystemLinker() {
...
+        MemoryAddress readPointer(MemoryLayout layout);
+
+        /**
+         * Reads a value into a {@code MemorySegment}
+         *
+         * @param layout the layout of the value
+         * @return the value read as an {@code MemorySegment}
+         */
+        MemorySegment readStructOrUnion(MemoryLayout layout);
+
```
mcimadamore, line 109: These two do not seem consistent with the builder methods

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CSupport.java
```diff
@@ -49,6 +50,163 @@ public static ForeignLinker getSystemLinker() {
...
+        interface Builder {
+
+            /**
+             * Adds a native value represented as an {@code int} to the va list.
+             *
+             * @param layout the native layout of the value.
+             * @param value the value, represented as an {@code int}.
+             * @return this builder.
+             */
+            Builder intArg(MemoryLayout layout, int value);
```
mcimadamore, line 170: Dual issue here - `int` again stands for Java `int`. Perthaps using `asInt` in the method name could help clarify.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/x64/sysv/SysVx64Linker.java
```diff
@@ -51,20 +54,73 @@
...
+    private static MethodType convertVaListCarriers(MethodType mt) {
+        Class<?>[] params = new Class<?>[mt.parameterCount()];
+        for (int i = 0; i < params.length; i++) {
+            Class<?> pType = mt.parameterType(i);
+            params[i] = ((pType == CSupport.VaList.class) ? SysVVaList.CARRIER : pType);
+        }
+        return MethodType.methodType(mt.returnType(), params);
+    }
+
+    private static MethodHandle unxboxVaLists(MethodType type, MethodHandle handle) {
```
mcimadamore, line 94: type here - unxbox -> unbox

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/x64/sysv/SysVx64Linker.java
```diff
@@ -51,20 +54,73 @@
...
+        handle = unxboxVaLists(type, handle);
+        return handle;
+    }
+
+    private static MethodHandle boxVaLists(MethodHandle handle) {
+        MethodType type = handle.type();
+        for (int i = 0; i < type.parameterCount(); i++) {
+            if (type.parameterType(i) == VaList.class) {
+               handle = MethodHandles.filterArguments(handle, i, MH_boxVaList);
+            }
```
mcimadamore, line 116: These looks a bit odd - in the sense that we have to hide valists from call arranger, but then add adaptation back in to make clients happy. I guess it seems as if this could be less error prone if CallArranger would also be in charge of dealing with valists.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/x64/windows/WinVaList.java
```diff
@@ -0,0 +1,239 @@
...
+                        }
+                        case STRUCT_REGISTER -> {
+                            MemorySegment slice = ms.asSlice(addr.segmentOffset(), VA_SLOT_SIZE_BYTES);
+                            slice.copyFrom(msArg);
+                        }
+                        default -> throw new IllegalStateException("Unexpected TypeClass: " + typeClass);
+                    }
+                } else {
+                    VarHandle writer = arg.varHandle();
+                    writer.set(addr, arg.value);
```
mcimadamore, line 231: From a performance perspective, both this and the reader code above are suboptimal, in two dimensions:
* the VH used is not constant
* the type of the get/set call is not exact
No need to dive in premature optimization territory now, but perhaps worth keeping in mind.

JornVernee: Thanks. I want to get the API right first, and then move on to tackling performance after that is finished, since diagnosing and solving performance problems usually takes a long time on it's own.

PR #198, opened 2020-06-09
CSupport::getSystemLinker should not be called from internal API points

mcimadamore:
Some jextract internal classes call the external getSystemLinker method, which then does a check on the `foreign.restricted` property (which is disabled when jextract runs).
Trusted internal API points should use the internal factory for the system linker instead.

---------
### Progress

### Reviewers
 * Athijegannathan Sundararajan ([sundar](@sundararajana) - Committer) ⚠️ Review applies to fa72d88f54a4730ff8a5a74547e81879d53fd39a

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/198/head:pull/198`
`$ git checkout pull/198`

### DIFF
--- a/src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/LayoutUtils.java
+++ b/src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/LayoutUtils.java
@@ -197,7 +197,7 @@ public final class LayoutUtils {
             case 8: return Primitive.Kind.Char;
             case 16: return Primitive.Kind.Short;
             case 32: return Primitive.Kind.Int;
-            case 64: return SharedUtils.getSystemLinker().name().equals(CSupport.Win64.NAME) ?
+            case 64: return abi.name().equals(CSupport.Win64.NAME) ?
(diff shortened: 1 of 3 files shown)


src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/LayoutUtils.java
```diff
@@ -197,7 +197,7 @@ static MemoryLayout getRecordLayout(Type type) {
             case 8: return Primitive.Kind.Char;
             case 16: return Primitive.Kind.Short;
             case 32: return Primitive.Kind.Int;
-            case 64: return CSupport.getSystemLinker().name().equals(CSupport.Win64.NAME) ?
+            case 64: return SharedUtils.getSystemLinker().name().equals(CSupport.Win64.NAME) ?
```
sundararajana, line 200: could use static field "abi" instead?

mcimadamore: Fixed in latest iteration

PR #199, opened 2020-06-12
8247439: NativeAllocationScope should have a way to register existing segments onto it

mcimadamore:
In extracted code, there are two cases where segments can be created outside of an allocation scope:

* structs passed returned by value
* callback segments

In these cases it is necessary for clients to manually handle the life-cycle of these segments with a separate (nested) try with resource, which makes the code harder to follow.

Jextract is currently bypassing this limitation by providing a "register" mechanism, but the provided mechanism is not very safe, since it is up to the user not to prematurely close the registered segment. 

This patch add a safer routine (called `NativeAllocationScope::claim`) which can be used to attach a segment onto an existing scope. This returns a _new_ segment with a brand _new_ memory scope (and the old segment is killed). The new segment only features mininal access modes (READ and WRITE). As such, it cannot be closed directly - it can only be closed when the allocation scope is.

Furthermore, there are some important restrictions on which segments can be attached to a scope:

* the thread owner of the segment must match that of the allocation scope (yes, now allocation scopes are confined too)
* the segment to be attached must feature the CLOSE access mode

The latter restriction is important for a number of reasons:

1. it prevents an attached segment to be re-attached somewhere else (since attached segment cannot feature CLOSE mode by definition)
2. it prevents acquired views to be attached - the lifecycle of acquired views is closely tied to that of the parent segment, and the spliterator they come from - so it doesn't seem a good idea to attach them somewhere else
3. it prevents, more generally, NativeAllocationScope::close to throw because some attached segment cannot be closed

Under the hood, NativeAllocationScope::claim uses the same logic as MemorySegment::withOwnerThread, so we know that logic is safe.

I'm open to suggestion for the method name (e.g. maybe `claimOwnership` could be more descriptive?); I'm also open to consider renaming `NativeAllocationScope` to just `NativeScope`.

---------
### Progress

### Issue
 * [JDK-8247439](https://bugs.openjdk.java.net/browse/JDK-8247439): NativeAllocationScope should have a way to register existing segments onto it

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/199/head:pull/199`
`$ git checkout pull/199`


src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/NativeAllocationScope.java
```diff
@@ -226,6 +240,25 @@ public MemoryAddress allocate(long bytesSize) {
...
+     * @param segment the segment whose ownership will be claimed by this allocation scope.
+     * @return a new, non closeable memory segment, backed by the same underlying region as {@code segment},
+     * but whose life-cycle is tied to that of this allocation scope.
+     * @throws IllegalStateException if {@code segment} is not <em>alive</em> (see {@link MemorySegment#isAlive()}).
+     * @throws NullPointerException if {@code segment == null}
+     * @throws IllegalArgumentException if {@code segment.ownerThread() != this.ownerThread()}, or if {@code segment}
+     * does not feature the {@link MemorySegment#CLOSE} access mode.
+     */
+    public abstract MemorySegment claim(MemorySegment segment);
+
```
sundararajana, line 261: I am not sure "claim" conveys the intent. Frankly I like "register" better.  In a way that name is consistent with ref   "Cleaner" API - user 'register's the reference for cleanup later.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/NativeAllocationScope.java
```diff
@@ -39,6 +39,14 @@
...
+ * <p>
+ * Allocation scopes are thread-confined (see {@link #ownerThread()}; as such, the resulting {@code MemoryAddress} instances
+ * returned by the allocation scope will be backed by memory segments confined by the same owner thread as the allocation scope.
+ * <p>
+ * To allow for more usability, it is possible for an allocation scope to claim ownership of an existing memory segment
+ * (see {@link #claim(MemorySegment)}). This might be useful to allow one or more segments which were independently
+ * created to share the same life-cycle as a given allocation scope - which in turns enables client to group all memory
+ * allocation and usage under a single <em>try-with-resources block</em>.
  */
 public abstract class NativeAllocationScope implements AutoCloseable {
```
sundararajana, line 52: Yes, NativeScope is better name - apart from being shorter name, it now does more than just allocation (manages a life time of a list of MemorySegments)

mcimadamore: Ok, I'll apply the renamings in a follow up patch.

PR #207, opened 2020-06-15
8247488: Add support for string helpers in CSupport (decoding support)

mcimadamore:
This patch add more symmetric support for charset-driven decoding of the contents of a native string.
That means we can no longer using a StringBuilder, but, instead, we have to get the byte array and create a `String` out of that using specified `Charset` instance.

Implementation-wise, following an offline discussion with Jorn, I've settled for an approach which finds string length, then does a bulk copy to a byte array. Not sure this is going to peform well on small strings. The alternative would have been to use `ByteArrayOutputStream` - but the issue there is that both `write` and `toString` methods there are `synchronized`.

In any case, I think the implementation can be made more efficient later, if we find that to be an issue.

---------
### Progress

### Issue
 * [JDK-8247488](https://bugs.openjdk.java.net/browse/JDK-8247488): Add support for string helpers in CSupport ⚠️ Title mismatch between PR and JBS.

### Reviewers
 * Henry Jen ([henryjen](@slowhog) - Committer) ⚠️ Review applies to f538ccb6dda7db8d3e5296d5b0b810a1daf56b0b

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/207/head:pull/207`
`$ git checkout pull/207`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CSupport.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CSupport.java
@@ -26,6 +26,8 @@
 package jdk.incubator.foreign;
 
 import jdk.internal.foreign.AbstractMemorySegmentImpl;
+import jdk.internal.foreign.MemoryAddressImpl;
+import jdk.internal.foreign.NativeMemorySegmentImpl;
 import jdk.internal.foreign.Utils;
 import jdk.internal.foreign.abi.SharedUtils;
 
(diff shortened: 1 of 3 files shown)


src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CSupport.java
```diff
@@ -647,11 +654,38 @@ public static MemoryAddress toCString(String str, Charset charset, NativeScope s
...
-     * Convert a null-terminated C string stored at given address into a Java string.
+     * Convert a null-terminated C string stored at given address into a Java string, using the given {@linkplain java.nio.charset.Charset charset}.
+     * <p>
+     * This method always replaces malformed-input and unmappable-character
+     * sequences with this charset's default replacement string.  The {@link
+     * java.nio.charset.CharsetDecoder} class should be used when more control
+     * over the decoding process is required.
+     * <p>
+     * This method is <em>restricted</em>. Restricted method are unsafe, and, if used incorrectly, their use might crash
+     * the JVM crash or, worse, silently result in memory corruption. Thus, clients should refrain from depending on
```
JornVernee, line 669: Seems like a type: "their use might crash the JVM crash"

PR #208, opened 2020-06-16
8247678: StdLibTest fails to create an empty VaList on Windows

JornVernee:
Hi,

This patch addresses a test failure in StdLibTest, caused by trying to allocate an empty VaList in the Windows implementation.

(After some discussion with Maurizio) this is fixed by special casing empty va lists in the implementation. VaList.Builder.build() now returns a special empty va list constant when the VaList is empty. There's also an added VaList.empty() method, to directly create an empty VaList.

To implement this I removed the getSegment methods in the VaList implementation classes, and instead rely on VaList::address for unboxing a VaList. This works better with the spacial case empty VaList. Along with that I moved some of the shared code for handling VaList in the linkers to SharedUtils.

I added tests for the empty() API to verify that the returned VaList is non-closeable, and also updated StdLibTest to close the VaLists it creates (it wasn't doing this yet).

Thanks,
Jorn

---------
### Progress

### Issue
 * [JDK-8247678](https://bugs.openjdk.java.net/browse/JDK-8247678): StdLibTest fails to create an empty VaList on Windows

### Reviewers
 * Maurizio Cimadamore ([mcimadamore](@mcimadamore) - Committer) ⚠️ Review applies to eb80703cdb22b648b02ddb8378a77b55ea595d5c

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/208/head:pull/208`
`$ git checkout pull/208`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/x64/sysv/SysVVaList.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/x64/sysv/SysVVaList.java
@@ -31,10 +31,10 @@ import jdk.incubator.foreign.MemoryAddress;
 import jdk.incubator.foreign.MemoryHandles;
 import jdk.incubator.foreign.MemoryLayout;
 import jdk.incubator.foreign.MemorySegment;
-import jdk.internal.foreign.AbstractMemorySegmentImpl;
 import jdk.internal.foreign.NativeMemorySegmentImpl;
 import jdk.internal.foreign.Utils;
 import jdk.internal.foreign.abi.SharedUtils;
+import jdk.internal.misc.Unsafe;
 
 import java.lang.invoke.VarHandle;
 import java.lang.ref.Cleaner;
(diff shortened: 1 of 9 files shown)


src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/x64/sysv/SysVVaList.java
```diff
@@ -121,6 +127,23 @@
         slices.add(regSaveArea);
     }
 
+    private static MemoryAddress emptyListAddress() {
+        MemorySegment ms = MemorySegment.allocateNative(LAYOUT);
+        cleaner.register(SysVVaList.class, ms::close);
```
mcimadamore, line 132: This will fail because of confinement

JornVernee: Thanks, I fixed it by switching to Unsafe for the allocation, so that we can free from any thread.

PR #212, opened 2020-06-19
8247937: Specialize downcall binding recipes using MethodHandle combinators

JornVernee:
Hi,

This PR adds specialization of downcall binding recipes using MethodHandles. The main addition to the code is the ProgrammableInvoker::specialize method, but there are also some refactorings which I will explain;

The current code has a single `invoke` method that takes the high-level input arguments of the down call as an Object[]. This method calls the BindingInterpreter to pre-process arguments, and also move any low-level values produced by the pre-processing directly into the intermediate argument buffer that is later passed to the native code that copies the values into registers and then calls the target function.

This `invoke` method is split into 2; `invokeInterpBindings`, which pre-processes the arguments and outputs any low-level values produced by the pre-processing into an Object[].

This Object[] is then passed to `invokeMoves`, which moves the low-level values into the intermediate argument buffer that is passed to native.

This split is done so that we can replace the invokeInterpBindings call with a specialized MethodHandle chain based on the binding recipe instead, which is built on top a type handle of invokeMoves.

This takes care of the binding recipe specialization. invokeMoves can later be replaced during C2 compilation with code that passes these low-level values into registers directly (but that is for another patch).

BindingInterpreter now doesn't write values to the intermediate buffer directly, and so instead of passing functors to obtain a pointer to write a low-level value to, it now takes functors that handle the reading or writing (see StoreFunc and LoadFunc).

Since the read/write methods in BindingInterpreter are now called from multiple places, I've moved them to SharedUtils.

---

The process of specializing the binding recipe is as follows:

We first calculate a low-level method type (the 'intrinsicType') based on the MOVE bindings of a particular recipe, this gives us a method type that takes only primitives as arguments, which represent the values to be copied into the various CPU registers before calling the native target function. We then get a low-level MethodHandle that calls invokeMoves, and adapt it to the intrinsic type. On top of that we build the specialized binding recipe method handle chain.

For each argument, we iterate through the bindings in reverse, and for each binding insert a new filter MH onto the handle we already have. At the end we will end up with a handle that features the high-level arguments of our native function (MemorySegment, MemoryAddress, etc.).

The same is done for the return bindings; the return value is repeatedly filtered according to the bindings.

This currently doesn't work for multiple return values, in which case invokeMoves will return an Object[] instead of just a plain Object. This case is not specialized with method handles, but is instead handled solely by invokeInterpBindings. (we detect this in getBoundMethodHandle)

In case the bindings need to do allocation, a NativeAllocationScope is created and passed in as the first argument, which is then forwarded to the various filter MH using mergeArguments (which is a wrapper for MethodHandles::permuteArguments that merges 2 arguments given their indices). The allocation and cleanup of the NativeAllocationScope are handled with a tryFinally combinator.

---

I've added a system property to disable the MethodHandle specialization so that the performance difference can be measured. The speed up I've seen on the CallOverhead benchmark between this and the current code is about 1.5x (but with the C2 intrinsics this gets us on par with JNI). The difference between specialization turned off and on is a little bigger than that, since the split of the 'invoke' method adds a bit of overhead as well, but it's still on overall gain compared to what we currently have.

Finally, I've also fixed a minor problems with TestUpcall where the expected and actual values were reversed when calling asserts.

---

If I've missed anything in the explanation, please feel free to ask!

Thanks,
Jorn

---------
### Progress

### Issue
 * [JDK-8247937](https://bugs.openjdk.java.net/browse/JDK-8247937): Specialize downcall binding recipes using MethodHandle combinators

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/212/head:pull/212`
`$ git checkout pull/212`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/ProgrammableInvoker.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/ProgrammableInvoker.java
@@ -22,11 +22,14 @@
  */
 package jdk.internal.foreign.abi;
 
-import jdk.incubator.foreign.FunctionDescriptor;
 import jdk.incubator.foreign.MemoryAddress;
 import jdk.incubator.foreign.MemoryHandles;
 import jdk.incubator.foreign.MemorySegment;
+import jdk.incubator.foreign.NativeAllocationScope;
+import jdk.internal.access.JavaLangInvokeAccess;
+import jdk.internal.access.SharedSecrets;
 import jdk.internal.foreign.MemoryAddressImpl;
+import jdk.internal.foreign.Utils;
 
 import java.lang.invoke.MethodHandle;
 import java.lang.invoke.MethodHandles;
@@ -34,10 +37,22 @@ import java.lang.invoke.MethodType;
 import java.lang.invoke.VarHandle;
 import java.nio.ByteOrder;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 
+import static java.lang.invoke.MethodHandles.collectArguments;
+import static java.lang.invoke.MethodHandles.dropArguments;
+import static java.lang.invoke.MethodHandles.empty;
+import static java.lang.invoke.MethodHandles.filterArguments;
+import static java.lang.invoke.MethodHandles.identity;
+import static java.lang.invoke.MethodHandles.insertArguments;
+import static java.lang.invoke.MethodHandles.permuteArguments;
+import static java.lang.invoke.MethodHandles.tryFinally;
+import static java.lang.invoke.MethodType.methodType;
 import static sun.security.action.GetBooleanAction.privilegedGetProperty;
 
 /**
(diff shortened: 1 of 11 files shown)


src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/ProgrammableInvoker.java
```diff
@@ -136,10 +360,56 @@ Object invoke(Object[] args) {
...
+            if (stackArgsSeg != null) {
+                stackArgsSeg.close();
+            }
+        }
+    }
+
+    Object invokeInterpBindings(Object[] args, MethodHandle leaf,
+                                Map<VMStorage, Integer> argIndexMap,
+                                Map<VMStorage, Integer> retIndexMap) throws Throwable {
+        List<MemorySegment> tempBuffers = new ArrayList<>();
```
mcimadamore, line 388: Should this also be expressed in terms of NativeScope?

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/ProgrammableInvoker.java
```diff
@@ -68,42 +111,225 @@
...
+         } else {
+             handle = specialize(handle);
+         }
+
+        return handle;
     }
 
-    Object invoke(Object[] args) {
-        List<MemorySegment> tempBuffers = new ArrayList<>();
+    private MethodHandle specialize(MethodHandle intrinsicHandle) {
```
mcimadamore, line 186: I think some renaming of variables would be helpful here - e.g. the parameter is a movesHandle, or a leafHandle - then we can use some other name, like specializedHandle for the thing we're building - intrinsics is a bit vague and is used to refer to two very different handles.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/ProgrammableInvoker.java
```diff
@@ -48,17 +63,45 @@
...
+                    methodType(Object.class, Object[].class, MethodHandle.class, Map.class, Map.class));
+            MH_UNBOX_ADDRESS = lookup.findStatic(ProgrammableInvoker.class, "toRawLongValue",
+                    methodType(long.class, MemoryAddress.class));
+            MH_BOX_ADDRESS = lookup.findStatic(ProgrammableInvoker.class, "ofLong",
+                    methodType(MemoryAddress.class, long.class));
+            MH_BASE_ADDRESS = lookup.findVirtual(MemorySegment.class, "baseAddress",
+                    methodType(MemoryAddress.class));
+            MH_COPY_BUFFER = lookup.findStatic(ProgrammableInvoker.class, "copyBuffer",
+                    methodType(MemorySegment.class, MemorySegment.class, long.class, long.class, NativeAllocationScope.class));
+            MH_MAKE_ALLOCATOR = lookup.findStatic(NativeAllocationScope.class, "boundedScope",
```
mcimadamore, line 99: This class has references to `NativeAllocationScope` which has been renamed to `NativeScope`

JornVernee: Thanks, fixed.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/ProgrammableInvoker.java
JornVernee: Good suggestion, I renamed them leafHandle and specializedHandle

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/ProgrammableInvoker.java
JornVernee: Done.

JornVernee:
I've applied your suggestion of putting the code for interpreting, verifying and specializing the Bindings in methods on Binding itself, instead of scattered throughout the switch statements. This removes a bit of code as well, since there is no more need for casting, and fields are directly accessible. There was one place where this didn't completely work out; in the specialization we need to update the argument index (insertPos) when we encounter a MOVE binding, so the code currently has to explicitly check for MOVE bindings.

I've also added a few more CallOverhead benchmarks, with a more diverse argument set. The results show that especially once you start adding more arguments, or the arguments become more complex (see the MemoryAddress case), that the MH specialization starts to make more of a difference.

```
Benchmark                                            Mode  Cnt     Score    Error  Units
CallOverhead.jni_blank                               avgt   30     8.415 □  0.076  ns/op
CallOverhead.jni_identity                            avgt   30    12.370 □  0.254  ns/op
CallOverhead.panama_args10                           avgt   30   583.976 □  5.895  ns/op
CallOverhead.panama_args10_NO_SPEC                   avgt   30  1319.897 □ 22.884  ns/op
CallOverhead.panama_args5                            avgt   30   502.865 □  8.452  ns/op
CallOverhead.panama_args5_NO_SPEC                    avgt   30   919.254 □ 11.101  ns/op
CallOverhead.panama_blank                            avgt   30   199.373 □  2.102  ns/op
CallOverhead.panama_blank_NO_SPEC                    avgt   30   317.260 □  4.067  ns/op
CallOverhead.panama_identity                         avgt   30   254.405 □  7.173  ns/op
CallOverhead.panama_identity_NO_SPEC                 avgt   30   494.224 □  2.359  ns/op
CallOverhead.panama_identity_memory_address          avgt   30   265.507 □  4.160  ns/op
CallOverhead.panama_identity_memory_address_NO_SPEC  avgt   30   592.381 □ 14.314  ns/op
CallOverhead.panama_identity_struct                  avgt   30   601.288 □ 16.147  ns/op
CallOverhead.panama_identity_struct_NO_SPEC          avgt   30   951.495 □ 15.450  ns/op
```

PR #221, opened 2020-06-29
8248487: Add static helpers to access segments

mcimadamore:
This patch addresses some usability issues in the foreign memory API. First, it adds several static accessors for all carrier types in MemoryLayouts (I'm thorn whether to add this in MemoryLayouts or MemoryHandles, suggestions welcome).

These accessors provide an access API that is similar to unsafe - all accessors expect a base address plus a byte offset. In most cases the accessors can be used and direct VH usage can be dropped, which might make approaching this API a bit easier.

Secondly, I've added a bunch of helpers to map a segment into _any_ array (as opposed to just `byte[]`). This makes the API more symmetric, given that, after all we _were_ able to create a segment from _any_ array (not just `byte[]`).

---------
### Progress

### Issue
 * [JDK-8248487](https://bugs.openjdk.java.net/browse/JDK-8248487): Add static helpers to access segments

### Reviewers
 * Athijegannathan Sundararajan ([sundar](@sundararajana) - Committer) ⚠️ Review applies to 2ca6bae33ae344b080440cacef262dd5f27c511d
 * Jorn Vernee ([jvernee](@JornVernee) - Committer) ⚠️ Review applies to e37b15a066d240d2acde9605652464477701c13d
 * Chris Hegarty ([chegar](@ChrisHegarty) - no project role) ⚠️ Review applies to 2ca6bae33ae344b080440cacef262dd5f27c511d
 * Henry Jen ([henryjen](@slowhog) - Committer) ⚠️ Review applies to 5a77d95609d6c910f6d013032a3efe1b1cd5f113

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/221/head:pull/221`
`$ git checkout pull/221`


mcimadamore:
Ideally, we could use varargs notation here, but one problem with that is that, if we ever add ability to allocate "empty" array - as in:

```java
MemoryAddress allocateArray(MemoryLayout elemLayout, long size)
```

Then we risk ambiguities.

If we are ok to never add a method like the one above - after all, one can allocate an empty array as follows:

```java
allocateArray(MemoryLayout.ofSequence(size, elemLayout))
```
then this is less of a problem.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryLayouts.java
```diff
@@ -61,6 +62,239 @@ private MemoryLayouts() {
...
+     * Read a char from given address and offset, with byte order set to {@link ByteOrder#LITTLE_ENDIAN}.
+     * <p>
+     * This is equivalent to the following code:
+     * <blockquote><pre>{@code
+    VarHandle handle = MemoryHandles.withStride(BITS_16_LE.withBitAlignment(8).varHandle(char.class), 1L);
+    char value = (char)handle.get(addr, offset);
+     * }</pre></blockquote>
+     * @param addr base address.
+     * @param offset offset (relative to {@code addr}). The final address of this read operation can be expressed as {@code addr.addOffset(offset)}.
+     * @return a char value read from {@code addr} at the offset specified by {@code index}.
```
JornVernee, line 116: Here, and similarly elsewhere:
```suggestion
     * @return a char value read from {@code addr} at the offset specified by {@code offset}.
```

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -403,6 +403,72 @@
      */
     byte[] toByteArray();
 
+    /**
+     * Copy the contents of this memory segment into a fresh short array.
+     * @return a fresh short array copy of this memory segment.
+     * @throws UnsupportedOperationException if this segment does not feature the {@link #READ} access mode, or if this
+     * segment's contents cannot be copied into a {@link short[]} instance, e.g. because {@code byteSize() % 4 != 0},
```
JornVernee, line 410: Typo
```suggestion
     * segment's contents cannot be copied into a {@link short[]} instance, e.g. because {@code byteSize() % 2 != 0},
```

test/jdk/java/foreign/TestByteBuffer.java
```diff
@@ -624,34 +615,34 @@ public void testDeadAccessOnClosedBufferSegment() {
...
+                (base) -> initBytes(base, ints, (addr, pos) -> MemoryLayouts.setInt_BE(addr, pos * 4, (int)(long)pos));
         Consumer<MemoryAddress> floatInitializer =
-                (base) -> initBytes(base, floats, (addr, pos) -> floatHandle.set(addr, pos, (float)(long)pos));
+                (base) -> initBytes(base, floats, (addr, pos) -> MemoryLayouts.setFloat_BE(addr, pos * 4, (float)(long)pos));
         Consumer<MemoryAddress> longInitializer =
-                (base) -> initBytes(base, longs, (addr, pos) -> longHandle.set(addr, pos, (long)pos));
+                (base) -> initBytes(base, longs, (addr, pos) -> MemoryLayouts.setLong_BE(addr, pos * 8, (long)pos));
         Consumer<MemoryAddress> doubleInitializer =
-                (base) -> initBytes(base, doubles, (addr, pos) -> doubleHandle.set(addr, pos, (double)(long)pos));
+                (base) -> initBytes(base, doubles, (addr, pos) -> MemoryLayouts.setDouble_BE(addr, pos * 8, (double)(long)pos));
```
JornVernee, line 630: Seems strange to me to switch to MemoryLayout.setXXX here, since we are actually accessing an element of a known layout, instead of a one-off value.

As a result, you now have to multiply the offset with a magic number that represents the stride. I don't think this is a good use-case...

mcimadamore:
Moved all accessors to a separate class (MemoryAccess). Also added a pair of accessors for memory address carriers.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryAccess.java
```diff
@@ -0,0 +1,733 @@
...
+     * @param addr base address.
+     * @param offset offset (relative to {@code addr}). The final address of this read operation can be expressed as {@code addr.addOffset(offset)}.
+     * @param value the byte value to be written.
+     */
+    public static void setByte_LE(MemoryAddress addr, long offset, byte value) {
+        byte_LE_handle.set(addr, offset, value);
+    }
+
+    /**
+     * Read a char from given address and offset, with byte order set to {@link ByteOrder#LITTLE_ENDIAN}.
```
slowhog, line 68: While the code indicates this is a 16-bit char, I wonder if it's better to clarify the address should have UTF-16 char, or Java char or wchar_t. Just to make it clear that the address is expected to have UTF-16 encoding?

Same comment applies to al [gs]etChar_[BL]E methods.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryLayouts.java
```diff
@@ -26,6 +26,7 @@
 
 package jdk.incubator.foreign;
 
+import java.lang.invoke.VarHandle;
```
slowhog, line 29:
```suggestion

File: src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryLayouts.java [hunk 6]
```diff
@@ -135,4 +136,5 @@ private MemoryLayouts() {
      * A value layout constant whose size is the same as that of a Java {@code double}, and byte order set to {@link ByteOrder#nativeOrder()}.
      */
     public static final ValueLayout JAVA_DOUBLE = MemoryLayout.ofValueBits(64, ByteOrder.nativeOrder());
+
```
slowhog, line 139:
```suggestion

File: src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryAccess.java (hunk 4 above)
Comment (mcimadamore) [reply src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryAccess.java:68]: I think it's tricky - a Java `char` is not _exactly_ a single unicode code point; as written in the javadoc, a Java char is a code _unit_ (which might be a code point, or half a code point), as per UTF16 encoding.
I agree this needs some clarification, just not sure 100% yet on what would be the best way to proceed.

File: src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryAccess.java [hunk 7]
```diff
@@ -0,0 +1,744 @@
...
+ * This class defines ready-made static accessors which can be used to dereference memory segments in many ways.
+ * Each accessor (see {@link #getInt(MemoryAddress, long)} takes a <em>base</em> address and an offset (expressed in bytes).
+ * The final address at which the dereference will occur will be computed by offsetting the base address by
+ * the specified offset, as if by calling {@link MemoryAddress#addOffset(long)} on the specified base address.
+ */
+public final class MemoryAccess {
+
+    public MemoryAccess() {
+        // just the one
+    }
```
JornVernee, line 16: Should be private I guess:
```suggestion
    private MemoryAccess() {
        // just the one
    }
```

test/jdk/java/foreign/TestByteBuffer.java
mcimadamore: most of our tests are like that (at least in the memory access domain - ABI tests are different). I just wanted at least _one_ test to use the accessor, and TestByteBuffer is one of the tests we have which is complex enough to provide good coverage for the accessors.

PR #222, opened 2020-06-29
8248499: Add methods to allocate off heap arrays from Java arrays

mcimadamore:
This patch adds methods to NativeScope which allow clients to allocate off-heap arrays from Java on-heap arrays.

While in most cases the code for copying is straightforward, we have to take into account endianness mismatches (in case the user wants to store the array elements off-heap in an endianness that doesn't match that of the platform).

To allow for that, I've added a new implementation method in AbstractMemorySegmentImpl, which is the same as `copyFrom`, but adds the byte swap (a wrapper around `Unsafe::copySwapMemory`).

---------
### Progress

### Issue
 * [JDK-8248499](https://bugs.openjdk.java.net/browse/JDK-8248499): Add methods to allocate off heap arrays from Java arrays

### Reviewers
 * Athijegannathan Sundararajan ([sundar](@sundararajana) - Committer) ⚠️ Review applies to ac6e37dc7c3eb91f4640db892096421d1a9f48f3
 * Jorn Vernee ([jvernee](@JornVernee) - Committer) ⚠️ Review applies to 38c81345427b02034d063d824ab6f982434f544a
 * Paul Sandoz ([psandoz](@PaulSandoz) - Committer) ⚠️ Review applies to 86c572f3281a16f5fa405f8e8027b71158d5f482
 * Henry Jen ([henryjen](@slowhog) - Committer) ⚠️ Review applies to 86c572f3281a16f5fa405f8e8027b71158d5f482

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/222/head:pull/222`
`$ git checkout pull/222`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/NativeScope.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/NativeScope.java
@@ -319,11 +319,10 @@ public abstract class NativeScope implements AutoCloseable {
     allocate(MemoryLayout.ofSequence(size, elementLayout));
      * }</pre>
      * @param elementLayout the array element layout.
-     * @param size the array size.
+     * @param size the array element count.
      * @return an address which points to the newly allocated memory block.
      * @throws OutOfMemoryError if there is not enough space left in this native scope, that is, if
      * {@code limit() - size() < (elementLayout.byteSize() * size)}.
-     * @throws IllegalArgumentException if {@code elementLayout.byteSize()} does not conform to the size of a double value.
      */
     public MemoryAddress allocateArray(MemoryLayout elementLayout, long size) {
         return allocate(MemoryLayout.ofSequence(size, elementLayout));
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/NativeScope.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/NativeScope.java
@@ -52,8 +52,6 @@ import java.util.OptionalLong;
  */
 public abstract class NativeScope implements AutoCloseable {
 
-    Unsafe unsafe = Unsafe.getUnsafe();
-
     /**
      * If this native scope is bounded, returns the size, in bytes, of this native scope.
      * @return the size, in bytes, of this native scope (if available).
@@ -354,6 +352,25 @@ public abstract class NativeScope implements AutoCloseable {
      */
     public MemoryAddress allocate(MemoryLayout layout) {
         return allocate(layout.byteSize(), layout.byteAlignment());
+    }
+
+    /**
+     * Allocate a block of memory corresponding to an array with given element layout and size.
+     * The address returned by this method is associated with a segment which cannot be closed.
+     * Moreover, the returned address must conform to the layout alignment constraints. This is equivalent to the
+     * following code:
+     * <pre>{@code
+    allocate(MemoryLayout.ofSequence(size, elementLayout));
+     * }</pre>
+     * @param elementLayout the array element layout.
+     * @param size the array size.
+     * @return an address which points to the newly allocated memory block.
+     * @throws OutOfMemoryError if there is not enough space left in this native scope, that is, if
+     * {@code limit() - size() < (elementLayout.byteSize() * size)}.
+     * @throws IllegalArgumentException if {@code elementLayout.byteSize()} does not conform to the size of a double value.
+     */
+    public MemoryAddress allocateArray(MemoryLayout elementLayout, long size) {
+        return allocate(MemoryLayout.ofSequence(size, elementLayout));
     }
 
     /**
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/NativeScope.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/NativeScope.java
@@ -26,9 +26,12 @@
 
 package jdk.incubator.foreign;
 
+import jdk.internal.foreign.AbstractMemorySegmentImpl;
 import jdk.internal.foreign.AbstractNativeScope;
+import jdk.internal.misc.Unsafe;
 
 import java.lang.invoke.VarHandle;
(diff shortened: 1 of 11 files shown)


src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/NativeScope.java
```diff
@@ -49,6 +52,8 @@
  */
 public abstract class NativeScope implements AutoCloseable {
 
+    Unsafe unsafe = Unsafe.getUnsafe();
+
```
sundararajana, line 56: should this field be private?

sundararajana: looks like unused. may be removed

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/NativeScope.java
```diff
@@ -176,31 +181,167 @@ public MemoryAddress allocate(MemoryLayout layout, double value) {
...
+     * @throws IllegalArgumentException if {@code elementLayout.byteSize()} does not conform to the size of a byte value.
      */
-    public MemoryAddress allocate(MemoryLayout layout, MemoryAddress value) {
-        VarHandle handle = MemoryHandles.asAddressVarHandle(layout.varHandle(carrierForSize(layout.byteSize())));
-        MemoryAddress addr = allocate(layout);
-        handle.set(addr, value);
+    public MemoryAddress allocateArray(ValueLayout elementLayout, byte[] array) {
+        if (elementLayout.byteSize() != 1) {
+            throw new IllegalArgumentException("Bad layout size");
+        }
```
JornVernee, line 197: Can maybe be changed to 
```suggestion
        Utils.checkPrimitiveCarrierCompat(byte.class, elementLayout);
```

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/NativeScope.java
```diff
@@ -176,31 +181,167 @@ public MemoryAddress allocate(MemoryLayout layout, double value) {
...
+    public MemoryAddress allocateArray(ValueLayout elementLayout, char[] array) {
+        if (elementLayout.byteSize() != 2) {
+            throw new IllegalArgumentException("Bad layout size");
+        }
+        MemoryAddress addr = allocate(MemoryLayout.ofSequence(array.length, elementLayout));
+        if (elementLayout.order() == ByteOrder.nativeOrder()) {
+            addr.segment().copyFrom(MemorySegment.ofArray(array));
+        } else {
+            ((AbstractMemorySegmentImpl)addr.segment()).copyFromSwap(MemorySegment.ofArray(array), elementLayout.byteSize());
+        }
```
JornVernee, line 247: This code seems to be the same in the other implementations (except for the call to MS::ofArray). Maybe you want to factor it out into a helper method?

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
```diff
@@ -132,6 +132,16 @@ public void copyFrom(MemorySegment src) {
                 base(), min(), size);
     }
 
+    public void copyFromSwap(MemorySegment src, long elemSize) {
```
JornVernee, line 135: Should this maybe be a public API in MemorySegment as well?

mcimadamore: > Should this maybe be a public API in MemorySegment as well?

I'll prefer to leave that discussion for another patch

mcimadamore:
I've addressed most of the comments; I also added an extra method `allocateArray(MemoryLayout, long)` to allocate an 'empty' array with given layout. This "just" a shortcut for `allocate(MemoryLayout.ofSequence(...))` but I think it's very handy - e.g. to create empty pointer arrays, as in:

```java
MemoryAddress addr = scope.allocateArray(C_POINTER, 42);
```

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/NativeScope.java
```diff
@@ -356,6 +354,25 @@ public MemoryAddress allocate(MemoryLayout layout) {
...
+    /**
+     * Allocate a block of memory corresponding to an array with given element layout and size.
+     * The address returned by this method is associated with a segment which cannot be closed.
+     * Moreover, the returned address must conform to the layout alignment constraints. This is equivalent to the
+     * following code:
+     * <pre>{@code
+    allocate(MemoryLayout.ofSequence(size, elementLayout));
+     * }</pre>
+     * @param elementLayout the array element layout.
+     * @param size the array size.
```
JornVernee, line 322: Maybe useful to mention that the size is the element count, and not in bytes.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/NativeScope.java
```diff
@@ -356,6 +354,25 @@ public MemoryAddress allocate(MemoryLayout layout) {
...
+     * @return an address which points to the newly allocated memory block.
+     * @throws OutOfMemoryError if there is not enough space left in this native scope, that is, if
+     * {@code limit() - size() < (elementLayout.byteSize() * size)}.
+     * @throws IllegalArgumentException if {@code elementLayout.byteSize()} does not conform to the size of a double value.
```
JornVernee, line 326: This looks like a remnant of copy-pasting?

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
JornVernee: Ok. Was just wondering if you had any thoughts about that already :)

PR #223, opened 2020-06-30
RFC: move factories and helper methods to new MemorySegments class

mcimadamore:
This patch explores moving several static factories and methods (as well as some instance ones such as fill/copy/mismatch) onto a separate "sidecar" class, namely `MemorySegments`.

I guess overall the approach is defensible, but the API feels a bit less tight:

* having segment factories in a different class than `MemorySegment` feels odd
* `segment.fill(42)` reads better than `MemorySegments.fill(segment, 42)`

There might be an argument that this patch is overrotating towards the static, helper class approach, but then the question is where to draw the line between essential and non-essential stuff.

One possibility would be to:

* keep MemorySegment.allocateNative where it is
* move MemorySegment.mapFromPath to MappedMemorySegment
* move all the view-y factories, as well as array and buffer projections, and accessors on the sidecar class
* perhaps, move the static spliterator method in the sidecar as well
* as for copy/mismatch/fill, we could either keep them where they are now (instance methods in MemorySegment), or move them onto the sidecar

Any thoughts?

Javadoc for this revision available here:
http://cr.openjdk.java.net/~mcimadamore/panama/memaccess-segments-helpers/javadoc/jdk/incubator/foreign/package-summary.html

---------
### Progress

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/223/head:pull/223`
`$ git checkout pull/223`


mcimadamore:
Closed, as we're not pursuing this strategy.

PR #224, opened 2020-07-01
8247993: AArch64: add support for VaList

nick-arm:
Based on the SysV x64 implementation as it's somewhat similar to AArch64.

I added some extra tests to cover some cases on AArch64 that weren't hit
by the existing tests. The new testHugeStructByValue fails on Linux
x86_64 because the HugePoint struct is too large to be passed in
registers in the SysV ABI. Made a minimal fix to pass it on the stack
instead.

---------
### Progress

### Issue
 * [JDK-8247993](https://bugs.openjdk.java.net/browse/JDK-8247993): AArch64: add support for VaList

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/224/head:pull/224`
`$ git checkout pull/224`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CSupport.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CSupport.java
@@ -291,7 +291,7 @@ public class CSupport {
     /**
      * The {@code va_list} native type.
      */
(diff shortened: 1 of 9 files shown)


JornVernee:
Hi Nick,

Thanks for the patch! And thanks for fixing the SysV problem. That use-case slipped my mind.

I plan to take a thorough look at this soon, but I'm currently juggling a few other things.

PR #232, opened 2020-07-03
8248487: Add static helpers to access segments (continued)

mcimadamore:
Following some feedback from jextract experiments, we have noted that, while the set of primitive accessors is the right one to expose, as it is the most primitive (and others can be derived from that), usability-wise we lose a bit when it comes to perform dereference at offset 0 (in which case an extra offset 0 has to be specified), or a dereference with a logical index (e.g. `p[i] = ...`).

To overcome these issues, I've added two additional variants of static accessors - now:

* `getByte_LE(MemoryAddress addr)` -> dereference at given address
* `getByteAtOffset(MemoryAddress, byte offset)` -> dereference at given address + offset (in bytes)
* `getByteAtIndex(MemoryAddress, long index)` -> dereference at given logical index (where offset = index * stride)

Looking at TestByteBuffer, this immediately seem to make things better.

---------
### Progress

### Issue
 * [JDK-8248487](https://bugs.openjdk.java.net/browse/JDK-8248487): Add static helpers to access segments ⚠️ Title mismatch between PR and JBS.

### Reviewers
 * Jorn Vernee ([jvernee](@JornVernee) - Committer) ⚠️ Review applies to 5e09a1350349ea5d06d7e039a26156861db907a9

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/232/head:pull/232`
`$ git checkout pull/232`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryAccess.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryAccess.java
@@ -5,9 +5,19 @@ import java.nio.ByteOrder;
 
 /**
  * This class defines ready-made static accessors which can be used to dereference memory segments in many ways.
- * Each accessor (see {@link #getInt(MemoryAddress, long)} takes a <em>base</em> address and an offset (expressed in bytes).
+ * <p>
+ * The most primitive accessors (see {@link #getInt(MemoryAddress, long)}) take a <em>base</em> address and an offset (expressed in bytes).
  * The final address at which the dereference will occur will be computed by offsetting the base address by
  * the specified offset, as if by calling {@link MemoryAddress#addOffset(long)} on the specified base address.
+ * <p>
+ * In cases where no offset is required, overloads are provided (see {@link #getInt(MemoryAddress)}) so that
+ * clients can omit the offset coordinate.
+ * <p>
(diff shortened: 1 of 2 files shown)


src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryAccess.java
```diff
@@ -5,9 +5,19 @@
...
  * The final address at which the dereference will occur will be computed by offsetting the base address by
  * the specified offset, as if by calling {@link MemoryAddress#addOffset(long)} on the specified base address.
+ * <p>
+ * In cases where no offset is required, overloads are provided (see {@link #getInt(MemoryAddress)}) so that
+ * clients can omit the offset coordinate.
+ * <p>
+ * To help dereferencing in array-like use cases (e.g. where the layout of a given memory segment is a sequence
+ * layout of given size an element count), higher-level overloads are also provided (see {@link #getIntAtIndex(MemoryAddress, long)}),
+ * which take an <em>base</em> address and a <em>logical</em> element index. The formula to obtain the byte offset {@code O} from an
+ * index {@code I} is given by {@code O = I * S} where {@code s} is the size (expressed in bytes) of the element to
```
JornVernee, line 19:
```suggestion
 * index {@code I} is given by {@code O = I * S} where {@code S} is the size (expressed in bytes) of the element to
```

PR #233, opened 2020-07-03
8248487: Add static helpers to access segments (continued) - benchmark fix

mcimadamore:
This patch fixes the benchmarks that were referring to the previous accessor names.

It also addresses a performance loophole with long multiplication being used even when index fits in an int.

---------
### Progress

### Issue
 * [JDK-8248487](https://bugs.openjdk.java.net/browse/JDK-8248487): Add static helpers to access segments ⚠️ Title mismatch between PR and JBS.

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/233/head:pull/233`
`$ git checkout pull/233`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryAccess.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryAccess.java
@@ -1,5 +1,6 @@
 package jdk.incubator.foreign;
 
+import jdk.internal.access.foreign.MemoryAddressProxy;
 import jdk.internal.vm.annotation.ForceInline;
 
 import java.lang.invoke.VarHandle;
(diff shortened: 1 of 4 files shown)


src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryAccess.java
```diff
@@ -2031,6 +2033,15 @@ public static MemoryAddress getAddressAtIndex(MemoryAddress addr, long index) {
...
+    }
+
+    @ForceInline
+    private static long scale(long index, int size) {
+        if (index < Integer.MAX_VALUE) {
+            return (int)index * size;
+        } else {
+            return index * size;
+        }
     }
```
JornVernee, line 2046: This looks problematic. `index * size` can overflow when casting to int, when it would not have overflown if the operands were `long`. For small segments we do something similar, but it's less of a problem there since we know the segment size fits in an int, so if there's overflow the access would have been out of bounds any ways. I think here you will have to pass in the `segment()` from the accessed address, and then use the isSmall() predicate as the criteria for doing the cast or not.

JornVernee: Maybe MemoryAddressProxy::multiplyOffsets can be used?

PR #238, opened 2020-07-09
AArch64 build fixes

JornVernee:
Hi,

This patch fixes several linker errors observed when building on aarch64, due to missing implementations of methods added for the C2 down call intrinsics.

Currently C2 compilation will just bail out when trying to generate a the native call, but I think another strategy is possible where we generate a call to the fallback MethodHandle instead. This patch is just fixing the build.

Thanks,
Jorn

---------
### Progress

### Reviewers
 * Paul Sandoz ([psandoz](@PaulSandoz) - Committer) ⚠️ Review applies to 1e8bfe3b4109024ed5e8ef4904a94d6ba6d011dc

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/238/head:pull/238`
`$ git checkout pull/238`

### DIFF
--- a/src/hotspot/cpu/aarch64/vmreg_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/vmreg_aarch64.cpp
@@ -26,7 +26,7 @@
 #include "precompiled.hpp"
 #include "asm/assembler.hpp"
 #include "code/vmreg.hpp"
-
+#include "vmreg_aarch64.inline.hpp"
 
 
 void VMRegImpl::set_regName() {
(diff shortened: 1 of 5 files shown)


src/hotspot/cpu/aarch64/vmreg_aarch64.cpp
```diff
@@ -51,3 +51,17 @@ void VMRegImpl::set_regName() {
...
+#define INTEGER_TYPE 0
+#define VECTOR_TYPE 1
+#define X87_TYPE 2
+#define STACK_TYPE 3
+
+VMReg VMRegImpl::vmStorageToVMReg(int type, int index) {
+  switch(type) {
+    case INTEGER_TYPE: return ::as_Register(index)->as_VMReg();
+    case VECTOR_TYPE: return ::as_FloatRegister(index)->as_VMReg();
+    case STACK_TYPE: return VMRegImpl::stack2reg(index LP64_ONLY(* 2)); // numbering on x64 goes per 64-bits
```
PaulSandoz, line 64: Referral to x64 in the comment

PR #244, opened 2020-07-14
8249300: jextract does not handle empty parameter list of a function pointer parameters

sundararajana:
Array type parameter in function callback is handled properly

---------
### Progress

### Issue
 * [JDK-8249300](https://bugs.openjdk.java.net/browse/JDK-8249300): jextract does not handle empty parameter list of a function pointer parameters

### Reviewers
 * Henry Jen ([henryjen](@slowhog) - Committer) ⚠️ Review applies to 6e6c762541fa402e5c69b169c59754a719e60d52

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/244/head:pull/244`
`$ git checkout pull/244`

### DIFF
--- a/test/jdk/tools/jextract/test8249300.h
+++ b/test/jdk/tools/jextract/test8249300.h
@@ -1,0 +1,39 @@
+/*
+ * Copyright (c) 2020, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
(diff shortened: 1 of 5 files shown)


test/jdk/tools/jextract/test8249300.h
```diff
@@ -0,0 +1,39 @@
...
+#endif // __cplusplus
+
+#ifdef _WIN64
+#define EXPORT __declspec(dllexport)
+#else
+#define EXPORT
+#endif
+
+typedef void MyVoid;
+EXPORT void func(MyVoid (*f)());
```
slowhog, line 35: Subtle test case relying C syntax without explicit void to make Object[] in function signature. Maybe add a comment to make that clear.

Not sure if we should make explicit varargs.

PR #246, opened 2020-07-15
8249536: jextract throw IllegalStateException for bitfields in nested anonymous structs

JornVernee:
Hi,

This patch fixes an issue with grouping of bitfields into a container for structs that contain anonymous nested structs that have bitfields in them. Such as

```C
struct Foo {
    struct {
        int a : 7;
        int b : 25;
    };
};
```

The current code iterates over the member layouts of a struct, and when it finds a bitfield storage unit, it tries to find matching bitfield Declarations and groups them into a synthetic nested BITFIELDS declaration.

But, the memory layout members also contain bitfields that are found in anonymous nested structs, since they are flattened into the parent layout. However, for the jextract Declaration tree this is not the case, so the current code can not find a matching declaration for these bitfield layouts. e.g. in the example `Foo` will have a bitfield storage unit which contains `a` and `b`, but the Declaration tree has the declarations for `a` and `b` nested inside the anonymous struct (layouts and declaration tree are different in this regard).

The fix is to ignore bitfields that come from nested anonymous structs. They will be handled when processing the nested anonymous struct instead.

Thanks,
Jorn

---------
### Progress

### Issue
 * [JDK-8249536](https://bugs.openjdk.java.net/browse/JDK-8249536): jextract throw IllegalStateException for bitfields in nested anonymous structs

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/246/head:pull/246`
`$ git checkout pull/246`

### DIFF
--- a/src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/TreeMaker.java
+++ b/src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/TreeMaker.java
@@ -31,13 +31,17 @@ import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.ConcurrentModificationException;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Set;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
 import jdk.incubator.foreign.GroupLayout;
 import jdk.incubator.foreign.MemoryLayout;
 import jdk.incubator.jextract.Declaration;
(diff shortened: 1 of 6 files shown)


JornVernee:
> 
> 
> What happens for the following?
> 
> ```c
> struct Foo {
>     struct {
>         int a : 7;
>         int b : 25;
>     };
>     struct Bar {
> 	int a : 12;
>     } bar;
> };
> ```

In that case there is no problem since the `bar` struct is not anonymous.

You get the following tree (On Windows):

```
Scoped: TOPLEVEL <toplevel>
    Scoped: STRUCT Foo layout = [b32[contents=[b7(a)[layout/name=a]b25(b)[layout/name=b]],abi/windows/class=INTEGER][b32[contents=[b12(a)[layout/name=a]x20],abi/windows/class=INTEGER]](bar)[layout/name=bar]](Foo)[layout/name=Foo]
        Scoped: STRUCT  layout = [b32[contents=[b7(a)[layout/name=a]b25(b)[layout/name=b]],abi/windows/class=INTEGER]]
            Scoped: BITFIELDS  layout = [b7(a)[layout/name=a]b25(b)[layout/name=b]]
                Variable: BITFIELD a type = Int(layout = b32[abi/windows/class=INTEGER]), layout = Optional[b7]
                Variable: BITFIELD b type = Int(layout = b32[abi/windows/class=INTEGER]), layout = Optional[b25]
        Scoped: STRUCT Bar layout = [b32[contents=[b12(a)[layout/name=a]x20],abi/windows/class=INTEGER]](Bar)[layout/name=Bar]
            Scoped: BITFIELDS  layout = [b12(a)[layout/name=a]x20]
                Variable: BITFIELD a type = Int(layout = b32[abi/windows/class=INTEGER]), layout = Optional[b12]
        Variable: FIELD bar type = Declared([b32[contents=[b12(a)[layout/name=a]x20],abi/windows/class=INTEGER]](Bar)[layout/name=Bar]), layout = Optional[[b32[contents=[b12(a)[layout/name=a]x20],abi/windows/class=INTEGER]](Bar)[layout/name=Bar]]
```

PaulSandoz:
Will the bit fields named "a" in the anonymous and named nested structs clash. so processing "a" in struct Bar will get filtered out because "a" is added to the `nestedBitfieldNames` set?

slowhog:
That won't have name clash as access would be foo.a and foo.bar.a in C, same should apply in Java code.

I am more concerned about, as we have two anonymous in parallel, we should have a test case for something like this,

```c
struct Foo {
    struct {
        int a : 7;
        int b : 25;
    };
    struct {
	int c : 12;
	int d : 12;
    };
};
```

PR #247, opened 2020-07-16
Test the Java VaList implementation on every platform

JornVernee:
Hi,

This patch re-writes the VaListTest to try and test all the implementations on the different platforms, instead of just the implementation for the current platform.

This results in 4 different versions of each test being run; once for each platform, where the reading of the VaList is done in Java as well, and then one native implementation (which corresponds to the status quo). The needed parameters are injected into tests using data providers. For the struct tests this was a little tricky, since we need to inject the platform specific C_INT/C_LONGLONG/C_FLOAT layouts into the struct layout. This resulted in the current code complexity. (As an aside, I think local interface and methods work from Amber could really help this use-case).

Upcalls are still only tested for the current platform, since the creation of the va_list happens in native code, and we only have access to 1 platform in that case :)

---

I've also cleaned up the sumStack test, which wasn't actually using any part of our VaList implementation (it was testing varargs), as well as reduced some copy-pasted code which was previously used for debugging, but never collapsed back into for-loops afterwards.

Thanks,
Jorn

---------
### Progress

### Reviewers
 * Paul Sandoz ([psandoz](@PaulSandoz) - Committer) ⚠️ Review applies to 0a4715a50c34ebce2738e26da9436cb90f86a37e

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/247/head:pull/247`
`$ git checkout pull/247`

### DIFF
--- a/test/jdk/java/foreign/valist/VaListTest.java
+++ b/test/jdk/java/foreign/valist/VaListTest.java
@@ -196,12 +196,12 @@ public class VaListTest {
 
     @DataProvider
     @SuppressWarnings("unchecked")
-    public static Object[][] getInt() {
+    public static Object[][] pointers() {
         Function<MemoryLayout, Function<VaList, Integer>> getIntJavaFact = layout ->
             list -> {
                 MemoryAddress ma = list.vargAsAddress(layout);
                 MemorySegment accessibleSegment = MemorySegment.ofNativeRestricted(ma,
-                        C_INT.byteSize(), Thread.currentThread(), null, null);
+                        4, Thread.currentThread(), null, null);
                 return MemoryAccess.getInt(accessibleSegment.baseAddress());
             };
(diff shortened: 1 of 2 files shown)


mcimadamore:
In addition to Paul's comment, the `getInt` test doesn't seem 100% appropriate, since `sumInt` is using `C_INT` but `getInt` is using `C_POINTER` - maybe rename to getPointer?
Overall the test is very comprehensive and I think it will spare us a lot of bugs going forwards, thanks.

JornVernee:
I've updated the naming. The previous idea with the win* tests was that each platform would add some tests for it's own corner cases with it's own name prefix, but it didn't end up working out that way in the end. I've reverted the names back to struct, bigStruct, hugeStruct, and floatStruct.

(There was also 1 case of a hard-coded platform specific layout being used to determine a segment's size, which I've swapped for a hard-coded '4', so we at least don't to rely on a platform specific layout)

PR #252, opened 2020-07-21
8249790: Add Addressable abstraction

mcimadamore:
This patch adds a new entity in the Foreign Memory Access API, namely `Addressable`. This is a simple functional interface which is implemented by all the entities which can be mapped onto a `MemoryAddress`. For now, only `MemoryAddress` and `MemorySegment` implements it (but on the foreign-abi branch we will repeat the exercise and add more implementations).

---------
### Progress

### Issue
 * [JDK-8249790](https://bugs.openjdk.java.net/browse/JDK-8249790): Add Addressable abstraction

### Reviewers
 * Athijegannathan Sundararajan ([sundar](@sundararajana) - Committer) ⚠️ Review applies to efceeed70af50502d1b0f0a9090ceebcc707468f
 * Jorn Vernee ([jvernee](@JornVernee) - Committer) ⚠️ Review applies to efceeed70af50502d1b0f0a9090ceebcc707468f

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/252/head:pull/252`
`$ git checkout pull/252`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/Addressable.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/Addressable.java
@@ -1,0 +1,21 @@
+package jdk.incubator.foreign;
+
+/**
+ * Represents a type which is <em>addressable</em>. An addressable type is one which can projected down to
+ * a memory address instance (see {@link #address()}. Examples of addressable types are {@link MemorySegment},
+ * and {@link MemoryAddress}.
+ *
+ * @apiNote In the future, if the Java language permits, {@link Addressable}
+ * may become a {@code sealed} interface, which would prohibit subclassing except by
+ * explicitly permitted types, such as {@link MemorySegment} and {@link MemoryAddress}.
+ *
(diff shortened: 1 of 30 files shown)


src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/Addressable.java
```diff
@@ -0,0 +1,21 @@
+package jdk.incubator.foreign;
```
JornVernee, line 1: Copyright header is missing here

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/Addressable.java
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Represents a type which is <em>addressable</em>. An addressable type is one which can projected down to
```
JornVernee, line 4:
```suggestion
 * Represents a type which is <em>addressable</em>. An addressable type is one which can be projected down to
```

PR #256, opened 2020-07-21
8249803: Add LibraryLookup.Symbol

mcimadamore:
This patch adds the LibraryLookup.Symbol, which implements Addressable, and that is returned on a successful symbol lookup. This patch also tweaks VaList to also implement the Addressable interface (VaList already had an `address()` method).

The `ForeignLinker::downcallHandle` method has been tweaked to accept an Addressable as a function entry point. This allows downcall handle to target not only native symbols, but also upcall stubs (since upcall stubs are segments, which are addressable) as well as native memory addresses.

Some tests needed minor tweaks.

---------
### Progress

### Issue
 * [JDK-8249803](https://bugs.openjdk.java.net/browse/JDK-8249803): Add LibraryLookup.Symbol

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/256/head:pull/256`
`$ git checkout pull/256`


src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/ProgrammableInvoker.java
```diff
@@ -107,12 +108,12 @@
...
     private final long bufferCopySize;
 
-    public ProgrammableInvoker(ABIDescriptor abi, MemoryAddress addr, CallingSequence callingSequence) {
+    public ProgrammableInvoker(ABIDescriptor abi, Addressable addr, CallingSequence callingSequence) {
         this.abi = abi;
         this.layout = BufferLayout.of(abi);
         this.stubAddress = adapterStubs.computeIfAbsent(abi, key -> generateAdapter(key, layout));
 
-        this.addr = addr;
+        this.addr = addr.address();
```
JornVernee, line 116: Don't think this will keep the reference to the LibraryLookup alive. I think you need to save the Addressable, and then convert to MA on demand, or use 2 fields (or maybe you have a better idea).

mcimadamore: Gosh - you are right; what I did in the addressable patch was to store the addressable as-is in the programmable invoker field; I should just do that.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/LibraryLookup.java
```diff
@@ -48,14 +48,33 @@
  */
 public interface LibraryLookup {
 
+    /**
+     * A symbol retrieved during a library lookup. A lookup symbol has a <em>name</em> and can be projected
+     * into a memory address (see {@link #name()} and {@link #address()}, respectively).
+     */
+    interface Symbol extends Addressable {
```
PaulSandoz, line 55: Sealed and implementations are values?
It might even make sense as a value-based record, perhaps a little early to tell right now.

PR #257, opened 2020-07-21
8249810: Jextract should generate parameter types with Addressable type

mcimadamore:
This patch tweaks jextract so that all foreign functions wrappers use the new Addressable carrier type, as opposed to the more specific MemoryAddress type. This is useful, as now clients can call such functions with either an address or a segment, w/o any need to call `MemorySegment::address`.

The bulk of the changes are in the TypeTranslator visitor, which now has a parameter which says as to whether an addressable should be used. There are also some changes in the HeaderBuilder class, as we now need to call `Addressable::address` on any incoming addressable parameter (since the method handle generated by the linker expects a `MemoryAddress`).

---------
### Progress

### Issue
 * [JDK-8249810](https://bugs.openjdk.java.net/browse/JDK-8249810): Jextract should generate parameter types with Addressable type

### Reviewers
 * Jorn Vernee ([jvernee](@JornVernee) - Committer) ⚠️ Review applies to e8abdab2ce0d9272fa8f4cc788041a89e546d688

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/257/head:pull/257`
`$ git checkout pull/257`

### DIFF
--- a/src/jdk.incubator.jextract/share/classes/jdk/incubator/jextract/HeaderBuilder.java
+++ b/src/jdk.incubator.jextract/share/classes/jdk/incubator/jextract/HeaderBuilder.java
@@ -24,6 +24,7 @@
  */
 package jdk.incubator.jextract;
 
+import jdk.incubator.foreign.Addressable;
 import jdk.incubator.foreign.FunctionDescriptor;
 import java.lang.invoke.MethodType;
 import java.util.ArrayList;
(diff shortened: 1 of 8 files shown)


mcimadamore:
> With code generation, I think that we can simply patch all MemoryAddress into Addressable in the static function wrapper to avoid conflict of choices.

What do you mean? Isn't that what the patch is doing? Note that we don't want to patch e.g. the MemoryAddress that represents a global variable, or the MemoryAddress which is used by a struct getter/setter to do a dereference - e.g. those should still be addresses?

src/jdk.incubator.jextract/share/classes/jdk/incubator/jextract/HeaderBuilder.java
```diff
@@ -60,20 +61,24 @@ public void addFunctionalInterface(String name, MethodType mtype,  FunctionDescr
...
+        List<String> pExprs = new ArrayList<>();
         final int numParams = paramNames.size();
         for (int i = 0 ; i < numParams; i++) {
             String pName = paramNames.get(i);
             if (pName.isEmpty()) {
                 pName = "x" + i;
             }
-            pNames.add(pName);
-            sb.append(delim + mtype.parameterType(i).getName() + " " + pName);
+            if (sigtype.parameterType(i).equals(Addressable.class)) {
```
slowhog, line 76: Here we assume the method handle have MemoryAddress type, while this is true in current implementation, it's possible that mhtype actually have Addressable if caller chose to do so otherwise.

mcimadamore:
> Looks good. I wonder if we should regenerate the libclang bindings for jextract at some point after this as well.

I was thinking the same - perhaps at the end of the various planned refactorings we can give it a go

src/jdk.incubator.jextract/share/classes/jdk/incubator/jextract/HeaderBuilder.java
mcimadamore: No, that's not possible. The ABI does not support Addressable as a carrier type for parameter/returns. In other words, Addressable is a fiction that's only valid in jextract - it is not really taken into account by the guts of the ABI support.

slowhog: It is true at ABI level, the method handle can only handle MemoryAddress. But there is nothing preventing something like ConstantHelper to adapt the ABI-level MH to take Addressable as you do here.

What I am suggesting is simply saying you don't need add sigtype, simply take the incoming method type and adapt MemoryAddress into Addressable.address().

mcimadamore: > What I am suggesting is simply saying you don't need add sigtype, simply take the incoming method type and adapt MemoryAddress into Addressable.address().

Not sure I get it. I think this same builder code is reused both for downcalls and upcalls; but we don't want to do the transformation for upcalls. So at the very least we'd need to pass a boolean to turn the transformation on/off. If we go down this path, then I guess TypeTranslator could be reverted back to what it was - is that what you are suggesting?

mcimadamore:
Folowing Henry's comments, I've simplified the patch; now all implementation changes are in HeaderBuilder, no other class is touched (apart from tests).

PR #260, opened 2020-07-22
8249879: Split MemorySegment and MemoryAddress

mcimadamore:
This patch splits, as the title suggests, the memory segment abstraction from the memory address abstraction, as described in this email:

https://mail.openjdk.java.net/pipermail/panama-dev/2020-July/009928.html

Aside from the obvious updates to the two interfaces (and their implementation), this patch greatly simplifies the way memory access var handles are generated; instead of spinning them on the fly (see `MemoryAccessVarHandleGenerator`, which is now removed) now all var handles can be derived from a primitive from which takes a segment and a byte offset. This caused the removal of several methods in the `JavaLangInvokeAccess` interface.

The new memory segment interface has some more methods to perform slicing - which can basically be used to give a new base to an existing segment and can therefore be used in a lot of cases as a replacement for `MemoryAddress::addOffset`.

The memory address interface is simplified, and features an additional `segmentOffset` method, which returns the byte offset of an address relative to the given segment (this is similar to the old `MemorySegment::rebase` method, which is now removed).

The `MemoryAccess` class needed a lot of tweaks in the various signatures, to take a segment instead of a base address.

I tried to update the documentation as best as I could, but it's possible I missed some references to the old relationship between segment and addresses, please double check.

Of course integrating this patch is gonna cause issues in foreign-abi and foreign-jextract, as these branches make (heavy) use of memory access var handles. I suspect that integration will trigger a merge failure, so we can fix followup issues as part of the merge process, so that we can guarantee that all the branches remain buildable.

---------
### Progress

### Issue
 * [JDK-8249879](https://bugs.openjdk.java.net/browse/JDK-8249879): Split MemorySegment and MemoryAddress

### Reviewers
 * Henry Jen ([henryjen](@slowhog) - Committer) ⚠️ Review applies to a2eaf4f4df4302702d9516df070c798ecfe0f141
 * Paul Sandoz ([psandoz](@PaulSandoz) - Committer) ⚠️ Review applies to 2afebee40099c83ff92ae2b0ee78f0b38751ac8b
 * Jorn Vernee ([jvernee](@JornVernee) - Committer) ⚠️ Review applies to 2afebee40099c83ff92ae2b0ee78f0b38751ac8b

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/260/head:pull/260`
`$ git checkout pull/260`


src/java.base/share/classes/java/lang/invoke/X-VarHandleMemoryAccess.java.template
```diff
@@ -66,305 +77,311 @@ final class MemoryAccessVarHandle$Type$Helper {
...
-            throw MemoryAccessVarHandleBase.newIllegalStateExceptionForMisalignedAccess(address);
+        if (skipOffsetCheck) {
+            //note: the offset portion has already been aligned-checked, by construction
+            if ((base & alignmentMask) != 0) {
+                throw MemoryAccessVarHandleBase.newIllegalStateExceptionForMisalignedAccess(address);
+            }
+        } else {
+            if ((address & alignmentMask) != 0) {
+                throw MemoryAccessVarHandleBase.newIllegalStateExceptionForMisalignedAccess(address);
+            }
```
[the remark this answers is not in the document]
mcimadamore: The final address is the sum of two components - a base B and an offset O. In a hot loop, B is typically constant (e.g. the base address of a segment) while O keeps changing. So, if the alignment check is only performed on B, the VM can prove that this check can be hoisted outside the loop. Of course that leaves the correctness problem - e.g. what if B is aligned but B + O is not? Luckily, by construction, this is not possible - skipOffsetCheck is only set when we construct a memory access var handle from a layout - and in that case the layout API make sure that the O part of the final address is aligned accordingly.

So, while it "looks" as if there's no performance gain, in reality the gain is quite big, because that move allows the alignment check to be moved outside the loop.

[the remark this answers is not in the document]
slowhog: Thanks for the explanation

test/micro/org/openjdk/bench/jdk/incubator/foreign/LoopOverNonConstantMapped.java
```diff
@@ -147,15 +146,15 @@ public int segment_loop() {
     public int segment_loop_static() {
         int res = 0;
         for (int i = 0; i < ELEM_SIZE; i ++) {
-            res += MemoryAccess.getIntAtIndex(segment.address(), i);
+            res += MemoryAccess.getIntAtIndex(segment, i * CARRIER_SIZE);
```
JornVernee, line 149: Why this multiply by CARRIER_SIZE? That should be done by the accessor.

test/jdk/java/foreign/TestVarHandleCombinators.java
```diff
@@ -176,16 +107,13 @@ public void testNestedSequenceAccess() {
...
         VarHandle vh = MemoryHandles.varHandle(int.class, ByteOrder.nativeOrder());
-        vh = MemoryHandles.withOffset(vh, 4);
-        VarHandle inner_vh = MemoryHandles.withStride(vh, 8);
-        VarHandle outer_vh = MemoryHandles.withStride(inner_vh, 5 * 8);
         int count = 0;
         try (MemorySegment segment = MemorySegment.allocateNative(inner_size * outer_size * 8)) {
             for (long i = 0; i < outer_size; i++) {
                 for (long j = 0; j < inner_size; j++) {
-                    outer_vh.set(segment.address(), i, j, count);
+                    vh.set(segment, i * 40 + j * 8, count);
```
JornVernee, line 114: Can this be done with combinators instead of manual offsets like before? Mostly curious about how this use-case changes.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryHandles.java
```diff
@@ -46,44 +46,22 @@
...
  * <p>
  * As an example, consider the memory layout expressed by a {@link SequenceLayout} instance constructed as follows:
  * <blockquote><pre>{@code
-SequenceLayout seq = MemoryLayout.ofSequence(5,
-    MemoryLayout.ofStruct(
+GroupLayout seq = MemoryLayout.ofStruct(
         MemoryLayout.ofPaddingBits(32),
         MemoryLayout.ofValueBits(32, ByteOrder.BIG_ENDIAN).withName("value")
-    ));
+));
```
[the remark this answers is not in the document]
mcimadamore: Will fix

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -278,6 +280,27 @@
...
+
+    /**
+     * Obtains a new memory segment view whose base address is the address passed as argument,
+     * and whose new size is computed by subtracting the address offset relative to this segment
+     * (see {@link MemoryAddress#segmentOffset(MemorySegment)}) from this segment size.
+     * @param address The new segment base offset (relative to the current segment base address), specified in bytes.
+     * @return a new memory segment view with updated base/limit addresses.
+     * @throws IndexOutOfBoundsException if {@code address.segmentOffset(this) < 0}, or {@code address.segmentOffset(this) > byteSize()}.
+     */
+    default MemorySegment asSlice(MemoryAddress address) {
```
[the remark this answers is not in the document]
mcimadamore: We can add that at a later stage

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryAddressImpl.java
```diff
@@ -37,100 +34,61 @@
...
         this.offset = offset;
     }
 
     // MemoryAddress methods
 
     @Override
-    public long segmentOffset() {
-        if (segment() == null) {
-            throw new UnsupportedOperationException("Address does not have a segment");
+    public long segmentOffset(MemorySegment segment) {
```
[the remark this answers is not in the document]
mcimadamore: Honestly I like it more the way it is, I'd prefer to consider after seeing some of the use cases.

test/micro/org/openjdk/bench/jdk/incubator/foreign/LoopOverNonConstantMapped.java
mcimadamore: mistake - will fix (but this has nothing to do with this patch, I think?)

test/jdk/java/foreign/TestVarHandleCombinators.java
mcimadamore: I can change this, for sure, but in real world, such use case would just go through the layout ABI, which does all the combination you want for free. Are you sure you want me to write what MemoryLayout::varHandle does inside the test?

JornVernee: Ok, just wanted to know if it should use the layout API or it couldn't for some reason.

test/micro/org/openjdk/bench/jdk/incubator/foreign/LoopOverNonConstantMapped.java
JornVernee: Well, this patch adds the multiply...

mcimadamore: > Well, this patch adds the multiply...

Ugh - that's a bug then

PR #261, opened 2020-07-23
8249949: jextract libclang Cursor children iteration creates a new up call stub every time

sundararajana:
Avoiding repeated upcall stub creation by caching. The array list updated is updated in a synchroinzed method.

---------
### Progress

### Issue
 * [JDK-8249949](https://bugs.openjdk.java.net/browse/JDK-8249949): jextract libclang Cursor children iteration creates a new up call stub every time

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/261/head:pull/261`
`$ git checkout pull/261`

### DIFF
--- a/src/jdk.incubator.jextract/share/classes/jdk/internal/clang/Cursor.java
+++ b/src/jdk.incubator.jextract/share/classes/jdk/internal/clang/Cursor.java
@@ -335,15 +335,26 @@ public final class Cursor {
         return new Cursor(Index_h.clang_getSpecializedCursorTemplate(cursor));
     }
 
-    public Stream<Cursor> children() {
-        final ArrayList<Cursor> ar = new ArrayList<>();
-        // FIXME: need a way to pass ar down as user data
-        Index_h.clang_visitChildren(cursor, Index_h.clang_visitChildren$visitor$make((c, p, d) -> {
+    private static class CursorChildren {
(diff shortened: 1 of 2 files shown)


src/jdk.incubator.jextract/share/classes/jdk/internal/clang/Cursor.java
```diff
@@ -335,15 +335,26 @@ public Cursor getSpecializedCursorTemplate() {
         return new Cursor(Index_h.clang_getSpecializedCursorTemplate(cursor));
     }
 
-    public Stream<Cursor> children() {
-        final ArrayList<Cursor> ar = new ArrayList<>();
-        // FIXME: need a way to pass ar down as user data
-        Index_h.clang_visitChildren(cursor, Index_h.clang_visitChildren$visitor$make((c, p, d) -> {
+    private static class CursorChildren {
```
mcimadamore, line 338: Wrapping the logic in a class is a good idea, but I think we can go further and avoid reusing the same static. E.g. we could make the list an _instance_ field, and then the static `get` method would create a new instance of `CursorChildren` (to do the visit) and then return the list attached to that instance.

mcimadamore: My bad - this is not really possible - we want a _shared_ address, so it has to live in a static field - and so we can't refer to instance stuff.

PR #264, opened 2020-07-24
8250518: Jextract should process macros in bulk

mcimadamore:
This patch adds bulk processing of macros; that is, instead of having a reparsing step for every macro constant, this patch tweaks jextract so that all unparsed macros are processed in one go.

The logic is quite convoluted and this work led to a major refactoring of MacroParserImpl; there is now a *table* where unparsed macros are kept - elements in this table are reparsed, and, based on the contents of the reparsed cursors, the table state is updated accordingly (the code has quite a bit of doc documenting the possible state transitions).

This results in some 2x performance boost in some of the heaviest extraction runs.

One thing to notice is that one of the test required some tweaks, as the new code seems to be able to infer a sharper type for a macro - e.g. `foo*` instead of `void*`. I suspect this is related to `TypeMaker::resolveTypeReferences` but I'm not 100% sure - @slowhog  can you please take a look?

---------
### Progress

### Issue
 * [JDK-8250518](https://bugs.openjdk.java.net/browse/JDK-8250518): Jextract should process macros in bulk

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/264/head:pull/264`
`$ git checkout pull/264`

### DIFF
--- a/src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/MacroParserImpl.java
+++ b/src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/MacroParserImpl.java
@@ -30,6 +30,7 @@ import jdk.incubator.jextract.Declaration;
 import jdk.incubator.jextract.JextractTask;
 import jdk.incubator.jextract.Position;
 import jdk.incubator.jextract.Type;
+import jdk.incubator.jextract.tool.Main;
 import jdk.internal.clang.Cursor;
 import jdk.internal.clang.CursorKind;
 import jdk.internal.clang.Diagnostic;
@@ -41,7 +42,6 @@ import jdk.internal.clang.TranslationUnit;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.util.ArrayList;
 import java.util.Collection;
 import java.util.LinkedHashMap;
 import java.util.List;
@@ -53,12 +53,14 @@ import java.util.stream.Stream;
 class MacroParserImpl implements JextractTask.ConstantParser {
 
     private Reparser reparser;
-    TreeMaker treeMaker = new TreeMaker();
+    TreeMaker treeMaker;
+    MacroTable macroTable;
 
     public MacroParserImpl(TreeMaker treeMaker, TranslationUnit tu, Collection<String> args) {
         try {
             this.reparser = new ClangReparser(tu, args);
             this.treeMaker = treeMaker;
+            this.macroTable = new MacroTable();
         } catch (IOException | Index.ParsingFailedException ex) {
             this.reparser = Reparser.DUMMY;
         }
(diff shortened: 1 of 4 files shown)


src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/MacroParserImpl.java
```diff
@@ -61,44 +72,26 @@ public MacroParserImpl(TranslationUnit tu, Collection<String> args) {
...
+                return Optional.empty();
+            } else if (tokens.length == 2) {
+                //check for fast path
+                Integer num = toNumber(tokens[1]);
+                if (num != null) {
+                    return Optional.of(treeMaker.createMacro(cursor, name, Type.primitive(Type.Primitive.Kind.Int), (long)num));
+                }
             }
+            macroTable.enterMacro(name, tokens, cursor);
             return Optional.empty();
```
sundararajana, line 91: Better to have a DEBUG message for skipped macros. Useful for debugging. Perhaps try..catch RuntimeException with DEBUG/rethrow in catch is good as well.

mcimadamore: this is not skipped though - it is just deferred

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/MacroParserImpl.java
```diff
@@ -230,34 +127,242 @@ public ClangReparser(TranslationUnit tu, Collection<String> args) throws IOExcep
...
+                    }
+                    case StrLiteral: {
+                        String value = result.getAsString();
+                        newEntry = entry.success(typeMaker.makeType(decl.type()), value);
+                        break;
+                    }
+                    default: {
+                        Type type = decl.type().equals(decl.type().canonicalType()) ?
+                                null : typeMaker.makeType(decl.type());
+                        newEntry = entry.failure(type);
```
JornVernee, line 323: Not sure what's going on here exactly, but AFAICS passing null to `failure` will either return null, making newEntry null, or throw an exception.

Would it be clearer to assign null to newEntry directly? (Also, maybe using a switch expression?)

mcimadamore: Not sure I understand. If the old entry is an UnparsedMacro, then failure(null) is != from failure(T) where T != null; the former returns null (meaning no entry), while the latter creates a new recoverable failure entry. The exception you mention is thrown regardless of the type - and occurs when failure is called on an entry for which no state transition associated to failure exists.

I can update the code to use a switch expression, but I think that encapsulating the state transition logic into the methods of the entry leads to better code - this is essentially an OO implementation of a finite state machine.

PR #267, opened 2020-07-27
8249879: Split MemorySegment and MemoryAddress (jextract edition)

mcimadamore:
This patch fixes uses of memory segments and address following the big refactoring introduced by https://git.openjdk.java.net/panama-foreign/pull/260 from sources and tests.

---------
### Progress

### Issue
 * [JDK-8249879](https://bugs.openjdk.java.net/browse/JDK-8249879): Split MemorySegment and MemoryAddress ⚠️ Title mismatch between PR and JBS.

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/267/head:pull/267`
`$ git checkout pull/267`


src/jdk.incubator.jextract/share/classes/jdk/internal/clang/Cursor.java
```diff
@@ -335,26 +335,15 @@ public Cursor getSpecializedCursorTemplate() {
         return new Cursor(Index_h.clang_getSpecializedCursorTemplate(cursor));
     }
```
sundararajana, line 337: This diff removes the callback cache fix for jextract performance. I believe this is unintentional merge artefact.

src/jdk.incubator.jextract/share/classes/jdk/internal/clang/Utils.java
```diff
@@ -27,46 +27,27 @@
...
     }
 
-    static void setInt(MemoryAddress addr, int i) {
-        INT_VH.set(addr, i);
+    static void setLong(MemorySegment addr, long i) {
+        MemoryAccess.setLong(addr, i);
     }
 
-    static int getLong(MemoryAddress addr) {
-        return (int)LONG_VH.get(addr);
```
sundararajana, line 53: Do we need this renaming wrapper? BTW most of this Utils class is just forwarding to MemoryAccess. Might as well use MemoryAccess directly in callsties?

mcimadamore: We could, yes

PR #269, opened 2020-07-28
Fix wrong ret_addr_offset value for some native calls.

JornVernee:
Hi,

This patch fixes a problem observed when running some of the jextract samples with intrinsics enabled.

The problem is that safepoint information is stored based on code offsets in C2, and for native calls, the code offset was wrong sometimes due to the assembler emitting different code for near and far calls.

The fix is to force the emission of a far call, so that the code offset is always correct. This follows what currently happens for normal runtime calls as well.

The patch also adds an assertion to verify that the given code offset is correct when emitting code. The assertion correctly fires on the current intrinsics tests without the fix.

Thanks,
Jorn

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/269/head:pull/269`
`$ git checkout pull/269`


iwanowww:
Looks good.

PR #278, opened 2020-08-04
Nio benchmark

slowhog:
This added a simple benchmark to get DirectoryStream implemented with Panama, compared to a simplified JNI version also the stock implementation. Currently MacOS and Linux are included, for other platform, default kicks in and abort with exception.

To enable platform dependant code needed by Panama, we add a folder, test/os, where we can put platform override implementation. This allows minimum change to the code layout, however, the colocation is less than desired.

Support package of jdk.incubator.jbind.core is added, as that's what needed for jbind generated code. This is basically what jextract generated RuntimeHelper.

LibC.java is generated by jbind. The command used to generate the file is gencode.sh, the patch also contains argument files.

When generate code with jbind, there are redundant files no needed for the benchmark, use `git status` to identify those and it's safe to remove them and only keep the LibC.java.

---------
### Progress

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/278/head:pull/278`
`$ git checkout pull/278`


slowhog:
It's just a reference, the pull request should be split up and create PRs for appropriate upstream

PR #279, opened 2020-08-05
8251047: GC stackwalking doesn't work when intrinsics are enabled

JornVernee:
Hi,

This patch fixes GC stack walking for optimized native calls.

For calls in compiled frames, the callee argument oops are processed using special case code in `CompiledMethod::preserve_callee_argument_oops`. This code relies on the original bytecode to find the signature of the callee, but for optimized native calls this doesn't make sense, since the original call is replaced, and some of the arguments (notably some that are oops) are discarded. In the case of an optimized native call, we can skip all callee argument oop processing, since we can't pass any oops to native calls any ways.

Since the current code was already looking at debug info to find the signature of the callee, I piggy-backed on that to add a flag for optimized native calls. If the flag is set, we return early, skipping any processing.

(I left in the assert I used to debug this, since it gives the failure a better error message, should we run into this again in the future)

Thanks,
Jorn

---------
### Progress

### Issue
 * [JDK-8251047](https://bugs.openjdk.java.net/browse/JDK-8251047): GC stackwalking doesn't work when intrinsics are enabled

### Reviewers
 * Maurizio Cimadamore ([mcimadamore](@mcimadamore) - Committer) ⚠️ Review applies to 67d7c0a5384ea4e3ad94f4bb27c8d9600b66c017

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/279/head:pull/279`
`$ git checkout pull/279`

### DIFF
--- a/src/hotspot/cpu/x86/frame_x86.cpp
+++ b/src/hotspot/cpu/x86/frame_x86.cpp
@@ -343,6 +343,11 @@ frame frame::sender_for_entry_frame(RegisterMap* map) const {
   assert(map->include_argument_oops(), "should be set by clear");
   vmassert(jfa->last_Java_pc() != NULL, "not walkable");
   frame fr(jfa->last_Java_sp(), jfa->last_Java_fp(), jfa->last_Java_pc());
+
+  if (jfa->saved_rbp_address()) {
+    update_map_with_saved_link(map, jfa->saved_rbp_address());
+  }
+
   return fr;
 }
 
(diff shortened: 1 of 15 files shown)


iwanowww:
Not sure it's the best way to fix the problem (need to look into the code more closely), but the fix looks good for now.

JornVernee:
Thanks for the reviews.

When running the Upcalls benchmark I ran into another crash that looks related, so I'm looking into that before integrating.

JornVernee:
@iwanowww Thanks for the review.

> Still, IMO the best solution would be to adjust CallNative so it always kills RBP in case state transition is present.

I'm already doing this, see the changes in lcm.cpp RBP seems to need special handling beyond killing it in the IR. Looking at some of the runtime stub code, these are also explicitly saving RBP and creating an oop map for it.

I have to look at this in more detail to figure out if we can refactor the current native call support to work more like that of the runtime stubs. (though that adds a lot of indirection, so maybe it's worth it to try and remove the special handling of RBP in the compiler, but that probably also touches more code and implicit assumptions).

I added a comment in the last revision that clarifies this, as well as only killing callee saved regs in case we are actually doing a thread state transition.

PR #280, opened 2020-08-06
8251256: jextrac code generated for nested structs, unions is incorrect

mcimadamore:
This patch rectifies the behavior of jextract with respect to nested structs and/or unions.

More specifically, jextract failed to recursively generate interfaces for anonymous structs and unioins; this led to a situation (as described in the JBS issue) where the only operation a client can do is to retrieve a segment for the nested struct/union, but all nested accessors are not available.

The fix allows jextract to recurse; more specifically, if a struct/union field type is itself an anonymous struct/union, the interface for the anonymous struct/union will be generated (nested inside the current struct interface).

So, in the example submitted in the JBS issue, it would be possible to poke at the nested union field, as follows:

```java
 MemorySegment rdma = ibv_send_wr.wr$addr(someSegment);
 ibv_send_wr.wr.rdma.remote_addr$get(rdma);
```

Note the nesting: `ibv_send_wr.wr.rdma` where:

* `ibv_send_wr` is the enclosing struct declaration
* `wr` is the union declaration, nested inside the struct
* `rdma` is a struct that is one of the union alternatives

To make the code more uniform, I've reworked JavaSourceBuilder a little - more specifically:

* JavaSourceBuilder is aware of the fact that builders can *nest* into each other
* Rather than dumping contents of a child builder into the parent, we use delegation, to make sure that all calls to `append` eventually hit the `StringBuilder` in the toplevel builder --- effectively removing the need to copy
* each builder has its own name context (to fixup duplicate names); this follows from the fact that each nested class introduces a fresh scope, so no clashes with names defined outside are possible

While I've added no test, I found out that an existing test (Test8244512) was already triggering the scenario described in the JBS issue. I've tweaked the test to take into account the fact that now jextract correctly nests inner structs, as well as the fact that jextract is now adding interfaces even for the anon structs/union attached to struct/union fields.

---------
### Progress

### Issue
 * [JDK-8251256](https://bugs.openjdk.java.net/browse/JDK-8251256): jextrac code generated for nested structs, unions is incorrect

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/280/head:pull/280`
`$ git checkout pull/280`


slowhog:
Overall looks good, I tested it on MacOS and seems like we get pretty good coverage. Some minor issue with indentation of generated code, but that's minor and can be fixed later. The only case missed seems to be when an anonymous type  in a function argument like following,
```code=c
void foo(struct {int x; int y; } *p);
```
I doubt that's important as while it's legal, sane programmer should not code like that as the struct type is not exposed. So it seems fit jextract not expose the layout either.

FWIW, there is a header file I used for testing anonymous types and dependency, the result from jextract did cover all anonymous types. There are still something both jbind/jextract cannot handle, but that's beyond this issue.

https://github.com/slowhog/panama-foreign/blob/jbind/test/jdk/java/jextract/anonymousDecl.h

sundararajana:
I tested the patch on Mac and ran all samples as well. All fine.

mcimadamore:
> Overall looks good, I tested it on MacOS and seems like we get pretty good coverage. Some minor issue with indentation of generated code, but that's minor and can be fixed later. 

Very good point - I knew about that and meaning to fix that (esp. after introducing the delegation scheme) - but for some reason I forgot - I will upload another iteration which fixes that.

The only case missed seems to be when an anonymous type in a function argument like following,
> 
> ```
> void foo(struct {int x; int y; } *p);
> ```
> 
> I doubt that's important as while it's legal, sane programmer should not code like that as the struct type is not exposed. So it seems fit jextract not expose the layout either.

Yeah - this seems a bit odd. To be clear, this patch doesn't have any aspiration to be a "full solution" - it merely allow jextract to look a bit deeper into anonymous struct/unions in field decls. We might need more tweaking - either in code generation, or in the how the declaration API represents this stuff.

> 
> FWIW, there is a header file I used for testing anonymous types and dependency, the result from jextract did cover all anonymous types. There are still something both jbind/jextract cannot handle, but that's beyond this issue.
> 
> [](https://github.com/slowhog/panama-foreign/blob/jbind/test/jdk/java/jextract/anonymousDecl.h)

Thanks - yes, it would be good to go over it and see what's missing.

sundararajana:
Additional change looks good.

PR #283, opened 2020-08-17
8251893: jextract does not generate source for ConstantHelper even when --source is used

sundararajana:
ConstantHelper.java is generated when --source is used

---------
### Progress

### Issue
 * [JDK-8251893](https://bugs.openjdk.java.net/browse/JDK-8251893): jextract does not generate source for ConstantHelper even when --source is used

### Reviewers
 * Maurizio Cimadamore ([mcimadamore](@mcimadamore) - Committer) ⚠️ Review applies to cca7ef8dbc8210963d0c31124742d7320d2978f0

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/283/head:pull/283`
`$ git checkout pull/283`


src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/SourceConstantHelper.java
```diff
@@ -0,0 +1,506 @@
...
+import java.net.URI;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Objects;
+import java.util.Set;
+
+import static java.lang.invoke.MethodType.methodType;
+
+// generates ConstantHelper as java source
+class SourceConstantHelper implements ConstantHelper {
```
mcimadamore, line 46: Shouldn't this extend JavaSourceBuilder, which has all the interesting source generation methods?

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/SourceConstantHelper.java
```diff
@@ -391,36 +396,54 @@ private String getConstantStringFieldName(String javaName) {
         return javaName + "$STR_CONSTANT_";
     }
 
-    private String emitConstantStringField(String javaName, Object value) {
+    private String emitConstantAddress(String javaName, Class<?> type, Object value) {
+        boolean isAddr = type == MemoryAddress.class;
         incrAlign();
         indent();
         String fieldName = getConstantStringFieldName(javaName);
```
mcimadamore, line 403: This should not be named "StringFieldName" since it now refers to both String constants and pointer constants?
Also, perhaps it would be better to just have two methods `emitXYZ` one for strings, and one for address constants, instead of having one which keeps checking `isAddr` ?

PR #285, opened 2020-08-19
8252016: jextract should handle va_list

sundararajana:
adding va_list support for jextract

---------
### Progress

### Issue
 * [JDK-8252016](https://bugs.openjdk.java.net/browse/JDK-8252016): jextract should handle va_list

### Reviewers
 * Maurizio Cimadamore ([mcimadamore](@mcimadamore) - Committer) ⚠️ Review applies to 6520f0aae72d494459224c5a19239ee3410d1e15

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/285/head:pull/285`
`$ git checkout pull/285`

### DIFF
--- a/src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/OutputFactory.java
+++ b/src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/OutputFactory.java
@@ -253,7 +253,7 @@ public class OutputFactory implements Declaration.Visitor<Void, Declaration> {
 
         if (isVaList(descriptor)) {
             MemoryLayout[] argLayouts = descriptor.argumentLayouts().toArray(new MemoryLayout[0]);
-            argLayouts[argLayouts.length - 1] = CSupport.C_VA_LIST.withName(VA_LIST_TAG);
+            argLayouts[argLayouts.length - 1] = CSupport.C_VA_LIST;
             descriptor = descriptor.returnLayout().isEmpty()?
                     FunctionDescriptor.ofVoid(argLayouts) :
                     FunctionDescriptor.of(descriptor.returnLayout().get(), argLayouts);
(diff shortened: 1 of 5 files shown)


src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/resources/RuntimeHelper.java.template
```diff
@@ -69,11 +70,21 @@ public class RuntimeHelper {
...
+        return false;
+    }
+
     public static final MethodHandle downcallHandle(LibraryLookup[] LIBRARIES, String name, String desc, FunctionDescriptor fdesc, boolean variadic) {
         return lookup(LIBRARIES, name).map(
                 addr -> {
                     MethodType mt = MethodType.fromMethodDescriptorString(desc, LOADER);
-                    return variadic ?
+                    return variadic || isVaList(fdesc)?
                         VarargsInvoker.make(addr, mt, fdesc) :
```
mcimadamore, line 88: I don't get this change - we don't want a vararg invoker for a valist call, right? The invoker will most surely not do the right thing here.

Also, the test seems to show that no Java varargs invocation is being applied (which is correct), so what is the reasons for the changes in RuntimeHelper?

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/OutputFactory.java
```diff
@@ -220,6 +222,22 @@ public Void visitScoped(Declaration.Scoped d, Declaration parent) {
         return null;
     }
 
+    private static final boolean isVaList(FunctionDescriptor desc) {
```
mcimadamore, line 225: Ugh - too bad there's no way to detect this in libclang - not your fault of course!

mcimadamore (commented):
Looks good, except for some chanegs in RuntimeHelper which are not very clear to me

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/resources/RuntimeHelper.java.template
mcimadamore: missng space between `variadic` and `?`

mcimadamore (approved):
Looks good

PR #287, opened 2020-08-20
8252121: jextract generated code fails with ABI for typedefed array type parameters

sundararajana:
TypeMaker now special cases typedef'ed arrays for argument types for functions.

---------
### Progress

### Issue
 * [JDK-8252121](https://bugs.openjdk.java.net/browse/JDK-8252121): jextract generated code fails with ABI for typedefed array type parameters

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/287/head:pull/287`
`$ git checkout pull/287`

### DIFF
--- a/src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/TypeMaker.java
+++ b/src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/TypeMaker.java
@@ -235,6 +235,14 @@ class TypeMaker {
         }
 
         @Override
+        public Type visitDelegated(Type.Delegated t, Void aVoid) {
+            if (t.kind() == Delegated.Kind.TYPEDEF && t.type() instanceof Type.Array) {
+                return visitArray((Type.Array)t.type(), aVoid);
+            }
+            return visitType(t, aVoid);
+        }
+
+        @Override
         public Type visitType(Type t, Void aVoid) {
             return t;
         }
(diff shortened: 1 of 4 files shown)


src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/TypeMaker.java
```diff
@@ -234,6 +234,14 @@ public Type visitArray(Type.Array t, Void aVoid) {
             return Type.pointer(t.elementType());
         }
 
+        @Override
+        public Type visitDelegated(Type.Delegated t, Void aVoid) {
+            if (t.kind() == Delegated.Kind.TYPEDEF && t.type() instanceof Type.Array) {
+                return visitArray((Type.Array)t.type(), aVoid);
```
mcimadamore, line 240: shouldn't we just recurse on the type() field? what if it's a typedef of a typedef of a typedef....

mcimadamore (commented):
I believe there are some recursion issues with multiple typedefs (we need also a test for that)

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/TypeMaker.java
sundararajana: type() seems to return canonical type. I can another test for typedef of typedef.

mcimadamore (approved):
That's ok for now - but I guess that, if this work, the declaration AST is wrong when multiple typedefs are involved.

PR #288, opened 2020-08-24
Unsafe ParallelSum test is broken

mcimadamore:
The numbers for the unsafe parallel sum tests are artificially low; after some debugging I realized that the test was not summing up all components.
I then plugged in the unsafe fork join action into our existing memory segment spliterator test, and verified that indeed that fork join action was violating all assertions.

I then fixed all the issues, verified that assertions were satisfied and then plugged the fixed action back into the microbenchmark.

Unsurprisingly, the perf numbers of the unsafe part are now in sync with those obtained from the segment counterpart.

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/288/head:pull/288`
`$ git checkout pull/288`

### DIFF
--- a/test/micro/org/openjdk/bench/jdk/incubator/foreign/ParallelSum.java
+++ b/test/micro/org/openjdk/bench/jdk/incubator/foreign/ParallelSum.java
@@ -190,7 +190,7 @@ public class ParallelSum {
 
     @Benchmark
     public int unsafe_parallel() {
-        return new SumUnsafe(address, 0, ALLOC_SIZE).invoke();
+        return new SumUnsafe(address, 0, ALLOC_SIZE / CARRIER_SIZE).invoke();
     }
 
     static class SumUnsafe extends RecursiveTask<Integer> {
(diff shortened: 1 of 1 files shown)


JornVernee (approved):

sundararajana (approved):

test/micro/org/openjdk/bench/jdk/incubator/foreign/ParallelSum.java
```diff
@@ -209,15 +209,19 @@ public int unsafe_parallel() {
...
         protected Integer compute() {
             if (length > SPLIT_THRESHOLD) {
-                SumUnsafe s1 = new SumUnsafe(address, start, length / 2);
-                SumUnsafe s2 = new SumUnsafe(address, length / 2, length / 2);
+                int rem = length % 2;
+                int split = length / 2;
+                int lobound = split;
+                int hibound = lobound + rem;
+                SumUnsafe s1 = new SumUnsafe(address, start, lobound);
+                SumUnsafe s2 = new SumUnsafe(address, start + lobound, hibound);
```
JornVernee, line 217: Judging by this code the third parameter is no longer a length, but an upper bound? Maybe you also want to change the field name in that case.

mcimadamore: Uhm - it's still the length (in elements) associated with the task. I'll leave the field name as is.

PR #292, opened 2020-08-25
updating samples. changed constants class to use package constructor

sundararajana:
fixed to use package private constructors for constants classes. piggybacking to
include latest samples.

---------
### Progress

### Reviewers
 * Jorn Vernee ([jvernee](@JornVernee) - Committer) ⚠️ Review applies to 69ce50de0e84310fff526d53e22bc5c72634485d

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/292/head:pull/292`
`$ git checkout pull/292`

### DIFF
--- a/doc/panama_jextract.md
+++ b/doc/panama_jextract.md
@@ -754,3 +754,137 @@ java -XstartOnFirstThread -Dforeign.restricted=permit --add-modules jdk.incubato
     -Djava.library.path=.:/System/Library/Frameworks/OpenGL.framework/Versions/Current/Libraries/ Teapot.java $*
 
 ```
+
+## Using time.h (Mac OS)
+
+### jextract sqlite3.h
+
+
+```sh
+
+jextract -t org.unix \
+  -I /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include \
+   /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/time.h
+
+```
+
+### Java program that uses POSIX time library
+
+```java
+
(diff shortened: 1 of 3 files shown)


doc/panama_jextract.md
```diff
@@ -754,3 +754,137 @@ java -XstartOnFirstThread -Dforeign.restricted=permit --add-modules jdk.incubato
...
+
+    public static void main(String[] args) {
+        if (args.length == 0) {
+            System.err.println("java ASTPrinter <C source or header>");
+            System.exit(1);
+        }
+
+        // parse the C header/source passed from the command line
+        var index = clang_createIndex(0, 0);
+        var tu = clang_parseTranslationUnit(index, toCString(args[0]),
```
JornVernee, line 842: This `toCString` should be allocated in the native scope below, so it gets cleaned up as well.

doc/panama_jextract.md
```diff
@@ -754,3 +754,137 @@ java -XstartOnFirstThread -Dforeign.restricted=permit --add-modules jdk.incubato
...
+            NULL, 0, NULL, 0, CXTranslationUnit_None());
+
+        try (var scope = NativeScope.unboundedScope()) {
+            // array trick to update within lambda
+            var level = new int[1];
+            var visitor = new MemorySegment[1];
```
JornVernee, line 848: FWIW, I think you could avoid both of these by using an anonymous class instead of a lambda

JornVernee (approved):

doc/panama_jextract.md
sundararajana: agree

doc/panama_jextract.md
sundararajana: While we can take care of level param that way, the function pointer address (visitor) is used inside to call recursively. Unfortunately we need array hack for that.

PR #294, opened 2020-08-25
8251047: GC stackwalking doesn't work when intrinsics are enabled (followup)

JornVernee:
Hi,

This PR is a followup on the earlier https://github.com/openjdk/panama-foreign/pull/279 PR. It adds a jtreg test that stresses the problematic stack walking fixed in the previous PR, and in particular the use-case that was failing.

This relies on a new WhiteBox API to manually (and reliably) trigger stack frame walking and verification.

Thanks,
Jorn

---------
### Progress

### Issue
 * [JDK-8251047](https://bugs.openjdk.java.net/browse/JDK-8251047): GC stackwalking doesn't work when intrinsics are enabled ⚠️ Title mismatch between PR and JBS.

### Reviewers
 * Vladimir Ivanov ([vlivanov](@iwanowww) - Committer) ⚠️ Review applies to 94157625b43ed3f89e0649e1f54af8a30b521569

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/294/head:pull/294`
`$ git checkout pull/294`


iwanowww (approved):
I suggest making tracing output in WB_VerifyFrames optional. 

Otherwise, looks good.

PR #297, opened 2020-09-01
8252563: Field VarHandle generation in --source mode is missing the PathElement.groupElement argument

JornVernee:
Hi,

This PR fixes an issue where var handles in structs were not being created based on the layout of the parent struct, and were not using a PathElement.groupElement(...) argument to look up the right struct field, in --source mode, resulting in the created var handle not having the right offset. The core change is found in SourceConstantHelper#emitVarHandleField.

To try and clear things up a bit, I've also split the addVarHandle method in ConstantHelper into addGlobalVarHandle, and addFieldVarHandle with only the applicable parameter types, at least at the ConstantHelper level, so that calling code doesn't have to pass a bunch of nulls when calling addVarHandle for a global variable (the nulls being for the parent layout and field name).

To test this change, I've re-used the existing LibStructTest by running the test twice; once under the existing jtreg jextract driver that uses the binary constant helper, and once under a new jtreg jextract driver that first generates sources and then compiles them.

Thanks,
Jorn

---------
### Progress

### Issue
 * [JDK-8252563](https://bugs.openjdk.java.net/browse/JDK-8252563): Field VarHandle generation in --source mode is missing the PathElement.groupElement argument

### Reviewers
 * Athijegannathan Sundararajan ([sundar](@sundararajana) - Committer) ⚠️ Review applies to 72071c025818b9b2073c29eb16ce5cc1f2cd2281

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/297/head:pull/297`
`$ git checkout pull/297`

### DIFF
--- a/src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/ClassConstantHelper.java
+++ b/src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/ClassConstantHelper.java
@@ -193,7 +193,7 @@ class ClassConstantHelper implements ConstantHelper {
     }
 
     @Override
-    public DirectMethodHandleDesc addFieldVarHandle(String javaName, String nativeName, MemoryLayout layout, Class<?> type, String __, MemoryLayout parentLayout) {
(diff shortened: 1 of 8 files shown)


sundararajana (approved):
__ may be avoided. Suggest "ignored" instead.

PR #300, opened 2020-09-02
8252634: jextract should generate type annotations for C types

sundararajana:
@C annotation generated for C types

---------
### Progress

### Issue
 * [JDK-8252634](https://bugs.openjdk.java.net/browse/JDK-8252634): jextract should generate type annotations for C types

### Reviewers
 * Maurizio Cimadamore ([mcimadamore](@mcimadamore) - Committer) ⚠️ Review applies to a8744715fbfe6c689db6124cb9e585f7a1ad6f53

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/300/head:pull/300`
`$ git checkout pull/300`


src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/AnnotationWriter.java
```diff
@@ -0,0 +1,126 @@
...
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
```
JornVernee, line 51: There are 2 copyright headers here!

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/AnnotationWriter.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    @Override
+    public String visitDelegated(Type.Delegated t, Void aVoid) {
+        if (t.kind() == Type.Delegated.Kind.TYPEDEF) {
+            return t.name().get();
+        } else if (t.kind() == Type.Delegated.Kind.POINTER) {
+            String typeStr = t.type().accept(this, null);
+            if (t.type() instanceof Type.Function) {
+                return typeStr.replace("(", "(*)(");
```
JornVernee, line 73: This might be problematic for multiple nested function types, e.g.

```Java
void (int, void (*)(void)) -> void (*)(int, void (*)(*)(*)(void))
```

The return type could also be a function pointer, so it's not always the first occurrence that needs to be replaced either.

JornVernee: Might just want to revisit the return type and argumentTypes, similar to what you do in visitFunction, but use `(*)(` as a joiner instead of `(`.

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/StructBuilder.java
```diff
@@ -138,37 +142,37 @@ void addLayoutGetter(String javaName, MemoryLayout layout) {
...
 
     @Override
-    void addSetter(String javaName, String nativeName, MemoryLayout layout, Class<?> type) {
+    void addSetter(String javaName, String nativeName, MemoryLayout layout, Class<?> type, String anno) {
         incrAlign();
         indent();
-        String param = MemorySegment.class.getName() + " seg";
-        append(PUB_MODS + "void " + javaName + "$set(" + param + ", " + type.getName() + " x) {\n");
+        String param = MemorySegment.class.getSimpleName() + " seg";
+        append(PUB_MODS + "void " + javaName + "$set(" + this.anno + " " + param + ", " + anno + " " + type.getSimpleName() + " x) {\n");
```
JornVernee, line 166: Maybe rename `this.anno` to `parentAnno` to avoid the name conflict.

test/jdk/tools/jextract/Test8252634.java
```diff
@@ -0,0 +1,88 @@
...
+            checkAnnotation(yGetter.getParameters()[0].getAnnotatedType(), "struct Point");
+            checkAnnotation(yGetter.getAnnotatedReturnType(), "int");
+            Method xSetter = findMethod(pointClass, "x$set", MemorySegment.class, int.class);
+            checkAnnotation(xSetter.getParameters()[0].getAnnotatedType(), "struct Point");
+            checkAnnotation(xSetter.getParameters()[1].getAnnotatedType(), "int");
+            Method ySetter = findMethod(pointClass, "y$set", MemorySegment.class, int.class);
+            checkAnnotation(ySetter.getParameters()[0].getAnnotatedType(), "struct Point");
+            checkAnnotation(ySetter.getParameters()[1].getAnnotatedType(), "int");
+            Method allocate = findMethod(pointClass, "allocate");
+            checkAnnotation(allocate.getAnnotatedReturnType(), "struct Point");
```
JornVernee, line 78: What about scoped allocate, allocate array, and indexed setters?

test/jdk/tools/jextract/test8252634.h
```diff
@@ -0,0 +1,38 @@
...
+#endif // __cplusplus
+
+struct Point {
+    int x; int y;
+};
+
+struct Point* make(int x, int y);
+
+void func(int (*callback)(int));
+
```
JornVernee, line 35: typedef is missing as well I guess.

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/resources/C.java.template
```diff
@@ -0,0 +1,21 @@
+// Generated by jextract
```
mcimadamore, line 1: Not sure about using a template for this. If we do this, we severely limit ability for 3rd parties to write an annotation processor (since the annotation name/package is not stable).

mcimadamore (approved):
In general this looks good. I commented on the location of the C annotation; I believe it's time we consider moving the jextract runtime support into a more blessed place (e.g. some exported `runtime` package inside the jextract module). If we do that, the drawback is that generated code would depend on the jextract module - but I think that's acceptable? Anyway, approving for now, what I'm requesting is a deeper change which should be handled separately.

JornVernee (approved):

PR #306, opened 2020-09-03
8252770: MemorySegment::allocateNative should honor direct memory limits

mcimadamore:
Currently, `MemorySegment::allocateNative` does not honor the memory limits set for direct buffers (`-XX:MaxDirectMemorySize` and `-XX:+PageAlignDirectMemory`).This patch rectifies that. The changes are pretty simple, the main issue was that most of the methods are defined away in the `Bits` class which belongs to the nio package, so the usual access dance was necessary.

Since segments can already be allocated with an alignment, I slightly tweaked the alignment logic to compute the max between page size and requested alignment, in case the option for aligning direct memory chunks to page boundaries is set.

Tweak NativeMemorySegmentImpl to respect direct memory settings

---------
### Progress

### Issue
 * [JDK-8252770](https://bugs.openjdk.java.net/browse/JDK-8252770): MemorySegment::allocateNative should honor direct memory limits

### Reviewers
 * [Paul Sandoz](https://openjdk.java.net/census#psandoz) (@PaulSandoz - Committer) ⚠️ Review applies to 015ee74868ce9c13494f3cc17938a9fb240293f0

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/306/head:pull/306`
`$ git checkout pull/306`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/NativeMemorySegmentImpl.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/NativeMemorySegmentImpl.java
@@ -48,7 +48,7 @@ public class NativeMemorySegmentImpl extends AbstractMemorySegmentImpl {
 
     // The maximum alignment supported by malloc - typically 16 on
     // 64-bit platforms and 8 on 32-bit platforms.
-    private final static long MAX_ALIGN = Unsafe.ADDRESS_SIZE == 4 ? 8 : 16;
+    private final static long MAX_MALLOC_ALIGN = Unsafe.ADDRESS_SIZE == 4 ? 8 : 16;
 
     private static final boolean skipZeroMemory = GetBooleanAction.privilegedGetProperty("jdk.internal.foreign.skipZeroMemory");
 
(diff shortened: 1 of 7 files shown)


src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/NativeMemorySegmentImpl.java
```diff
@@ -85,18 +83,24 @@ Object base() {
...
 
     public static MemorySegment makeNativeSegment(long bytesSize, long alignmentBytes) {
-        long alignedSize = bytesSize;
-
-        if (alignmentBytes > MAX_ALIGN) {
-            alignedSize = bytesSize + (alignmentBytes - 1);
+        if (VM.isDirectMemoryPageAligned()) {
+            alignmentBytes = Math.max(alignmentBytes, nioAccess.pageSize());
         }
+        long alignedSize = alignmentBytes > MAX_ALIGN ?
```
PaulSandoz, line 89: Minor comment, it might be a little clearer if `MAX_ALIGN` was renamed to `MALLOC_MAX_ALIGN`.

PaulSandoz (approved):

PR #307, opened 2020-09-04
8252700: jextract ignores typedef to undefined structs

sundararajana:
A class with annotated allocatePointer methods is generated

---------
### Progress

### Issue
 * [JDK-8252700](https://bugs.openjdk.java.net/browse/JDK-8252700): jextract ignores typedef to undefined structs

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/307/head:pull/307`
`$ git checkout pull/307`


mcimadamore (approved):
Looks good.
I note a mild inconsistencies in the sense that, in most emitXYZ the annotation is passed from outside, while in emitTypeDef the annotation is calculated internally.

I believe this way is totally the right way to go and that, currently, the various source builders are too "stringy" so, perhaps in the future, we can clean up some of the code by always genertating metadata in the source builders rather than output factory, and pass more types/declarations around.

PR #308, opened 2020-09-04
JDK-8248809: Consider supporting address dereference/allocation for platforms where pointers are not 64 bits

mcimadamore:
This patch makes support for addresses more uniform, in a way that scales to platforms whose pointer size might be != 64 bits. After considering a number of alternatives, we settled on something which, I think is pleasing for a number of rteasons.

First, we added a new layout, namely `MemoryLayouts.ADDRESS`. This is the layout of a machine pointer, in the platform in which the JVM runs. This is a useful piece of information to have regardless.

This new layout is then used in two places:

* In MemoryAccess, to decide whether we need a 32 or a 64 dereference handle to get/set addresses

* In NativeScope, again, to decide to which NativeScope factory we should delegate (the long vs. the int one)

This should slightly generalize what we have and also adds some extra helper methods to NativeScope which were dropped in an earlier iteration, waiting for this issue to be addressed.

I've tweaked the NativeScope test in order to stress the new address-accepting factories.

---------
### Progress

### Issue
 * [JDK-8248809](https://bugs.openjdk.java.net/browse/JDK-8248809): Consider supporting address dereference/allocation for platforms where pointers are not 64 bits

### Reviewers
 * [Paul Sandoz](https://openjdk.java.net/census#psandoz) (@PaulSandoz - Committer) ⚠️ Review applies to 4ac3aa43354dad968d608217e540b3a052c3a206
 * [Jorn Vernee](https://openjdk.java.net/census#jvernee) (@JornVernee - Committer) ⚠️ Review applies to 4ac3aa43354dad968d608217e540b3a052c3a206

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/308/head:pull/308`
`$ git checkout pull/308`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/NativeScope.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/NativeScope.java
@@ -185,7 +185,7 @@ public interface NativeScope extends AutoCloseable {
     /**
      * Allocate a block of memory in this native scope with given layout and initialize it with given address value.
      * The address value might be narrowed according to the platform address size (see {@link MemoryLayouts#ADDRESS}).
-     * The segment returned by this method is associated with a segment which cannot be closed. Moreover, the returned
+     * The segment returned by this method cannot be closed. Moreover, the returned
      * segment must conform to the layout alignment constraints.
      * @param layout the layout of the block of memory to be allocated.
      * @param value the value to be set on the newly allocated memory block.
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/NativeScope.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/NativeScope.java
@@ -36,6 +36,7 @@ import java.lang.reflect.Array;
 import java.nio.ByteOrder;
 import java.util.OptionalLong;
 import java.util.function.Function;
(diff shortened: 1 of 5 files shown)


PaulSandoz (approved):
Our exposure to 32-bit platforms is getting less and less, making it tricky to test. I suspect most users will hard code to the long carrier type without realizing it.

I admit to being a little torn over this. Unsafe either ignores the 32 MSBs of long values, or requires they are zeroed. Not ideal either.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/NativeScope.java
```diff
@@ -181,6 +182,29 @@ default MemorySegment allocate(MemoryLayout layout, double value) {
         return addr;
     }
 
+    /**
+     * Allocate a block of memory in this native scope with given layout and initialize it with given address value.
+     * The address value might be narrowed according to the platform address size (see {@link MemoryLayouts#ADDRESS}).
+     * The segment returned by this method is associated with a segment which cannot be closed. Moreover, the returned
```
JornVernee, line 188: This sentence seems strange.
```suggestion
     * The segment returned by this method cannot be closed. Moreover, the returned
```

JornVernee (approved):

PR #314, opened 2020-09-09
8252931: Extend multiple constant classes support to ClassConstantHelper

JornVernee:
Hi,

This PR extracts the multiple source file functionality from SourceConstantHelper into a decorator class, and then uses it for ClassConstantHelper as well.

(I've also done a minor cleanup where I simplified the way booleans are turned into constants, since we can now just call describeConstable() on a Boolean)

Thanks,
Jorn

---------
### Progress

### Issue
 * [JDK-8252931](https://bugs.openjdk.java.net/browse/JDK-8252931): Extend multiple constant classes support to ClassConstantHelper

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/314/head:pull/314`
`$ git checkout pull/314`

### DIFF
--- a/src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/MultiFileConstantHelper.java
+++ b/src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/MultiFileConstantHelper.java
@@ -1,0 +1,127 @@
+/*
+ * Copyright (c) 2020, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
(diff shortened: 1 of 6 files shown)


src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/MultiFileConstantHelper.java
```diff
@@ -0,0 +1,127 @@
...
+
+import javax.tools.JavaFileObject;
+import java.lang.constant.DirectMethodHandleDesc;
+import java.lang.invoke.MethodType;
+import java.util.ArrayList;
+import java.util.List;
+
+public class MultiFileConstantHelper implements ConstantHelper {
+
+    private static final int CONSTANTS_PER_CLASS = Integer.getInteger("jextract.constants.per.class", 1000);
```
sundararajana, line 38: This limit was empirically found by running source constant helper on samples (like lapack, python, libgit2). I'd recommend doing similar experiment with binary constant helper and arrive at suitable threshold.

JornVernee: I've updated the code. The limit for binary generation is set to 10000

sundararajana (approved):

PR #320, opened 2020-09-10
8253025: Add support for registering a segment with a Cleaner

mcimadamore:
This patch adds an extra API point to register a segment against a cleaner.

The basic idea is that, when a segment is registered against a cleaner, the segment's scope is register, and the scope's cleanup action is passed to the cleaner.

In practice, things are more subtle than it seems. First, we need to make sure that we cannot have races - which could result in cleanup action being called multiple times. The possible races are:

* close vs. cleaner
* cleaner vs. cleaner

We can get rid of the first type of race by adding reachability fences around the scope terminal operations. If we do that, we are guaranteed that the Cleaner cannot kick in before we are actually done with the scope.

Now, even if reachability fences can give us some guarantee that close will never run concurrently with cleaner, we're not out of the woods yet. The cleaner can in fact still be called _after_ a segment has been closed explicitly.

Now, the simplest thing to do on paper would be to check the scope liveness bit, and only call the cleanup action if the scope is still alive. But this is not possible, becaue the scope liveness bit is, well, inside the scope and, to access it, the Cleaner action must keep a strong reference to the scope being registered (which means the cleaner will never work).

To address this, I came up with a level of indirection: I replaced the references to Runnable with a new interface, called CleanupAction. This interface has methods to cleanup and to dup onto a new cleanup action, killing the previous one.

Now, segment factories create cleanup action, rather than simple Runnable. There is a static "dummy" cleanup action, which does nothing; and there is also a more complex and stateful cleanup action, which supports atomicity (this cleanup action has a boolean flag which is chcked with CAS).

So, all terminal operation in memory scope must either call cleanup() or dup() on the cleanup action, to ensure that the cleanup action attached to the scope is no longer usable upon return.

Since the logic inside the stateful cleanup action uses CAS, it is possible for multiple threads to race to cleanup the same action, and only one of them is guaranteed to win.

While the logic is tricky, and partly replicates the liveness bit in scope, the implementation strategy tries hard not to allocate more than previosly, especially when a new segment is created. The only drawback is that we need to create a new cleanup action every time we change scope ownership - but these operations are not performance critical, so I assume this is an acceptable cost.

Note that, given all our core memory access operation (in the newly define ScopeMemoryAccess) already take care of adding the require reachability fence on the scopes being accessed, it is not possible for a cleaner to kick in while a segment is being actively accessed.

---------
### Progress

### Issue
 * [JDK-8253025](https://bugs.openjdk.java.net/browse/JDK-8253025): Add support for registering a segment with a Cleaner

### Reviewers
 * [Jorn Vernee](https://openjdk.java.net/census#jvernee) (@JornVernee - Committer) ⚠️ Review applies to 49312e456aa355021bd83f02540cdb81f557a30e
 * [Paul Sandoz](https://openjdk.java.net/census#psandoz) (@PaulSandoz - Committer) ⚠️ Review applies to e5935955b9192fc26b788c5093a2dfbc45a07095

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/320/head:pull/320`
`$ git checkout pull/320`


src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
```diff
@@ -102,9 +105,11 @@ static MemoryScope createShared(Object ref, Runnable cleanupAction) {
      * a confined scope and this method is called outside of the owner thread.
      */
     final void close() {
-        justClose();
-        if (cleanupAction != null) {
-            cleanupAction.run();
+        try {
+            justClose();
+            cleanupAction.cleanup();
```
JornVernee, line 110: Won't explicitly calling close() end up calling disable() on BasicCleanupAction, and then when the cleaner calls close it will end up throwing an IllegalStateException because disable() was already called?

I guess the code that calls the cleanup action is guarding against exceptions, so this isn't a problem?

mcimadamore: Exactly - I'm relying on the fact that exceptions during cleaner actions are ok and ignored.

JornVernee (approved):

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
```diff
@@ -254,4 +267,88 @@ void justClose() {
...
+
+    /**
+     * A functional interface modelling the cleanup action associated with a scope.
+     */
+    interface CleanupAction {
+        void cleanup();
+        CleanupAction dup();
+
+        /** Dummy cleanup action */
+        CleanupAction DUMMY = new CleanupAction() {
```
PaulSandoz, line 279: Recommend renaming to `NO_ACTION`

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
```diff
@@ -254,4 +267,88 @@ void justClose() {
...
+                return this;
+            }
+        };
+    }
+
+    /**
+     * A stateful cleanup action; this action can only be called at most once. The implementation
+     * guarantees this invariant even when multiple threads race to call the {@link #cleanup()} method.
+     */
+    static abstract class BasicCleanupAction implements CleanupAction {
```
PaulSandoz, line 296: Rename to `AtMostOnceCleanupAction` or some other snappier name that conveys it's purpose. Embedded in `CleanupAction` then `CleanupAction.AtMostOnceOnly` might work.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
```diff
@@ -254,4 +267,88 @@ void justClose() {
...
+        }
+
+        final void disable() {
+            if (!CALLED.compareAndSet(this, false, true)) {
+                // This can never happen under normal circumstances. The only case where this can happen is when
+                // when two cleaners race to cleanup the same scope. It is never possible to have a race
+                // between explicit/implicit close because all the scope terminal operations have
+                // reachability fences which prevent a scope to be deemed unreachable before we are done
+                // marking the original cleanup action as "dead".
+                throw new IllegalStateException("Already cleaned");
```
PaulSandoz, line 335: I recommend not throwing, it may cause noise for the developer when they observe through say flags or events that exceptions are being thrown.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
```diff
@@ -254,4 +267,88 @@ void justClose() {
...
+            run();
+        };
+
+        @Override
+        public CleanupAction dup() {
+            disable();
+            return new BasicCleanupAction() {
+                @Override
+                void run() {
+                    BasicCleanupAction.this.run();
```
PaulSandoz, line 323: This has the effect of chaining calls, that might accumulate over time. I don't see how to avoid this unless a delegating model is used. But IIUC you want to avoid creating two objects, the "clean up" action as a `Runnable` then the "scope close action" encapsulating the "clean up" action. Perhaps when inline types are available we can do better.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
```diff
@@ -56,35 +59,35 @@ private MemoryScope(Object ref, Runnable cleanupAction) {
...
      * @return a shared memory scope
      */
-    static MemoryScope createShared(Object ref, Runnable cleanupAction) {
+    static MemoryScope createShared(Object ref, CleanupAction cleanupAction) {
         return new SharedScope(ref, cleanupAction);
     }
 
     protected Object ref;
-    protected Runnable cleanupAction;
+    protected CleanupAction cleanupAction;
```
PaulSandoz, line 90: Make `final`

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
```diff
@@ -294,6 +295,14 @@ public MemorySegment withOwnerThread(Thread newOwner) {
         }
     }
 
+    @Override
+    public void registerCleaner(Cleaner cleaner) {
+        checkAccessModes(CLOSE);
+        checkValidState();
+        MemoryScope.CleanupAction cleanupAction = scope.cleanupAction;
+        cleaner.register(this.scope, cleanupAction::cleanup);
```
PaulSandoz, line 303: If `CleanupAction` extended from `Runnable` you could pass it in directly. e.g rename `cleanup` to `run`, and flip the two methods on `BasicCleanupAction`

PaulSandoz (commented):
Nice utility.

mcimadamore:
Thanks @PaulSandoz for all the comments; some of the stuff you pointed out (e.g. chaining of calls) I was aware of - I'm trying badly to avoid allocations as much as possible - but I'll see if I can do something else.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
mcimadamore: I think this is addresses in latest iteration; I used a local class which keeps track of the "root" action, and keeps propagating it. So calls always hit the root action, no chaining.

test/jdk/java/foreign/TestCleaner.java
```diff
@@ -0,0 +1,135 @@
...
+        assertEquals(segmentState.cleanupCalls(), 1);
+    }
+
+    MemorySegment makeSegment(SegmentState segmentState) {
+        return MemorySegment.ofNativeRestricted(MemoryAddress.NULL, 10, Thread.currentThread(), segmentState::cleanup, null);
+    }
+
+    enum SegmentFunction implements Function<MemorySegment, MemorySegment> {
+        IDENTITY(Function.identity()),
+        CLOSE(s -> { s.close(); return s; }),
```
PaulSandoz, line 97: Won't that just skip cleaner registration, making the test redundant for that case? I guess for this case it should be called after registration?

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
```diff
@@ -254,4 +267,106 @@ void justClose() {
...
+            public final void cleanup() {
+                if (disable()) {
+                    doCleanup();
+                }
+            };
+
+            @Override
+            public CleanupAction dup() {
+                disable();
+                class DupAction extends AtMostOnceOnly {
```
PaulSandoz, line 325: Nice approach. Shame we cannot declare static classes within a method, since the inner class does not need to implicitly capture "this" of the outer class. I suppose we could make it static for some marginal efficiency gain reducing dependency chains.

PaulSandoz (approved):

test/jdk/java/foreign/TestCleaner.java
mcimadamore: Not really - note that the function is called after registration:
```
MemorySegment segment = makeSegment(segmentState);
        // register cleaners before
        for (int i = 0 ; i < n ; i++) {
            segment.registerCleaner(cleanerFactory.get());
        }
        segment = segmentFunction.apply(segment); // <-------------------------
```
So, the test here is useful: it tests that the cleaner registered before close() can't really call the cleanup action multiple times if the segment has also been closed explicitly.

PaulSandoz: Oh, doh!, i missed the code above.

PR #323, opened 2020-09-11
8253063: ScopedAccessError is sometimes thrown spuriously

mcimadamore:
After running the test for shared segment handshake during unrelated work, I noted that the test exhibit two kind of failure modes:

1. First, the test can sometimes fail with a raw ScopedAccessError
2. More rarely, the test sometimes even crash (this only happens in the vectorized mismatch test)

After some investigation carried out by @fisk, turns out that (2) was caused by the fact that when we exit from deoptimization we don't check for pending async exception (while we check for regulart ones). This condition was not triggered in other tests because, it seems, plain unsafe accesses are effectively atomic once intrinisfied - e.g. there can be no safepoint between check and access. But the vectorized mismatch routine is a biggie Java routine which gets some initial compilation (by C1) - and this is when the bug hit.

The problem (2) was more difficult to pinpoint - basically, there are some routines in the VM which are hostile w.r.t. async exceptions - and are marked with the special `JRT_ENTRY_NO_ASYNC` entry. One of those routines happens to be the one that throws exceptions, and this routine can safepoint too. Now, since the scope memory access can also throw (if the scope is not alive) it is possible to end up in a situation where we try to install an async exception while the VM code was already in the middle of installing one. We thought we covered this case by checking for *pending exceptions* - but the check we had was too weak. After trying with different approaches we agreed to address this more explicitly - by having the exception code set a flag on the running thread - so that we can precisely determine, when we do the handshake, as to whether the async exception should be installed or not.

With these fixes, the test passes and no more spurious failures can be observed (I left it running for a long time ;-)).
I've beefed up the test by always having it use max number of available processors, and also by lowering the delay time associated with the closing thread. The longer the wait, the less frequently transient issues such as (1) can be observed.

---------
### Progress

### Issue
 * [JDK-8253063](https://bugs.openjdk.java.net/browse/JDK-8253063): ScopedAccessError is sometimes thrown spuriously

### Reviewers
 * [Erik Österlund](https://openjdk.java.net/census#eosterlund) (@fisk - no project role)

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/323/head:pull/323`
`$ git checkout pull/323`

### DIFF
--- a/src/hotspot/share/c1/c1_Runtime1.cpp
+++ b/src/hotspot/share/c1/c1_Runtime1.cpp
@@ -638,6 +638,7 @@ address Runtime1::exception_handler_for_pc(JavaThread* thread) {
   {
     // Enter VM mode by calling the helper
     ResetNoHandleMark rnhm;
+    ExceptionHandlingMark ehm(thread);
     continuation = exception_handler_for_pc_helper(thread, exception, pc, nm);
   }
   // Back in JAVA, use no oops DON'T safepoint
--- a/src/hotspot/share/jvmci/jvmciRuntime.cpp
+++ b/src/hotspot/share/jvmci/jvmciRuntime.cpp
@@ -375,6 +375,7 @@ address JVMCIRuntime::exception_handler_for_pc(JavaThread* thread) {
   {
     // Enter VM mode by calling the helper
     ResetNoHandleMark rnhm;
+    ExceptionHandlingMark ehm(thread);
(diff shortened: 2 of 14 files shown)


fisk (approved):
This looks good. Sorry I didn't catch this earlier.

PaulSandoz (approved):
ScopedAccessException -> ScopedAccessError looks good.

The HotSpot changes look ok to me, although i lack understanding of async exceptions.

PR #327, opened 2020-09-15
8248560: Specify the behaviour of the ForeignLinker returned by CSupport::getSystemLinker

JornVernee:
Hi,

This patch creates a new class called CLinker, which extends ForeignLinker. The new class documents how linking works in the class's javadoc. This is done in a separate class since the specification is only relevant to linkers that support the C ABI.

With the new class there was no more need to have CSupport, so the contents of CSupport has all been move to CLinker.

After some offline discussion with Maurizio, we realized that it doesn't make much sense to expose platform specific layout constants, since it is not possible to combine them with a platform specific linker any way, so for a particular platform, there is only one set of layouts that can be used, namely the ones that are dynamically picked based on the current platform. For that reason, all platform specific layouts have instead been moved to an internal PlatformLayouts class. In order for downstream APIs (like jextract) to be able to filter out the ABI attributes from layouts, I've added a String constant to CLinker that is used as a name prefix for all the ABI attribute names, so clients can for instance filter attribute names using `.startsWith(ABI_ATTR_NAME)`.

The ForeignLinker::name method has been dropped in favor of an internal enum.

The rest of the patch is just mechanical CSupport -> CLinker renames.

Thanks,
Jorn

---------
### Progress

### Issue
 * [JDK-8248560](https://bugs.openjdk.java.net/browse/JDK-8248560): Specify the behaviour of the ForeignLinker returned by CSupport::getSystemLinker

### Reviewers
 * [Athijegannathan Sundararajan](https://openjdk.java.net/census#sundar) (@sundararajana - Committer) ⚠️ Review applies to bbd7d1cb2e88ab06d4048989b3728fc65cc21dff

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/327/head:pull/327`
`$ git checkout pull/327`


src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
```diff
@@ -54,132 +96,430 @@ private CSupport() {}
      * @throws IllegalAccessError if the runtime property {@code foreign.restricted} is not set to either
      * {@code permit}, {@code warn} or {@code debug} (the default value is set to {@code deny}).
      */
-    public static ForeignLinker getSystemLinker() {
-        Utils.checkRestrictedAccess("CSupport.getSystemLinker");
+    static CLinker getSystemLinker() {
```
mcimadamore, line 99: Should be `getSystemLinker` here? Also, the string passed to the restricted check doesn't match the name of the method

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
```diff
@@ -26,23 +26,65 @@
...
+
 /**
- * A set of utilities for working with libraries using the C language/ABI
+ * This interface is a used as a super class for all ForeinLinker implementations that
+ * implement the system's C ABI and calling convention. As such, instances of this interface
+ * can be used to link foreign functions in native libraries that follow the JVM's target
+ * platform C application binary interface (ABI).
+ *
+ * <p>There are two components that go into linking a foreign function: a method type, and
+ * a function descriptor. The method type, consisting of a set of 'carrier' types constitutes
```
mcimadamore, line 48: I'd suggest doing `<em>` around `carrier`, instead of using `'`

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
```diff
@@ -26,23 +26,65 @@
...
+ * the Java side of a call to a foreign function, while the function descriptor constitutes
```
mcimadamore, line 49: I'd replace "constitutes the Java side of the call" with something more explicit like "...set of carrier types, which, together, specify the Java signature which clients must adhere to when e.g. calling the underlying foreign function."

Similarly "constitutes the native side of the call"; I'd replace that with "the function descriptor contains a set of memory layouts (link) which, together, uniquely specify the foreign function signature and classification information (via custom layout attributes), so that linking can take place."

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
```diff
@@ -26,23 +26,65 @@
...
+ * the native side of the call. Memory layout attributes are used in the function descriptor
+ * to attach ABI classification meta-data to memory layouts, which are required for linking.
+ * Clients of this API should use the prepared memory layout constants found in this interface
```
mcimadamore, line 52: Clients of this API should build function descriptors using the predefined memory layout constants (based on a subset of the built-in types provided by the C language), found in this interface; a failure to do so might result in linkage errors, given that this linker requires additional classification information to determine e.g. how arguments should be shuffled during a foreign function call.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
```diff
@@ -26,23 +26,65 @@
...
+ * carrier type can be used to match the native {@code va_list} type.</p>
+ *
+ * <p>The function descriptor used in linking contains a memory layout to match each carrier type.
+ * There are some restrictions on the carrier type and memory layout combinations that are allowed:</p>
+ * <ul>
+ *   <li>If a primitve type is used as a carrier type, the corresponding
+ *   memory layout must be a {@code ValueLayout}, and the bit size of the layout must match that of the carrier type
+ *   (see {@link Integer#SIZE} and similar fields in other primitive wrapper classes).</li>
+ *
+ *   <li>If the carrier type is {@code MemoryAddress} or {@code VaList}, then the corresponding memory layout must be a
```
mcimadamore, line 68: I would split this into two, one for pointers and one for VaList since the carrier and layout is different

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
```diff
@@ -26,23 +26,65 @@
...
+ *   and {@link CLinker#C_VA_LIST} can be used for {@code VaList} carriers.</li>
+ *
+ *   <li>If the carrier type is {@code MemorySegment}, then the corresponding memory layout must be a
+ *   {@code GroupLayout}</li>
+ * </ul>
+ *
+ * <p>Variadic functions, declared in C either with a trailing ellipses ({@code ...}) at the end of the formal parameter
+ * list or with an empty formal parameter list, are not supported directly. It is not possible to create a method handle
+ * that takes a variable number of arguments, and neither is it possible to create an upcall stub wrapping a method
+ * handle that accepts a variable number of arguments. However, for down calls only, it is possible to link a native
```
mcimadamore, line 80: down calls or downcall?

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
```diff
@@ -26,23 +26,65 @@
...
+ * variadic function by using a 'specialized' method type and function descriptor: for each argument that is to be
```
mcimadamore, line 81: specialized should be in `<em>`

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
```diff
@@ -26,23 +26,65 @@
...
+ * passed as a variadic argument, an explicit carrier type and memory layout must be present in the method type and
+ * function descriptor when linking the function. Furthermore, the memory layouts of variadic arguments must
```
mcimadamore, line 83: Furthermore, as memory layouts corresponding to variadic arguments in a function descriptor must contain additional classification information, it is required that the function {@link #asVarArg(MemoryLayout)} is called on each parameter layout that is used to compose a specialized variadic function descriptor.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
```diff
@@ -54,132 +96,430 @@ private CSupport() {}
...
-         *
-         * @param layout the layout of the value
-         * @return the value read as an {@code double}
-         * @throws IllegalStateException if the C {@code va_list} associated with this instance is no longer valid
-         * (see {@link #close()}).
-         * @throws IllegalArgumentException if the given memory layout is not compatible with {@code double}
-         */
-        double vargAsDouble(MemoryLayout layout);
+    /**
+     * The {@code _Bool} native type.
```
mcimadamore, line 140: I'd rewrite these comments as `The layout for the {@code XYZ} C type`

mcimadamore: Also - what about `Bool` ? Shouldn't we just remove it?

mcimadamore (commented):
Added some initial API comments - impl comments to follow soon

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
```diff
@@ -26,23 +26,65 @@
...
+import java.lang.invoke.MethodType;
 import java.nio.charset.Charset;
 import java.util.Objects;
 import java.util.function.Consumer;
 
+import static jdk.internal.foreign.PlatformLayouts.*;
+
 /**
- * A set of utilities for working with libraries using the C language/ABI
+ * This interface is a used as a super class for all ForeinLinker implementations that
```
PaulSandoz, line 42: Suggest as the first paragraph: "A foreign linker specializing for C Application Binary Interface (ABI) calling conventions. Instances ... platform C ABI."

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
JornVernee: This is not quite correct, as it must only be called on variadic arguments, not fixed ones. I went with:

Furthermore, as memory layouts corresponding to variadic arguments in a function descriptor must contain additional classification information, it is required that {@link #asVarArg(MemoryLayout)} is used to create the memory layouts for each parameter corresponding to a variadic arguments in a specialized function descriptor.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
JornVernee: Not sure what you mean here? The method is named getSystemLinker?

I've changed the message text.

JornVernee: Discussed offline: getSystemLinker implies too much that a particular system has One True Linker™. Changing this name to getInstance().

JornVernee:
@mcimadamore & @PaulSandoz Thanks for the review, I've applied the suggested changes: https://github.com/openjdk/panama-foreign/pull/327/commits/24f77111f5817c22decbeef5bee91218784158e9

I've also fixed a test that I noticed was no longer doing what is was supposed to (pre-existing to this PR).

JornVernee:
> By "de facto" do you mean the default? If so how does one get a non-default one? Do we expect that a developer or third-party can implement this interface or is it only the jDK that provides the implementations? (i.e. ideally we would like to seal it.)

By 'de facto' I mean that there is no official single C ABI for a particular platform, but in _practice_ there is only one. Maybe I'm being too pedantic here?

It is only intended for internal JDK classes to implement this Interface. Sealing might be an option, though I'm not sure if the `permits` list of a sealed class counts as part of the specification, i.e. will adding or removing implementations need a CSR? That might be prohibitive.

PaulSandoz:
Perhaps just say something like "Returns the C linker for the native platform"?

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
```diff
@@ -26,26 +26,75 @@
...
+ * set of carrier types, which, together, specify the Java signature which clients must adhere to when
+ * calling the underlying foreign function. The function descriptor contains a set of memory layouts which,
+ * together, specify the foreign function signature and classification information (via custom layout attributes),
+ * so that linking can take place. Memory layout attributes are used in the function descriptor
+ * to attach ABI classification meta-data to memory layouts, which are required for linking.
+ * Clients of this API should build function descriptors using the predefined memory layout constants
+ * (based on a subset of the built-in types provided by the C language), found in this interface;
+ * a failure to do so might result in linkage errors, given that linking requires additional classification
+ * information to determine, for instance, how arguments should be loaded into registers during a
+ * foreign function call.</p>
```
mcimadamore, line 57: I think this paragraph is general enough that it would work very well on the top ForeignLinker interface (replace C with any other language). I also wonder if we should add an example here - e.g. for `strlen` ?

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
```diff
@@ -54,11 +103,305 @@ private CSupport() {}
...
+     * @return the native stub segment.
+     * @throws IllegalArgumentException in the case of a carrier type and memory layout mismatch.
+     */
+    @Override
+    MemorySegment upcallStub(MethodHandle target, FunctionDescriptor function);
+
+    /**
+     * The prefix of all ABI attribute names.
+     */
+    String ABI_ATTR_PREFIX = "abi/";
```
mcimadamore, line 144: does this go here? Not sure what it buys to clients.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/ForeignLinker.java
```diff
@@ -32,7 +32,7 @@
  * This class can be used to link native functions as a {@link MethodHandle}, or to link Java
  * methods as a native function pointer (modelled as a {@link MemorySegment}).
  *
- * Instances of this interface can be obtained for instance by calling {@link CSupport#getSystemLinker()}
+ * Instances of this interface can be obtained for instance by calling {@link CLinker#getSystemLinker()}
```
mcimadamore, line 35: This comments should says something else - e.g. ```
Subclasses of this interface provide support for different linking strategies. For instance, the {@link CLinker} interface defines a foreign linker which supports the C ABI.```

mcimadamore: We are in a situation where we have an interface and only one implementation (`CLinker`) - should we just have the latter, and maybe reserve the room to add the general interface at a later time, as we have more implementations? @PaulSandoz - what do you think?

We started off in this direction because we thought that it might have been useful to have different places where to document general foreign linker functionality - as opposed to functionalities that are only applicable to the C linker - but the interface is so general that I'm not sure exactly what we can say at this level.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
```diff
@@ -26,26 +26,75 @@
...
+ * list or with an empty formal parameter list, are not supported directly. It is not possible to create a method handle
+ * that takes a variable number of arguments, and neither is it possible to create an upcall stub wrapping a method
+ * handle that accepts a variable number of arguments. However, for downcalls only, it is possible to link a native
+ * variadic function by using a <em>specialized</em> method type and function descriptor: for each argument that is to be
+ * passed as a variadic argument, an explicit carrier type and memory layout must be present in the method type and
+ * function descriptor when linking the function. Furthermore, as memory layouts corresponding to variadic arguments in
+ * a function descriptor must contain additional classification information, it is required that
+ * {@link #asVarArg(MemoryLayout)} is used to create the memory layouts for each parameter corresponding to a variadic
+ * argument in a specialized function descriptor</p>
  */
```
mcimadamore, line 93: There should be the usual comment about sealedness and immutability - see MemoryLayout

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
```diff
@@ -72,7 +415,7 @@ public static ForeignLinker getSystemLinker() {
      * As such, this interface only supports reading {@code int}, {@code double},
      * and any other type that fits into a {@code long}.
      */
-    public interface VaList extends Addressable, AutoCloseable {
+    interface VaList extends Addressable, AutoCloseable {
```
mcimadamore, line 418: Comment about sealedness and immutability here - probably for the Builder too

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/ForeignLinker.java
mcimadamore: We also need a comment about sealedness and immutability here

mcimadamore (commented):
Looks better -  a lot tighter than it ever was - still a couple of comments and a question to @PaulSandoz as to whether, from an API perspective, it makes sense to have an interface with only one supported impl.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/ForeignLinker.java
PaulSandoz: > We are in a situation where we have an interface and only one implementation (`CLinker`) - should we just have the latter, and maybe reserve the room to add the general interface at a later time, as we have more implementations? @PaulSandoz - what do you think?
> 

Now that i have looked more closely `ForeignLinker` does not seem weighty enough, even though the concepts of `downcall` and `upcall` are important. I suspect a common super interface could be defined if another kind of linker becomes a realization, and even then it's not clear to me that would be useful given the following on the `CLinker`:

> * Clients of this API should build function descriptors using the predefined memory layout constants
> * (based on a subset of the built-in types provided by the C language), found in this interface;

which indicates some tighter coupling.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
```diff
@@ -26,26 +26,75 @@
...
+ *
+ * <p>The function descriptor used in linking contains a memory layout to match each carrier type.
+ * There are some restrictions on the carrier type and memory layout combinations that are allowed:</p>
+ * <ul>
+ *   <li>If a primitve type is used as a carrier type, the corresponding
+ *   memory layout must be a {@code ValueLayout}, and the bit size of the layout must match that of the carrier type
+ *   (see {@link Integer#SIZE} and similar fields in other primitive wrapper classes).</li>
+ *
+ *   <li>If the carrier type is {@code MemoryAddress}, then the corresponding memory layout must be a
+ *   {@code ValueLayout}, and its bit size must match the platform's address size (see {@link MemoryLayouts#ADDRESS}).
```
brcolow, line 73: super minor grammatical thing, I think "and its bit size" is more correct as "and its' bit size"

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/ForeignLinker.java
JornVernee: I agree. Thanks for the fresh perspective :)

CLinker was created as a place to put the C specific documentation, but there's probably no need for different linkers to have a common supertype (ForeignLinker). Maybe general doc about what 'linking' means can be put in package-info?

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
JornVernee: I added this for fixing jextract. Jextract needs to filter out these attributes before making condys, since otherwise you get an access error when resolving. So, I thought that specifying a common name prefix would be the best way to make that possible, without exposing the actual attribute names that need to be filtered.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
JornVernee: Can do sealed, but neither VaList and Builder impls are immutable.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
JornVernee: "its" already is the possessive form, "its'"seems to be a misspelling: https://en.wiktionary.org/wiki/its%27

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/ForeignLinker.java
mcimadamore: I think putting general linking info in the package-level doc sounds totally the right thing to do.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
mcimadamore: ah sure - just sealed

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
mcimadamore: not super sure I get this. I imagine that jextact knows well when it needs to emit a layout corresponding to a primitive type, so, if that's the case, it should just emit an access to a static constant in Clinker and be done, rather than trying to dissect whatever intermediate layout jextract might have generated internally. I guess what I'm saying is - let's say that jextract has some layout with 4 attributes for C_INT; then its backend should just emit an access for CLinker.C_INT, and re-add the name attribute, which is really the only important part we need?

JornVernee: Yes, the name and alignment are important, but maybe we'd want to add other attributes in the future? Or maybe not us, but another tool that wants to emit more attributes. It seemed fair that if we added some attributes that clients were not allowed to look at, at least we could help them with stripping if needed.

I guess another approach is to not let jextract's Type.Primitive use the CLinker layout constants, but let it generate it's own ValueLayouts, attach name and alignment attributes, and then copy them over to the condy/source using withAttribute calls when emitting classes/sources.

JornVernee: I'll remove it for now, and then we can figure out what to do for jextract separately.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
```diff
@@ -54,11 +109,298 @@ private CSupport() {}
...
      */
-    public static ForeignLinker getSystemLinker() {
-        Utils.checkRestrictedAccess("CSupport.getSystemLinker");
+    static CLinker getInstance() {
+        Utils.checkRestrictedAccess("CLinker.getInstance");
         return SharedUtils.getSystemLinker();
     }
 
+    /**
+     * Obtain a method handle which can be used to call a given native function.
```
mcimadamore, line 118: This feels a bit thin. I would replace with:
```
Obtain a foreign method handle, with given type, which can be used to call a target foreign function at a given address and featuring a given function descriptor.
```

Let's also consider renaming `downcallHandle` to `foreignHandle`

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
```diff
@@ -54,11 +109,298 @@ private CSupport() {}
...
+
+    /**
+     * Allocates a native stub segment which contains executable code to upcall into a given method handle.
+     *
+     * <p>The base address of the returned stub segment can be passed to other foreign functions
+     * (as a function pointer). The returned segment is <em>not</em> thread-confined, and it only features
+     * the {@link MemorySegment#CLOSE} access mode. When the returned segment is closed,
+     * the corresponding native stub will be deallocated.</p>
+     *
+     * <p>The method type of the target method handle is used for linking</p>
```
mcimadamore, line 136: I think we can leave this out

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
```diff
@@ -54,11 +109,298 @@ private CSupport() {}
...
+     * @param symbol   downcall symbol.
+     * @param type     the method type.
+     * @param function the function descriptor.
+     * @return the downcall method handle.
+     * @throws IllegalArgumentException in the case of a carrier type and memory layout mismatch.
+     */
+    MethodHandle downcallHandle(Addressable symbol, MethodType type, FunctionDescriptor function);
+
+    /**
+     * Allocates a native stub segment which contains executable code to upcall into a given method handle.
```
mcimadamore, line 129: Suggestion:

```
Allocates a native segment whose base address (see {@link MemorySegment#address} can can be passed to other foreign functions (as a function pointer); calling such a function pointer from native code will result in the execution of the provided method handle.
<p>
The returned segment is <em>not</em> thread-confined, and it only features
the {@link MemorySegment#CLOSE} access mode. When the returned segment is closed, the corresponding native stub will be deallocated.
```

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
mcimadamore: @PaulSandoz mentioned (and I agree) that the terms upcalls and downcalls are colloquial, in the sense that it's technically possible to pass an upcall stub to `downcallHandle`, so that it's no longer really making a downcall and, conversely, it's possible to pass a native method handle to `upcallStub` so that it will not, strictly speaking, execute an upcall to Java.

I think the real nature of what we're describing here is:

* given an address and a function descriptor, obtain a MethodHandle which can be used to execute the code at that address from Java code

* given a MethodHandle and a function descriptor, obtain a memory segment, whose base address can be passed to native code where a function pointer is expected

I don't know if the terms downcallHandle and upcallStub are the best to describe what these do.

mcimadamore (commented):
Some more comments, now mostly focused on downcallHandle/upcallStub

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
JornVernee: Maybe we can incorporate 'link' in the name? Since we're using CLinker as a name as well. Maybe we can call them: `linkForeignFunction` and `linkMethodHandle`?

sundararajana (approved):
Tests add @module clauses. Perhaps not needed - earlier code too used those classes and imports. We had global config for those modules? Perhaps that could be removed now that we've specific tests having @module. Not super critical though.

JornVernee:
@sundararajana The @modules is needed to export the internal jdk.internal.foreign package, since the tests are now relying on the internal CABI enum for determining the current platform (previously they were using the ForeignLinker::name) method.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
JornVernee: I'll save the rename of the methods for later, and integrate this so that the refactor can go through.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
brcolow: Wow, sorry for the noise.

PR #328, opened 2020-09-16
8253243: Investigate ways to make MemorySegment::ofNativeRestricted more composable

mcimadamore:
This patch simplifies the MemorySegment::ofNativeRestricted API, based on a number of observations:

* now that we have shared segment support, passing an explicit owner thread isn't that useful (e.g. this is no longer an escape hatch, and can - and should - be done with public, safe API)
* if we had a public safe API to add cleanup action to existing segments, we could also get rid of the cleanup parameter
* if a custom cleanup action embedded some other object, that object will be kept reachable, so no attachment API is needed
* if we moved the method to MemoryAddress, we would streamline uses of this API quite a bit, as in `address.asSegmentRestricted(100)` - e.g. only size is needed

Adding the new API for attaching cleanup action is relatively straightforward - as usual we have to dup the segment and the scope, and create a new segment, with a new scope whose cleanup action calls the custom runnable before the original action (that original action should still be called, since otherwise we would fail to free/unmap memory).
Any exception thrown by the cleanup action is caught and ignored, so as not to mess with the critical cleanup which has to occur.

A nice consequence of this work is that the surface for unsafety is much reduced, and the various methods compose much better. For instance, the "everything" segment can be expressed as follows:

```
MemoryAddress.NULL
    .asSegmentRestricted(Long.MAX_VALUE)
    .withOwnerThread(null)
    .withAccessModes(READ | WRITE);
```

Which is, I think, much nicer. The changes in tests also suggest that the new API is easier on clients, and in one case (TestCleaner) the need for opting into unsafe operation actually vanished.

I also did a bit of cleaning on the javadoc, which appeared to be off-sync when it comes to confinement.

---------
### Progress

### Issue
 * [JDK-8253243](https://bugs.openjdk.java.net/browse/JDK-8253243): Investigate ways to make MemorySegment::ofNativeRestricted more composable

### Reviewers
 * [Athijegannathan Sundararajan](https://openjdk.java.net/census#sundar) (@sundararajana - Committer) ⚠️ Review applies to 3a186d46a27ba4da4b51d70e6c45df3626eb3662
 * [Paul Sandoz](https://openjdk.java.net/census#psandoz) (@PaulSandoz - Committer) ⚠️ Review applies to 3a186d46a27ba4da4b51d70e6c45df3626eb3662

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/328/head:pull/328`
`$ git checkout pull/328`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryAddress.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryAddress.java
@@ -93,6 +93,11 @@ public interface MemoryAddress extends Addressable {
      * (often as a plain {@code long} value). The returned segment will feature all <a href="#access-modes">access modes</a>
      * (see {@link MemorySegment#ALL_ACCESS}), and its confinement thread is the current thread (see {@link Thread#currentThread()}).
      * <p>
+     * Calling {@link MemorySegment#close()} on the returned segment will <em>not</em> result in releasing any
+     * memory resources which might implicitly be associated with the segment. If the client wants to specify
+     * a cleanup action to be executed when the returned segment is closed, the {@link MemorySegment#withCleanupAction(Runnable)}
+     * method should be used.
+     * <p>
      * This method is <em>restricted</em>. Restricted methods are unsafe, and, if used incorrectly, their use might crash
      * the JVM or, worse, silently result in memory corruption. Thus, clients should refrain from depending on
      * restricted methods, and use safe and supported functionalities, where possible.
@@ -105,10 +110,10 @@ public interface MemoryAddress extends Addressable {
      * {@code permit}, {@code warn} or {@code debug} (the default value is set to {@code deny}).
      */
     default MemorySegment asSegmentRestricted(long bytesSize) {
+        Utils.checkRestrictedAccess("MemoryAddress.asSegmentRestricted");
         if (bytesSize <= 0) {
             throw new IllegalArgumentException("Invalid size : " + bytesSize);
         }
-        Utils.checkRestrictedAccess("MemoryAddress.ofSegmentRestricted");
         return NativeMemorySegmentImpl.makeNativeSegmentUnchecked(this, bytesSize);
     }
 
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryAddress.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryAddress.java
@@ -27,6 +27,10 @@
 package jdk.incubator.foreign;
 
 import jdk.internal.foreign.MemoryAddressImpl;
+import jdk.internal.foreign.NativeMemorySegmentImpl;
+import jdk.internal.foreign.Utils;
+
+import java.util.Objects;
 
 /**
  * A memory address models a reference into a memory location. Memory addresses are typically obtained using the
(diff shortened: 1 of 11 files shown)


sundararajana (approved):

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryAddress.java
```diff
@@ -83,9 +87,35 @@ default MemoryAddress address() {
...
+     * @throws IllegalArgumentException if {@code bytesSize <= 0}.
+     * @throws UnsupportedOperationException if this address is an heap address.
+     * @throws IllegalAccessError if the runtime property {@code foreign.restricted} is not set to either
+     * {@code permit}, {@code warn} or {@code debug} (the default value is set to {@code deny}).
+     */
+    default MemorySegment asSegmentRestricted(long bytesSize) {
+        if (bytesSize <= 0) {
+            throw new IllegalArgumentException("Invalid size : " + bytesSize);
+        }
+        Utils.checkRestrictedAccess("MemoryAddress.ofSegmentRestricted");
```
PaulSandoz, line 111: Suggest this be the first check, so it fails independently of the arguments when permission is not granted.

Might be worth stating that closing does not free up any memory or resources. We can refer to the use of `withCleanupAction` if such action is required.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -257,13 +257,33 @@
...
      * or if {@code newOwner == null} and this segment does not support the {@link #SHARE} access mode.
      */
     MemorySegment withOwnerThread(Thread newOwner);
 
+    /**
+     * Obtains a new memory segment backed by the same underlying memory region as this segment, and featuring
+     * the same confinement thread (see {@link #ownerThread()} and access modes (see {@link #accessModes()}),
+     * but with a different cleanup action. More specifically, the cleanup action associated with the returned segment will
+     * first call the user-provided action, before delegating back to the original cleanup action associated with
+     * this segment. Any exception thrown by the user-provided action will be discarded, and will not prevent the
```
PaulSandoz, line 271: "this segment (if any). Any ..." ?

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -257,13 +257,33 @@
...
+     * release of any memory resources associated with this segment.
+     * <p>
+     * As a side-effect, this segment will be marked as <em>not alive</em>,
+     * and subsequent operations on this segment will result in runtime errors.
```
PaulSandoz, line 275: errors/exceptions?

PaulSandoz (approved):

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
```diff
@@ -325,6 +332,20 @@ public CleanupAction dup() {
                 return new DupAction(this);
             }
 
+            @Override
+            public CleanupAction wrap(Runnable runnable) {
+                disable();
+                return AtMostOnceOnly.of(() -> {
+                    try {
+                        runnable.run();
+                    } catch (Throwable t) {
```
JornVernee, line 341: Maybe the exception should be re-thrown after doing critical cleanup?

test/jdk/java/foreign/TestNoForeignUnsafeOverride.java
```diff
@@ -40,6 +40,6 @@
 
     @Test(expectedExceptions = IllegalAccessError.class)
     public void testUnsafeAccess() {
-        MemorySegment.ofNativeRestricted(MemoryAddress.ofLong(42), 10, null, null, null);
+        MemorySegment.ofNativeRestricted();
```
JornVernee, line 43: Should be:
```suggestion
        MemoryAddress.ofLong(42).asSegmentRestricted(10);
```
Right? To retain the same behaviour?

JornVernee (approved):

JornVernee:
Also wanted to say. I think this is a very nice cleanup :) You can see a definite positive improvement in the test code.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
mcimadamore: We could do that - but I'd prefer to leave this as is.

test/jdk/java/foreign/TestNoForeignUnsafeOverride.java
mcimadamore: This test is, if I understand correctly, simply testing that you can't calla restricted method if the JDK property has not been set on the command line. It's not important WHICH restricted method is called. At the time the test was written we didn't have the more succint ofNativeRestricted - now we do, so I thought to just use that, for brevity.

PR #330, opened 2020-09-17
8239780: FunctionDescriptor factories and combinators should check whether return layout is null

JornVernee:
Hi,

I'm going through the foreign-abi JBS issues and taking care of some of the open ones.

This PR adds null checking to the MemoryLayouts passed to the FunctionDescriptor factories and adapter methods, together with a simple test.

I've also added an adapter for dropping the return layout, since with the null check, it is no longer possible to pass `null` to drop the return layout. I think having the explicit check, and a separate combinator for dropping the return layout instead will be better for avoiding accidental nulls being used.

Thanks,
Jorn

---------
### Progress

### Issue
 * [JDK-8239780](https://bugs.openjdk.java.net/browse/JDK-8239780): FunctionDescriptor factories and combinators should check whether return layout is null

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/330/head:pull/330`
`$ git checkout pull/330`

### DIFF
--- a/test/jdk/java/foreign/TestFunctionDescriptor.java
+++ b/test/jdk/java/foreign/TestFunctionDescriptor.java
@@ -28,13 +28,99 @@
  */
 
 import jdk.incubator.foreign.FunctionDescriptor;
+import jdk.incubator.foreign.MemoryLayout;
 import org.testng.annotations.Test;
+
+import java.lang.constant.Constable;
+import java.lang.constant.ConstantDesc;
+import java.lang.invoke.MethodHandles;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
 
 import static jdk.incubator.foreign.CLinker.C_DOUBLE;
 import static jdk.incubator.foreign.CLinker.C_INT;
 import static jdk.incubator.foreign.CLinker.C_LONGLONG;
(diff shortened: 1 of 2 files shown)


test/jdk/java/foreign/TestFunctionDescriptor.java
```diff
@@ -0,0 +1,61 @@
...
+        FunctionDescriptor fd = FunctionDescriptor.ofVoid(C_INT, C_LONGLONG);
+        fd.appendArgumentLayouts(C_DOUBLE, null); // should throw
+    }
+
+    @Test(expectedExceptions = NullPointerException.class)
+    public void testNullReturnLayoutChange() {
+        FunctionDescriptor fd = FunctionDescriptor.ofVoid(C_INT, C_LONGLONG);
+        fd.changeReturnLayout(null); // should throw
+    }
+
```
sundararajana, line 60: Test for dropReturnLayout missing?

JornVernee: I can add that. We don't have any tests that just test FunctionDescriptor currently, so I went only with the tests that look for the NPE.

I'll add some basic API tests as well.

sundararajana (approved):

PR #335, opened 2020-09-17
8253290: Tweak jextract code to use new MemoryAddress::asSegmentRestricted

mcimadamore:
This fixes references to the now-removed MemorySegment::ofNativeRestricted(...).

The only tricky issue is in RunimeHelper - which has to create an unchecked segment for a global variable and,
in doing so, it has to keep a reference of the lookup symbol in there, to prevent the library from being unloaded.

Since the new API doesn't have an explicit way to add an attachment, I've added a simple Runnable impl which wraps a symbol,
and pass that as a no-op cleanup action to the segment.

For the records, I believe that creating segments for global vars is tricky enough that there should be a blessed (restricted) way to do that in the LibraryLookup.Symbol API.

---------
### Progress

### Issue
 * [JDK-8253290](https://bugs.openjdk.java.net/browse/JDK-8253290): Tweak jextract code to use new MemoryAddress::asSegmentRestricted

### Reviewers
 * [Athijegannathan Sundararajan](https://openjdk.java.net/census#sundar) (@sundararajana - Committer) ⚠️ Review applies to e9b47df56d2222568ba71670435f39cd58889a4d

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/335/head:pull/335`
`$ git checkout pull/335`

### DIFF
--- a/src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/resources/RuntimeHelper.java.template
+++ b/src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/resources/RuntimeHelper.java.template
@@ -63,14 +63,14 @@ public class RuntimeHelper {
             return lookup(LIBRARIES, name).map(s ->
                 nonCloseableNonTransferableSegment(s.address().asSegmentRestricted(layout.byteSize())
                         .withOwnerThread(null)
-                        .withCleanupAction(new RunnableHolder(s))
+                        .withCleanupAction(new SymbolHolder(s))
                 )).orElse(null);
     }
 
-    static class RunnableHolder implements Runnable {
+    static class SymbolHolder implements Runnable {
         LibraryLookup.Symbol symbol;
 
-        RunnableHolder(LibraryLookup.Symbol symbol) {
+        SymbolHolder(LibraryLookup.Symbol symbol) {
             this.symbol = symbol;
         }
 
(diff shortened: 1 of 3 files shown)


src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/resources/RuntimeHelper.java.template
```diff
@@ -60,10 +60,24 @@ public class RuntimeHelper {
...
-                 s.address(), layout.byteSize(), null, null, s
-            ))).orElse(null);
+            return lookup(LIBRARIES, name).map(s ->
+                nonCloseableNonTransferableSegment(s.address().asSegmentRestricted(layout.byteSize())
+                        .withOwnerThread(null)
+                        .withCleanupAction(new RunnableHolder(s))
+                )).orElse(null);
+    }
+
+    static class RunnableHolder implements Runnable {
```
sundararajana, line 70: Should this be called SymbolHolder?

sundararajana (approved):

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/resources/RuntimeHelper.java.template
mcimadamore: yeah... will fix

PR #339, opened 2020-09-17
8253307: Simplify code in RuntimeHelper::lookupGlobalVariable

mcimadamore:
Earlier today, I added a workaround on RuntimeHelper when I tweaked it to use the new MemoryAddress::asSegmentRestricted API.
The old code was adding the lookup symbol as an attachment to the returned segment, so I did a workaround to do just that under the new API.

Turns out that the workaround wasn't needed (I think). Upon more thinking, I realized that we always store an array of LibraryLookup inside the very constant holder class - either as a static field, or as a dynamically loaded constant.

I think in both cases that should be enough to prevent the lookup object from going unreachable.

---------
### Progress

### Issue
 * [JDK-8253307](https://bugs.openjdk.java.net/browse/JDK-8253307): Simplify code in RuntimeHelper::lookupGlobalVariable

### Reviewers
 * [Henry Jen](https://openjdk.java.net/census#henryjen) (@slowhog - Committer)

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/339/head:pull/339`
`$ git checkout pull/339`

### DIFF
--- a/src/jdk.incubator.jextract/share/classes/jdk/internal/clang/libclang/RuntimeHelper.java
+++ b/src/jdk.incubator.jextract/share/classes/jdk/internal/clang/libclang/RuntimeHelper.java
@@ -89,22 +89,7 @@ public class RuntimeHelper {
     public static final MemorySegment lookupGlobalVariable(LibraryLookup[] LIBRARIES, String name, MemoryLayout layout) {
         return lookup(LIBRARIES, name).map(s ->
             nonCloseableNonTransferableSegment(s.address().asSegmentRestricted(layout.byteSize())
-                    .withOwnerThread(null)
-                    .withCleanupAction(new SymbolHolder(s))
-            )).orElse(null);
-    }
-
-    static class SymbolHolder implements Runnable {
-        LibraryLookup.Symbol symbol;
-
-        SymbolHolder(LibraryLookup.Symbol symbol) {
-            this.symbol = symbol;
-        }
-
-        @Override
-        public void run() {
-            // do nothing
-        }
+                    .withOwnerThread(null))).orElse(null);
     }
 
     public static final MemorySegment nonCloseableNonTransferableSegment(MemorySegment seg) {
--- a/src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/resources/RuntimeHelper.java.template
+++ b/src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/resources/RuntimeHelper.java.template
@@ -65,19 +65,6 @@ public class RuntimeHelper {
                     .withOwnerThread(null))).orElse(null);
     }
 
-    static class SymbolHolder implements Runnable {
-        LibraryLookup.Symbol symbol;
-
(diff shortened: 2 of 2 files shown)


slowhog (approved):
LGTM!

JornVernee (approved):

sundararajana (approved):
While this solves the issue for for jextract which keeps LinkerLookups, as a general API usage, this should be documented then i.e, it is up to the user to keep linker lookup objects alive

In the JNI world this problem does not arise because NativeLibrary is always loaded in the context of the class loader that loaded and so native libraries are alive as long as the class loader is alive.

mcimadamore:
> While this solves the issue for for jextract which keeps LinkerLookups, as a general API usage, this should be documented then i.e, it is up to the user to keep linker lookup objects alive
> 
> In the JNI world this problem does not arise because NativeLibrary is always loaded in the context of the class loader that loaded and so native libraries are alive as long as the class loader is alive.

We do have some javadoc covering that already.

Not super sure about JNI - will creating a BB from JNI code (from a random point) keep the library alive? How?

mcimadamore:
> > While this solves the issue for for jextract which keeps LinkerLookups, as a general API usage, this should be documented then i.e, it is up to the user to keep linker lookup objects alive
> > In the JNI world this problem does not arise because NativeLibrary is always loaded in the context of the class loader that loaded and so native libraries are alive as long as the class loader is alive.
> 
> We do have some javadoc covering that already.
> 
> Not super sure about JNI - will creating a BB from JNI code (from a random point) keep the library alive? How?

To be more explicit about the case I have in mind for JNI:

1.  start classloader
2.  load library
3.  obtain BB from pointer in library
4.  share the BB with some other class loader
5.  drop the original classloader (and the library)
6. problem?

sundararajana:
> > > While this solves the issue for for jextract which keeps LinkerLookups, as a general API usage, this should be documented then i.e, it is up to the user to keep linker lookup objects alive
> > > In the JNI world this problem does not arise because NativeLibrary is always loaded in the context of the class loader that loaded and so native libraries are alive as long as the class loader is alive.
> > 
> > 
> > We do have some javadoc covering that already.
> > Not super sure about JNI - will creating a BB from JNI code (from a random point) keep the library alive? How?
> 
> To be more explicit about the case I have in mind for JNI:
> 
> 1. start classloader
> 2. load library
> 3. obtain BB from pointer in library
> 4. share the BB with some other class loader
> 5. drop the original classloader (and the library)
> 6. problem?

Agree.

NewDirectByteBuffer
jobject NewDirectByteBuffer(JNIEnv* env, void* address, jlong capacity);

document says:

"Allocates and returns a direct java.nio.ByteBuffer referring to the block of memory starting at the memory address address and extending capacity bytes.
Native code that calls this function and returns the resulting byte-buffer object to Java-level code should ensure that the buffer refers to a valid region of memory that is accessible for reading and, if appropriate, writing. An attempt to access an invalid memory location from Java code will either return an arbitrary value, have no visible effect, or cause an unspecified exception to be thrown.”

Perhaps a warning like that for "MemoryAddress.asSegmentRestricted" as well? As a separate PR

PR #348, opened 2020-09-21
8253430: Improve the LibraryLookup API

mcimadamore:
This patch implements a couple of suggestion which were made a couple of weeks ago by @BlueGoliath - here:

https://mail.openjdk.java.net/pipermail/panama-dev/2020-July/010140.html

Two ideas here:
* Change the signature of `LibraryLookup::ofPath` to actually take a `Path` (rathar than a `String`)
* Tweak `LibraryLookup::lookup` to return `Optional` (rather than throw)

Both suggestions seem to change the API for the better. The second one is probably more controversial: on the one hand it removes the need for having exception handling logic and makes it for more composable logic. On the other hand clients have to manually call `get` on the lookup result. But overall seems a win.

---------
### Progress

### Issue
 * [JDK-8253430](https://bugs.openjdk.java.net/browse/JDK-8253430): Improve the LibraryLookup API

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/348/head:pull/348`
`$ git checkout pull/348`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/LibraryLookup.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/LibraryLookup.java
@@ -108,12 +108,17 @@ public interface LibraryLookup {
 
     /**
      * Obtain a library lookup object corresponding to a library identified by given path.
-     * @param path the library path.
+     * @param path the library absolute path.
      * @return a library lookup object for given path.
+     * @throws IllegalArgumentException if the specified path does not correspond to an absolute path,
+     * e.g. if {@code !path.isAbsolute()}.
      */
     static LibraryLookup ofPath(Path path) {
         Objects.requireNonNull(path);
-        String absolutePath = path.toAbsolutePath().toString();
+        if (!path.isAbsolute()) {
+            throw new IllegalArgumentException("Not an absolute path: " + path.toString());
+        }
+        String absolutePath = path.toString();
         SecurityManager security = System.getSecurityManager();
         if (security != null) {
             security.checkLink(absolutePath);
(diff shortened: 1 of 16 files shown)


src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/LibraryLookup.java
```diff
@@ -110,17 +111,14 @@ static LibraryLookup ofDefault() {
      * @param path the library path.
      * @return a library lookup object for given path.
      */
-    static LibraryLookup ofPath(String path) {
+    static LibraryLookup ofPath(Path path) {
         Objects.requireNonNull(path);
+        String absolutePath = path.toAbsolutePath().toString();
```
sundararajana, line 116: Path.toAbsolutePath itself is a secure operation (could potentially throw SecurityException).  Should insist on absolute Path (that way it is up to the client to manage absolute Path object)?

JornVernee (approved):

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/LibraryLookup.java
mcimadamore: I think it's nice, from an usability perspective, that any path is accepted - but I'm open to change here

JornVernee (approved):

sundararajana (approved):

BlueGoliath:
Thanks for making these changes.

Would it be possible to also modify LibraryLookup.ofLibrary() so that it takes a var-arg string list? Currently, I'm using ofPath only because LibraryLookup.ofLibrary() only accepts a single name and NVML has different library names on Windows Vs. Linux.:

Path.of(System.getenv("WINDIR"), "/system32/nvml.dll")

vs.

Path.of("/usr/lib/", "libnvidia-ml.so")

mcimadamore:
> Path.of(System.getenv("WINDIR"), "/system32/nvml.dll")
> 
> vs.
> 
> Path.of("/usr/lib/", "libnvidia-ml.so")

I think using a varargs of library names does not adhere with my (obviously subjective) API bar.

If you have a library that has name X in one system and name Y in a different system, it seems to me that you should have a constant that is initialized with the right name depending on the system and then load that. Passing a list of all possible alternatives, while doable, seems an odd way to get there.

PR #354, opened 2020-09-22
Turn on downcall intrinsics by default

JornVernee:
Hi,

This PR turns on down call intrinsics by default.

Since the time they were turned off some fixes have been done in the area of GC handling of optimized native calls, and no further problems in that area have been found. What currently works is:

1. All TestDowncall and TestUpcall tests, where each function is invoked in a 20_000 iteration loop with -Xbatch, making sure these calls are intrinsified.
2. TestIntrinsics tests which intrinsifies calls the same way.
3. All jextract samples with intrinsics turned on, which from the logs also leads to calls being intrinsified, especially on large extraction runs like Windows.h.

There is one more case that is known to be problematic: TestUpcallHighArity crashes/fails when the call is intrinsified (using same method as 1 and 2). This seems to be caused by a known problem with the handling on stack arguments for intrinsified calls. So, for now, intrinsification of calls that pass arguments on the stack will be turned off.

If there are other cases you would like to make sure work, please take the test for a spin.

Thanks,
Jorn

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/354/head:pull/354`
`$ git checkout pull/354`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/ProgrammableInvoker.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/ProgrammableInvoker.java
@@ -161,7 +161,8 @@ public class ProgrammableInvoker {
                                             .asType(leafType);
 
         boolean isSimple = !(retMoves.length > 1);
-        if (USE_INTRINSICS && isSimple) {
+        boolean usesStackArgs = stackArgsBytes != 0;
+        if (USE_INTRINSICS && isSimple && !usesStackArgs) {
             NativeEntryPoint nep = NativeEntryPoint.make(
                 addr.address().toRawLongValue(),
                 "native_call",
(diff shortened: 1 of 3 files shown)


mcimadamore (approved):
Looks good - I wonder if we should add (either here, or in a separate changeset) a test for high arity downcalls - we can verify correctness by generating all combinations:

```
m1(a1, a2 ..... an) { return a1; }
...
mn(a1, a2 ..... an) { return an; } 
```

What do you think?

JornVernee:
@mcimadamore Thanks for the suggestion. I've refactored the test a bit and implemented it, but if I re-enable stack arguments some of the test cases fail; exactly the ones that are testing the arguments passed on the stack.

mcimadamore (approved):
Looks good!

PR #356, opened 2020-09-28
Transition back to java on safepoint poll slow-path when returning from an optimized native call

JornVernee:
Hi,

Due to a bug report last week, it was found that threads do not transition back to the _thread_in_java state on the safepoint poll slow path after returning from an optimized native call.

While I was not able to reproduce the crash in question so far, the current code is obviously wrong, since the thread state transition back only happens on the fast path.

This patch rectifies that issue.

Thanks,
Jorn

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/356/head:pull/356`
`$ git checkout pull/356`


mcimadamore (approved):
Looks good

iwanowww (commented):
Looks good.

PR #358, opened 2020-09-28
8253255: Investigate replacing ABI layout attributes with custom subtypes of ValueLayout

JornVernee:
Hi,

This patch removes ABI attributes, and replaces them with a custom ValueLayout subtype, called CValueLayout.

This, for one, allows us to override the describeConstable method to return a dynamic constant descriptor that will load the public constant field in CLinker. This avoids getting illegal access errors when trying to resolve the internal ABI attributes. The illegal access error was forcing clients, like jextract, to basically re-implement the describeConstable methods for FunctionDescriptor, GroupLayout, SequenceLayout and ValueLayout in order the 'cannonicalize' these constant descriptors to reference the fields.

Most of the needed changes were in the TypeClass classes, which now need to map the CValueLayout.Kind into the appropriate ABI classes (though doing this was pretty straight forward).

I've also added a test to make sure it is possible to resolve constant descriptors with these CValueLayouts inside with a public Lookup object.

Thanks,
Jorn

---------
### Progress

### Issue
 * [JDK-8253255](https://bugs.openjdk.java.net/browse/JDK-8253255): Investigate replacing ABI layout attributes with custom subtypes of ValueLayout

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/358/head:pull/358`
`$ git checkout pull/358`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/AbstractLayout.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/AbstractLayout.java
@@ -43,6 +43,7 @@ import java.util.OptionalLong;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import static java.lang.constant.ConstantDescs.BSM_GET_STATIC_FINAL;
 import static java.lang.constant.ConstantDescs.BSM_INVOKE;
 import static java.lang.constant.ConstantDescs.CD_String;
 import static java.lang.constant.ConstantDescs.CD_long;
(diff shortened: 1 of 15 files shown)


mcimadamore (commented):
Looks very neat - what about the equals/hashcode behavior of the new class? Will CValueLayout be considered equals with another value layout with same attributes and charateristics?

I think so - code looks like (in ValueLayout):

```
@Override
 91     public boolean equals(Object other) {
 92         if (this == other) {
 93             return true;
 94         }
 95         if (!super.equals(other)) {
 96             return false;
 97         }
 98         if (!(other instanceof ValueLayout)) {
 99             return false;
100         }
101         ValueLayout v = (ValueLayout)other;
102         return order.equals(v.order) &&
103             bitSize() == v.bitSize() &&
104             alignment == v.alignment;
105     }
```

JornVernee:
> ```
> Layout v = (ValueLayout)other;
> 102         return order.equals(v.order) &&
> 103             bitSize() == v.bitSize() &&
> 104             alignment == v.alignment;
> 105     }
> ```
> 
> 
> Let's maybe add a test on this.

Thanks, for catching this! I had forgotten to add equals & hashCode methods to the new class. Will add a test as well.

mcimadamore (approved):
Looks good!

PR #359, opened 2020-09-29
8253796: Consider making CValueLayout public.

JornVernee:
Hi,

This PR makes CValueLayout a public nested class of CLinker, meant to be extended. To quote the motivation from the JBS issue:

> Jextract uses an attribute to attach temporary information to layouts regarding the canonical layout field in CLinker they correspond to.
> 
> But, creating a dynamic constant out of these layouts is a complex process, since we need to recursively filter out this internal jextract attribute, which requires us to recreate GroupLayouts, SequenceLayouts, FunctionDescriptors, and ValueLayouts.
> 
> It would be easier if we could extend CValueLayout with a custom jextract layout class, which has an extract field to indicate which CLinker constant they correspond to. In that case e.g. SourceConstantHelper can cast to this custom class and retrieve this extra field, removing the need to use attributes, and the difficulties associated with them. 

Thanks,
Jorn

---------
### Progress

### Issue
 * [JDK-8253796](https://bugs.openjdk.java.net/browse/JDK-8253796): Consider making CValueLayout public.

### Reviewers
 * [Maurizio Cimadamore](https://openjdk.java.net/census#mcimadamore) (@mcimadamore - Committer) ⚠️ Review applies to 16da67a765db0a21c50c63f2a98f8139d8d6d510

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/359/head:pull/359`
`$ git checkout pull/359`


src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
```diff
@@ -700,4 +706,143 @@ static VaList empty() {
         }
     }
 
+    class CValueLayout extends ValueLayout {
```
mcimadamore, line 709: Some javadoc is missing here

mcimadamore (commented):
Looks good - except for missing javadoc

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
```diff
@@ -700,4 +706,143 @@ static VaList empty() {
...
+        public enum Kind {
```
mcimadamore, line 710: not sure we want the Kind exposed

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
```diff
@@ -700,4 +706,143 @@ static VaList empty() {
...
+
+        public static CValueLayout ofLongDouble(ByteOrder order, long bitSize) {
+            return new CValueLayout(Kind.LONGDOUBLE, order, bitSize, bitSize, Map.of());
+        }
+
+        public static CValueLayout ofPointer(ByteOrder order, long bitSize) {
+            return new CValueLayout(Kind.POINTER, order, bitSize, bitSize, Map.of());
+        }
+
+        public final Kind kind() {
```
mcimadamore, line 794: Same here - should this exposed?

JornVernee: Yes, kind should be exposed. The classification logic needs to access it, and if we don't expose it it's not possible to extend this class, since the kind is a constructor argument and we need to call the constructor.

mcimadamore: I guess the fact that classification logic needs it can be handle by reorganizing the code so that the impl private.
But your concern about subclasses needing to pass a kind is true.

JornVernee:
I've added javadoc now.

I made some Kind.bsm() private since it could be, and moved the factory methods to PlatformLayouts, since that's the only place they were being used (had to make the constructor public for that, but that doesn't seem like a problem?). Finally, I also added an isFloat() function to Kind in order to complete the set of predicates (should this also be implemented using a boolean flag?)

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
```diff
@@ -700,4 +706,197 @@ static VaList empty() {
         }
     }
 
+    /**
+     * Subclass of ValueLayout that contains information needed when linking
```
mcimadamore, line 710: Use `@link` ?

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
```diff
@@ -700,4 +706,197 @@ static VaList empty() {
...
+     * downcalls or upcalls.
+     */
+    class CValueLayout extends ValueLayout {
+        /**
+         * The kind of CValueLayout. Each kind corresponds to a particular
```
mcimadamore, line 715: Use `@link` ?

mcimadamore (approved):
Good moves - I missed the fact that exposing factories was not super desirable. Added some minor comments around the javadoc

PR #360, opened 2020-09-29
8253798: Simplify ClassConstantHelper

JornVernee:
Hi,

This patch removes a whole bunch of workarounds in jextract code, which are made possible by upstream changes to foreign-abi.

ClassConstantHelper no longer has to manually canonicalize MemoryLayout constant that represent layout constants in CLinker. This is now handled by the CValueLayout.describeConstable() override.

Making CValueLayout public also means we can remove the CanonicalField enum from jextract, and rely on the CValueLayout.Kind instead. (note that since the enum was internal, complex recursive filtering would be needed if we had kept it).

Thanks,
Jorn

---------
### Progress

### Issue
 * [JDK-8253798](https://bugs.openjdk.java.net/browse/JDK-8253798): Simplify ClassConstantHelper

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/360/head:pull/360`
`$ git checkout pull/360`


mcimadamore (approved):
Looks good!

PR #361, opened 2020-09-29
8253823: Investigate ways to make handoff-like operation more explicit

mcimadamore:
The API currently doesn't do a good job at differentiating between simple wither-like methods (e.g. withAccessModes) and deep side-effecting operations (e.g. withOwnerThread).

This patch explores a way to make handoff more of a first-class concept of the API - you can take apart an existing segment and reconstruct it with a given `HandoffTransform` (a recipe which specifies what properties should be associated with the reconstructed segment). Turns out that `handoff` and `HandoffTransform` is the ideal place where to put several of the *advanced* operations we used to have exposed directly by the memory segment API (`withCleanupAction`, `registerCleaner`). We can also add na overload for `handoff` which takes a native scope and bounds the segment to it.

By reorganizing the API in this way, it is now clear that memory segment feature _two_ terminal operations: `close` and `handoff`. We can document this in the API in a more centralized fashion, without having to repeat the same text across seemingly unrelated methods. For clients, it is also clearer that doing an `handoff` is a big reconstruction, which is different from a simple wither.

I've also somewhat simplified the access mode story; there's now a single HANDOFF mode which covers all types of handoff - this was clearly needed, since the access modes we had before were very fine-grained, and it was very painful for client to prevent against *all* kinds of terminal operations (for instance, I realized that the current impl has no access modes check for `withCleanupAction`).

Impl note: for now the HandoffTransform implementation is mutable and not thread-safe - the javadoc calls this out. I'm open to make this more immutable/value oriented if this is desirable.

Note: this PR is against the `foreign-abi` so that it shows how `NativeScope` is affected by the changes. If the changes brought in this patch looks good, it would be better to push the memory access changes in the `memory-access` branch in a separate PR, and then follow up with the remaining `NativeScope` changes here.

Javadoc  link:
http://cr.openjdk.java.net/~mcimadamore/panama/segment-rebuilder-javadoc_v2/javadoc/jdk/incubator/foreign/package-summary.html

---------
### Progress

### Issue
 * [JDK-8253823](https://bugs.openjdk.java.net/browse/JDK-8253823): Investigate ways to make handoff-like operation more explicit

### Reviewers
 * [Paul Sandoz](https://openjdk.java.net/census#psandoz) (@PaulSandoz - Committer) ⚠️ Review applies to 75d3ba2a3b19aa57b2f028f756d8885ebdb870d0

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/361/head:pull/361`
`$ git checkout pull/361`


src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -849,25 +873,106 @@ static MemorySegment ofNativeRestricted() {
...
+     * may become a {@code sealed} interface, which would prohibit subclassing except by
+     * explicitly permitted subtypes.
+     *
+     * @implSpec
+     * Implementations of this interface are not thread-safe.
+     *
+     * @see MemorySegment#handoff(HandoffTransform)
      */
-    int ALL_ACCESS = READ | WRITE | CLOSE | SHARE | HANDOFF;
+    interface HandoffTransform {
```
PaulSandoz, line 909: There might be a use-case for starting out with the existing confinement properties of a segment, say if just registering a cleaner or adding an attachment e.g. `HandoffTransform.ofSegment`, which is not a good name since it would imply the acceptance of a segment parameter, whereas it would be applied to the segment that accepts the handoff transform. That way it's not necessary to query a segment before constructing the correct transform.

I suspect there will only be one implementation of `HandoffTransform`, since it's really just a dumb holder of accumulated state, so it could be made into a `final class` (value-based) with a non-public constructor.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
```diff
@@ -288,38 +284,91 @@ private void checkAccessModes(int accessModes) {
...
 
-    @Override
-    public void registerCleaner(Cleaner cleaner) {
-        checkAccessModes(CLOSE);
-        checkValidState();
-        cleaner.register(this.scope, scope.cleanupAction);
+    public static class HandoffTransformImpl implements HandoffTransform {
+
+        final Thread ownerThread;
+        List<Runnable> cleanupActions = new ArrayList<>();
```
PaulSandoz, line 317: Mark all fields as `final`

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -141,37 +137,75 @@
...
  * to directly manipulate the memory segment (e.g. close the memory segment) or access the underlying memory associated with
  * the segment using a memory access var handle. Any attempt to perform such operations from a thread other than the
  * owner thread will result in a runtime failure.
  * <p>
- * Memory segments support <em>serial thread confinement</em>; that is, ownership of a memory segment can change (see
- * {@link #withOwnerThread(Thread)}). This allows, for instance, for two threads {@code A} and {@code B} to share
- * a segment in a controlled, cooperative and race-free fashion.
+ * The {@link #handoff(HandoffTransform)} method can be used to change the thread-confinement properties (as well
+ * as other properties) of a memory segment. This method is, like {{@link #close()}} a <em>terminal operation</em>
+ * which marks the original segment as not alive (see {@link #isAlive()} and create a <em>new</em> segment with the
```
PaulSandoz, line 150: s/create/creates

PaulSandoz (changes requested):
I think this provides much more clarity with a minimal increment in API surface.

Unsure if `MemorySegment` should closely couple with `NativeScope`, there is an intention to separate out into base and/or other modules. Perhaps  `NativeScope` is independently useful from the ABI that it should be pushed down into the memory-access branch and part of the core memory API?

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
mcimadamore: `HandoffTransform.identity()` ?

mcimadamore: Yes - this is a dumb holder for accumulated state - my question in the review header was as to whether we're fine with it being dumb and mutable and thread unsafe, since that is reflected in the javadoc

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
mcimadamore: will do

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
mcimadamore: will change

mcimadamore:
> Unsure if `MemorySegment` should closely couple with `NativeScope`, if there is an intention to separate out into base and/or other modules. Perhaps `NativeScope` is independently useful from the ABI that it should be pushed down into the memory-access branch and part of the core memory API?

I believe that NativeScope, although it was born as an ABI creature, makes sense as a standalone facility provided by the memory access API. So, even if we decided to split ABI into a separate module, NativeScope would go inside the foreign memory access API, as there's nothing ABI specific in there. It's at its core, an allocation helper.

JornVernee (approved):
Looks really good! Left some minor comments inline

test/jdk/java/foreign/TestNative.java
```diff
@@ -173,7 +173,7 @@ public void testNativeCapacity(Function<ByteBuffer, Buffer> bufferFunction, int
     public void testDefaultAccessModes() {
         MemoryAddress addr = allocate(12);
         MemorySegment mallocSegment = addr.asSegmentRestricted(12)
-                .withCleanupAction(() -> free(addr));
+                .handoff(HandoffTransform.ofShared().addCleanupAction(() -> free(addr)));
```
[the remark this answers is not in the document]
mcimadamore: I can change this so that it matches exactly the old code

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -418,6 +394,54 @@ default MemorySegment asSlice(MemoryAddress newBase) {
...
+     * spatial bounds and access modes (see {@link #accessModes()}) as this segment. Moreover, the
+     * confinement thread, and cleanup action of the new segment will be specified by the provided
+     * {@link HandoffTransform} instance. This is a <em>terminal operation</em>; as a side-effect, this segment will be
+     * marked as <em>not alive</em>, and subsequent operations on this segment will fail with {@link IllegalStateException}.
+     * <p>
+     * In case where the owner thread of the returned segment differs from that of this segment, write accesses to this
+     * segment's content <a href="../../../java/util/concurrent/package-summary.html#MemoryVisibility"><i>happens-before</i></a>
+     * hand-over from the current owner thread to the new owner thread, which in turn <i>happens before</i> read accesses
+     * to the returned segment's contents on the new owner thread.
+     *
```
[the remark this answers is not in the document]
mcimadamore: So, mentioning that cleanup actions are called on `MemorySegment#close` is a good thing.
Regarding Cleaners - your comment makes me wonder - cleaners are the only thing that is NOT propagated, should we propagate them for consistency; this way what you get out of handoff is a segment that has SAME characteristics as old segment - except some _additional_ stuff and, possibly, a different owner thread.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/NativeScope.java
```diff
@@ -29,12 +29,12 @@
 import jdk.internal.foreign.AbstractMemorySegmentImpl;
 import jdk.internal.foreign.AbstractNativeScope;
 import jdk.internal.foreign.Utils;
-import jdk.internal.misc.Unsafe;
 
 import java.lang.invoke.VarHandle;
 import java.lang.reflect.Array;
 import java.nio.ByteOrder;
 import java.util.OptionalLong;
+import java.util.function.Consumer;
```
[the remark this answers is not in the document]
mcimadamore: will fix

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -418,6 +394,54 @@ default MemorySegment asSlice(MemoryAddress newBase) {
...
+     * @param nativeScope the native scope.
+     * @return a new memory segment backed by the same underlying memory region as this segment, but whose life-cycle
+     * is tied to that of {@code nativeScope}.
+     * @throws IllegalStateException if this segment is not <em>alive</em>, or if access occurs from a thread other than the
+     * thread owning this segment.
+     * @throws UnsupportedOperationException if this segment does not support the {@link #HANDOFF} access mode.
```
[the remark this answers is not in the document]
mcimadamore: will fix

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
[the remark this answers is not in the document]
JornVernee: I think if we can propagate them (technically), we should. I don't know if there's a way to unregister them from the old scope as well? (i.e. prevent unecessary cleaner thread churn)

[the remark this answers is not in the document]
mcimadamore: when I looked into it - seemed like there was no way to unregister a cleaner. You can downcast the Cleanable when you register back to a PhantomCleanable, and that has a remove() method, but I think throws UOE.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -849,25 +873,106 @@ static MemorySegment ofNativeRestricted() {
...
+         */
+        static HandoffTransform ofConfined() {
+            return ofConfined(Thread.currentThread());
+        }
+
+        /**
+         * Specifies an additional cleanup action to be associated with the target segment. More specifically, the target
+         * segment produced by this transform will feature a cleanup action which will first call the user-provided action,
+         * before delegating back to cleanup action associated with the source segment.
+         * Any errors and/or exceptions thrown by the user-provided action will be discarded, and will not prevent the
```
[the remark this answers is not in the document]
mcimadamore: Will fix

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -849,25 +873,106 @@ static MemorySegment ofNativeRestricted() {
...
+         * @param action the new cleanup action
+         * @return this handoff transform.
+         * @throws NullPointerException if {@code action == null}
+         */
+        HandoffTransform addCleanupAction(Runnable action);
+
+        /**
+         * Specifies an additional attachment object which will be kept alive by the target segment, in addition
+         * to any attachment object which might also be associated with the source segment.
+         * This can be useful in cases where the lifecycle of the segment is dependent on that of some other external resource.
```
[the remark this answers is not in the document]
mcimadamore: I guess what I wanted to mean here is if the segment depends on some memory source which is maintained externally and must be kept alive. That is the bytebuffer case - you have a segment which has to keep the BB alive, otherwise access with segfault.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
PaulSandoz: `HandoffTransform.of()`, or `ofSource`, seems clear, such that `segment.handOff(HandoffTransform.of())` returns a new segment with exactly the same properties as the source.

I think being a dumb holder of mutable state is fine. It's simple and there is no binding/reuse issue as, say, with a builder scoped to a method call.

mcimadamore:
I ended up doing a pretty big U-turn here. After a long (and useful) offline discussion with Jorn, we agreed that:

1. stacking up multiple cleaners/attachments/cleanup action is a very obscure use cases
2. similarly, being able to replace cleaners/attachments/cleanup action is also a very obscure use cases
3. the real reason as to why we want attachments and cleanup actions in the first place is to support unchecked segment creation

So, I realized that my attempt at consolidating all handoff-like operation under the same umbrella was, perhaps, a bit of a siren song; we can deal with (3) by adding an overload on `MemoryAddress` (as we had once), and we can then worry only about changing ownership, sharing and registering cleaners. Now, since there are only 3 operations remaining, it feels like adding a new abstraction like `HandoffTransform` would be overkill; instead, let's just reshuffle the names of the terminal operations, as follows:

* `MemorySegment::handoff(Thread)` -> return new confined segment
* `MemorySegment::handoff(NativeScope)` -> new confined, non-closeable segment controlled by native scope
* `MemorySegment::share()` -> return new shared segment
* `MemorySegment::registerCleaner()` -> return new segment with cleaner attached - throws if there's one already

In other words, let's just use *verbs* for the terminal operations, and avoid overlap with `withXYZ` methods, which assume a more innocuous wither-like behavior. This patch is also preserving all the javadoc improvement introduced yesterday which, I think, make it very clear that certain operations are now terminal operations.

I've also simplified the code quite a bit - internally, there's no CleanupAction abstraction anymore; that's because, I realized, I could use a `PhantomCleanable` directly to achieve my needs; this class support de-registration, so now we can, instead of adding state to cleanup actions, just make sure that the cleanable is unregistered before returning. In fact, all terminal operation will do that - but in case of `share` and `handoff` we also make sure that the previous cleaner is re-registered onto the newly created scope.

I've also beefed up the cleaner test to check that things were working as expected,

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
```diff
@@ -134,18 +146,28 @@ final MemoryScope confineTo(Thread newOwner, boolean strict) {
...
     }
 
-    final MemoryScope wrapAction(Runnable runnable) {
+    MemoryScope cleanable(Cleaner cleaner) {
+        if (scopeCleanable != null) {
+            throw new IllegalStateException("Already registered with a cleaner");
+        }
         try {
             justClose();
+            if (scopeCleanable != null) {
```
PaulSandoz, line 165: `scopeCleanable` is checked above and if non-`null` throws.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
```diff
@@ -264,127 +281,19 @@ void justClose() {
...
-    /**
-     * A functional interface modelling the cleanup action associated with a scope.
-     */
-    interface CleanupAction extends Runnable {
-        void cleanup();
-        CleanupAction dup();
-        CleanupAction wrap(Runnable runnable);
+    static class ScopeCleanable extends PhantomCleanable<MemoryScope> {
+        final Cleaner cleaner;
+        Runnable cleanupAction;
```
PaulSandoz, line 286: Mark field and class as `final`

PaulSandoz (approved):
This is indeed simpler (esp. also for implementation of the clean up actions).

I am not entirely convinced with clean up action being only associated with unchecked segment creation. For example the [BLIS object API](https://github.com/flame/blis/blob/master/docs/BLISObjectAPI.md) may be used to create (`bli_obj_create`) and free (`bli_obj_free`) the allocated region of memory representing a matrix with elements of a certain type. The BLIS `obj_t` would be allocated separately as a segment and it may be useful to attach a cleanup action to that segment to free the region using the native method call.

JornVernee:
> I am not entirely convinced with clean up action being only associated with unchecked segment creation. For example the BLIS object API may be used to create (bli_obj_create) and free (bli_obj_free) the allocated region of memory representing a matrix with elements of a certain type. The BLIS obj_t would be allocated separately as a segment and it may be useful to attach a cleanup action to that segment to free the region using the native method call.

This is exactly the use-case we have in mind, which is supported by allowing attachments when creating unchecked segments. e.g. you would call bli_obj_create, get back a MemoryAddress that points to the object. Then you could turn this MemoryAddress into an unchecked MemorySegment, passing bli_obj_free as the cleanup action.

What we're saying here is that there is no need to change and/or attach _more_ cleanup actions when doing a handoff/share/registerCleaner operation on that segment later.

mcimadamore:
> This is indeed simpler (esp. also for implementation of the clean up actions).
> 
> I am not entirely convinced with clean up action being only associated with unchecked segment creation. For example the [BLIS object API](https://github.com/flame/blis/blob/master/docs/BLISObjectAPI.md) may be used to create (`bli_obj_create`) and free (`bli_obj_free`) the allocated region of memory representing a matrix with elements of a certain type. The BLIS `obj_t` would be allocated separately as a segment and it may be useful to attach a cleanup action to that segment to free the region using the native method call.

I understand the concern - however, the cleanup action associated to an unchecked memory segment has always been more about supporting low-level pieces of machinery, such as custom memory allocators, which might want to allocate memory in C/C++ and then exposing said memory using a plain MemorySegment.

For more convoluted idioms such as the one you describe, where the library mandates some setup/destroy actions, I feel that adding an explicit construct (which might also be auto closeable) _wrapping_ a segment, might be a better way to go, rather than exposing segment _as is_.

JornVernee (approved):

PR #362, opened 2020-09-30
8253841: jextract --source mode is broken

sundararajana:
* left constant name is specially modified which is problematic.
* access flags for source constant classes should not be just ""

---------
### Progress

### Issue
 * [JDK-8253841](https://bugs.openjdk.java.net/browse/JDK-8253841): jextract --source mode is broken

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/362/head:pull/362`
`$ git checkout pull/362`


sundararajana:
"left constant name is specially modified which is problematic"

should have been

"leaf constant name is specially modified which is problematic."

mcimadamore (approved):
Thanks for fixing this - forgot to revert these when I went from header-based generation back to the previous model

PR #363, opened 2020-10-01
Investigate ways to make handoff-like operation more explicit (foreign-memacess edition)

mcimadamore:
This is a subset of the patch proposed here:
https://git.openjdk.java.net/panama-foreign/pull/361

I'd like to integrate the memory access bits first and then merge the changes into PR 361.

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/363/head:pull/363`
`$ git checkout pull/363`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryAddress.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryAddress.java
@@ -31,6 +31,8 @@ import jdk.internal.foreign.MemoryAddressImpl;
 import jdk.internal.foreign.NativeMemorySegmentImpl;
 import jdk.internal.foreign.Utils;
 
+import java.lang.ref.Cleaner;
+
 /**
  * A memory address models a reference into a memory location. Memory addresses are typically obtained using the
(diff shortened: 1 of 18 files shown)


JornVernee (approved):
Looks good. Though, I think we should mention somewhere on handoff/share/registerCleaner, or maybe on the class, that attachments, cleanup actions, and cleaners are transferred to the new segment when doing these operations.

mcimadamore:
> Looks good. Though, I think we should mention somewhere on handoff/share/registerCleaner, or maybe on the class, that attachments, cleanup actions, and cleaners are transferred to the new segment when doing these operations.

Not sure I fully agree - in the sense that cleanupAction/attachment are no longer first class properties of the segment - as much as they are property of segments created unsafely.

For cleaner, it's a different story, I'll work on something.

JornVernee:
> > Looks good. Though, I think we should mention somewhere on handoff/share/registerCleaner, or maybe on the class, that attachments, cleanup actions, and cleaners are transferred to the new segment when doing these operations.
> 
> Not sure I fully agree - in the sense that cleanupAction/attachment are no longer first class properties of the segment - as much as they are property of segments created unsafely.

Could also mention this on the asSegmentRestricted factory maybe?

JornVernee (approved):

PR #369, opened 2020-10-02
8253937: Simplify support for shared memory segment

mcimadamore:
This patch removs the dependency between shared memory segment support and VM async exceptions support. This dependency has proven to be problematic, since not all the code paths in the VM handle async exceptions correctly (this can also be seen by the number of places which have to be touched in the VM code - which are now reverted by this patch).

Instead, we can adopt a much simpler approach where, if when closing a segment, we detect concurrent access inside the handshake, we simply fail on close. I've added extra javadoc to clarify that, if a client sees an exception when closing a segment, that should be perceived as a bug in the code (lack of synchronization) rather than be tempted to catch the exception and rethrow.

Overall, the support requires a lot less from the VM and I think it's a fairly reasonable compromise. Note that now shared scopes have three states:

* ALIVE
* CLOSING
* CLOSED

The liveness check upon segment access fails if state != ALIVE; on the other hand, the public API `MemorySegment::isAlive` returns true, as long as the state is != CLOSED. This means that clients sees only one transition, from `isAlive() == true` to `isAlive() == false`. 

We will keep experimenting with the async exception support in a separate branch, as a future implementation improvement (note that relaxing the semantics of `close` so that it will *never* fail, is both source- and binary- compatible with the approach proposed here).

---------
### Progress

### Issue
 * [JDK-8253937](https://bugs.openjdk.java.net/browse/JDK-8253937): Simplify support for shared memory segment

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/369/head:pull/369`
`$ git checkout pull/369`

### DIFF
--- a/src/hotspot/share/c1/c1_Runtime1.cpp
+++ b/src/hotspot/share/c1/c1_Runtime1.cpp
@@ -636,7 +636,6 @@ address Runtime1::exception_handler_for_pc(JavaThread* thread) {
   {
     // Enter VM mode by calling the helper
     ResetNoHandleMark rnhm;
-    ExceptionHandlingMark ehm(thread);
     continuation = exception_handler_for_pc_helper(thread, exception, pc, nm);
   }
   // Back in JAVA, use no oops DON'T safepoint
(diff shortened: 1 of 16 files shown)


PaulSandoz (approved):
Exceptions on `close` seems a reasonable compromise for now.

PR #371, opened 2020-10-07
8254156: Simplify ABI classification logic

mcimadamore:
This patch reverts recent changes to introduce a public CValueLayout class; the realization here is that, after all, CValueLayout exposes a public `Type` enum which is used by ABI classification. So, instead of using subclassing (which is messy and tedious with immutable data structures with covariant overrides), let's just double down on layout attributes, get rid of CValueLayout and simply expose the Type enum (now renamed to TypeKind).

---------
### Progress

### Testing

| Build | ✔️ (3/3 passed) | ✔️ (2/2 passed) | ✔️ (2/2 passed) |
| Test (tier1) | ✔️ (9/9 passed) | ⏳ (5/9 running) | ✔️ (9/9 passed) |

### Issue
 * [JDK-8254156](https://bugs.openjdk.java.net/browse/JDK-8254156): Simplify ABI classification logic

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/371/head:pull/371`
`$ git checkout pull/371`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
@@ -31,44 +31,42 @@ import jdk.internal.foreign.Utils;
 import jdk.internal.foreign.abi.SharedUtils;
 
 import java.lang.constant.Constable;
-import java.lang.constant.DynamicConstantDesc;
 import java.lang.invoke.MethodHandle;
 import java.lang.invoke.MethodType;
-import java.nio.ByteOrder;
 import java.nio.charset.Charset;
-import java.util.Map;
 import java.util.Objects;
-import java.util.Optional;
 import java.util.function.Consumer;
 
-import static java.lang.constant.ConstantDescs.BSM_GET_STATIC_FINAL;
 import static jdk.internal.foreign.PlatformLayouts.*;
 
 /**
  * A foreign linker specializing for C Application Binary Interface (ABI) calling conventions.
  * Instances of this interface can be used to link foreign functions in native libraries that
  * follow the JVM's target platform C ABI.
- *
- * <p>There are two components that go into linking a foreign function: a method type, and
+ * <p>
+ * There are two components that go into linking a foreign function: a method type, and
  * a function descriptor. The method type, consists of a set of <em>carrier</em> types, which, together,
  * specify the Java signature which clients must adhere to when calling the underlying foreign function.
  * The function descriptor contains a set of memory layouts which, together, specify the foreign function
- * signature and classification information (via custom layout attributes), so that linking can take place.
- * Memory layout attributes are used in the function descriptor to attach ABI classification meta-data to
- * memory layouts, which are required for linking. Clients of this API should build function descriptors
- * using the predefined memory layout constants (based on a subset of the built-in types provided by the C language),
- * found in this interface; a failure to do so might result in linkage errors, given that linking requires additional
- * classification information to determine, for instance, how arguments should be loaded into registers during a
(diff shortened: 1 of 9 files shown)


src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/PlatformLayouts.java
```diff
@@ -52,47 +50,56 @@ public static MemoryLayout asVarArg(MemoryLayout ml) {
...
+     * Creates a new ValueLayout with the {@code POINTER} kind
      *
      * @param order the byte order of the layout
      * @param bitSize the size, in bits, of the layout
-     * @return the newly created CValueLayout
+     * @return the newly created ValueLayout
      */
-    public static CValueLayout ofPointer(ByteOrder order, long bitSize) {
-        return new CValueLayout(CValueLayout.Kind.POINTER, order, bitSize, bitSize, Map.of());
+    public static ValueLayout ofPointer(ByteOrder order, long bitSize) {
```
JornVernee, line 100: Looks like this can be private? (and in that case I think it's okay to drop the javadoc as well, since none of the other factories have it either.)

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/aarch64/TypeClass.java
```diff
@@ -42,11 +42,10 @@
     private static final int MAX_AGGREGATE_REGS_SIZE = 2;
 
     private static TypeClass classifyValueType(ValueLayout type) {
-        if (!(type instanceof CLinker.CValueLayout)) {
-            throw new IllegalStateException("Unexpected value layout: could not determine ABI class");
-        }
+        CLinker.TypeKind kind = (CLinker.TypeKind)type.attribute(CLinker.TypeKind.ATTR_NAME).orElseThrow(
+                () -> { throw new IllegalStateException("Unexpected value layout: could not determine ABI class"); });
```
JornVernee, line 46: Note that orElseThrow takes a `Supplier<Throwable>` (basically), so this can be simplified to return the new exception instance instead of throwing it.

Also, this code is copied in the other TypeClass classes, so might be good to create a helper method in SharedUtils for it instead?

JornVernee (approved):

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/aarch64/TypeClass.java
mcimadamore: Done

JornVernee (approved):

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
```diff
@@ -707,196 +705,90 @@ static VaList empty() {
...
-        }
+        DOUBLE(false),
+        /**
+         * A kind corresponding to the C {@code long double} type
+         */
+        LONGDOUBLE(false),
+        /**
+         * A kind corresponding to the a C pointer type
+         */
+        POINTER(false);
```
chrisvest, line 749: Pointer or pointer-sized? We'd also use this one for `size_t` for instance, right?

mcimadamore: This is pointer; e.g. values of this layout will be associated with the MemoryAddress.class carrier. So, no, for size_t, you need something different (either LONG or LONGLONG, depending on the platform). We could probably add a SIZE_T in there, since it's frequent enough.

PR #375, opened 2020-10-08
8254260: Consider splitting binding recipe operators that serve a dual role

JornVernee:
Hi,

This patch splits the binding recipe operators that currently serve a dual purpose into 2 operators each:
- MOVE -> VM_STORE and VM_LOAD
- DEREFERENCE -> BUFFER_STORE and BUFFER_LOAD
- CONVERT_ADDRESS -> BOX_ADDRESS and UNBOX_ADDRESS

Note that I also added a TO_SEGMENT operator which converts a MemoryAddress into a MemorySegment. This was previously done as part of the COPY operator, but only for upcalls, so I had to split out this functionality so that COPY could have only one interpretation as well. This will also enable some more optimizations down the line, such as merely wrapping a by-value struct passed as a pointer in a MemorySegment, instead of having to make a copy.

This also regularized the use of allocators by the operators. Now an allocator is always passed as an argument, and that's what is used to do allocations, instead of relying implicitly on MemorySegment::allocateNative. (The index of the allocator argument is now also always passed when specializing an operator, to indicate the dependency, and so that operator impls don't have to 'guess' which argument index is the allocator (see for instance the implementation of Copy::specialize).

The rest of the patch is more or less a mechanical renaming, and updating the tests/CallArrangers to use the different operator names.

Thanks,
Jorn

---------
### Progress

### Testing

| Build | ❌ (2/5 failed) | ✔️ (2/2 passed) | ✔️ (2/2 passed) |
| Build / test | ⏳ (1/1 running) |    |     | 
| Test (tier1) |    |  ⏳ (3/9 running) | ✔️ (9/9 passed) |

**Failed test tasks**
- [Linux x64 (build hotspot minimal)](https://github.com/JornVernee/panama-foreign/runs/1231457476)
- [Linux x64 (build hotspot zero)](https://github.com/JornVernee/panama-foreign/runs/1231457450)

### Issue
 * [JDK-8254260](https://bugs.openjdk.java.net/browse/JDK-8254260): Consider splitting binding recipe operators that serve a dual role

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/375/head:pull/375`
`$ git checkout pull/375`


src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/Binding.java
```diff
@@ -281,22 +283,47 @@ private static void checkType(Class<?> type) {
...
         return copy;
     }
 
-    public static Move move(VMStorage storage, Class<?> type) {
+    private static MemorySegment allocateBuffer(long size, long allignment, SharedUtils.Allocator allocator) {
+        return allocator.allocate(size, allignment);
+    }
+
+    // FIXME should register with scope
+    private static MemorySegment toSegment(MemoryAddress operand, long size) {
```
mcimadamore, line 303: This seems to be only used inside the `ToSegment` binding, so, perhaps this could should be moved there (also because this is an interpreter function, not a binding factory - e.g. the overload with the `toSegment` binding factory is confusing)

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/Binding.java
```diff
@@ -324,19 +359,42 @@ public static Dup dup() {
         return new Binding.Builder();
     }
 
+    @Override
+    public boolean equals(Object o) {
```
mcimadamore, line 363: General question - where do we use equals/hash? Just for testing purposes right?

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/ProgrammableInvoker.java
```diff
@@ -323,7 +340,9 @@ Object invokeMoves(Object[] args, Binding.Move[] argBindings, Binding.Move[] ret
     Object invokeInterpBindings(Object[] args, MethodHandle leaf,
                                 Map<VMStorage, Integer> argIndexMap,
                                 Map<VMStorage, Integer> retIndexMap) throws Throwable {
-        NativeScope scope = bufferCopySize != 0 ? NativeScope.boundedScope(bufferCopySize) : null;
+        boolean hasBufferCopies = bufferCopySize != 0;
+        NativeScope scope = hasBufferCopies ? NativeScope.boundedScope(bufferCopySize) : null;
```
mcimadamore, line 344: If Allocator had a close() method, this could be simplified a little?

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/ProgrammableInvoker.java
```diff
@@ -200,58 +216,59 @@ private MethodHandle specialize(MethodHandle leafHandle) {
 
         MethodHandle specializedHandle = leafHandle; // initial
 
-        int insertPos = -1;
+        int argInsertPos = -1;
```
mcimadamore, line 219: The more we tweak this routine, the more I'm convinced we should just spin specialized bytecode :-) - it's becoming increasingly hard to follow what's happening here. (not for this review - more longer term)

test/jdk/java/foreign/NativeTestHelper.java
```diff
@@ -33,7 +33,7 @@
     }
 
     public static boolean isIntegral(MemoryLayout layout) {
-        return kind(layout).isIntergral();
+        return kind(layout).isIntegral();
```
mcimadamore, line 36: Whoops - sorry - I guess I fixed the API and have not run the tests?

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/Binding.java
```diff
@@ -218,21 +216,27 @@
...
+                    methodType(MemorySegment.class, MemoryAddress.class, long.class));
         } catch (ReflectiveOperationException e) {
             throw new RuntimeException(e);
         }
     }
 
     enum Tag {
-        MOVE,
-        DEREFERENCE,
+        VM_STORE,
```
mcimadamore, line 230: Should these be called REG_STORE and REG_LOAD? Ah, I guess we also have stack, right? Ok, I agree with the name - these are "moves" that only the VM can perform, makes sense.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/Binding.java
```diff
@@ -476,77 +562,111 @@ public long offset() {
...
+        }
+    }
+
+    /**
+     * BUFFER_STORE([offset into memory region], [type])
+     * Pops a MemorySegment from the operand stack, loads a [type] from
+     * [offset into memory region] from it, and pushes it onto the operand stack.
+     * The [type] must be one of byte, short, char, int, long, float, or double
+     */
+    public static class BuffferStore extends Dereference {
```
mcimadamore, line 591: Typo, there are 3 `f`

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/Binding.java
```diff
@@ -476,77 +562,111 @@ public long offset() {
...
+            MethodHandle setter = varHandle().toMethodHandle(VarHandle.AccessMode.SET);
+            setter = setter.asType(methodType(void.class, MemorySegment.class, type()));
+            return collectArguments(specializedHandle, insertPos + 1, setter);
         }
 
-        private VarHandle varHandle() {
-            return MemoryHandles.insertCoordinates(MemoryHandles.varHandle(type, ByteOrder.nativeOrder()), 1, offset);
+        @Override
+        public String toString() {
+            return "BuffferStore{" +
```
mcimadamore, line 622: Typo here too

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/CallingSequenceBuilder.java
```diff
@@ -90,7 +90,12 @@ private static void verifyUnboxBindings(Class<?> inType, List<Binding> bindings)
         stack.push(inType);
 
         for (Binding b : bindings) {
-            b.verifyUnbox(stack);
+            if (b.tag() == Binding.Tag.VM_LOAD
```
mcimadamore, line 93: We could have static EnumSet for what's an box vs. unbox binding, so that you can filter by checking for presence in the right enum set?

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/CallingSequenceBuilder.java
```diff
@@ -102,7 +107,11 @@ private static void verifyBoxBindings(Class<?> expectedReturnType, List<Binding>
         Deque<Class<?>> stack = new ArrayDeque<>();
```
mcimadamore, line 107: I've seen in the code that the verifyUnbox method speaks about a `returnType` parameter - should it be an `outType` for symmetry (since, I guess, we use this more than just for returns - e.g. we use it for args in upcalls).

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/ProgrammableUpcallHandler.java
```diff
@@ -98,7 +99,7 @@ private void invoke(MemoryAddress buffer) {
                                 ? stackArgsBase.asSlice(storage.index() * abi.arch.typeSize(abi.arch.stackType()))
                                 : bufferBase.asSlice(layout.argOffset(storage));
                             return SharedUtils.read(ptr, type);
-                        });
+                        }, DEFAULT_ALLOCATOR);
```
mcimadamore, line 102: For upcalls, couldn't we use a `NativeScope` and close at the end of the call? If the callback leaks segments outside... well, too bad.

mcimadamore (approved):
Looks a great cleanup. Left some comments.
I also ran all tests here and everything looks good.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/Binding.java
JornVernee: Yeah, good idea. This and 2 others are used for the `specialize` implemantion, but they can be moved inside the binding classes.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/Binding.java
JornVernee: Yeah, we use them at least equals for the CallArranger tests, which is comparing arrays of Bindings.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/ProgrammableInvoker.java
JornVernee: This is not really a problem with Allocator, it's a problem with having either a NativeScope, or having none (null), which will result in an NPE when trying to reference `scope::allocateNative`. That's why this code is conditional.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/ProgrammableInvoker.java
JornVernee: Yeah, especially the indexing is tricky since we have to walk the recipe in reverse, but we shouldn't have that problem when spinning bytecode.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/Binding.java
JornVernee: REG_LOAD and REG_STORE also works for me. The name is derived from VMReg in the HotSpot code -> VMStorage in our code -> VM_LOAD/VM_STORE. (So, REG works equally well).

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/Binding.java
JornVernee: Thanks!

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/CallingSequenceBuilder.java
JornVernee: Yeah, that's a good idea.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/ProgrammableUpcallHandler.java
JornVernee: Yes, exactly. But, seemed like a separate patch.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/CallingSequenceBuilder.java
JornVernee: I'll change the name. FWIW, `returnType` refers to the type 'returned' by the chain of operations.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/ProgrammableInvoker.java
JornVernee: I'm experimenting with this a bit. Seems like there is some room for improvement.

JornVernee:
@mcimadamore Thanks for the review!

I've added the suggested close method to SharedUtils.Allocator (see: https://github.com/openjdk/panama-foreign/pull/375/commits/9ffebc0016672e273a066c84f8cafdae55bb10c4). Looks like an improvement?

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/Binding.java
```diff
@@ -215,11 +215,11 @@
                     methodType(MemoryAddress.class, long.class));
             MH_BASE_ADDRESS = lookup.findVirtual(MemorySegment.class, "address",
                     methodType(MemoryAddress.class));
-            MH_COPY_BUFFER = lookup.findStatic(Binding.class, "copyBuffer",
+            MH_COPY_BUFFER = lookup.findStatic(Binding.Copy.class, "copyBuffer",
```
mcimadamore, line 218: I'm assuming these are only really required in the specialize() method of the various subclasses - should we move those constants there too?

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/ProgrammableInvoker.java
```diff
@@ -340,10 +341,10 @@ Object invokeMoves(Object[] args, Binding.VMStore[] argBindings, Binding.VMLoad[
     Object invokeInterpBindings(Object[] args, MethodHandle leaf,
                                 Map<VMStorage, Integer> argIndexMap,
                                 Map<VMStorage, Integer> retIndexMap) throws Throwable {
-        boolean hasBufferCopies = bufferCopySize != 0;
-        NativeScope scope = hasBufferCopies ? NativeScope.boundedScope(bufferCopySize) : null;
-        SharedUtils.Allocator unboxAllocator = hasBufferCopies ? SharedUtils.Allocator.ofScope(scope) : null;
-        try {
+        SharedUtils.Allocator unboxAllocator = bufferCopySize != 0
```
mcimadamore, line 344: Nice! I also like THROWING_ALLOCATOR more than `null`

mcimadamore (approved):
Very nice - just added another minor stylistic comment on binding MH constants

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/Binding.java
JornVernee: We can, though we end up duplicating the try/catch & lookup code if we do, or would have to add some helper method, for both findStatic and findVirtual. That's why I held off on doing that.

PR #378, opened 2020-10-12
8254343: Revisit API for supporting mapped memory segments

mcimadamore:
This patch contains an alternate approach to support mapped memory segment operations,
effectively addressing the issues described in PR/377.

Here, support for mapped segments is reduced to bare minimum in the public API; a mapped segment is a segment whose file descriptor is set.

A separate helper class `MappedMemorySegments` exposes the same routines which are now provided by `MappedMemorySegment`.

---------
### Progress

### Testing

| Build | ❌ (2/5 failed) | ⏳ (2/2 running) | ✔️ (2/2 passed) |
| Test (tier1) |    |     |  ⏳ (9/9 running) |

**Failed test tasks**
- [Linux x64 (build hotspot minimal)](https://github.com/mcimadamore/panama-foreign/runs/1243493214)
- [Linux x64 (build hotspot zero)](https://github.com/mcimadamore/panama-foreign/runs/1243493198)

### Issue
 * [JDK-8254343](https://bugs.openjdk.java.net/browse/JDK-8254343): Revisit API for supporting mapped memory segments

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/378/head:pull/378`
`$ git checkout pull/378`


PaulSandoz:
I like this approach, very minimal on `MemorySegment`. Since `MappedByteBuffer` has such operations on mapped memory it's somewhat of a forced move to provide something specific for `MemorySegment`, rather than asking developers to transition to a `ByteBuffer`. We could enhance the documentation on`MappedMemorySegments` when the Foreign Linker API has sedimented, to refer to more platform-specific use cases.

PaulSandoz (approved):

PR #391, opened 2020-11-10
8256080: AArch64: fatal error with unexpected intrinsic id _linkToNative

nick-arm:
AArch64 doesn't implement linkToNative yet. However during VM startup
several of the MethodHandles methods are called with the _linkToNative
intrinsic ID. Add minimal handling of this (taken from x86) to avoid
fatal errors.

---------
### Progress

### Testing

| Build | ❌ (1/5 failed) | ❌ (2/2 failed) | ✔️ (2/2 passed) | ✔️ (2/2 passed) |
| Test (tier1) |    |     |  ✔️ (9/9 passed) | ✔️ (9/9 passed) |

**Failed test tasks**
- [Linux x64 (build hotspot zero)](https://github.com/nick-arm/panama-foreign/runs/1388288453)
- [Linux x86 (build debug)](https://github.com/nick-arm/panama-foreign/runs/1388288490)
- [Linux x86 (build release)](https://github.com/nick-arm/panama-foreign/runs/1388288485)

### Issue
 * [JDK-8256080](https://bugs.openjdk.java.net/browse/JDK-8256080): AArch64: fatal error with unexpected intrinsic id _linkToNative

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/391/head:pull/391`
`$ git checkout pull/391`

### DIFF
--- a/src/hotspot/cpu/aarch64/methodHandles_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/methodHandles_aarch64.cpp
@@ -177,6 +177,13 @@ address MethodHandles::generate_method_handle_interpreter_entry(MacroAssembler* 
     return NULL;
   }
 
+  // No need in interpreter entry for linkToNative for now.
+  // Interpreter calls compiled entry through i2c.
+  if (iid == vmIntrinsics::_linkToNative) {
+    __ hlt(0);
+    return NULL;
+  }
+
   // r13: sender SP (must preserve; see prepare_to_jump_from_interpreted)
   // rmethod: Method*
   // r3: argument locator (parameter slot count, added to rsp)
@@ -271,7 +278,10 @@ void MethodHandles::generate_method_handle_dispatch(MacroAssembler* _masm,
   assert_different_registers(temp1, temp2, temp3, receiver_reg);
   assert_different_registers(temp1, temp2, temp3, member_reg);
 
-  if (iid == vmIntrinsics::_invokeBasic) {
+  if (iid == vmIntrinsics::_invokeBasic || iid == vmIntrinsics::_linkToNative) {
+    if (iid == vmIntrinsics::_linkToNative) {
+      assert(for_compiler_entry, "only compiler entry is supported");
+    }
     // indirect through MH.form.vmentry.vmtarget
     jump_to_lambda_form(_masm, receiver_reg, rmethod, temp1, for_compiler_entry);
 
--- a/src/hotspot/cpu/aarch64/sharedRuntime_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/sharedRuntime_aarch64.cpp
@@ -1194,7 +1194,7 @@ static void gen_special_dispatch(MacroAssembler* masm,
     member_arg_pos = method->size_of_parameters() - 1;  // trailing MemberName argument
     member_reg = r19;  // known to be free at this point
     has_receiver = MethodHandles::ref_kind_has_receiver(ref_kind);
-  } else if (iid == vmIntrinsics::_invokeBasic) {
+  } else if (iid == vmIntrinsics::_invokeBasic || iid == vmIntrinsics::_linkToNative) {
     has_receiver = true;
   } else {
     fatal("unexpected intrinsic id %d", iid);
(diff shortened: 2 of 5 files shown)


JornVernee (approved):
I'm surprised to hear that this is what's causing the problems. We don't have any handling for _linkToNative in e.g. sharedRuntime_x86_32.cpp in gen_special_dispatch either, but that doesn't seem to be a problem for a 32-bit build (with some additional fixes).

I also didn't expect this code to be called before linking a _linkToNative call. I'm curious to know where the calls during startup are coming from.

Any ways, thanks for debugging and fixing. The only comment I have is that this should go into the foreign-abi branch, so please change the target branch.

mcimadamore:
Thanks for this! I'm currently testing the patch. One thing I noted is that this PR has been filed against the `foreign-jextract` branch and not the `foreign-abi` one which would be more appropriate.

mcimadamore:
The following tests seem to fail:

* java/foreign/TestUpcallHighArity.java
* java/foreign/TestIntrinsics.java
* java/foreign/StdLibTest.java
* java/foreign/stackwalk/TestStackWalk.java

The last three of them fail with a crash:

```
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  Internal Error (doCall.cpp:662), pid=6557, tid=6572
#  guarantee(failing()) failed: call failed to generate:  calls should work
#
# JRE version: Java(TM) SE Runtime Environment (16.0) (build 16-internal+0-2020-11-10-1203454.maurizio.cimadamore.dev)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (16-internal+0-2020-11-10-1203454.maurizio.cimadamore.dev, mixed mode, tiered, compressed oops, g1 gc, linux-aarch64)
# Problematic frame:
# V  [libjvm.so+0x5ee0ec]  Parse::do_call()+0x72c
#
```
`TestHighArity` seems to fail because of a bad method handle adaptation. I believe in all cases the problem is related with the specialization and intrinsifics support.

Let me know what you want to do; I'm ok to approve these changes as is (as they obviously make the world a lot better) and fix the other issues as followup - or wait until we have fixes for these tests as well.

nick-arm:
I've changed the target branch to `foreign-abi`.

@JornVernee the stack trace looks like this:

```
V  [libjvm.so+0xeed41c]  MethodHandles::generate_method_handle_interpreter_entry(MacroAssembler*, vmIntrinsics::ID)+0x290
V  [libjvm.so+0xee4edc]  MethodHandlesAdapterGenerator::generate()+0x68
V  [libjvm.so+0xee4e30]  MethodHandles::generate_adapters()+0x130
V  [libjvm.so+0xab6954]  init_globals()+0xe4
V  [libjvm.so+0x121b424]  Threads::create_vm(JavaVMInitArgs*, bool*)+0x34c
V  [libjvm.so+0xbd1120]  JNI_CreateJavaVM_inner(JavaVM_**, void**, void*)+0x100
V  [libjvm.so+0xbd1408]  JNI_CreateJavaVM+0x2c
```

`MethodHandlesAdapterGenerator::generate()` calls `MethodHandles::generate_method_handle_interpreter_entry()` once for each intrinsic ID. (I think the second change in `methodHandles_aarch64.cpp` is perhaps not strictly necessary.)

@mcimadamore The C2 crash should be fixed with the extra commit I pushed just now. We need to call `C->record_failure(..)` if the native invoker is not implemented otherwise a C2 assertion fails later.

TestIntrinsics.java was crashing because the `CallNative` opcode is not implemented. I've restricted this test to just x86 for now.

The failure of TestUpcallHighArity.java seems unrelated to linkToNative and this PR - I'll look at that separately.

JornVernee:
>  (I think the second change in methodHandles_aarch64.cpp is perhaps not strictly necessary.)
It would be needed for AArch64, since we actually use native method handles on that platform, they just don't get intrinsified currently.

For other platforms it should not be needed, since we don't have a CLinker implementation for those yet.

Thanks, I'll push that fix to other platforms as well.

JornVernee:
With that test requirement, the test also gets disabled on Mac, which is not what we want. You can add Mac as well using:

    @requires (os.arch == "x86_64") | (os.arch == "amd64")

nick-arm:
> With that test requirement, the test also gets disabled on Mac, which is not what we want.  

Thanks, fixed.

JornVernee (approved):

PR #395, opened 2020-11-18
8256466: MemoryLayout factories do not guard against null layouts

mcimadamore:
This is a trivial, but a bit bigger than expected patch which fixes mentions of NPEs in the foreign memory access API javadoc. Almost all API points are affected, from memory layouts, to memory segments, etc.   
This patch does a number of things:
* Document the various possible NPEs with corresponding `@throws` javadoc tags
* Add `Objects::requireNotNull calls` where required
* Added tests to check for nulls
* Added a brand new test for `MemoryAccess`; this is both to check normal behavior, and also to check behavior when it comes to nulls

---------
### Progress

### Testing

| Build | ✔️ (5/5 passed) | ✔️ (2/2 passed) | ✔️ (2/2 passed) | ✔️ (2/2 passed) |
| Test (tier1) | ✔️ (9/9 passed) | ✔️ (9/9 passed) | ✔️ (9/9 passed) | ✔️ (9/9 passed) |

### Issue
 * [JDK-8256466](https://bugs.openjdk.java.net/browse/JDK-8256466): MemoryLayout factories do not guard against null layouts

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/395/head:pull/395`
`$ git checkout pull/395`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryLayout.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryLayout.java
@@ -41,6 +41,7 @@ import java.util.OptionalLong;
 import java.util.Set;
 import java.util.function.Function;
 import java.util.function.UnaryOperator;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 /**
@@ -250,6 +251,7 @@ public interface MemoryLayout extends Constable {
      *
      * @param name the layout name.
      * @return a new layout which is the same as this layout, except for the <em>name</em> associated to it.
+     * @throws NullPointerException if {@code name == null}.
      * @see MemoryLayout#name()
      */
     MemoryLayout withName(String name);
@@ -308,6 +310,7 @@ public interface MemoryLayout extends Constable {
      *
      * @param name the attribute name
      * @return the attribute with the given name (if it exists).
+     * @throws NullPointerException if {@code name == null}.
      */
     Optional<Constable> attribute(String name);
 
@@ -319,6 +322,7 @@ public interface MemoryLayout extends Constable {
      * @param name the attribute name.
      * @param value the attribute value.
      * @return a new memory layout which features the same attributes as this layout, plus the newly specified attribute.
+     * @throws NullPointerException if {@code name == null}.
      */
     MemoryLayout withAttribute(String name, Constable value);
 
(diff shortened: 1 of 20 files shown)


src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryLayout.java
```diff
@@ -432,16 +441,18 @@ default MemoryLayout select(PathElement... elements) {
...
      */
     default MemoryLayout map(UnaryOperator<MemoryLayout> op, PathElement... elements) {
+        Objects.requireNonNull(op);
         return computePathOp(LayoutPath.rootPath(this, l -> 0L), path -> path.map(op),
                 EnumSet.of(PathKind.SEQUENCE_ELEMENT_INDEX, PathKind.SEQUENCE_RANGE), elements);
     }
 
     private static <Z> Z computePathOp(LayoutPath path, Function<LayoutPath, Z> finalizer,
                                        Set<LayoutPath.PathElementImpl.PathKind> badKinds, PathElement... elements) {
         for (PathElement e : elements) {
```
JornVernee, line 454: What if `elements` itself is `null`? Doesn't seem to be covered currently in either impl or doc.

test/jdk/java/foreign/TestSpliterator.java
```diff
@@ -228,6 +228,11 @@ protected Long compute() {
         };
     }
 
+    @Test(expectedExceptions = NullPointerException.class)
+    public void testNullSpliteratoorLayout() {
```
JornVernee, line 232: typo:
```suggestion
    public void testNullSpliteratorLayout() {
```

JornVernee (changes requested):

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryLayout.java
JornVernee: e.g. a case like 
```
MemoryLayouts.BITS_8_BE.map(UnaryOperator.identity(), null);
```

mcimadamore: You are right, I got confused between `null` and zero-element array (which we do support with varargs call).

JornVernee (approved):
LGTM! Thanks for the thorough testing!

PR #396, opened 2020-11-19
8252504: Add a method to MemoryLayout which returns a offset-computing method handle

JornVernee:
This PR adds two new methods: `bitOffsetHandle` and `byteOffsetHandle` to `MemoryLayout`, that can be used to create method handles that can compute the effective offset of a layout element given a set of additional `long` indexes.

I also cleaned up a leftover in the javadoc of bitOffset and byteOffset where an API note was stating that `sequenceElement()` layout path elements would be interpreted as `sequenceElement(0)`, but they were actually being rejected outright (as they should).

I also added some tests to cover the corner case where an UnsupportedOperationException was being thrown. (some of the churn in the tests is from introducing static imports. I hope it's not too noisy).

---------
### Progress

### Issue
 * [JDK-8252504](https://bugs.openjdk.java.net/browse/JDK-8252504): Add a method to MemoryLayout which returns a offset-computing method handle

### Reviewers
 * [Chris Hegarty](https://openjdk.java.net/census#chegar) (@ChrisHegarty - no project role)

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/396/head:pull/396`
`$ git checkout pull/396`


src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryLayout.java
```diff
@@ -350,16 +348,36 @@ default long byteAlignment() {
...
-        return computePathOp(LayoutPath.rootPath(this, MemoryLayout::bitSize), LayoutPath::offset, EnumSet.of(PathKind.SEQUENCE_ELEMENT, PathKind.SEQUENCE_RANGE), elements);
+        return computePathOp(LayoutPath.rootPath(this, MemoryLayout::bitSize), LayoutPath::offset,
+                EnumSet.of(PathKind.SEQUENCE_ELEMENT, PathKind.SEQUENCE_RANGE), elements);
+    }
+
+    /**
+     * Creates a method handle that can be used to compute the offset, in bits, of the layout selected
+     * by a given layout path, where the path is considered rooted in this layout.
+     *
+     * <p>This is like {@link #bitOffset(PathElement...)}, except that free dimensions
```
mcimadamore, line 359: I'd replace this text, or maybe move it after we explain what the signature of the returned method handle is. E.g.

```
The returned method handle has a return type of `long`, and features as many `long` parameter types as there are free dimensions in the provided layout path. In other words, this method can be used as a replacements in cases where {@link #bitOffset(PathElement...)} does not work.
```
Or something similar.
```

File: test/jdk/java/foreign/TestLayoutPaths.java [hunk 2]
```diff
@@ -343,22 +396,76 @@ public void testSequencePaths() {
...
+            long actualByteOffset = (long) byteOffsetHandle.invokeExact(indexes);
+            assertEquals(actualByteOffset, expectedBitOffset / 8);
+        }
+    }
+
+    @DataProvider
+    public static Object[][] offsetHandleCases() {
+        return new Object[][]{
+            {
+                MemoryLayout.ofSequence(10, JAVA_INT),
```
mcimadamore, line 436: The test is good, but there's no case which features more than one long index. I think it is worth adding at least a few, since that's where all the combination logic (and ordering) can go wrong?

mcimadamore (commented):
Looks good - added some javadoc/test suggestions

test/jdk/java/foreign/TestLayoutPaths.java
JornVernee: Yes, makes sense. Thanks for the suggestion.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/LayoutPath.java
```diff
@@ -185,6 +192,27 @@ public VarHandle dereferenceHandle(Class<?> carrier) {
...
+        return base + (stride * index);
+    }
+
+    public MethodHandle offsetHandle() {
+        MethodHandle mh = MethodHandles.identity(long.class);
+        LayoutPath cur = this;
+        while (cur.enclosing != null) {
+            if (cur.enclosing.layout instanceof SequenceLayout
+                && cur.elementIndex == UNSPECIFIED_ELEM_INDEX) { // sequenceElement()
+                MethodHandle collector = MethodHandles.insertArguments(MH_ADD_SCALED_OFFSET, 2, cur.layout.bitSize());
```
mcimadamore, line 205: You already have the `strides` array for doing this? E.g. I don't think there's a need to walk the layout paths upwards and to get their size - all you need is in the stride array.

ChrisHegarty:
This looks good.  It is true to say that `bitOffset`/`byteOffset` do not accept any path elements with free dimensions? If so, then it would be good to call this out explicitly - I cannot find it in the API.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryLayout.java
```diff
@@ -350,16 +348,36 @@ default long byteAlignment() {
...
+     * ({@link PathElement#sequenceElement()}) are accepted, and translated into {@code long} parameters
+     * of the returned method handle, to be specified later.
```
ChrisHegarty, line 361: "specified later" -> "specified to the invocation of the method handle"

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/LayoutPath.java
JornVernee: We need to walk to inspect the element indices. Could replace the `cur.layout.bitSize` with a lookup into the strides array maybe, but I think that would require keeping track of a separate index as well. (I'll take a look).

mcimadamore: My recollection of this code is that when you get to generating the handle, you already have the unspecified paths in the stride array. Look at the places for when a new stride is set - I think it should match the places where you need the info as well (that is `PathElement::sequenceElement()` and `PathElement::sequenceElement(long, long)`).

JornVernee: Ah, right. The strides array only contains elements for unspecified indexes, not all sequence layouts. I'm confusing myself

JornVernee:
I've updated this PR to:
- Add tests for multiple sequences: https://github.com/openjdk/panama-foreign/pull/396/commits/798cf6ffeb9bbf6908cae1e984d8b20dbad1aaba
- Simplify the implementation to use the existing `strides` array: https://github.com/openjdk/panama-foreign/pull/396/commits/8ccdb94bf3ed476c88a6182095ebb13a1fa72093
- Touch up the javadoc: https://github.com/openjdk/panama-foreign/pull/396/commits/e8ceb119a77fab16d4072dcd371d6d74f1d2d305
- Merge with upstream, which replaces the explicit `null` tests with TestNulls.

mcimadamore (approved):
Looks very good!

ChrisHegarty (approved):

PR #397, opened 2020-11-19
8256466: MemoryLayout factories do not guard against null layouts

mcimadamore:
After starting some of the work to improve NPE handling on foreign-abi, I realized that what I was doing was indeed mechanical, and very amenable to a single, centralized test which sends null into all reference-accepting API methods.

The new test, `TestNull` takes care of this task; it is much easier to maintain, as, moving forward, we only need to add new classes to the tests, or add methods to the exclude list.

For each method featuring at least one reference type, the test creates some default values for the non-reference parameters, and then a bunch of replacements for the reference parameter. The replacement could simply be `null`, but in the case of arrays and collections more complex replacements are provided (e.g. arrays and collections with null elements).

Both replacements and default values can be easily added by tweaking some of the test static initializers. An exclusion list is also provided for those APIs which are deliberately null-friendly.

This patch also reverts the test changes from yesterday's PR, and also some of the null-related tests which has been added in the past.

---------
### Progress

### Testing

| Build | ✔️ (5/5 passed) | ✔️ (2/2 passed) | ✔️ (2/2 passed) | ✔️ (2/2 passed) |
| Test (tier1) | ✔️ (9/9 passed) | ✔️ (9/9 passed) | ✔️ (9/9 passed) | ✔️ (9/9 passed) |

### Issue
 * [JDK-8256466](https://bugs.openjdk.java.net/browse/JDK-8256466): MemoryLayout factories do not guard against null layouts ⚠️ Issue is not open.

### Reviewers
 * [Chris Hegarty](https://openjdk.java.net/census#chegar) (@ChrisHegarty - no project role)

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/397/head:pull/397`
`$ git checkout pull/397`


test/jdk/java/foreign/TestNulls.java
```diff
@@ -0,0 +1,223 @@
...
+
+    static final Set<String> EXCLUDE_LIST = Set.of(
+            "jdk.incubator.foreign.MemoryLayout/withAttribute(java.lang.String,java.lang.constant.Constable)/1/0",
+            "jdk.incubator.foreign.MemoryAddress/asSegmentRestricted(long,java.lang.Runnable,java.lang.Object)/1/0",
+            "jdk.incubator.foreign.MemoryAddress/asSegmentRestricted(long,java.lang.Runnable,java.lang.Object)/2/0",
+            "jdk.incubator.foreign.SequenceLayout/withAttribute(java.lang.String,java.lang.constant.Constable)/1/0",
+            "jdk.incubator.foreign.ValueLayout/withAttribute(java.lang.String,java.lang.constant.Constable)/1/0",
+            "jdk.incubator.foreign.GroupLayout/withAttribute(java.lang.String,java.lang.constant.Constable)/1/0",
+            "jdk.incubator.foreign.MemoryHandles/insertCoordinates(java.lang.invoke.VarHandle,int,java.lang.Object[])/2/1"
+    );
```
ChrisHegarty, line 93: This is the set of API points, and parameter positions, that accept null. Cool to have this in one place.

mcimadamore: Yeah - it's a bit "stringy" and maybe can be improved, but compared to what we had before, I think it's a massive improvement :-)

If you are curious about the encoding it's:

<className> `/` <method signature> `/` <index of the parameter to be replaced> `/` <index of the replacement to be used>

JornVernee: Could you add this in a comment as well?

JornVernee (approved):
LGTM! Very cool.

ChrisHegarty (approved):

PR #413, opened 2020-12-11
8257692: Using both heap and native segments can degrade performance

mcimadamore:
This patch fixes a problem with type profile pollution when segments of different kinds are used on the same memory access var handle, or on the same `MemoryAccess` static method.

In principle, argument profiling should kick in for VarHandles and MethodHandles, and that should be enough at least to avoid pollution when using var handles directly. In reality, this is not the case; as Vlad discovered after relatively intense debugging session (thanks!), the VarHandle implementation code has to cast the incoming segment to the `MemorySegmentProxy` internal interface. This creates problems for C2, as concrete segment implementations have _two_ interface types: `MemorySegment` and the internal `MemorySegmentProxy` class. Side casts from one to the other are not supported well, and can cause loss of type profiling infomation.

To solve this problem we realized, in hindisght, that `MemorySegmentProxy` didn't really needed to be an interface and that it could easily be converted to an abstract class. Alone this solves 50% of the issues, since that makes direct var handle access robust to pollution issues. The remaining problems (using accessors in `MemoryAccess` class) can be addressed the usual way, by adding argument type profiling for the methods in that class (similarly to what we've done for `ScopedMemoryAccess`).

Here are some numbers before the patch:

```
Benchmark                                            Mode  Cnt   Score   Error  Units
LoopOverPollutedSegments.heap_segment_floats_VH      avgt   30  11.535 ? 0.039  ms/op
LoopOverPollutedSegments.heap_segment_floats_static  avgt   30  10.860 ? 0.162  ms/op
LoopOverPollutedSegments.heap_segment_ints_VH        avgt   30  11.479 ? 0.202  ms/op
LoopOverPollutedSegments.heap_segment_ints_static    avgt   30  10.562 ? 0.027  ms/op
LoopOverPollutedSegments.heap_unsafe                 avgt   30   0.240 ? 0.003  ms/op
LoopOverPollutedSegments.native_segment_VH           avgt   30  11.603 ? 0.154  ms/op
LoopOverPollutedSegments.native_segment_static       avgt   30  10.613 ? 0.128  ms/op
LoopOverPollutedSegments.native_unsafe               avgt   30   0.243 ? 0.003  ms/op
```

As you can see there is quite a big difference between unsafe access and all the other modes. Here are the results after the patch:

```
Benchmark                                            Mode  Cnt  Score   Error  Units
LoopOverPollutedSegments.heap_segment_floats_VH      avgt   30  0.244 ? 0.002  ms/op
LoopOverPollutedSegments.heap_segment_floats_static  avgt   30  0.301 ? 0.001  ms/op
LoopOverPollutedSegments.heap_segment_ints_VH        avgt   30  0.245 ? 0.003  ms/op
LoopOverPollutedSegments.heap_segment_ints_static    avgt   30  0.302 ? 0.004  ms/op
LoopOverPollutedSegments.heap_unsafe                 avgt   30  0.242 ? 0.003  ms/op
LoopOverPollutedSegments.native_segment_VH           avgt   30  0.246 ? 0.004  ms/op
LoopOverPollutedSegments.native_segment_static       avgt   30  0.295 ? 0.006  ms/op
LoopOverPollutedSegments.native_unsafe               avgt   30  0.245 ? 0.003  ms/op
```

That is, the situation is back to normal. Thanks to @JornVernee and @iwanowww for the help!

---------
### Progress

### Issue
 * [JDK-8257692](https://bugs.openjdk.java.net/browse/JDK-8257692): Using both heap and native segments can degrade performance

### Reviewers
 * [Jorn Vernee](https://openjdk.java.net/census#jvernee) (@JornVernee - Committer) ⚠️ Review applies to 6c77935b0cbe18eee3a396927caea7d8f1eb4cc9
 * [Paul Sandoz](https://openjdk.java.net/census#psandoz) (@PaulSandoz - Committer) ⚠️ Review applies to 6c77935b0cbe18eee3a396927caea7d8f1eb4cc9

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/413/head:pull/413`
`$ git checkout pull/413`


src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/Utils.java
```diff
@@ -52,7 +52,7 @@
     private static final String foreignRestrictedAccess = Optional.ofNullable(VM.getSavedProperty("foreign.restricted"))
             .orElse("deny");
 
-    private static final MethodHandle SEGMENT_FILTER;
+    public static final MethodHandle SEGMENT_FILTER;
```
JornVernee, line 55: Why this change?

JornVernee (approved):

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/Utils.java
mcimadamore: this is a leftover from other experiments - will remove

JornVernee (approved):

PaulSandoz (approved):
Glad you all managed to resolve this.

It would be nice to mark the classes that have special profiling characteristics, otherwise it's easy to forgot or be unaware of the implications. Perhaps comments on each are sufficient. An annotation would be nice but is more work.

JornVernee:
@PaulSandoz FWIW, I also looked into using an annotation to drive the profiling as well, instead of the special cases based on the name, but the problem is that we can't see the annotations of an unresolved method, so would have to resolve everything in the interpreter just to check for annotations. The current mechanism is solely based on the symbolic reference found in the caller's bytecode.

iwanowww (approved):
Looks good.

PR #417, opened 2020-12-14
8258238: Some jextract samples crash with: "updating map that does not need updating"

JornVernee:
This fixes the failing assert by checking if the register map needs to be updated.

This gets the jextract samples working again, but some followup is still need to add a test case that can catch this problem.

---------
### Progress

### Issue
 * [JDK-8258238](https://bugs.openjdk.java.net/browse/JDK-8258238): Some jextract samples crash with: "updating map that does not need updating"

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/417/head:pull/417`
`$ git checkout pull/417`


mcimadamore (approved):
Looks good

PR #423, opened 2020-12-16
Add stack walk test case that doesn't verify oops

JornVernee:
This is a followup for the earlier quickfix of: https://github.com/openjdk/panama-foreign/pull/417

This adds another test case to TestStackWalk that catches the failing code path.

This re-uses the previously added `WhiteBox::verifyFrames`, and adds a flag to turn off the verification, which triggers the problematic code path. (I've also renamed this to just `walkFrames`, since now we don't always do verification anymore.

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/423/head:pull/423`
`$ git checkout pull/423`

### DIFF
--- a/src/hotspot/share/prims/whitebox.cpp
+++ b/src/hotspot/share/prims/whitebox.cpp
@@ -2284,23 +2284,21 @@ WB_ENTRY(void, WB_CheckThreadObjOfTerminatingThread(JNIEnv* env, jobject wb, job
   }
 WB_END
 
-WB_ENTRY(void, WB_WalkFrames(JNIEnv* env, jobject wb, jboolean log, jboolean update_reg_map_and_verify_oops))
+WB_ENTRY(void, WB_VerifyFrames(JNIEnv* env, jobject wb, jboolean log, jboolean update_map))
   intx tty_token = -1;
   if (log) {
     tty_token = ttyLocker::hold_tty();
-    tty->print_cr("[WhiteBox::WalkFrames] Walking Frames");
+    tty->print_cr("[WhiteBox::VerifyFrames] Walking Frames");
   }
-  for (StackFrameStream fst(JavaThread::current(), update_reg_map_and_verify_oops, true); !fst.is_done(); fst.next()) {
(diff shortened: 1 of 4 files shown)


src/hotspot/share/prims/whitebox.cpp
```diff
@@ -2284,21 +2284,23 @@ WB_ENTRY(void, WB_CheckThreadObjOfTerminatingThread(JNIEnv* env, jobject wb, job
...
-WB_ENTRY(void, WB_VerifyFrames(JNIEnv* env, jobject wb, jboolean log))
+WB_ENTRY(void, WB_WalkFrames(JNIEnv* env, jobject wb, jboolean log, jboolean verify_oops))
   intx tty_token = -1;
   if (log) {
     tty_token = ttyLocker::hold_tty();
-    tty->print_cr("[WhiteBox::VerifyFrames] Walking Frames");
+    tty->print_cr("[WhiteBox::WalkFrames] Walking Frames");
   }
-  for (StackFrameStream fst(JavaThread::current(), true, true); !fst.is_done(); fst.next()) {
+  for (StackFrameStream fst(JavaThread::current(), verify_oops, true); !fst.is_done(); fst.next()) {
```
iwanowww, line 2293: Can you elaborate, please, why `verify_oops` is equivalent to `RegisterMap::_update_map`?

JornVernee: If the register map is not updated, we can't do the call to `frame::verify` below that, which currently only verifies oops in the frame. I thought that the name `verify_oops` would best describe the functionality of the parameter in general, but maybe I've got that backwards... I use it specifically to turn off `RegisterMap::_update_map` after all.

JornVernee: Renaming this to `update_reg_map_and_verify_oops` a bit of a mouth full maybe, but want to make clear that oop verification only happens when the flag is set.

JornVernee:
I've integrated all open comments now (including offline discussion), please have another look. Thanks.

iwanowww (approved):
Looks good.

PR #424, opened 2020-12-17
8258605: regression: jextract can not handle function prototypes as function arguments

sundararajana:
Old fix is not applicable in the new code. But test brought and modified.

---------
### Progress

### Issue
 * [JDK-8258605](https://bugs.openjdk.java.net/browse/JDK-8258605): regression: jextract can not handle function prototypes as function arguments

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/424/head:pull/424`
`$ git checkout pull/424`

### DIFF
--- a/src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/OutputFactory.java
+++ b/src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/OutputFactory.java
@@ -361,6 +361,14 @@ public class OutputFactory implements Declaration.Visitor<Void, Declaration> {
                 default:
                     return getAsFunctionPointer(((Type.Delegated) type).type());
             }
+        } else if (type instanceof Type.Function) {
+            /*
+             * // pointer function declared as function like this
+             *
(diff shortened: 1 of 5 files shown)


src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/OutputFactory.java
```diff
@@ -361,6 +361,14 @@ public Void visitFunction(Declaration.Function funcTree, Declaration parent) {
                 default:
                     return getAsFunctionPointer(((Type.Delegated) type).type());
             }
+        } else if (type instanceof Type.Function) {
```
mcimadamore, line 364: I think the check before this can be simplified. Essentially, in this routine we should have three cases:

* we have found a function type -> return it
* we have found a delegate type -> recurse on the delegatee
* none of the above, return null

mcimadamore (commented):
Added some comments to simplify the logic in OutputFactory

mcimadamore (approved):
Looks good

PR #432, opened 2021-01-08
8259473: jextract generated code should throw exception for unfound native symbols from calls, variable access, set immediately

sundararajana:
NPE with proper error message is thrown immediately (on call, variable access/set)

---------
### Progress

### Issue
 * [JDK-8259473](https://bugs.openjdk.java.net/browse/JDK-8259473): jextract generated code should throw exception for unfound native symbols from calls, variable access, set immediately

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/432/head:pull/432`
`$ git checkout pull/432`

### DIFF
--- a/src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/HeaderFileBuilder.java
+++ b/src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/HeaderFileBuilder.java
@@ -99,23 +99,25 @@ class HeaderFileBuilder extends JavaSourceBuilder {
         builder.append(") {\n");
         builder.incrAlign();
         builder.indent();
+        builder.append("var mh$ = RuntimeHelper.requireNonNull(");
+        builder.append(methodHandleGetCallString(javaName, nativeName, mtype, desc, varargs));
+        builder.append(", \"unresolved symbol: ");
+        builder.append(nativeName);
+        builder.append("\");\n");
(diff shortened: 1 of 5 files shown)


src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/HeaderFileBuilder.java
```diff
@@ -105,13 +105,17 @@ void addStaticFunctionWrapper(String javaName, String nativeName, MethodType mty
         if (!mtype.returnType().equals(void.class)) {
             builder.append("return (" + mtype.returnType().getName() + ")");
         }
-        builder.append(methodHandleGetCallString(javaName, nativeName, mtype, desc, varargs) + ".invokeExact(" + String.join(", ", pExprs) + ");\n");
+        builder.append("Objects.requireNonNull(");
+        builder.append(methodHandleGetCallString(javaName, nativeName, mtype, desc, varargs));
+        builder.append(", \"no such native function: ");
```
mcimadamore, line 110: Nice! I wonder if here we should use "no such" vs. some other form, e.g. "unresolved". I also wonder if NPE is the right exception here - maybe a NoSuchMethodException, or LinkageError might be better?

mcimadamore (approved):
Looks good

JornVernee (approved):

PR #433, opened 2021-01-11
8258516: Allow passing a target address as an argument to a native MethodHandle 

JornVernee:
Hi,

This patch adds the ability to CLinker to link a method handle that takes the address to the target function separately as a leading argument, instead of having to supply it during linking.

This will avoid the need to re-link method handles for 'virtual' call sites, where the type of the call is always the same, but the address of the target function can differ.

I've updated the C2 code to intrinsify the current case when the address argument is a constant, but there is still more work to be done to add intrinsification of virtual calls. Though, not having to re-link a method handle on every call should already speed things up significantly.

There are other possible enhancements imaginable as well, such as caching linked method handles inside the linker, and re-using them with different addresses (but that is for a later patch).

I added a benchmark similar to CallOverhead that measures the virtual variants. Some refactoring has been done there to put shared code in a helper class, and the old benchmark has been renamed to CallOverheadConstant so it's possible to select either class to run independently of the other with JMH. I've also added several more variants that test different numbers of arguments.

I added a TestVirtualCalls test as a smoke test for simple virtual calls, and updated (as well as simplified a little) TestIntrinsics to make sure nothing crashes in the C2 code when encountering a virtual call.

Thanks,
Jorn

---------
### Progress

### Issue
 * [JDK-8258516](https://bugs.openjdk.java.net/browse/JDK-8258516): Allow passing a target address as an argument to a native MethodHandle

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/433/head:pull/433`
`$ git checkout pull/433`

### DIFF
--- a/src/hotspot/share/ci/ciNativeEntryPoint.cpp
+++ b/src/hotspot/share/ci/ciNativeEntryPoint.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2020, 2021, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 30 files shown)


mcimadamore (approved):
Very solid piece of work. The adapter changes seem good, and I see some nice consolidation of some of the helper routines - but it made my head hurt to follow through some of the adaptations :-)

(I did not review the HS part)

iwanowww (approved):
Looks good.

PR #434, opened 2021-01-14
8252759: Jextract flattens nested anonymous structs and unions when computing layouts

JornVernee:
Hi,

This patch changes the way jextract handles nested structs and unions.

Currently, when encountering an anonymous struct or union, jextract folds it's fields into the parent layout. But, this can lead to incorrect layouts, especially for unions. e.g. see the following example:

    struct Foo {
        int x;
        union {
            int y;
            int z;
        };
    }; 
    
becomes:

    static final MemoryLayout Foo$struct$LAYOUT_ = MemoryLayout.ofStruct(
        C_INT.withName("x"),
        C_INT.withName("y"),
        C_INT.withName("z")
    ).withName("Foo"); 
    
In this case, the offset of the `z` field will not be correct.

This patch changes the jextract behaviour to preserve the nested layouts of anonymous unions and structs, so the example becomes:

    static final MemoryLayout Foo$struct$LAYOUT_ = MemoryLayout.ofStruct(
        C_INT.withName("x"),
        MemoryLayout.ofUnion(
            C_INT.withName("y"),
            C_INT.withName("z")
        ).withName("$anon$0")
    ).withName("Foo"); 
    
Here, the anonymous layout gets a generated name which is needed so that the `y` and `z` fields can still be reached from the root layout with layout paths to create var handles for instance.

Just the layout is changed, the Foo java class that we generate still has getters and setters for `y` and `z`.

In the patch this behaviour is implemented by temporarily adding a 'prefix' (`"$anon$0"` in this case) to the `StructBuilder`, which is used in the generated layout paths to reach the nested fields.

Thanks,
Jorn

---------
### Progress

### Issue
 * [JDK-8252759](https://bugs.openjdk.java.net/browse/JDK-8252759): Jextract flattens nested anonymous structs and unions when computing layouts

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/434/head:pull/434`
`$ git checkout pull/434`


src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/RecordLayoutComputer.java
```diff
@@ -101,12 +105,16 @@ void addFieldLayout(MemoryLayout MemoryLayout) {
     }
 
     void addFieldLayout(long offset, Type parent, Cursor c) {
-        if (c.isAnonymousStruct()) {
-            GroupLayout layout = (GroupLayout) compute(offset, parent, c.type());
-            fieldLayouts.addAll(layout.memberLayouts());
-        } else {
-            addFieldLayout(fieldLayout(c));
-        }
+        MemoryLayout memoryLayout = c.isAnonymousStruct()
```
mcimadamore, line 108: So, am I right that this avoid flattening of the anon structs fields into enclosing layout - e.g. layouts coming out of the parser now are non-flattened, but we tag them with an attribute so that we can emit flattened accessors later?

mcimadamore (commented):
I like the approach, because not only it fixes code generation, but it also makes the AST "right" by avoiding early flattening. I think the test is a bit light - we might want to add more cases of deeply nested anon structs, just in case - as this area is a bit fragile.

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/RecordLayoutComputer.java
JornVernee: Yeah. The layout is only tagged if it is nested inside a parent record layout, since we can only assign a synthetic unique name to it in the context of the parent layout. But, the layout attached to the Declaration object for the anonymous struct will not have this name, since it is created separately outside of the context of the parent struct.

That means that when generating code, and when we counter an anonymous nested struct/union decl in visitScoped, we have to look up it's layout in the layout of the parent in order to find the context dependent name we attached here. The attribute is used to find anonymous layouts. Technically we could also check if the name starts with `$anon$` instead of looking for the attribute, but doing it this way seemed more robust.

JornVernee:
Note that we already have a test that more broadly tests nested structs; https://github.com/openjdk/panama-foreign/blob/foreign-jextract/test/jdk/tools/jextract/nested.h

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/RecordLayoutComputer.java
JornVernee: The name on the anonymous struct itself is needed to create a layout path to the nested field.

mcimadamore:
> Note that we already have a test that more broadly tests nested structs; https://github.com/openjdk/panama-foreign/blob/foreign-jextract/test/jdk/tools/jextract/nested.h
> 
> It just wasn't catching this particular case, so I added something separate. I can enhance that test to also check offsets of fields.

Yes, but that test is super simple:
https://github.com/openjdk/panama-foreign/blob/foreign-jextract/test/jdk/tools/jextract/Test8244512.java

(unless I'm missing something).

I'd suggest to unify both test and verify the emitted layouts (and field accessors too), not just the classes that come out.

JornVernee:
I've unified the two tests, and expanded the existing test of nested structs to also verify field offsets and accessors.

mcimadamore (approved):
Looks great! Thanks for making the test more portable too (e.g. `long long`).

PR #437, opened 2021-01-14
Performance improvement to unchecked segment ofNativeRestricted

rsmogura:
This changes removes (by making no-ops) range and temporal checks for `ofNativeRestricted` segment. As this segment is global, above checks are not needed.

Generated native code is smaller, and execution outperforms Java native arrays (depending on CPU)
Changed
```
Benchmark                           Mode  Cnt          Score        Error  Units
AccessBenchmark.foreignAddress     thrpt    5  128946129.691 ± 317433.113  ops/s
AccessBenchmark.foreignAddressRaw  thrpt    5  136883439.221 ± 749390.255  ops/s
AccessBenchmark.target             thrpt    5  125325586.957 ±  32129.931  ops/s
```
Base
```
Benchmark                           Mode  Cnt          Score        Error  Units
AccessBenchmark.foreignAddress     thrpt    5  125257424.876 ± 230508.169  ops/s
AccessBenchmark.foreignAddressRaw  thrpt    5  128818591.434 ± 241806.765  ops/s
AccessBenchmark.target             thrpt    5  125083379.819 ± 184070.467  ops/s
```
---
This PR is replacement for https://github.com/openjdk/panama-foreign/pull/431 (OCA)
and was partially discussed (before changes) in https://mail.openjdk.java.net/pipermail/panama-dev/2021-January/011747.htm

---
Benchmark
```
@State(Scope.Thread)
public class AccessBenchmark {
    static final MemorySegment ms = MemorySegment.ofNativeRestricted();
    static final VarHandle intHandle = MemoryHandles.varHandle(int.class, ByteOrder.nativeOrder());

    int[] intData = new int[12];
    volatile int intDataOffset = 0;

    volatile MemoryAddress address;
    volatile long addressRaw;

    @Setup
    public void setup() {
        var ms = MemorySegment.allocateNative(256);
        address = ms.address();
        addressRaw = address.toRawLongValue();
    }

    @Benchmark
    public void target(Blackhole bh) {
        int[] local = intData;
        int localOffset = intDataOffset;
        bh.consume(local[localOffset]);
        bh.consume(local[localOffset + 1]);
    }

    @Benchmark
    public void foreignAddress(Blackhole bh) {
        var a = address;
        bh.consume((int) intHandle.get(ms, a.addOffset(0).toRawLongValue()));
        bh.consume((int) intHandle.get(ms, a.addOffset(4).toRawLongValue()));
    }

    @Benchmark
    public void foreignAddressRaw(Blackhole bh) {
        var a = addressRaw;
        bh.consume((int) intHandle.get(ms, a));
        bh.consume((int) intHandle.get(ms, a + 4));
    }
}
```

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/437/head:pull/437`
`$ git checkout pull/437`

### DIFF
--- a/test/micro/org/openjdk/bench/jdk/incubator/foreign/LoopOverNonConstant.java
+++ b/test/micro/org/openjdk/bench/jdk/incubator/foreign/LoopOverNonConstant.java
@@ -51,7 +51,7 @@ import static jdk.incubator.foreign.MemoryLayouts.JAVA_INT;
 @Measurement(iterations = 10, time = 500, timeUnit = TimeUnit.MILLISECONDS)
 @State(org.openjdk.jmh.annotations.Scope.Thread)
 @OutputTimeUnit(TimeUnit.MILLISECONDS)
-@Fork(value = 3, jvmArgsAppend = { "--add-modules=jdk.incubator.foreign" })
+@Fork(value = 3, jvmArgsAppend = { "--add-modules=jdk.incubator.foreign", "-Dforeign.restricted=permit" })
 public class LoopOverNonConstant {
 
     static final Unsafe unsafe = Utils.unsafe;
@@ -60,8 +60,11 @@ public class LoopOverNonConstant {
     static final int CARRIER_SIZE = (int)JAVA_INT.byteSize();
     static final int ALLOC_SIZE = ELEM_SIZE * CARRIER_SIZE;
 
+    static final MemorySegment globalRestrictedSegment = MemorySegment.ofNativeRestricted();
+
     static final VarHandle VH_int = MemoryLayout.ofSequence(JAVA_INT).varHandle(int.class, sequenceElement());
     MemorySegment segment;
+    long segment_addr_idx; // The segment address divided by carrier size, as it's sequence layout
     long unsafe_addr;
 
     ByteBuffer byteBuffer;
@@ -72,10 +75,15 @@ public class LoopOverNonConstant {
         for (int i = 0; i < ELEM_SIZE; i++) {
             unsafe.putInt(unsafe_addr + (i * CARRIER_SIZE) , i);
         }
-        segment = MemorySegment.allocateNative(ALLOC_SIZE);
+
+        // Allocate bigger size for aligning
+        segment = MemorySegment.allocateNative(ALLOC_SIZE + CARRIER_SIZE);
         for (int i = 0; i < ELEM_SIZE; i++) {
             VH_int.set(segment, (long) i, i);
         }
+        // Get index of segment's address aligned up
+        segment_addr_idx = (segment.address().toRawLongValue() + CARRIER_SIZE - 1) / CARRIER_SIZE;
+
         byteBuffer = ByteBuffer.allocateDirect(ALLOC_SIZE).order(ByteOrder.nativeOrder());
         for (int i = 0; i < ELEM_SIZE; i++) {
             byteBuffer.putInt(i * CARRIER_SIZE , i);
@@ -103,6 +111,14 @@ public class LoopOverNonConstant {
 
     @Benchmark
     @OutputTimeUnit(TimeUnit.NANOSECONDS)
+    public int global_segment_get() {
+        // The segment used inside this benchmarks should have range & temporal checks
+        // removed. Results should be better than for other kinds of segments
+        return (int) VH_int.get(globalRestrictedSegment, segment_addr_idx);
+    }
+
+    @Benchmark
+    @OutputTimeUnit(TimeUnit.NANOSECONDS)
     public int BB_get() {
         return byteBuffer.getInt(0);
     }
@@ -112,6 +128,16 @@ public class LoopOverNonConstant {
         int res = 0;
         for (int i = 0; i < ELEM_SIZE; i ++) {
             res += unsafe.getInt(unsafe_addr + (i * CARRIER_SIZE));
+        }
+        return res;
+    }
+
+    @Benchmark
+    public int global_segment_loop() {
+        // In fact, we operate inside `segment`
+        int res = 0;
+        for (int i = 0; i < ELEM_SIZE; i ++) {
+            res += (int) VH_int.get(globalRestrictedSegment, segment_addr_idx + i);
         }
         return res;
     }
--- a/test/micro/org/openjdk/bench/jdk/incubator/foreign/LoopOverNonConstant.java
+++ b/test/micro/org/openjdk/bench/jdk/incubator/foreign/LoopOverNonConstant.java
@@ -23,7 +23,7 @@
 package org.openjdk.bench.jdk.incubator.foreign;
 
 import jdk.incubator.foreign.MemoryAccess;
-import jdk.incubator.foreign.MemoryAddress;
+import jdk.incubator.foreign.MemoryHandles;
 import jdk.incubator.foreign.MemoryLayout;
 import jdk.incubator.foreign.MemorySegment;
 import org.openjdk.jmh.annotations.Benchmark;
@@ -63,8 +63,11 @@ public class LoopOverNonConstant {
     static final MemorySegment globalRestrictedSegment = MemorySegment.ofNativeRestricted();
 
     static final VarHandle VH_int = MemoryLayout.ofSequence(JAVA_INT).varHandle(int.class, sequenceElement());
+    /** Used to test access using raw pointers. */
+    static final VarHandle VH_int_ptr = MemoryHandles.varHandle(int.class, ByteOrder.nativeOrder());
+    static final VarHandle VH_int_ptr_noalign = MemoryHandles.varHandle(int.class, 1, ByteOrder.nativeOrder());
+
     MemorySegment segment;
-    long segment_addr_idx; // The segment address divided by carrier size, as it's sequence layout
     long unsafe_addr;
 
     ByteBuffer byteBuffer;
@@ -77,12 +80,10 @@ public class LoopOverNonConstant {
         }
 
         // Allocate bigger size for aligning
-        segment = MemorySegment.allocateNative(ALLOC_SIZE + CARRIER_SIZE);
+        segment = MemorySegment.allocateNative(ALLOC_SIZE);
         for (int i = 0; i < ELEM_SIZE; i++) {
             VH_int.set(segment, (long) i, i);
         }
-        // Get index of segment's address aligned up
-        segment_addr_idx = (segment.address().toRawLongValue() + CARRIER_SIZE - 1) / CARRIER_SIZE;
 
         byteBuffer = ByteBuffer.allocateDirect(ALLOC_SIZE).order(ByteOrder.nativeOrder());
         for (int i = 0; i < ELEM_SIZE; i++) {
@@ -114,7 +115,7 @@ public class LoopOverNonConstant {
     public int global_segment_get() {
         // The segment used inside this benchmarks should have range & temporal checks
         // removed. Results should be better than for other kinds of segments
-        return (int) VH_int.get(globalRestrictedSegment, segment_addr_idx);
+        return (int) VH_int_ptr.get(globalRestrictedSegment, unsafe_addr);
     }
 
     @Benchmark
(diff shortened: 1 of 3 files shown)


mcimadamore (commented):
Looks good! Thanks for investigating the issues with lack of static on Unsafe.

Could we please add the benchmark under test/micro/org/openjdk/bench/jdk/incubator/foreign ? We try to add benchmark for each case we find as we go along, to make sure we don't regress. This seems like another interesting case.

rsmogura:
Hi. I've added the benchmarks:
* benchmarks used by me;
* added a new benchmarks to `LoopOverNonConstant` to compare performance of this approach with existing methods for getting values.

mcimadamore:
> Hi. I've added the benchmarks:
> 
>     * benchmarks used by me;
> 
>     * added a new benchmarks to `LoopOverNonConstant` to compare performance of this approach with existing methods for getting values.

It's strange to see the performances of global segment to be worse than that of a regular segment?

rsmogura:
> > Hi. I've added the benchmarks:
> 
> > 
> 
> >     * benchmarks used by me;
> 
> > 
> 
> >     * added a new benchmarks to `LoopOverNonConstant` to compare performance of this approach with existing methods for getting values.
> 
> 
> 
> It's strange to see the performances of global segment to be worse than that of a regular segment?

Indeed. I can't understand this (I'll make additional research). I suspect that's because the access is indexed (so global memory is treated as big array), or it's more about test itself - there's additional additions and we use longs instead of ints.

mcimadamore:
> > Hi. I've added the benchmarks:
> > ```
> > * benchmarks used by me;
> > 
> > * added a new benchmarks to `LoopOverNonConstant` to compare performance of this approach with existing methods for getting values.
> > ```
> 
> It's strange to see the performances of global segment to be worse than that of a regular segment?

Ah! I think I know what's happening here. In principle, the original implementation should have been just as good - but there's one issue: since the everything segment has length of Long.MAX_VALUE it is, by definition, not a SMALL segment - which means the internal optimizations that are aimed at working around the hotspot limitations around int vs. long loops will not work for the everything segment. That's why it was 2x slower. Your patch removes _some_ of these factors, but not _all_ of them. And I think the updates to LoopOverNonConstants are wrong - as you are essentially trying to use a strided VarHandle in an absolute way (see code comment).

test/micro/org/openjdk/bench/jdk/incubator/foreign/LoopOverNonConstant.java
```diff
@@ -60,8 +60,11 @@
     static final int CARRIER_SIZE = (int)JAVA_INT.byteSize();
     static final int ALLOC_SIZE = ELEM_SIZE * CARRIER_SIZE;
 
+    static final MemorySegment globalRestrictedSegment = MemorySegment.ofNativeRestricted();
+
     static final VarHandle VH_int = MemoryLayout.ofSequence(JAVA_INT).varHandle(int.class, sequenceElement());
```
mcimadamore, line 65: This is a strided VarHandle - e.g. it takes a logical index (the sequence element index inside the segment) and it dereferences that element. So, you use it like this:

```
VH_int.get(segment, 0); // first int element
VH_int.get(segment, 1); // second int element (segment base address + 4)
VH_int.get(segment, 2); // second int element (segment base address + 8)
...
```

test/micro/org/openjdk/bench/jdk/incubator/foreign/LoopOverNonConstant.java
```diff
@@ -116,6 +132,16 @@ public int unsafe_loop() {
         return res;
     }
 
+    @Benchmark
+    public int global_segment_loop() {
+        // In fact, we operate inside `segment`
+        int res = 0;
+        for (int i = 0; i < ELEM_SIZE; i ++) {
+            res += (int) VH_int.get(globalRestrictedSegment, segment_addr_idx + i);
```
mcimadamore, line 140: This looks wrong. you are passing an absolute address to a "logical index" argument.  I see that you are attempting to divide the segment address by the carrier size, and that kind offset things, but still leaves you with suboptimal performances.

mcimadamore: If you want to access memory this way, you need a different VarHandle:

```
VarHandle absolute_int_VH = MemoryHandles.varHandle(int.class, ByteOrder.nativeOrder());
```
Or use one of the static accessors:

```
MemoryAccess.getIntAtOffset(globalRestrictedSegment, segment_addr_idx + i);
```

rsmogura:
> > > Hi. I've added the benchmarks:
> > > ```
> > > * benchmarks used by me;
> > > 
> > > * added a new benchmarks to `LoopOverNonConstant` to compare performance of this approach with existing methods for getting values.
> > > ```
> > 
> > 
> > It's strange to see the performances of global segment to be worse than that of a regular segment?
> 
> Ah! I think I know what's happening here. In principle, the original implementation should have been just as good - but there's one issue: since the everything segment has length of Long.MAX_VALUE it is, by definition, not a SMALL segment - which means the internal optimizations that are aimed at working around the hotspot limitations around int vs. long loops will not work for the everything segment. That's why it was 2x slower. Your patch removes _some_ of these factors, but not _all_ of them. And I think the updates to LoopOverNonConstants are wrong - as you are essentially trying to use a strided VarHandle in an absolute way (see code comment).
> 
> Note that an indexed VarHandle will, internally, need to perform additions and multiplications - so if your segment is "big" these will be long additions and multiplications - which you will pay for (at least for the time being).
> 
> Overall I'm hopeful that, by the time we fix the long vs. int optimization problem in hotspot, the code we have today should just get faster, without a real need for a global segment.

Hi,

That's interesting how changing just one type can introduce such a big performance results.

I changed the access to `globalRestrictedSegment`. I completely agree previous version was a bit wired.

I analyzed both methods, it looks like that (I'm not the expert), segment / int versions do loop unrolling
```
  0x00007fbcd80c8220:   add    0x0(%r13),%eax                                                                                                                               
  0x00007fbcd80c8224:   add    0x4(%r13),%eax                                                                                                                               
  0x00007fbcd80c8228:   add    0x8(%r13),%eax                                                                                                                               
// Removed similar lines                                                                                                                           
  0x00007fbcd80c8254:   add    0x34(%r13),%eax                                                                                                                              
  0x00007fbcd80c8258:   add    0x38(%r13),%eax                                                                                                                              
  0x00007fbcd80c825c:   add    0x3c(%r13),%eax              ;*iadd {reexecute=0 rethrow=0 return_oop=0}                                                                     
                                                            ; - org.openjdk.bench.jdk.incubator.foreign.LoopOverNonConstant::segment_loop@23 (line 165)                     
  0x00007fbcd80c8260:   add    $0x10,%ebx                   ;*iinc {reexecute=0 rethrow=0 return_oop=0}                                                                     
                                                            ; - org.openjdk.bench.jdk.incubator.foreign.LoopOverNonConstant::segment_loop@25 (line 164) 
```
I could not find such optimization in case of global segment and long numbers. 

From the other hand I have put together this PR and previous with Unsafe access, disabled alignment checks by settings alignment to 1, and global memory segment rocks. I added benchmark method to show performance of it.

mcimadamore:
>  could not find such optimization in case of global segment and long numbers.

The optimization happens on segment construction - if the segment size is < Integer.MAX_VALUE, the AbstractMemorySegment::defaultAccessMode method appends the SMALL flag to the set of segment flags. This flag is then consulted many times, typically before long addition/multiplications have to be performed.
> 
> From the other hand I have put together this PR and previous with Unsafe access, disabled alignment checks by settings alignment to 1, and global memory segment rocks. I added benchmark method to show performance of it.

Alignment check is an outstanding issue for the API - hotspot doesn't optimize those very well. See also:

https://mail.openjdk.java.net/pipermail/panama-dev/2021-January/011794.html

mcimadamore:
Would you mind pasting the benchmark results before/after the patch, now that the benchmark has been fixed? Thanks!

rsmogura:
After patch
```
Benchmark                                         Mode  Cnt  Score    Error  Units
LoopOverNonConstant.BB_get                        avgt   30  3.915 ?  0.012  ns/op
LoopOverNonConstant.BB_loop                       avgt   30  0.230 ?  0.001  ms/op
LoopOverNonConstant.global_segment_get            avgt   30  3.878 ?  0.001  ns/op
LoopOverNonConstant.global_segment_loop           avgt   30  0.400 ?  0.002  ms/op
LoopOverNonConstant.global_segment_loop_no_align  avgt   30  0.248 ?  0.001  ms/op
LoopOverNonConstant.segment_get                   avgt   30  5.497 ?  0.015  ns/op
LoopOverNonConstant.segment_loop                  avgt   30  0.229 ?  0.001  ms/op
LoopOverNonConstant.segment_loop_readonly         avgt   30  0.239 ?  0.001  ms/op
LoopOverNonConstant.segment_loop_slice            avgt   30  0.241 ?  0.001  ms/op
LoopOverNonConstant.segment_loop_static           avgt   30  0.229 ?  0.001  ms/op
LoopOverNonConstant.unsafe_get                    avgt   30  3.434 ?  0.006  ns/op
LoopOverNonConstant.unsafe_loop                   avgt   30  0.229 ?  0.001  ms/op
```
before patch
```
Benchmark                                         Mode  Cnt  Score    Error  Units
LoopOverNonConstant.BB_get                        avgt   30  3.906 ?  0.012  ns/op
LoopOverNonConstant.BB_loop                       avgt   30  0.231 ?  0.001  ms/op
LoopOverNonConstant.global_segment_get            avgt   30  4.125 ?  0.012  ns/op
LoopOverNonConstant.global_segment_loop           avgt   30  0.596 ?  0.004  ms/op
LoopOverNonConstant.global_segment_loop_no_align  avgt   30  0.489 ?  0.002  ms/op
LoopOverNonConstant.segment_get                   avgt   30  5.507 ?  0.017  ns/op
LoopOverNonConstant.segment_loop                  avgt   30  0.231 ?  0.001  ms/op
LoopOverNonConstant.segment_loop_readonly         avgt   30  0.238 ?  0.001  ms/op
LoopOverNonConstant.segment_loop_slice            avgt   30  0.242 ?  0.001  ms/op
LoopOverNonConstant.segment_loop_static           avgt   30  0.230 ?  0.001  ms/op
LoopOverNonConstant.unsafe_get                    avgt   30  3.442 ?  0.008  ns/op
LoopOverNonConstant.unsafe_loop                   avgt   30  0.230 ?  0.001  ms/op

```

rsmogura:
Hi Maurizio,

I added benchmarks, in previous comment. If you would like more, please let me know!

Have a great day,
Rado

mcimadamore (approved):
Looks good for now - we can reassess after the hotspot improvements for long in loops start to have visible effects. Thanks!

rsmogura:
Hi all,

I realized that my benchmarks contains small issue. I think @mcimadamore mentioned it, but I did not understood.

The loops are counted using long instead of int. With int I've seen the loop get unrolled and the aligned to 1 loop is as fast as other tests.

I'm not sure if at this stage I c an add modified benchmarks or should I opens new PR?

mcimadamore:
> Hi all,
> 
> I realized that my benchmarks contains small issue. I think @mcimadamore mentioned it, but I did not understood.
> 
> The loops are counted using long instead of int. With int I've seen the loop get unrolled and the aligned to 1 loop is as fast as other tests.
> 
> I'm not sure if at this stage I c an add modified benchmarks or should I opens new PR?

Hi,
this PR is still open, since I forgot to integrate it (sorry). But this means that we can decide what to do. I see the issue with the benchmark - well spotted. At this point, to be honest, I'm unsure about the value of the proposed patch. It helps a bit for long loops, yes, but in the idiomatic use (e.g. with int) performances are the same. Given that better VM support for long loops is on the way, I'm not sure we should add more workarounds for longs (as it's proving already difficult to keep an inventory of all the hacks we have in the implementation in order to eliminate this hotspot issue).

Would you be ok with just closing the PR?

Thanks!

rsmogura:
Hi Maurizio,

I hope this message finds you well.

Maybe let me add a benchmark first. In fact I never targeted this for loops, but for use cases where the access will be random, so when the chance to put checks outside loop are minimal.

mcimadamore:
> Hi Maurizio,
> 
> I hope this message finds you well.
> 
> Maybe let me add a benchmark first. In fact I never targeted this for loops, but for use cases where the access will be random, so when the chance to put checks outside loop are minimal.

Sure - let me know how that goes! Cheers.

PR #450, opened 2021-02-12
8261642: improve jextract source generation mode and remove class generation

mcimadamore:
This patch replaces the existing code generators used by jextract by a single, unified, source generation strategy.

As discussed in [1], there are some trade-offs between source generation and classfile generation; for instance, when using classfile translation, we can take advantage of certain features of the JVM bytecode, such as dynamic constants, which makes it easy to generate big files full of constants which are only initialized on a *by-need* basis.

Classfile generation is not without issue though; first, it is almost always easier to deal with sourcefiles rather than classes - both in terms of building environment and also in terms of sharing (e.g. uploading generated bindings on a GitHub project). This is for instance why we have always used source-based bindings for libclang to power jextract. We feel that users will probably be subject to similar limitations.

Secondly, IDEs can sometimes be picky about what's in the classfile - and I have seen cases of IntelliJ failing to decompile classes containing dynamic constants.

Last, but not least, classfiles cannot contain information which would be helpful for the user, such as javadoc. Adding javadoc comments, containing at the very least the original signatures of the C functions/structs would be very helpful when developing applications which use jextract generated code.

For these reasons, I have decided to explore a source-based strategy for jextract. Note that this does not alter the set of options available in the command line, or the default behavior of jextract: that is, jextract will still emit classfiles if you do not specify --sources; but instead of having _two_ different backends (as we do now), this patch replaces both backends with a single source-based backend. We can discuss, as followup changes, as to whether we should just make source generation the default and drop classfile generation, but that's for another day.

The basic idea behind the approach is to stash constants in a separate class; that is, when we need a native method handle, we can generate a class which contains all the constants relative to that native method handle. This allows us to only load the constants that are used for a given native invocation.

One obvious problem of this approach is the sheer number of classes that it generates (although only a subset of them are truly loaded and initialized). To counteract that, this patch adopts several strategies:

1. first, it reuses a container class, if that's already available, for storing the constants. For instance, if we are generating a struct class, we can just emit the constants inside the struct class, since the constants are private to that struct anyway (we cannot do that trick for functional interfaces, unfortunately, as these are interfaces, and all fields in interfaces must be public - so there's no way to *hide* the constant fields in there - we could do that if Java supported `private static` fields in interfaces, which is probably not so unreasonable to consider).

2. The new code avoids some duplication which was present in the old approach when it comes to `#define` constants; these had getters both in the header Java API and in the internal constant source files, therefore almost doubling the size of code required to support a single constant. This is now gone in the new scheme; since constants are pretty uniquitous in header files, this move saves quite a bit of footprint.

3. To further reduce the number of constant holder classes, I added a very simple grouping logic, so that the same holder class can be reused across N different API points (where N can be tweaked using a JDK property). In other words, if you turn up N to a very high value (e.g. 1000) you end up with the same generation scheme we had before, where all constants end up in the same class (or very few different classes). The default is set to 5, which, in my experiments has proven to be a good balance between fast startup (negligible difference when compared to the dynamic constant classfile approach) and artifact footprint (what comes out of jextract has basically the same size as before).

In terms of implementation, the main idea was to double down on the XYZBuilder interface and, at the same time, simplify it. We now have the following abstractions:

* JavaSourceBuilder: basic abstraction which deals with creating java sources, maintains a backing string buffer, etc. It has a main API, which is used by `OutputFactory` to add variables, functions, constants, structs, functional interfaces, typedefs. This probably represents the biggest architectural shift in the code: previously, OutputFactory was responsible for low-level operation, such as starting and ending source generation - now all these details are encapsulated in the builders and OutputFactory deals with a much simplified API (with the result that OutputFactory became a lot simpler in places, such as `visitVariable` and `visitScoped`). Note that, since this now includes the former StringSourceBuilder featuring the string buffer, a lot of indirections of the like `builder.xyz...` have been replaced with just `xyz` which makes the code easier to follow.

* HeaderFileBuilder: as before, this is the builder that is responsible for holding the contents of an header file. It can be used to add vars (globals), functions, structs, typedefs and functional interfaces. 

* NestedClassBuilder: as before, this is a builder for all things that are nested inside the main header file.

* ConstantBuilder: this is a new class which partly plays the role played by the former `ConstantHelper` - it is now a type of `NestedClassBuilder` which has logic to emit constants. Note that JavaSourceBuilder has a method, namely `emitWithConstantClass` which spins a new constant holder on the fly, and allows some code generation to depends on the constants being added to that holder. ConstantBuilder also has a main API for adding method/var handles, segment/address constants, etc. Each of these methods return a `Constant` instance, which has few methods that can be useful when generating code (e.g. to generate the access expression needed to get the constant, or to emit a getter for the constant inside another builder).

* StructBuilder, FunctionalInterfaceBuilder, TypedefBuilder; as before - the only notable difference being that now StructBuilder extends ConstantBuilder, so it can be the holder of its own constants.

I think this covers the main things that have been touched by this patch - feel free to ask questions, in case I missed something!

Cheers

[1] - https://mail.openjdk.java.net/pipermail/panama-dev/2021-January/011887.html

---------
### Progress

### Issue
 * [JDK-8261642](https://bugs.openjdk.java.net/browse/JDK-8261642): improve jextract source generation mode and remove class generation

### Reviewers
 * [Athijegannathan Sundararajan](https://openjdk.java.net/census#sundar) (@sundararajana - Committer) ⚠️ Review applies to 8d424baf6f7ddbe156b7c2eb589e1eec02c1a4b4

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/450/head:pull/450`
`$ git checkout pull/450`


mcimadamore:
I did some experiments with removing parameter info and debugging information - honestly the gain were much less than what I was hoping for (around 100K max). I ended up disabling parameter info (which has been added to support the now gone annotation) and remove `g:lines` which is the default for javac anyway (e.g. line number tables will *still be emitted*) as I did not find a significant bloat coming out of that.

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/JavaSourceBuilder.java
```diff
@@ -259,12 +203,74 @@ protected String addressGetCallString(String javaName, String nativeName, Memory
...
+    static final int CONSTANTS_PER_CLASS = Integer.getInteger("jextract.constants.per.class", 5);
+    ConstantBuilder constantBuilder;
+
+    protected void emitWithConstantClass(String javaName, Consumer<ConstantBuilder> constantConsumer) {
+        if (constant_counter > CONSTANTS_PER_CLASS || constantBuilder == null) {
+            if (constantBuilder != null) {
+                constantBuilder.classEnd();
+            }
+            constant_counter = 0;
+            constantBuilder = new ConstantBuilder(this, Kind.CLASS, javaName + "_constants");
```
JornVernee, line 270: Using the java name of some constant to uniqueify the name of the constant class seems a bit strange to me... Couldn't there be collisions? (e.g. if struct and unction are named the same?) Why not just use to id/number here like before?

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/StructBuilder.java
```diff
@@ -49,104 +54,117 @@
         this.structLayout = structLayout;
         this.structType = structType;
         prefixElementNames = new ArrayDeque<>();
+        classBegin();
```
JornVernee, line 57: Kinda strange that the constructor here call `classBegin()` but OutputFactory calls `classEnd()`. Could both calls be moved to `OutputFactory`?

JornVernee (approved):
LGTM, some minor comments inline

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/ConstantBuilder.java
```diff
@@ -0,0 +1,395 @@
...
+                l -> l.get(2);
+    }
+
+    // private generators
+
+    public Constant emitIfAbsent(String name, Constant.Kind kind, Supplier<Constant> constantFactory) {
+        String lookupName = kind.fieldName(name);
+        Constant constant = namesGenerated.get(lookupName);
+        if (constant == null) {
+            constant = constantFactory.get();
```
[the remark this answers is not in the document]
mcimadamore: good idea, will do

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/ConstantBuilder.java
```diff
@@ -0,0 +1,395 @@
...
+            append(typeToLayoutName((ValueLayout) l));
+        } else if (l instanceof SequenceLayout) {
+            append("MemoryLayout.ofSequence(");
+            if (((SequenceLayout) l).elementCount().isPresent()) {
+                append(((SequenceLayout) l).elementCount().getAsLong() + ", ");
+            }
+            emitLayoutString(((SequenceLayout) l).elementLayout());
+            append(")");
+        } else if (l instanceof GroupLayout) {
+            if (((GroupLayout) l).isStruct()) {
```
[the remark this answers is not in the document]
mcimadamore: true

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/JavaSourceBuilder.java
mcimadamore: there could be collisions, yes - not sure I get what you propose by using "id/number"

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/StructBuilder.java
mcimadamore: I can fix this

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/ConstantBuilder.java
```diff
@@ -0,0 +1,395 @@
...
+            append("FunctionDescriptor.ofVoid(");
+        }
+        if (!noArgs) {
+            append("\n");
+            incrAlign();
+            String delim = "";
+            for (MemoryLayout e : desc.argumentLayouts()) {
+                append(delim);
+                indent();
+                emitLayoutString(e);
```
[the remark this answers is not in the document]
mcimadamore: Yes - this duplicates code for layout creation. We can improve this further - e.g. by referring to the struct/functional interface layout, but I'd prefer if we moved that to a separate issue. Also I've played with things like e.g. sharing function descriptors with same shape and didn't really observed any significant gain in footprint.

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/JavaSourceBuilder.java
mcimadamore: I see what you mean - I thought you meant collisions in case an header contains `foo` and `foo_constants` - but you mean conflicts between members that have same native name.

sundararajana (approved):

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/JavaSourceBuilder.java
JornVernee: Yes. With id/number I mean what [MultiFileConstantHelper::getConstantClassName()](https://github.com/openjdk/panama-foreign/pull/450/files#diff-0c06a52b2e1c4375f51e101bbbd60d6dc8405e9ace3d6d6a8608e2b184146263L62) was doing before, i.e. using the number of constant classes emitted so far as a suffix id.

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/StructBuilder.java
mcimadamore: forgot about this - I cannot really do this, as nested structs don't return a new builder - in other words, the OutputFactory code doesn't really know if it should begin a class or not. I could do `if (newBuilder != currentBuilder) newBuilder.classBegin() ...` but I don't think that improves much?

JornVernee: Yeah, I was looking at `classEnd()` where you use `if (prefixElementNames.isEmpty())` to check if the struct builder is for an anonymous nested struct or not. So I thought you could do the same for `classBegin()` and have:

```
    @Override
    void classBegin() {
        if (prefixElementNames.isEmpty()) {
            super.classBegin();
            addLayout(layoutField(), ((Type.Declared)structType).tree().layout().get())
                    .emitGetter(this, MEMBER_MODS, Constant.SUFFIX_ONLY);
        }
    }
```

?

JornVernee: Also, it might make things clearer to create an `isAnonymousNested()` predicate for that (just for reader's sake).

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/ConstantBuilder.java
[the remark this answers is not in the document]
JornVernee: You missed the ValueLayout and SequenceLayout cases?

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/StructBuilder.java
mcimadamore: Not sure about the predicate - in the sense that the builder is not an anonymous nested - it is just being reused/recycled for an anonymous nested.

JornVernee (approved):
Looks good. Thanks for addressing.

PR #454, opened 2021-02-17
Add a benchmark for strlen using Foreign Linker API

mcimadamore:
I've been spending some time looking into this issue:

https://bugs.openjdk.java.net/browse/JDK-8261828

And, to understand better the problem, I put together an hopefully comprehensive benchmark of the strlen function; it turns out that the strlen call itself is fast, and it's the conversion from Java to native string where the benchmark spends most of its time.

While playing with the benchmark, I came up with alternative ways to do this conversion which greatly speed up the benchmark results, even surpassing (at least on my machine) what's possible with JNI. Jorn and I think that, for future references, it would be a good idea to include this benchmark in our suite.

For the curious, there are many factors which make the default `CLinker::toCString` go slower than expected: 

* allocating a fresh segment on each iteration is expensive, because it takes two native call (malloc, memset), plus a bunch of CAS to reserve memory in the Java runtime
* freeing the segment on each iteration is equally expensive - one native call (free), plus again, some CAS to unreserve memory 
* bulk copy is fast, but again requires a native call
* all in all, we need 4 native calls per iteration (malloc, memset, copy, free) each adding cost when it comes to state transitions

In other words, the advantage of JNI here is that (i) the level of safety provided by JNI is lower (e.g. the runtime doesn't need to track e.g. allocated memory, which segments do); also (ii) when we call the JNI-ified strlen function, the malloc, free, copy happen when we're in native code already - which means less state transitions are required.

Note that we can completely eliminate (i) basically by creating restricted segments using CLinker::allocateMemoryRestricted (which does a plain malloc). We can also eliminate (ii) by creating *trivial* function descriptors for the calls to malloc/free/strlen, thereby removing cost associated with state transitions there. Both routes are tested in the benchmark (note that they both requires some willingness to embrace restricted methods). I have put together a variant which shows how NativeScope can be used to speed allocation up (which works really well for small strings, and is _not_ restricted).

What are the lessons learned for plain `CLinker::toCString` ? 

* While the logic is generally fast, all state transitions and unsafe calls are killing performance in such a tight scenario; perhaps worth considering intrinsifying Unsafe::allocateMemory/copyMemory/setMemory/freeMemory.
* The way to go, performance-wise is not to rely on the default malloc-based allocation. This is where Panama has a big edge over JNI, whose allocation logic is _fixed_. Proposals such as the one described in [1] will make passing custom allocators to `CLinker::toCString` easier, so that clients can decide which allocation strategy best fits their use case.

[1] - https://inside.java/2021/01/25/memory-access-pulling-all-the-threads/

---------
### Progress

### Reviewers
 * [Jorn Vernee](https://openjdk.java.net/census#jvernee) (@JornVernee - Committer) ⚠️ Review applies to 6c4a49816e618b62590b35b14187f806474d3052

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/454/head:pull/454`
`$ git checkout pull/454`

### DIFF
--- a/test/micro/org/openjdk/bench/jdk/incubator/foreign/libStrLen.c
+++ b/test/micro/org/openjdk/bench/jdk/incubator/foreign/libStrLen.c
@@ -29,7 +29,7 @@
 
 JNIEXPORT jint JNICALL Java_org_openjdk_bench_jdk_incubator_foreign_StrLenTest_strlen(JNIEnv *const env, const jclass cls, const jstring text) {
     const char *str = (*env)->GetStringUTFChars(env, text, NULL);
-	int len = strlen(str);
-	(*env)->ReleaseStringUTFChars(env, text, str);
-	return len;
+    int len = strlen(str);
+    (*env)->ReleaseStringUTFChars(env, text, str);
+    return len;
 }
--- a/test/micro/org/openjdk/bench/jdk/incubator/foreign/libStrLen.c
+++ b/test/micro/org/openjdk/bench/jdk/incubator/foreign/libStrLen.c
@@ -1,0 +1,35 @@
+/*
+ * Copyright (c) 2021, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+#include <jni.h>
+#include <stdlib.h>
+#include <string.h>
+
+JNIEXPORT jint JNICALL Java_org_openjdk_bench_jdk_incubator_foreign_StrLenTest_strlen(JNIEnv *const env, const jclass cls, const jstring text) {
+    const char *str = (*env)->GetStringUTFChars(env, text, NULL);
+	int len = strlen(str);
+	(*env)->ReleaseStringUTFChars(env, text, str);
+	return len;
+}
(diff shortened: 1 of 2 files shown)


mcimadamore:
For the records, benchmark results on my machine look as below:

```
Benchmark                                (size)  Mode  Cnt    Score   Error  Units
StrLenTest.jni_strlen                         5  avgt   30   47.952 ? 0.899  ns/op
StrLenTest.jni_strlen                        20  avgt   30   61.918 ? 2.668  ns/op
StrLenTest.jni_strlen                       100  avgt   30  135.449 ? 1.454  ns/op
StrLenTest.panama_strlen                      5  avgt   30  115.883 ? 2.194  ns/op
StrLenTest.panama_strlen                     20  avgt   30  114.238 ? 1.896  ns/op
StrLenTest.panama_strlen                    100  avgt   30  133.208 ? 2.056  ns/op
StrLenTest.panama_strlen_scope                5  avgt   30   34.467 ? 0.596  ns/op
StrLenTest.panama_strlen_scope               20  avgt   30   50.872 ? 2.357  ns/op
StrLenTest.panama_strlen_scope              100  avgt   29   89.604 ? 3.675  ns/op
StrLenTest.panama_strlen_unsafe               5  avgt   30   52.222 ? 3.626  ns/op
StrLenTest.panama_strlen_unsafe              20  avgt   30   55.937 ? 2.125  ns/op
StrLenTest.panama_strlen_unsafe             100  avgt   30   67.084 ? 0.671  ns/op
StrLenTest.panama_strlen_unsafe_trivial       5  avgt   30   29.762 ? 0.443  ns/op
StrLenTest.panama_strlen_unsafe_trivial      20  avgt   30   36.156 ? 0.369  ns/op
StrLenTest.panama_strlen_unsafe_trivial     100  avgt   30   59.830 ? 4.222  ns/op
```

test/micro/org/openjdk/bench/jdk/incubator/foreign/libStrLen.c
```diff
@@ -0,0 +1,35 @@
...
+ * questions.
+ */
+
+#include <jni.h>
+#include <stdlib.h>
+#include <string.h>
+
+JNIEXPORT jint JNICALL Java_org_openjdk_bench_jdk_incubator_foreign_StrLenTest_strlen(JNIEnv *const env, const jclass cls, const jstring text) {
+    const char *str = (*env)->GetStringUTFChars(env, text, NULL);
+    int len = strlen(str);
```
JornVernee, line 32: I'm seeing a warning here: `conversion from 'size_t' to 'int', possible loss of data` which makes the compilation fail. An explicit cast to `int` fixes that.

JornVernee (approved):
Looks good, besides a warning in the native code (see inline comment).

Numbers on my Windows machine look like this:

```
Benchmark                                (size)  Mode  Cnt    Score     Error  Units
StrLenTest.jni_strlen                         5  avgt   30  158.912 �  11.886  ns/op
StrLenTest.jni_strlen                        20  avgt   30  206.928 �   3.980  ns/op
StrLenTest.jni_strlen                       100  avgt   30  405.025 �  10.837  ns/op
StrLenTest.panama_strlen                      5  avgt   30  211.031 �   3.851  ns/op
StrLenTest.panama_strlen                     20  avgt   30  234.639 �  27.133  ns/op
StrLenTest.panama_strlen                    100  avgt   30  235.827 �   4.062  ns/op
StrLenTest.panama_strlen_scope                5  avgt   30   65.311 �   4.032  ns/op
StrLenTest.panama_strlen_scope               20  avgt   30   96.457 �   6.595  ns/op
StrLenTest.panama_strlen_scope              100  avgt   24  157.490 �  24.683  ns/op
StrLenTest.panama_strlen_unsafe               5  avgt   30  135.766 �   1.252  ns/op
StrLenTest.panama_strlen_unsafe              20  avgt   30  154.713 �   4.510  ns/op
StrLenTest.panama_strlen_unsafe             100  avgt   30  152.058 �   2.003  ns/op
StrLenTest.panama_strlen_unsafe_trivial       5  avgt   30  119.331 � 6.429  ns/op
StrLenTest.panama_strlen_unsafe_trivial      20  avgt   30  127.749 � 1.932  ns/op
StrLenTest.panama_strlen_unsafe_trivial     100  avgt   30  129.092 � 0.967  ns/op
```

(timings seem a bit jittery)

PR #456, opened 2021-02-18
8261906: Improve jextract support for virtual functions

mcimadamore:
This patch add support for function pointer calls. Whenever jextract encounters a function pointer variable (be it a global variable, or a struct field), it will additionally emit code to _call_ the function pointer - by means of a static wrapper around a virtual method handle.

The implementation is relatively straightforward, although there is some duplication in the code that is being emitted. I played with this a bit, to see if duplication could be removed, but I ended up with a single routine to generate wrappers which worked across the axis { virtual, non virtual } x { struct, global } - which in the end I found too complex for my taste (note that, e.g. when we're inside a struct, the logic to get the virtual address depends on what else is generated inside the struct, so there's a lot of ad-hocness).

For these reason I ended up with separate emit function for virtual wrappers (e.g. I did not reuse the code for native functions) - the code which emit virtual function wrappers is also overridden in StructBuilder. My sense is that if we move the code to use text blocks, the duplication will be much less problematic.

In OutputFactory, I cleaned up things a bit - since I realized that support for valist doesn't work - this is now removed. I will file a followup issue to add proper VaList support, both in downcalls and upcalls.

---------
### Progress

### Issue
 * [JDK-8261906](https://bugs.openjdk.java.net/browse/JDK-8261906): Improve jextract support for virtual functions

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/456/head:pull/456`
`$ git checkout pull/456`


src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/JavaSourceBuilder.java
```diff
@@ -256,6 +263,147 @@ protected void emitGetter(String mods, Class<?> type, String name, String access
         emitGetter(mods, type, name, access, false, null);
     }
 
+    protected void emitFunctionWrapper(String mods, MethodType mtype, String javaName, String access,
```
mcimadamore, line 266: This was just moved "as is" from HeaderFileBuilder, so that ConstantBuilder could access it

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/Writer.java
```diff
@@ -51,7 +51,7 @@ public Writer(Path dest, List<? extends JavaFileObject> files) {
         } else {
             return InMemoryJavaCompiler.compile(sources,
                 "--add-modules", "jdk.incubator.foreign",
-                "--release", "16",
+                "--release", "17",
```
mcimadamore, line 54: Note that this is required to see the new linker method to create virtual handles - but this creates a dependency with 17 - e.g. the generated API will no longer work against 16.

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/OutputFactory.java
```diff
@@ -434,13 +427,41 @@ public Void visitVariable(Declaration.Variable tree, Declaration parent) {
...
+        if (descriptor == null) {
+            //abort
+            return;
+        }
+
+        //generate functional interface
+        if (funcPtr.varargs() && !funcPtr.argumentTypes().isEmpty()) {
+            warn("varargs in callbacks is not supported: " + funcPtr);
+            return;
+        }
```
JornVernee, line 452: Seems the same as in `generateFunctionalInterface` ? Might be worth it to unify?

test/jdk/tools/jextract/funcPointerInvokers/TestFuncPointerInvokers.java
```diff
@@ -0,0 +1,69 @@
+/*
+ * Copyright (c) 2020, Oracle and/or its affiliates. All rights reserved.
```
JornVernee, line 2:
```suggestion
 * Copyright (c) 2021, Oracle and/or its affiliates. All rights reserved.
```

test/jdk/tools/jextract/funcPointerInvokers/TestFuncPointerInvokers.java
```diff
@@ -0,0 +1,69 @@
...
+ * @run testng/othervm -Dforeign.restricted=permit TestFuncPointerInvokers
+ */
+public class TestFuncPointerInvokers {
+    @Test
+    public void testStructField() {
+        try (NativeScope scope = NativeScope.unboundedScope()) {
+            AtomicInteger val = new AtomicInteger(-1);
+            MemorySegment bar = Bar.allocate(scope);
+            Bar.foo$set(bar, Foo.allocate((i) -> val.set(i), scope).address());
+            Bar.foo(bar, 42);
```
JornVernee, line 55: Nice! :)

JornVernee (approved):
LGTM. Minor comments inline

test/jdk/tools/jextract/funcPointerInvokers/func.h
```diff
@@ -0,0 +1,23 @@
+#include <stdarg.h>
```
JornVernee, line 1: Didn't notice until now, but this is missing a copyright header.

test/jdk/tools/jextract/funcPointerInvokers/libFunc.c
```diff
@@ -0,0 +1 @@
+#include "func.h"
```
JornVernee, line 1: Same here

JornVernee (approved):

mcimadamore:
I've addressed the various comments; I ended up simplifying the code a bit and remove the Constant.MethodHandle abstraction which was, I think, a bit too convoluted. Now virtual function invokers are generated as part of JavaSourceBuilder::addVar, which I think makes the most sense from the perspective of OutputFactory. To facilitate passing multiple arguments from OutputFactory to the various builders, I've added a couple of records (VarInfo and FunctionInfo) which store various info about variables and functions (e.g. carrier, method types, layouts, descriptors). This also allowed me to reuse most of the code in OutputFactory between visitVariable/visitFunction and generateFunctionalInteraface. Now all the three methods call a factory which returns an optional FunctionInfo (if missing, it means that the function/variable/func interface cannot be generated).

mcimadamore:
Question: I realized there are two ways to support function pointers: one would be to do what this patch does, e.g. add an extra accessor to "call" the underlying function pointer. The other would be to have an accessor which, given a memory address encoding a function pointer, returns an instance of the functional interface that can be used to call it. In other words, we have a choice here:

```
typedef void (*func)(int,float);

func f;
```

One option is to generate a toplevel accessor, like this patch does:

```
void f(int x0, float x1) { ... }
```

another option is to piggy back on the functional interface that jextract emits for the typedef:

```
func f();
```

Both are valid approaches. The latter seem to depend on JDK-8259054

JornVernee:
I think both options look good. The latter is maybe a little more flexible, but the usage of the former looks more like what you would see in C:
```
foo->f(10, 21F)
```
In Java:
```
foo.f(segment, 10, 21F)
```
vs.
```
foo.f(segment).apply(10, 21F)
```

Maybe we can just do both? Otherwise, I think out of these 2 I prefer the first. Both in terms of syntax, and since it's the more primitive variant of the 2 (i.e. no need to create an intermediate wrapper instance).

mcimadamore:
> Maybe we can just do both? Otherwise, I think out of these 2 I prefer the first, since it's the more primitive variant of the 2 (i.e. no need to create an intermediate wrapper instance).

I have no strong preference, but the currently implemented option does have the advantage of not requiring an intermediate allocation for the wrapper (although not 100% how that matters in practice: I'd expect the object to be escape-analyzed if used only once).

Having two is problematic in terms of adding the required overloads (e.g. we'd need to make up a new name for both options, which might be hard to decipher).

It also occurs to me, that, with the currently implemented option, one could still do this:

```
func f = header::f;
```
E.g. if a functional interface is expected, it should be possible to instantiate it using a method reference.

JornVernee:
> Having two is problematic in terms of adding the required overloads (e.g. we'd need to make up a new name for both options, which might be hard to decipher).

Ah right. Maybe we could call the direct accessor `f$call` instead ?
 
> It also occurs to me, that, with the currently implemented option, one could still do this:
> 
> ```
> func f = header::f;
> ```
> 
> E.g. if a functional interface is expected, it should be possible to instantiate it using a method reference.

That's an interesting idea, though I think it only works for global variables? (struct fields require a MemorySegment as well).

I think in general, when JDK-8259054 is implemented it should be easy enough to get a function interface instance:

```
func x = func.ofAddress(foo.f$get(segment));
```

So, maybe that's another reason to pick the simpler accessor...

mcimadamore:
> > Having two is problematic in terms of adding the required overloads (e.g. we'd need to make up a new name for both options, which might be hard to decipher).
> 
> Ah right. Maybe we could call the direct accessor `f$call` instead ?
> 
> > It also occurs to me, that, with the currently implemented option, one could still do this:
> > ```
> > func f = header::f;
> > ```
> > 
> > 
> > E.g. if a functional interface is expected, it should be possible to instantiate it using a method reference.
> 
> That's an interesting idea, though I think it only works for global variables? (struct fields require a MemorySegment as well).
> 
> I think in general, when JDK-8259054 is implemented it should be easy enough to get a function interface instance:
> 
> ```
> func x = func.ofAddress(foo.f$get(segment));
> ```
> 
> So, maybe that's another reason to pick the simpler accessor...

I don't know. I think the argument cuts both ways. On the one hand, the simpler accessor seems more direct, but requires also more ad-hoc logic in code generation. On the other hand, if we fix JDK-8259054 and then we generate a function pointer accessor which does what you wrote above, we reuse more of the bits we already generate.

For instance, I like the fact that, by returning a functional interface, we stick to a principle where each bit of generate code does one thing: the accessor creates the functional interface; the apply method on the functional interface does the call (as opposed to have a single method which does both).

In other words, I find this:

```
foo.f(segment).apply(10, 21F)
```

*more* readable (not less) than (and less magic than):

```
foo.f(segment, 10, 21F)
```

JornVernee:
> I don't know. I think the argument cuts both ways. On the one hand, the simpler accessor seems more direct, but requires also more ad-hoc logic in code generation. On the other hand, if we fix JDK-8259054 and then we generate a function pointer accessor which does what you wrote above, we reuse more of the bits we already generate.
> 
> For instance, I like the fact that, by returning a functional interface, we stick to a principle where each bit of generate code does one thing: the accessor creates the functional interface; the apply method on the functional interface does the call (as opposed to have a single method which does both).
> 
> In other words, I find this:
> 
> ```
> foo.f(segment).apply(10, 21F)
> ```
> 
> _more_ readable (not less) than (and less magic than):
> 
> ```
> foo.f(segment, 10, 21F)
> ```

Yeah, that's a good point. It also gets rid of the somewhat awkward segment prefix argument. Though I'm not sure I'd say that I find the former example less clunky-looking in comparison, it seems clearer that way, which I think is more important. So, you've convinced me ;) Let's go for the functional interface instance accessor.

mcimadamore:
I've uploaded a new iteration which emits a functional interface getter instead of the direct invoker we were generating before. The code got a bit simpler as the functional interface getter is small to generate compared to the mh invoker wrapper. There's a bit of dance in OutputFactory to get the name of the typedef which corresponds to the functional interface we have generated; the logic is not super robust, in the sense that we're just going to assume that if we see a function pointer type def, there's gonna be a functional interface with same name.

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/FunctionalInterfaceBuilder.java
```diff
@@ -49,6 +51,7 @@
     JavaSourceBuilder classEnd() {
         emitFunctionalInterfaceMethod();
         emitFunctionalFactories();
+        emitFunctionaRestrictedFactory();
```
JornVernee, line 54: Typo
```suggestion
        emitFunctionalRestrictedFactory();
```

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/OutputFactory.java
```diff
@@ -274,74 +245,59 @@ public Void visitFunction(Declaration.Function funcTree, Declaration parent) {
...
+
+        Optional<FunctionInfo> functionInfo = functionInfo(funcTree.type(), funcTree.name(), true,
+                (mtype, desc) -> FunctionInfo.ofFunction(mtype, desc, funcTree.type().varargs(), paramNames));
+
+        if (functionInfo.isPresent()) {
+            int i = 0;
+            for (Declaration.Variable param : funcTree.parameters()) {
+                Type.Function f = getAsFunctionPointer(param.type());
+                if (f != null) {
+                    String name = funcTree.name() + "$" + (param.name().isEmpty() ? "x" + i : param.name());
```
JornVernee, line 264: This seems to let the name of the functional interface class depend on the parameter name? I thought we did _not_ want to do that?

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/OutputFactory.java
```diff
@@ -427,9 +383,28 @@ public Void visitVariable(Declaration.Variable tree, Declaration parent) {
...
+            fiName = generateFunctionalInterface(func, fieldName);
+            if (fiName != null) {
+                varInfo = VarInfo.ofFunctionalPointerVar(clazz, layout, fiName);
+            }
+        } else {
+            Optional<String> funcTypedef = getAsFunctionPointerTypedef(type);
+            if (funcTypedef.isPresent()) {
+                varInfo = VarInfo.ofFunctionalPointerVar(clazz, layout, Utils.javaSafeIdentifier(funcTypedef.get()));
+            }
         }
```
JornVernee, line 408: I see what you mean here. I guess we expect these functional interface classes to be generated when visiting the typedef?

test/jdk/tools/jextract/funcPointerInvokers/TestFuncPointerInvokers.java
```diff
@@ -0,0 +1,89 @@
...
+ * @run testng/othervm -Dforeign.restricted=permit TestFuncPointerInvokers
+ */
+public class TestFuncPointerInvokers {
+    @Test
+    public void testStructField() {
+        try (NativeScope scope = NativeScope.unboundedScope()) {
+            AtomicInteger val = new AtomicInteger(-1);
+            MemorySegment bar = Bar.allocate(scope);
+            Bar.foo$set(bar, Foo.allocate((i) -> val.set(i), scope).address());
+            Bar.foo(bar).apply(42);
```
JornVernee, line 54: Cool! :)

JornVernee (approved):
Changes look good! Though there is a merge failure it looks like.

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/OutputFactory.java
JornVernee: Ah, I see it's per-existing, sorry. I guess we can take care of it in another patch.

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/OutputFactory.java
mcimadamore: yeah - not great - as the functional interface name might differ significantly from the typedef name in case of clashes...

mcimadamore: Can we rely on typedef names being unique? If so, we could build a map String -> FunctionBuilder when we construct a functional interface for a typedef.

In a header file this should perhaps always be the case - but I've seen that you can have suspicious stuff like

```
{
typedef int a;
}

{
typedef float a;
}
```
But maybe we shouldn't worry about this with C headers?

JornVernee: I'm not sure if typedefs have any scope in C, or if they're just global always. C++ has scoped typedefs, but I think in C every name period is global.

JornVernee: Godbolt says I'm wrong though :) The example you give is allowed, and the typedef name's scope is restricted to the block

mcimadamore: The snippet I typed above compiles (e.g. no error about redefinition of same name, which you do get w/o the braces).

JornVernee: C Standard:

An identifier can denote an object; a function; a tag or a member of a structure, union, orenumeration;  a  typedef  name; a  label  name;  a  macro  name;  or  a  macro  parameter. The same identifier can denote different entities at different points in the program.

JornVernee: > But that could only happen inside a function body - unless I'm missing something?
Yeah, I think that's correct.

mcimadamore: Then I think I'll settle on a map from String to FunctionalIterfaceBuilder

mcimadamore: Actually, I did better and added a map from the delegated type to the functional class name - since both declaration and type support deep equality.

JornVernee (approved):
Looks good!

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/OutputFactory.java
JornVernee: That's an even better solution!

PR #457, opened 2021-02-22
8262118: Specialize upcalls

JornVernee:
Hi,

This patch adds specialization of upcalls, which includes both the specialization of the binding recipe using MethodHandle combinators, as for downcalls, as well as VM support for generating a customized wrapper for a 'low-level' method handle taking and returning only primitive types, which can be used as a base handle in the binding specialization.

The latter is based on Vladimir Ivanov's earlier linkToNative work, with a few changes to the frame layout, exception handling, and added support for attaching non-java threads, as well as re-writing it to support an arbitrary caller ABI.

The current implementation supports x86 only for now - though I've made sure all the tests still pass on AArch64. Stack argument passing is disabled, as for downcalls - there is a mismatch between the encoding for stack slots we use on the Java side, and the encoding we use in the VM, which needs some more consideration. And finally, as with downcalls, multi-register returns are disabled as well. We need to figure out what the right protocol for those should be, maybe taking inspiration from Valhalla.

In the code I've moved some things from ProgrammableInvoker to SharedUtils in order to reuse for upcalls. I also added a bit of debugging code to BufferLayout for dumping stack arguments. Note that there is quite a bit of motion in ProgrammableUpcallHandler as well. This is mostly to untangle the part of an upcall that is specialized using MethodHandle combinators (invokeInterpBindings), from the part that is replaced by the VM's wrapper stub (invokeMoves). I've for instance introduced a simple helper class, called InterpretedHandler, which acts as a wrapper around the target MethodHandle for doing the old-style interpreted calls, instead of using the whole ProgrammableUpcallHandler instance.

I've also removed some of the test configurations. We now only test the default settings of various specialization and intrinsification flags when running TestDowncall, TestUpcall, and TestUpcallHighArity. This patch 2 more of these flags, making a total of 16 combinations, which would take way too long to test, and is probably not necessary. I gave the full testing matrix a run, and found no issues. We can repeat that process periodically to check that all the combinations still work. I've also added some more benchmark cases for the upcalls benchmark.

I also gave the Windows jextract samples a try with this patch and found no issues.

Here are some of the benchmark numbers (from Linux-x64):

```
Benchmark                Mode  Cnt    Score   Error  Units
Upcalls.jni_args10       avgt   30  134.220 ? 1.553  ns/op
Upcalls.jni_args5        avgt   30   87.340 ? 1.753  ns/op
Upcalls.jni_blank        avgt   30   57.590 ? 0.877  ns/op
Upcalls.jni_identity     avgt   30  109.859 ? 0.930  ns/op
Upcalls.panama_args10    avgt   30   30.606 ? 0.088  ns/op
Upcalls.panama_args5     avgt   30   23.908 ? 0.121  ns/op
Upcalls.panama_blank     avgt   30   20.314 ? 0.137  ns/op
Upcalls.panama_identity  avgt   30   24.894 ? 0.111  ns/op
```
While these are  the numbers with the optimizations disabled (quite a bit worse):

```
Upcalls.panama_args10    avgt   30  886.885 ? 22.133  ns/op
Upcalls.panama_args5     avgt   30  566.600 ? 12.442  ns/op
Upcalls.panama_blank     avgt   30  212.962 ?  2.386  ns/op
Upcalls.panama_identity  avgt   30  361.358 ?  3.212  ns/op
```

We beat JNI across the board, and also seem to scale a lot better for larger numbers of arguments. So, this seems like a success :)

Thanks,
Jorn

---------
### Progress

### Issue
 * [JDK-8262118](https://bugs.openjdk.java.net/browse/JDK-8262118): Specialize upcalls

### Reviewers
 * [Maurizio Cimadamore](https://openjdk.java.net/census#mcimadamore) (@mcimadamore - Committer) ⚠️ Review applies to 638f8a10c76436abafd7688f52d5627713ee52a6
 * [Vladimir Ivanov](https://openjdk.java.net/census#vlivanov) (@iwanowww - Committer) ⚠️ Review applies to 638f8a10c76436abafd7688f52d5627713ee52a6

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/457/head:pull/457`
`$ git checkout pull/457`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/ProgrammableInvoker.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/ProgrammableInvoker.java
@@ -240,7 +240,7 @@ public class ProgrammableInvoker {
         }
 
         if (bufferCopySize > 0) {
-            specializedHandle = SharedUtils.wrapWithAllocatorForDowncall(specializedHandle, argAllocatorPos, bufferCopySize);
+            specializedHandle = SharedUtils.wrapWithAllocator(specializedHandle, argAllocatorPos, bufferCopySize, false);
         }
         return specializedHandle;
     }
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/ProgrammableInvoker.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/ProgrammableInvoker.java
@@ -240,7 +240,7 @@ public class ProgrammableInvoker {
         }
 
         if (bufferCopySize > 0) {
-            specializedHandle = SharedUtils.wrapWithAllocator(specializedHandle, argAllocatorPos, bufferCopySize);
+            specializedHandle = SharedUtils.wrapWithAllocatorForDowncall(specializedHandle, argAllocatorPos, bufferCopySize);
         }
         return specializedHandle;
     }
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/ProgrammableInvoker.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/ProgrammableInvoker.java
@@ -28,7 +28,6 @@ import jdk.incubator.foreign.Addressable;
 import jdk.incubator.foreign.MemoryAddress;
 import jdk.incubator.foreign.MemoryLayouts;
 import jdk.incubator.foreign.MemorySegment;
-import jdk.incubator.foreign.NativeScope;
 import jdk.internal.access.JavaLangInvokeAccess;
 import jdk.internal.access.SharedSecrets;
 import jdk.internal.foreign.abi.SharedUtils.Allocator;
@@ -44,19 +43,14 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.stream.Collectors;
-import java.util.stream.IntStream;
 import java.util.stream.Stream;
 
 import static java.lang.invoke.MethodHandles.collectArguments;
 import static java.lang.invoke.MethodHandles.dropArguments;
-import static java.lang.invoke.MethodHandles.empty;
 import static java.lang.invoke.MethodHandles.filterArguments;
 import static java.lang.invoke.MethodHandles.identity;
 import static java.lang.invoke.MethodHandles.insertArguments;
-import static java.lang.invoke.MethodHandles.tryFinally;
 import static java.lang.invoke.MethodType.methodType;
-import static jdk.internal.foreign.abi.SharedUtils.Allocator.THROWING_ALLOCATOR;
 import static jdk.internal.foreign.abi.SharedUtils.DEFAULT_ALLOCATOR;
 import static sun.security.action.GetBooleanAction.privilegedGetProperty;
 
@@ -79,12 +73,7 @@ public class ProgrammableInvoker {
 
     private static final MethodHandle MH_INVOKE_MOVES;
     private static final MethodHandle MH_INVOKE_INTERP_BINDINGS;
-
     private static final MethodHandle MH_ADDR_TO_LONG;
-    private static final MethodHandle MH_MAKE_SCOPE;
-    private static final MethodHandle MH_CLOSE_SCOPE;
-    private static final MethodHandle MH_WRAP_SCOPE;
-
     private static final Map<ABIDescriptor, Long> adapterStubs = new ConcurrentHashMap<>();
 
     private static final MethodHandle EMPTY_OBJECT_ARRAY_HANDLE = MethodHandles.constant(Object[].class, new Object[0]);
@@ -96,12 +85,6 @@ public class ProgrammableInvoker {
                     methodType(Object.class, long.class, Object[].class, Binding.VMStore[].class, Binding.VMLoad[].class));
             MH_INVOKE_INTERP_BINDINGS = lookup.findVirtual(ProgrammableInvoker.class, "invokeInterpBindings",
                     methodType(Object.class, Addressable.class, Object[].class, MethodHandle.class, Map.class, Map.class));
-            MH_MAKE_SCOPE = lookup.findStatic(NativeScope.class, "boundedScope",
-                    methodType(NativeScope.class, long.class));
-            MH_CLOSE_SCOPE = lookup.findVirtual(NativeScope.class, "close",
-                    methodType(void.class));
-            MH_WRAP_SCOPE = lookup.findStatic(Allocator.class, "ofScope",
-                    methodType(Allocator.class, NativeScope.class));
             MethodHandle MH_Addressable_address = lookup.findVirtual(Addressable.class, "address",
                     methodType(MemoryAddress.class));
             MethodHandle MH_MemoryAddress_toRawLongValue = lookup.findVirtual(MemoryAddress.class, "toRawLongValue",
@@ -176,8 +159,8 @@ public class ProgrammableInvoker {
         if (USE_SPEC && isSimple) {
             handle = specialize(handle);
          } else {
-            Map<VMStorage, Integer> argIndexMap = indexMap(argMoves);
-            Map<VMStorage, Integer> retIndexMap = indexMap(retMoves);
+            Map<VMStorage, Integer> argIndexMap = SharedUtils.indexMap(argMoves);
+            Map<VMStorage, Integer> retIndexMap = SharedUtils.indexMap(retMoves);
 
             handle = insertArguments(MH_INVOKE_INTERP_BINDINGS.bindTo(this), 2, handle, argIndexMap, retIndexMap);
             MethodHandle collectorInterp = makeCollectorHandle(callingSequence.methodType());
@@ -257,31 +240,9 @@ public class ProgrammableInvoker {
         }
 
         if (bufferCopySize > 0) {
-            // insert try-finally to close the NativeScope used for Binding.Copy
-            MethodHandle closer = leafType.returnType() == void.class
-                  // (Throwable, Addressable, NativeScope) -> void
-                ? collectArguments(empty(methodType(void.class, Throwable.class, Addressable.class)), 2, MH_CLOSE_SCOPE)
-                  // (Throwable, V, Addressable, NativeScope) -> V
-                : collectArguments( // (Throwable, V, Addressable, NativeScope) -> V
-                    dropArguments( // (Throwable, V, Addressable) -> V
-                        dropArguments( // (Throwable, V) -> V
-                            identity(specializedHandle.type().returnType()), // (V) -> V
-                            0, Throwable.class),
-                        2, Addressable.class),
-                                   3, MH_CLOSE_SCOPE);
-            // Handle takes a SharedUtils.Allocator, so need to wrap our NativeScope
-            specializedHandle = filterArguments(specializedHandle, argAllocatorPos, MH_WRAP_SCOPE);
-            specializedHandle = tryFinally(specializedHandle, closer);
-            MethodHandle makeScopeHandle = insertArguments(MH_MAKE_SCOPE, 0, bufferCopySize);
-            specializedHandle = collectArguments(specializedHandle, argAllocatorPos, makeScopeHandle);
+            specializedHandle = SharedUtils.wrapWithAllocator(specializedHandle, argAllocatorPos, bufferCopySize);
         }
         return specializedHandle;
-    }
-
-    private static Map<VMStorage, Integer> indexMap(Binding.Move[] moves) {
-        return IntStream.range(0, moves.length)
-                        .boxed()
-                        .collect(Collectors.toMap(i -> moves[i].storage(), i -> i));
     }
 
     /**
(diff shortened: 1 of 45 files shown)


test/jdk/java/foreign/TestDowncall.java
```diff
@@ -30,23 +30,6 @@
  *
  * @run testng/othervm
  *   -Dforeign.restricted=permit
- *   -Djdk.internal.foreign.ProgrammableInvoker.USE_SPEC=false
```
mcimadamore, line 33: Shouldn't we test all combinations? Why removing these sections?

test/jdk/java/foreign/TestUpcall.java
```diff
@@ -30,23 +30,6 @@
  *
  * @run testng/othervm
  *   -Dforeign.restricted=permit
- *   -Djdk.internal.foreign.ProgrammableInvoker.USE_SPEC=false
```
mcimadamore, line 33: Here, in addition to reverting these changes, we should also test with all the new flavors of upcall specialization/intrinsics?

mcimadamore (commented):
Not reviewed in full - added some comments on tests

test/jdk/java/foreign/TestUpcall.java
JornVernee: I've tried to explain this in the PR body. Testing the current configuration already takes 15 minutes. Adding the new configurations would multiply the number of configurations by 4. This seems excessive. The full matrix can be tested periodically (maybe as part of a higher tier), but for regular testing runs we could only run the default configuration?

mcimadamore: Oops - I missed that part. On linux it doesn't take long to execute (1 min maybe) so I was surprised. But I understand it might be an issue in other platforms. That said, in the recent ResourceScope rewrite, executing using all modes has been helpful in shaking out issues, so I'm a bit worried to see this going completely. I'm pretty sure in a month I'll forget that I have to go in a manually edit to add the missing combo...

mcimadamore: One wacky idea: I assume these properties are not read-only - what if, the tests had an additional parameter which dictated the specific setup for generating upcall/downcalls? That way, we can decide how many stuff should be included when generating the various test combinations (and maybe just keep the main mode for standard executions).

JornVernee: Yes, for such big patches it seems useful, but if the linker internals are untouched not so much. I can look a bit more at this and separate the full configuration out into a separate test group perhaps, e.g. we'd have like `run-test-jdk_foreign` and `run-test-jdk_foreign-full` (or with the default reversed).

JornVernee: Isn't the additional parameter pretty much what we do now?

JornVernee: Okay, I've added another test file called TestMatrix that contains all the configurations. This is not run by default, but could be run in isolation to shake out issues.

mcimadamore: I meant having a different data provider which tests every combo, by first setting JDK parameter, then creating downcall/upcall

JornVernee: Hmm, would need to re-write the linker to query the property every time when linking, instead of initializing it once when the class is initialized. (I don't think it's possible to run tests in different forks with just TestNG). Seems like using a separate test file is more straight-forward? Maybe I'm misunderstanding but the result seems more or less the same.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/aarch64/CallArranger.java
```diff
@@ -324,7 +322,7 @@ protected void spillStructBox(Binding.Builder bindings, MemoryLayout layout) {
                         for (int i = 0; i < group.memberLayouts().size(); i++) {
                             VMStorage storage = regs[i];
                             final long size = group.memberLayouts().get(i).byteSize();
-                            Class<?> type = SharedUtils.primitiveCarrierForSize(size);
+                            Class<?> type = SharedUtils.primitiveCarrierForSize(size, false);
```
mcimadamore, line 325: Not sure this is ok - this is a struct of floats passed by register?

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/aarch64/CallArranger.java
```diff
@@ -423,7 +421,7 @@ protected void spillStructBox(Binding.Builder bindings, MemoryLayout layout) {
...
```
mcimadamore, line 424: Same here - aren't HFA passed by (vector) registers?

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/x64/sysv/CallArranger.java
```diff
@@ -270,10 +268,11 @@ protected BindingCalculator(boolean forArguments) {
                     while (offset < layout.byteSize()) {
                         final long copy = Math.min(layout.byteSize() - offset, 8);
                         VMStorage storage = regs[regIndex++];
-                        Class<?> type = SharedUtils.primitiveCarrierForSize(copy);
                         if (offset + copy < layout.byteSize()) {
                             bindings.dup();
                         }
+                        boolean useFloat = storage.type() == StorageClasses.VECTOR;
```
mcimadamore, line 274: So , my understanding here is that we had issues all along - e.g. this is unrelated to the upcall work?

test/jdk/TEST.groups
```diff
@@ -337,7 +337,8 @@ jdk_svc = \
     :svc_tools
 
 jdk_foreign = \
-    java/foreign
+    java/foreign \
+    -java/foreign/TestMatrix.java
```
mcimadamore, line 341: Thanks for setting this up!

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/ProgrammableUpcallHandler.java
```diff
@@ -28,115 +28,287 @@
...
+            JLI.ensureCustomized(doBindings);
+            VMStorage[] args = Arrays.stream(argMoves).map(Binding.Move::storage).toArray(VMStorage[]::new);
+            VMStorage[] rets = Arrays.stream(retMoves).map(Binding.Move::storage).toArray(VMStorage[]::new);
+            CallRegs conv = new CallRegs(args, rets);
+            entryPoint = allocateOptimizedUpcallStub(doBindings, abi, conv);
+        } else {
+            BufferLayout layout = BufferLayout.of(abi);
+            MethodHandle doBindingsErased = doBindings.asSpreader(Object[].class, doBindings.type().parameterCount());
+            MethodHandle invokeMoves = insertArguments(MH_invokeMoves, 1, doBindingsErased, argMoves, retMoves, abi, layout);
+            InterpretedHandler handler = new InterpretedHandler(invokeMoves);
```
mcimadamore, line 134: In principle, this could be just replaced by the method handle no? Then we could have a static entry point in this class which can be used by the VM to "call" the method handle with given args. E.g. do we need the `InterpreterHandler` wrapper?

mcimadamore (commented):
Looks a very solid piece of work (I've mostly checked the Java side). I like how the code has been consolidated between upcall and downcall paths. I've added some minor comments around for your consideration.

mcimadamore:
When trying this on my machine (Ubuntu 20.04) I get this:
```
In file included from ../../src/hotspot/share/prims/foreign_globals.cpp:25:
../../src/hotspot/share/prims/foreign_globals.hpp:40:46: error: 'VMRegPair' has not been declared
   40 |   void calling_convention(BasicType* sig_bt, VMRegPair *parm_regs, uint argcnt) const;
      |                                              ^~~~~~~~~
../../src/hotspot/share/prims/foreign_globals.cpp:102:6: error: no declaration matches 'void CallRegs::calling_convention(BasicType*, VMRegPair*, uint) const'
  102 | void CallRegs::calling_convention(BasicType* sig_bt, VMRegPair *parm_regs, uint argcnt) const {
      |      ^~~~~~~~
In file included from ../../src/hotspot/share/prims/foreign_globals.cpp:25:
../../src/hotspot/share/prims/foreign_globals.hpp:40:8: note: candidate is: 'void CallRegs::calling_convention(BasicType*, int*, uint) const'
   40 |   void calling_convention(BasicType* sig_bt, VMRegPair *parm_regs, uint argcnt) const;
      |        ^~~~~~~~~~~~~~~~~~
../../src/hotspot/share/prims/foreign_globals.hpp:33:8: note: 'struct CallRegs' defined here
   33 | struct CallRegs {
      |        ^~~~~~~~
```

I compile with disabled precompiled headers, which typically causes issues like this - is there some include missing somewhere?

mcimadamore:
Other includes seem to be missing - here's the patch I needed to get it working:

```
+++ b/src/hotspot/cpu/x86/universalUpcallHandler_x86_64.cpp
@@ -31,6 +31,8 @@
 #include "memory/resourceArea.hpp"
 #include "prims/universalUpcallHandler.hpp"
 #include "runtime/sharedRuntime.hpp"
+#include "runtime/signature.hpp"
+#include "utilities/formatBuffer.hpp"
 #include "utilities/globalDefinitions.hpp"
 
 #define __ _masm->
diff --git a/src/hotspot/share/prims/foreign_globals.hpp b/src/hotspot/share/prims/foreign_globals.hpp
index 4f02ee625cb..44c65771e7c 100644
--- a/src/hotspot/share/prims/foreign_globals.hpp
+++ b/src/hotspot/share/prims/foreign_globals.hpp
@@ -27,6 +27,7 @@
 #include "oops/oopsHierarchy.hpp"
 #include "utilities/growableArray.hpp"
 #include "utilities/macros.hpp"
+#include "code/vmreg.hpp"
 
 #include CPU_HEADER(foreign_globals)
```

mcimadamore:
Executed benchmark on my machine:

```
Benchmark                Mode  Cnt    Score   Error  Units
Upcalls.jni_args10       avgt   30  102.245 ? 1.952  ns/op
Upcalls.jni_args5        avgt   30   59.715 ? 1.488  ns/op
Upcalls.jni_blank        avgt   30   43.189 ? 0.642  ns/op
Upcalls.jni_identity     avgt   30   90.224 ? 1.775  ns/op
Upcalls.panama_args10    avgt   30   24.633 ? 0.260  ns/op
Upcalls.panama_args5     avgt   30   18.708 ? 0.314  ns/op
Upcalls.panama_blank     avgt   30   15.925 ? 0.262  ns/op
Upcalls.panama_identity  avgt   30   19.769 ? 0.266  ns/op
```
Seems to be from 2x to 4x faster - awesome!

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/aarch64/CallArranger.java
JornVernee: Yes, I found that that the native code passes float-only structs in float registers, but we were trying to load it using an int carrier, which maps to an int register in the Java calling convention. Besides being inefficient to have to move between these registers, it is not possible to move between float and int registers directly in x86, which would complicate the shuffling process by having to go through memory.

By using float/double as a carrier here instead, the upcall handle expects it's argument to be in a float register as well, and we don't have to move it between registers.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/ProgrammableInvoker.java
```diff
@@ -257,33 +240,11 @@ private MethodHandle specialize(MethodHandle leafHandle) {
...
-                            identity(specializedHandle.type().returnType()), // (V) -> V
-                            0, Throwable.class),
-                        2, Addressable.class),
-                                   3, MH_CLOSE_SCOPE);
-            // Handle takes a SharedUtils.Allocator, so need to wrap our NativeScope
-            specializedHandle = filterArguments(specializedHandle, argAllocatorPos, MH_WRAP_SCOPE);
-            specializedHandle = tryFinally(specializedHandle, closer);
-            MethodHandle makeScopeHandle = insertArguments(MH_MAKE_SCOPE, 0, bufferCopySize);
-            specializedHandle = collectArguments(specializedHandle, argAllocatorPos, makeScopeHandle);
+            specializedHandle = SharedUtils.wrapWithAllocator(specializedHandle, argAllocatorPos, bufferCopySize, false);
```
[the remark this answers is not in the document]
JornVernee: Thanks, will remove the leafType variable.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/aarch64/CallArranger.java
JornVernee: Ah, wait this is AArch64. I've kept this as is so far (it was previously using ints as a carrier as well). You are probably right that this is incorrect.

Currently it doesn't matter though since we always go through an intermediate buffer (AArch64 doesn't support specialized upcalls yet). I simply didn't attempt to pick the correct type here yet.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/x64/sysv/CallArranger.java
JornVernee: Not in issue so far since we were using an intermediate buffer. We simply write the bits as an integer, but the native code reads it as a long.

Since the specialized version doesn't use an intermediate buffer though, the types have to be aligned properly.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/aarch64/CallArranger.java
mcimadamore: Yes - this is Aarch, the STRUCT_HFA case which AFAIK should be structs passed by FP registers.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/ProgrammableUpcallHandler.java
JornVernee: Probably not. I think I had more fields in there before, so it's a leftover.

JornVernee:
Thanks for testing. I forgot that precompiled headers  can be an issue, and was testing on multiple Linux machines, one of which doesn't test with PCH disabled. Will give it another try here as well.

JornVernee:
I've address review comments.

I've update the AArch64 CallArranger as well as the test to use float as a carrier type when also using vector registers, to avoid the same mismatch.

mcimadamore (approved):
Looks good!

iwanowww (approved):
Looks good!

PR #458, opened 2021-02-23
8262117: jextract crashes with javac compilation error "class u is already defined"

sundararajana:
enclosing names are checked for clash now

---------
### Progress

### Issue
 * [JDK-8262117](https://bugs.openjdk.java.net/browse/JDK-8262117): jextract crashes with javac compilation error "class u is already defined"

### Reviewers
 * [Maurizio Cimadamore](https://openjdk.java.net/census#mcimadamore) (@mcimadamore - Committer) ⚠️ Review applies to bea8e2d43e5dec5fc9f127c1adc7ede3c6e377f8

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/458/head:pull/458`
`$ git checkout pull/458`

### DIFF
--- a/src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/JavaSourceBuilder.java
+++ b/src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/JavaSourceBuilder.java
@@ -212,7 +212,7 @@ abstract class JavaSourceBuilder {
 
     // is the name enclosed enclosed by a class of the same name?
     boolean isEnclosedBySameName(String name) {
-        return false;
+        return className().equals(name);
     }
 
     protected void emitPackagePrefix() {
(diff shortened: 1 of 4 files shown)


src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/JavaSourceBuilder.java
```diff
@@ -205,7 +205,14 @@ String build() {
...
         name = Utils.javaSafeIdentifier(name);
-        return nestedClassNames.add(name.toLowerCase()) ? name : (name + "$" + nestedClassNameCount++);
+        var notSeen = nestedClassNames.add(name.toLowerCase());
+        var notEnclosed = !isEnclosedBySameName(name.toLowerCase());
+        return notSeen && notEnclosed? name : (name + "$" + nestedClassNameCount++);
+    }
+
+    // is the name enclosed enclosed by a class of the same name?
+    boolean isEnclosedBySameName(String name) {
+        return false;
```
mcimadamore, line 215: Should we add className().equals(name) in here - just to be on the safe side>

I guess it's never possible for a nested class to have the same name of a "toplevel" class (e.g. generated by HeaderFileBuilder), because we explicitly mangle the names (e.g. `toplevel$nested`). But I think adding the name check at the toplevel is the correct thing to do code-wise.

Then in the subclass you can do `super.isEnclosedBySameName(name) || enclosing.isEnclosedBySameName(name))`

mcimadamore (approved):
Looks good, added a minor comment

PR #459, opened 2021-02-23
8262198: Overhaul bitfield parsing logic

mcimadamore:
This parsing stratgey cannot (by definition) fail - so, moving forward there will be no more exceptions coming from this side of the code. Of course, since we no longer have container classification, when bitfield support is added to the backend, things might be a bit tricker; however I realized that, since the layout is well-formed, the backend can always use the layout API to query the offset and size of a given bitfield, and decide on the best strategy to get there (e.g. get the closed 8/16/32/64-bit word and apply some bitmasking). So, perhaps things don't change much in terms of code generation support.

This patch fixes a small bug in Parser, where we where creating the same tree twice (!!). There are some small tweaks to TreeMaker, to work with the new bitfield info coming out from RecordLayoutComputer, but overall nothing too difficult here.

Note that the `BadBitfieldTest`, which was previously asserting a jextract crash, has now turned into a positive test (and I've dropped the exclusion on Windows, since the test should now pass on all platforms).

[1] - https://github.com/rust-lang/rust-bindgen/issues/743

---------
### Progress

### Issue
 * [JDK-8262198](https://bugs.openjdk.java.net/browse/JDK-8262198): Overhaul bitfield parsing logic

### Reviewers
 * [Athijegannathan Sundararajan](https://openjdk.java.net/census#sundar) (@sundararajana - Committer) ⚠️ Review applies to e1c2e7fefbaeb86c874028492eba1e960e295e11
 * [Jorn Vernee](https://openjdk.java.net/census#jvernee) (@JornVernee - Committer) ⚠️ Review applies to e1c2e7fefbaeb86c874028492eba1e960e295e11

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/459/head:pull/459`
`$ git checkout pull/459`

### DIFF
--- a/src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/RecordLayoutComputer.java
+++ b/src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/RecordLayoutComputer.java
@@ -137,13 +137,8 @@ abstract class RecordLayoutComputer {
         return c.isBitField() ? c.getBitFieldWidth() : c.type().size() * 8;
     }
 
-    ValueLayout bitfield(ValueLayout container, List<MemoryLayout> sublayouts) {
-        return Utils.addContents(container, MemoryLayout.ofStruct(sublayouts.toArray(new MemoryLayout[0])));
-    }
-
-    ValueLayout bitfield(long containerSize, List<MemoryLayout> sublayouts) {
-        return bitfield((ValueLayout)LayoutUtils.valueLayoutForSize(containerSize)
-                        .layout().orElseThrow(() -> new IllegalStateException("Unsupported size: " + containerSize)), sublayouts);
+    MemoryLayout bitfield(List<MemoryLayout> sublayouts) {
+        return MemoryLayout.ofStruct(sublayouts.toArray(new MemoryLayout[0])).withAttribute("BITFIELDS", true);
     }
 
     long offsetOf(Type parent, Cursor c) {
--- a/src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/RecordLayoutComputer.java
+++ b/src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/RecordLayoutComputer.java
@@ -26,9 +26,7 @@
 
 package jdk.internal.jextract.impl;
 
-import jdk.incubator.foreign.GroupLayout;
 import jdk.incubator.foreign.MemoryLayout;
-import jdk.incubator.foreign.ValueLayout;
 import jdk.internal.clang.Cursor;
 import jdk.internal.clang.CursorKind;
 import jdk.internal.clang.Type;
@@ -138,7 +136,7 @@ abstract class RecordLayoutComputer {
     }
 
     MemoryLayout bitfield(List<MemoryLayout> sublayouts) {
-        return MemoryLayout.ofStruct(sublayouts.toArray(new MemoryLayout[0])).withAttribute("BITFIELDS", true);
+        return LayoutUtils.setBitfields(MemoryLayout.ofStruct(sublayouts.toArray(new MemoryLayout[0])));
     }
 
     long offsetOf(Type parent, Cursor c) {
--- a/src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/UnionLayoutComputer.java
+++ b/src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/UnionLayoutComputer.java
@@ -33,6 +33,7 @@ import jdk.internal.clang.Cursor;
 import jdk.internal.clang.Type;
 import jdk.internal.clang.TypeKind;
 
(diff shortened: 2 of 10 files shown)


sundararajana (approved):
LGTM

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/RecordLayoutComputer.java
```diff
@@ -137,13 +135,8 @@ long fieldSize(Cursor c) {
...
 
-    ValueLayout bitfield(ValueLayout container, List<MemoryLayout> sublayouts) {
-        return Utils.addContents(container, MemoryLayout.ofStruct(sublayouts.toArray(new MemoryLayout[0])));
-    }
-
-    ValueLayout bitfield(long containerSize, List<MemoryLayout> sublayouts) {
-        return bitfield((ValueLayout)LayoutUtils.valueLayoutForSize(containerSize)
-                        .layout().orElseThrow(() -> new IllegalStateException("Unsupported size: " + containerSize)), sublayouts);
+    MemoryLayout bitfield(List<MemoryLayout> sublayouts) {
+        return LayoutUtils.setBitfields(MemoryLayout.ofStruct(sublayouts.toArray(new MemoryLayout[0])));
```
JornVernee, line 139: Does this work correctly for unions? What if we have many bitfields that together end up being large than the union itself?

```
union Foo {
    int x;
    int bitfield1: 24;
    int bitfield2: 24;
};
```
Doesn't this get laid out like:
```
union Foo {
    int x;
    struct {
        int bitfield1: 24;
        int bitfield2: 24;
    }
};
```

With this patch? The layouts seem to differ between those two.

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/UnionLayoutComputer.java
```diff
@@ -77,15 +77,19 @@ MemoryLayout fieldLayout(Cursor c) {
...
     }
 
     @Override
     MemoryLayout finishLayout() {
-        // size mismatch indicates anonymous bitfield used for padding
+        // size mismatch indicates use of bitfields in union
         long expectedSize = type.size() * 8;
         if (actualSize < expectedSize) {
+            // emit an extra padding of expected size to make sure union layout size is computed correctly
             addFieldLayout(MemoryLayout.ofPaddingBits(expectedSize));
```
JornVernee, line 93: Maybe an assert could be added here to check that `actualSize` is exactly `expectedSize` at the end. I think we have a problem otherwise.

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/RecordLayoutComputer.java
mcimadamore: My understanding is that bitfields in union are just disjoint. If you want them to be adjacent, then you need to put them inside a struct.

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/UnionLayoutComputer.java
mcimadamore: actual size can be less if there are bitfields - in which case we need to emit a padding word to make sure the union layout has right size.

JornVernee: Yes, I'm wondering if there can be cases where the actualSize is larger than the expected size, which isn't checked here currently.

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/RecordLayoutComputer.java
JornVernee: I guess that's what I'm saying. The patch seems to be collecting all the bitfields at once and then emitting a struct layout that holds them right? But, since in a union bitfields overlap, it seems that by putting them adjacent in a struct like this has the potential to create a layout that is larger then the actual expected size.

JornVernee: Ah, never mind. I see you are only collecting all the bitfields in the Struct case, but in the union case are just emitting a Struct layout with a single bitfield inside, more like:

```
union Foo {
    int x;
    struct {
        int bitfield1: 24;
    }
    struct {
        int bitfield2: 24;
    }
};
```

JornVernee (approved):

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/UnionLayoutComputer.java
mcimadamore: I'll add an assertion

PR #461, opened 2021-02-24
8262301: Reorganize jextract package structure

mcimadamore:
Despite the appearances, this is a straightforward change which has mostly been generated by moving things around in the IDE.

Over the last few days, I noticed that most jextract implementation classes belong to the same package; as the size of the implementation is growing, this flat organization is starting to create issues.

For instance, while most of the classes are organized in logically related lumps, the failure to capture those lumps in the package hierarchy is starting to have repercussion on the code, and creating some redundancies - the primary example is the Utils class which has become some kind of a dumping ground for all static functions needed throughout the implementation. In reality, on a closer look, most of the functions in there are related to code generation, but there's no way to cleanly separate between e.g. utlity functions for code generation and utility functions for layout computations.

Using the IDE, I've reorganized classes as follows:

```
jdk.internal.jextract
\- clang
\- parser
\- resources
\- layout
\- gen
\- util
```

In the `clang` package we find the autogenerated clang impl (as before). `parser` contains all the stuff related to parsing clang cursor and turnin them into declaration/types. `resources` contains templates for localized messages as well as for generated source. `layout` contains utilities for generating layouts from clang types/cursors. `gen` contains all we need to generate sources/classes.

Note that the `util` package just has 3 classes which don't fit anywhere else (`Options`, `Filter` and `PrettyPrinter`) - as long as we keep this under control, I think it's ok to have it (we have a similar package in the javac implementation and it has not created too many issues).

I only have done a very small number of manual changes, listed below:

* Renamed `LayoutUtils` to `Layouts`
* Merged `UnsupportedLayouts` into `Layouts`
* Renamed `MacroParserImpl` to `MacroParser`
* Removed a lot of unused method from `gen.Utils`
* Moved `flattenableChildren` from `Utils` to `RecordLayoutComputer` (as it was only used there)
* Fixed up a couple of issues with resource loading (as the package location changed), in `JextractTool` and `OutputFactory` respectively

Let me know what you think.

---------
### Progress

### Issue
 * [JDK-8262301](https://bugs.openjdk.java.net/browse/JDK-8262301): Reorganize jextract package structure

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/461/head:pull/461`
`$ git checkout pull/461`


mcimadamore:
Closed for now - we will revisit at later point.

PR #462, opened 2021-03-01
8262733: jextract generates clashing names which results in compilation error with javac

sundararajana:
name clash is avoided by checking surrounding class names

---------
### Progress

### Issue
 * [JDK-8262733](https://bugs.openjdk.java.net/browse/JDK-8262733): jextract generates clashing names which results in compilation error with javac

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/462/head:pull/462`
`$ git checkout pull/462`

### DIFF
--- a/src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/StructBuilder.java
+++ b/src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/StructBuilder.java
@@ -48,14 +48,12 @@ class StructBuilder extends ConstantBuilder {
     private final GroupLayout structLayout;
     private final Type structType;
     private final Deque<String> prefixElementNames;
-    private final String setterParameterName;
 
     StructBuilder(JavaSourceBuilder enclosing, String className, GroupLayout structLayout, Type structType) {
         super(enclosing, Kind.CLASS, className);
         this.structLayout = structLayout;
         this.structType = structType;
         prefixElementNames = new ArrayDeque<>();
-        setterParameterName = safeParameterName("x");
     }
 
     private String safeParameterName(String paramName) {
@@ -153,11 +151,11 @@ class StructBuilder extends ConstantBuilder {
     private void emitFieldGetter(Constant vhConstant, String javaName, Class<?> type) {
         incrAlign();
         indent();
-        append(MEMBER_MODS + " " + type.getSimpleName() + " " + javaName + "$get(MemorySegment seg) {\n");
+        append(MEMBER_MODS + " " + type.getSimpleName() + " " + javaName + "$get(MemorySegment " + safeParameterName("seg") + ") {\n");
         incrAlign();
         indent();
         append("return (" + type.getName() + ")"
-                + vhConstant.accessExpression() + ".get(seg);\n");
+                + vhConstant.accessExpression() + ".get(" + safeParameterName("seg") + ");\n");
         decrAlign();
         indent();
         append("}\n");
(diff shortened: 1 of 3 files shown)


src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/StructBuilder.java
```diff
@@ -48,12 +48,14 @@
...
     private final Type structType;
     private final Deque<String> prefixElementNames;
+    private final String safeParameterName;
 
     StructBuilder(JavaSourceBuilder enclosing, String className, GroupLayout structLayout, Type structType) {
         super(enclosing, Kind.CLASS, className);
         this.structLayout = structLayout;
         this.structType = structType;
         prefixElementNames = new ArrayDeque<>();
+        safeParameterName = isEnclosedBySameName("x")? "x$" : "x";
```
mcimadamore, line 58: I'm not sure by this. It seems very tailored to the issue you have found. I guess I'd prefer a function like:
```
String safeParameterName(String paramName)
```

Which can be called by clients. Note that when we'll add support for calling function pointers (https://git.openjdk.java.net/panama-foreign/pull/456) structs will also have "functions" (which allow to call a function pointer); these function have more than one parameter, each of which could probably clash with any of the enclosing classes. So, in short, I think all parameter names in nested classes should probably be guarded by similar code.

mcimadamore (commented):
hardwiring the choice of the parameter name to the builder seems odd, and will probably not scale when structs will feature functions with multiple parameters

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/StructBuilder.java
```diff
@@ -48,14 +48,18 @@
...
-    private final String safeParameterName;
+    private final String setterParameterName;
 
     StructBuilder(JavaSourceBuilder enclosing, String className, GroupLayout structLayout, Type structType) {
         super(enclosing, Kind.CLASS, className);
         this.structLayout = structLayout;
         this.structType = structType;
         prefixElementNames = new ArrayDeque<>();
-        safeParameterName = isEnclosedBySameName("x")? "x$" : "x";
+        setterParameterName = safeParameterName("x");
```
mcimadamore, line 58: Can we avoid the caching and just call `safeParameterName` where required? I think caching like this makes code generation more brittle than it has to, as it introduces an implicit dependency. In other words, in a way it is an "accident" that all methods generated in struct builder accept a parameter named "x" - I'd like not to rely on this assumption too much, and make sure that _all_ parameters of _all_ methods in any of the nested builders (StructBuilder, FunctionalInterfaceBuilder) are validated accordingly.

For instance, in `emitIndexedFieldSetter` the problem is not just `x` - the problem is also the `seg` and the `index` parameter (as the struct could be called `seg` or `index`). In other words, I think the parameter name should be mangled in a lot more places.

mcimadamore (commented):
Added more comments

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/StructBuilder.java
```diff
@@ -147,11 +151,11 @@ public void addVar(String javaName, String nativeName, MemoryLayout layout, Clas
     private void emitFieldGetter(Constant vhConstant, String javaName, Class<?> type) {
         incrAlign();
         indent();
-        append(MEMBER_MODS + " " + type.getSimpleName() + " " + javaName + "$get(MemorySegment seg) {\n");
+        append(MEMBER_MODS + " " + type.getSimpleName() + " " + javaName + "$get(MemorySegment " + safeParameterName("seg") + ") {\n");
```
mcimadamore, line 154: In this, and other methods, we could at least cache the safe name into a local and reuse it later?

mcimadamore (approved):
Looks good!

PR #464, opened 2021-03-02
8262851: jextract crashes with "Cannot compute size of a layout which is, or depends on a sequence layout with unspecified size"

sundararajana:
problematic fields are skipped with a warning

---------
### Progress

### Issue
 * [JDK-8262851](https://bugs.openjdk.java.net/browse/JDK-8262851): jextract crashes with "Cannot compute size of a layout which is, or depends on a sequence layout with unspecified size"

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/464/head:pull/464`
`$ git checkout pull/464`

### DIFF
--- a/src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/StructBuilder.java
+++ b/src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/StructBuilder.java
@@ -136,6 +136,14 @@ class StructBuilder extends ConstantBuilder {
 
     @Override
     public void addVar(String javaName, String nativeName, MemoryLayout layout, Class<?> type) {
+        try {
+            structLayout.byteOffset(elementPaths(nativeName));
+        } catch (UnsupportedOperationException uoe) {
+            // bad layout - do nothing
+            OutputFactory.warn("skipping '" + className() + "." + nativeName + "' : " + uoe.toString());
+            return;
+        }
+
         if (type.equals(MemorySegment.class)) {
             emitSegmentGetter(javaName, nativeName, layout);
         } else {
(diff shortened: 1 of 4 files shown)


src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/StructBuilder.java
```diff
@@ -136,6 +136,14 @@ private String findAnonymousStructName(GroupLayout parentLayout, GroupLayout lay
 
     @Override
     public void addVar(String javaName, String nativeName, MemoryLayout layout, Class<?> type) {
+        try {
+            structLayout.byteOffset(elementPaths(nativeName));
```
mcimadamore, line 140: Now that I think about it - calling byteSize on the structLayout should be enough. I also think that, if we do this, we can remove the logic in OutputFactory which checks whether size is available.

sundararajana: hmm.. calling byteSize on structLayout elsewhere? i.e., before addVar being called?

sundararajana: hmm.. byteSize() does not result in exception being thrown (the previous struct field has undermined size because of unsized array at the end. but current offending field has proper size). Am I missing something?

mcimadamore: I mean byteSize on `structLayout`. If there is any variable-length array, structLayout (which is supposedly storing the _enclosing_ layout of all nested anon structs inside it as well) should have its size undefined too.

mcimadamore (approved):
After conducting some more experiments, it seems like calling bitsSize casts too wide of a net - as that would fail also for fields *preceding* the VLA.

PR #468, opened 2021-03-10
8261511: jextract fails to handle name clashes between unrelated constants

sundararajana:
constants class is nested properly. piggybacking to fix long whitespace string in pretty printer

---------
### Progress

### Issue
 * [JDK-8261511](https://bugs.openjdk.java.net/browse/JDK-8261511): jextract fails to handle name clashes between unrelated constants

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/468/head:pull/468`
`$ git checkout pull/468`

### DIFF
--- a/src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/NestedClassBuilder.java
+++ b/src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/NestedClassBuilder.java
@@ -86,10 +86,11 @@ public class NestedClassBuilder extends JavaSourceBuilder {
 
     @Override
     protected void emitWithConstantClass(String javaName, Consumer<ConstantBuilder> constantConsumer) {
-        if (kind == Kind.INTERFACE) {
-            enclosing.emitWithConstantClass(javaName, constantConsumer);
+        if (this instanceof ConstantBuilder cb) {
+            // use this class to emit constants
+            constantConsumer.accept(cb);
         } else {
-            super.emitWithConstantClass(javaName, constantConsumer);
+            enclosing.emitWithConstantClass(javaName, constantConsumer);
         }
     }
 }


src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/NestedClassBuilder.java
```diff
@@ -86,6 +86,10 @@ protected void emitImportSection() {
 
     @Override
     protected void emitWithConstantClass(String javaName, Consumer<ConstantBuilder> constantConsumer) {
-        enclosing.emitWithConstantClass(javaName, constantConsumer);
+        if (kind == Kind.INTERFACE) {
```
mcimadamore, line 89: Not super sure about this fix.

If I undrerstand correctly, this will generate a nested constant class inside the enclosing struct, right (rather than going back to the enclosing header class). This is ok for clashes between header functions and function pointers, but how about cases like:

```
struct Foo {
    int (*sum)(int x, int y);
    struct Bar {
      int (*sum)(float x, float y);
    } bar;
};
```

I tried this, and it seems like we're probably ok, there are only two possible cases here:

* the innermost struct is anon (e.g. doesn't have a name, and doesn't introduce a new struct field) - in this case it is an error if the anon struct declares a field that clashes with some enclosing field

* the innermost struct is not anon (e.g. has a name, or introduces a struct field name); in this case a nested struct class is generated, and constants are added in there instead.

So, I believe the logic works, but I think it generates a redundant constant class (note that the struct class already acts as a constant holder - in fact, StructBuilder extends ConstantBuilder).

It would be nice if, instead, this method would do something like:

```
public void emitWithConstantClass(String javaName, Consumer<ConstantBuilder> constantConsumer) {
if (this instanceof ConstantBuilder cb) {
   // use this class to emit constants
   constantConsumer.accept(cb);
   
} else {
    enclosing.emitWithConstantClass(javaName, constantConsumer);
}
}
```
By doing so, we should skip the generation of the extra constant class?

PR #471, opened 2021-03-19
8263459: Add better support for restricted methods

sundararajana:
--enable-native-access command line option

---------
### Progress

### Issue
 * [JDK-8263459](https://bugs.openjdk.java.net/browse/JDK-8263459): Add better support for restricted methods

### Download
To checkout this PR locally:
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/471/head:pull/471`
`$ git checkout pull/471`

`$ git checkout pull/471`
`$ git pull https://git.openjdk.java.net/panama-foreign pull/471/head`


src/java.base/share/classes/jdk/internal/module/IllegalNativeAccessChecker.java
```diff
@@ -0,0 +1,100 @@
+/*
+ *  Copyright (c) 2019, Oracle and/or its affiliates. All rights reserved.
```
mcimadamore, line 2: Copyright year seems off ?

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/Utils.java
```diff
@@ -107,22 +106,6 @@ private static MemorySegmentProxy filterSegment(MemorySegment segment) {
         return (AbstractMemorySegmentImpl)segment;
     }
 
-    public static void checkRestrictedAccess(String method) {
-        switch (foreignRestrictedAccess) {
```
mcimadamore, line 111: is this "foreignRestrictedAccess" field removed in the patch?

mcimadamore (approved):
I've already reviewed the code separately and provided some commments which have been addressed here. It would be great if somebody from the VM team could go over the VM changes. Added some minor comments.

mcimadamore:
For the records, the approach implemented in this patch works as follows: when a Panama native restricted method is called, we will no require that the caller module (which could be ALL-UNNAMED) has been whitelisted accordingly, on the command line, using this flag:

```
--enable-native-access=<modulename>
```

In other words, this work replaces the hacky `-Dforeign.restricted=permit` property with a system which works in tandem with the Java module system.

While the support is relatively minimal, we have spent some cycles to add a new jlink plugin which will search a module for all calls to restricted methods and, if any is found, will mark the module-info.class as "native" with a special classfile attribute.

At startup, if the runtime sees that the module graph contains one or more modules that are marked as "native" but whose command line flag has not been set, an error will occur immediately - otherwise, w/o jlink, an error will only occur when a restricted method is called.

Changes to javadoc and jshell will come as a separate patch - the idea is to for for native restricted methods something similar as to what has been done in javadoc for preview API methods.

src/hotspot/share/classfile/modules.cpp
```diff
@@ -780,3 +780,32 @@ void Modules::add_module_exports_to_all_unnamed(Handle module, jstring package_n
...
 }
+
+// Mark module as native
+void Modules::add_module_enable_native_access(Handle module, TRAPS) {
+  if (module == NULL) {
+    THROW_MSG(vmSymbols::java_lang_NullPointerException(),
+              "module is null");
+  }
+  ModuleEntry* module_entry = get_module_entry(module, CHECK);
+  if (module_entry == NULL) {
```
dholmes-ora, line 791: I can't see how you can get NULL out of `get_module_entry`.

src/hotspot/share/classfile/modules.cpp
```diff
@@ -780,3 +780,32 @@ void Modules::add_module_exports_to_all_unnamed(Handle module, jstring package_n
...
+    Symbol* name = module_entry->name();
+    log_debug(module)("add_module_enable_native_access(): module"
+                      " %s is marked as a native module",
+                       name == NULL ? UNNAMED_MODULE : name->as_C_string());
+  }
+}
+
+volatile bool Modules::_native_access_all_unnamed = false;
+
+void Modules::enable_native_access_all_unnamed(TRAPS) {
```
dholmes-ora, line 809: This method doesn't need the TRAPS argument.

src/hotspot/share/interpreter/linkResolver.cpp
```diff
@@ -806,6 +823,8 @@ Method* LinkResolver::resolve_method(const LinkInfo& link_info,
   if (link_info.check_loader_constraints()) {
     // check loader constraints
     check_method_loader_constraints(link_info, resolved_method, "method", CHECK_NULL);
+
+    check_restricted_method(link_info, resolved_method, CHECK_NULL);
```
dholmes-ora, line 827: Not at all clear why this is guarded by `check_loader_constraints`. ?? Otherwise the placement of the check seems fine.

src/hotspot/share/interpreter/linkResolver.cpp
```diff
@@ -902,6 +921,8 @@ Method* LinkResolver::resolve_interface_method(const LinkInfo& link_info, Byteco
   }
   if (link_info.check_loader_constraints()) {
     check_method_loader_constraints(link_info, resolved_method, "interface method", CHECK_NULL);
+
+    check_restricted_method(link_info, resolved_method, CHECK_NULL);
```
dholmes-ora, line 925: Same comment as in `resolve_method`.

src/hotspot/share/prims/jvm.cpp
```diff
@@ -1099,6 +1099,15 @@ JVM_ENTRY(void, JVM_AddModuleExportsToAllUnnamed(JNIEnv *env, jobject from_modul
   Modules::add_module_exports_to_all_unnamed(h_from_module, package, CHECK);
 JVM_END
 
+JVM_ENTRY(void, JVM_AddModuleEnableNativeAccess(JNIEnv *env, jobject from_module))
+  Handle h_module (THREAD, JNIHandles::resolve(from_module));
+  Modules::add_module_enable_native_access(h_module, CHECK);
+JVM_END
+
+JVM_ENTRY(void, JVM_EnableNativeAccessAllUnnamed(JNIEnv *env))
+  Modules::enable_native_access_all_unnamed(CHECK);
```
dholmes-ora, line 1108: As no exceptions can actually be thrown this could probably be a JVM_LEAF method

mcimadamore:
> _Mailing list message from [Chris Vest](mailto:mr.chrisvest@gmail.com) on [panama-dev](mailto:panama-dev@openjdk.java.net):_
> 
> The PR uses a mix of terminology for what sounds like the same concept;
> restricted, native, restricted_native.
> Are there subtle differences between these or are they the same thing?
> 
> On Fri, 19 Mar 2021 at 14:39, Athijegannathan Sundararajan <
> sundar at openjdk.java.net> wrote:

The concepts are meant to be the same, although the terminology used is not consistent, you are right. I think that, since the option is "enable-native-access" the various bits of code should probably reflect that a bit more; but note that there's a risk of clashing with regular JNI native methods here, for which this patch does nothing.

Note also that, in a followup patch, the "restricted" prefix from method declarations will be removed (as javadoc will render these methods in a special way anyway).

src/hotspot/share/interpreter/linkResolver.cpp
```diff
@@ -823,9 +823,8 @@ Method* LinkResolver::resolve_method(const LinkInfo& link_info,
   if (link_info.check_loader_constraints()) {
     // check loader constraints
     check_method_loader_constraints(link_info, resolved_method, "method", CHECK_NULL);
-
-    check_restricted_method(link_info, resolved_method, CHECK_NULL);
   }
+  check_restricted_method(link_info, resolved_method, CHECK_NULL);
```
mcimadamore, line 827: Can we please rename this as "check_native_access" to make terminology consistent

make/jdk/src/classes/build/tools/panama/RestrictedMethodsFinder.java
```diff
@@ -0,0 +1,178 @@
...
+import javax.tools.Diagnostic;
+import javax.lang.model.type.ArrayType;
+import javax.lang.model.type.DeclaredType;
+import javax.lang.model.type.ExecutableType;
+import javax.lang.model.type.TypeKind;
+import javax.lang.model.type.TypeMirror;
+import javax.lang.model.util.Types;
+import static java.nio.file.StandardOpenOption.*;
+
+public class RestrictedMethodsFinder implements Plugin {
```
mcimadamore, line 57: Rename to "NativeAccessMethodFinder" (and fix up naming of associated build parameters)

src/java.base/share/classes/jdk/internal/vm/annotation/RestrictedNative.java
```diff
@@ -0,0 +1,39 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * FIXME
+ */
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface RestrictedNative {
```
mcimadamore, line 38: Rename to @NativeAccess

src/jdk.jlink/share/classes/jdk/tools/jlink/internal/plugins/RestrictedNativeMarkerPlugin.java
```diff
@@ -0,0 +1,283 @@
...
+
+import jdk.tools.jlink.plugin.PluginException;
+import jdk.tools.jlink.plugin.ResourcePool;
+import jdk.tools.jlink.plugin.ResourcePoolBuilder;
+import jdk.tools.jlink.plugin.ResourcePoolEntry;
+
+/**
+ * Jlink plugin makes modules that use RestrictedNative methods.
+ */
+public final class RestrictedNativeMarkerPlugin extends AbstractPlugin {
```
mcimadamore, line 61: Rename to NativeAccessMarkerPlugin (and fixup related build variables)

src/java.base/share/classes/jdk/internal/module/SystemModules.java
```diff
@@ -95,4 +95,9 @@
      * module name, the value is the set of exported packages to open.
      */
     Map<String, Set<String>> exportedPackagesToOpen();
+
+    /**
+     * Returns the restricted native flags for modules
+     */
+    boolean[] restrictedNativeModules();
```
mcimadamore, line 102: rename to `nativeAccessModules`

src/java.base/share/classes/jdk/internal/module/ModuleReferenceImpl.java
```diff
@@ -132,6 +136,11 @@ public ModuleResolution moduleResolution() {
         return moduleResolution;
     }
 
+    public boolean usesRestrictedNative() {
+        return usesRestrictedNative;
```
mcimadamore, line 140: rename to `nativeAccessEnabled`, or `usesNativeAccess`

src/java.base/share/classes/jdk/internal/module/ClassFileConstants.java
```diff
@@ -42,6 +42,7 @@ private ClassFileConstants() { }
     public static final String MODULE_TARGET      = "ModuleTarget";
     public static final String MODULE_HASHES      = "ModuleHashes";
     public static final String MODULE_RESOLUTION  = "ModuleResolution";
+    public static final String MODULE_RESTRICTED_NATIVE = "ModuleRestrictedNative";
```
mcimadamore, line 45: Rename to ModuleNativeAccess/MODULE_NATIVE_ACCESS

src/java.base/share/classes/java/lang/reflect/Method.java
```diff
@@ -491,6 +499,25 @@ void specificToGenericStringHeader(StringBuilder sb) {
         sb.append(getName());
     }
 
+    private boolean isRestrictedNative() {
+        return (flags & RESTRICTED_NATIVE) != 0;
+    }
+
+    private final void checkRestrictedNative(Class<?> caller) throws IllegalAccessException {
```
mcimadamore, line 506: isNativeAccess/checkNativeAccess

mcimadamore (commented):
I've suggested a bunch of renamings. I probably missed some, but in general let's make sure that we use "nativeAccess" everywhere and that the "restricted" word is never present in the implementation (but we'll keep that in API method names, for now).

make/Images.gmk
```diff
@@ -70,12 +70,15 @@ JLINK_ORDER_RESOURCES += \
...
 
+JLINK_RESTRICTED_NATIVE := --restricted-native-marker=@$(SUPPORT_OUTPUTDIR)/link_opt/restricted_native_methods.txt
+
 JLINK_TOOL := $(JLINK) -J-Djlink.debug=true \
     --module-path $(IMAGES_OUTPUTDIR)/jmods \
     --endian $(OPENJDK_TARGET_CPU_ENDIAN) \
     --release-info $(BASE_RELEASE_FILE) \
     --order-resources=$(call CommaList, $(JLINK_ORDER_RESOURCES)) \
     --dedup-legal-notices=error-if-not-same-content \
+    $(JLINK_RESTRICTED_NATIVE) \
```
AlanBateman, line 81: How does this impact jlink usage outside of the JDK?

sundararajana: The native methods text goes as "conf" file of jlink tool. That plugin is always on (and loads conf file). Only during jdk build we need to pass the location of the the native methods text file as an argument.

sundararajana:
A new PR with reduced changes will be raised.

PR #472, opened 2021-03-23
8263459: Add better support for restricted methods

sundararajana:
When a Panama native restricted method is called, we will no require that the caller module (which could be ALL-UNNAMED) has been whitelisted accordingly, on the command line, using this flag:

--enable-native-access=<modulename>

In other words, this work replaces the hacky -Dforeign.restricted=permit property with a system which works in tandem with the Java module system.

The current PR includes subset of the earlier PR 

https://github.com/openjdk/panama-foreign/pull/471

The 'fail fast' aspect (jlink plugin to mark "NativeAccess" calling modules and issuing error if those modules were not mentioned in --enable-native-access) has been removed in this new PR.

---------
### Progress

### Issue
 * [JDK-8263459](https://bugs.openjdk.java.net/browse/JDK-8263459): Add better support for restricted methods

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/472/head:pull/472` \
`$ git checkout pull/472`

`$ git checkout pull/472` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/472/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 472`

`$ git pr show -t 472`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/472.diff">https://git.openjdk.java.net/panama-foreign/pull/472.diff</a>

</details>


src/hotspot/share/classfile/classFileParser.cpp
```diff
@@ -2128,6 +2129,10 @@ AnnotationCollector::annotation_index(const ClassLoaderData* loader_data,
       if (!privileged)              break;  // only allow in privileged code
       return _method_Scoped;
     }
+    case VM_SYMBOL_ENUM_NAME(jdk_internal_vm_annotation_NativeAccess_signature): {
```
mcimadamore, line 2132: I don't think this stuff is needed anymore? The annotation should just be there for javadoc puprposes.

mcimadamore (commented):
Looks good - but there's still some VM code left which I don't think is needed (if we go for this approach).

test/jdk/java/foreign/TestNoForeignUnsafeOverride.java
```diff
@@ -21,23 +21,14 @@
...
-    public void testUnsafeAccess() {
-        MemorySegment.ofNativeRestricted();
-    }
+public class PanamaMain {
+   public static void main(String[] args) {
+       System.out.println("Trying to get CLinker");
+       CLinker.getInstance();
+       System.out.println("Got CLinker");
+   }
 }
```
JornVernee, line 34: This file doesn't seem used? You already have the PanamaMain in org.openjdk.foreigntest.PanamaMain as well

JornVernee (commented):
Empty file was added: test/jdk/java/foreign/enablenativeaccess/panama_module/org/openjdk/foreigntest/PanamaMainInvoke.java I guess that should contain a test?

src/hotspot/share/classfile/classFileParser.cpp
sundararajana: Yes. pushed to remove most hotspot changes

src/hotspot/share/classfile/javaClasses.cpp
```diff
@@ -2917,7 +2917,7 @@ void java_lang_LiveStackFrameInfo::set_mode(oop element, int value) {
 }
 
 
-// java_lang_reflect_AccessibleObject
+// java_lang_AccessibleObject
```
JornVernee: Spurious change?

sundararajana: No. I was like that before I made any hotspot change. I'd like to leave what it was -- because we're making only minimal hotspot changes.

JornVernee: Ah, nvm, it only shows up in the diff of the commit.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryAddressImpl.java
```diff
@@ -94,9 +97,11 @@ public String toString() {
     }
 
     @Override
+    @CallerSensitive
+    @NativeAccess
```
mcimadamore, line 101: I don't think we need @NativeAccess here - this is not an API method

sundararajana: removed. Added @NativeAccess for an API method (missing earlier)

mcimadamore (commented):
One last change occured to me. I'd like the downcall/upcall methods in the ABI to be marked as @NativeAccess, instead of getInstance (since obtaining an instance is totally safe).

mcimadamore (approved):
Looks good - on a second thoughts, let's separate between adding the new mechanics for restricted checks and changes to which API is restricted or not (and whether to drop the "Restricted" suffixes.

PR #474, opened 2021-03-24
8263018: Improve API for lifecycle of native resources

mcimadamore:
This contains the jextract changes required to support the new changes to the foreign memory access API.

The main change is that now NativeScope is co-generated by jextract (although, moving forward, I'm not sure how much value it adds, but that's for another day).

I had to fix the extracted libclang in few places (probably better to re-generate this code once this is pushed).

Tests needed changes to fix imports so that they pick up the generated NativeScope class.

---------
### Progress

### Issue
 * [JDK-8263018](https://bugs.openjdk.java.net/browse/JDK-8263018): Improve API for lifecycle of native resources ⚠️ Issue is not open.

### Reviewers
 * [Jorn Vernee](https://openjdk.java.net/census#jvernee) (@JornVernee - Committer) ⚠️ Review applies to 65f7de7bf70ef92d37d5999fce232ddd8037aa3a

### Download
To checkout this PR locally:
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/474/head:pull/474`
`$ git checkout pull/474`

`$ git checkout pull/474`
`$ git pull https://git.openjdk.java.net/panama-foreign pull/474/head`


src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/HeaderFileBuilder.java
```diff
@@ -122,61 +130,115 @@ private void emitFunctionWrapper(Constant mhConstant, String javaName, String na
...
         indent();
         append(MEMBER_MODS + " ");
-        append(functionInfo.methodType().returnType().getSimpleName() + " " + javaName + " (");
+        MethodType declType = functionInfo.methodType();
+        List<String> paramNames = functionInfo.parameterNames().get();
+        if (functionInfo.methodType().returnType().equals(MemorySegment.class)) {
+            // needs allocator parameter
+            declType = declType.insertParameterTypes(0, SegmentAllocator.class);
+            paramNames = new ArrayList<>(paramNames);
+            paramNames.add("allocator");
```
JornVernee, line 139: Shouldn't this be added as the first element rather than the last?

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/HeaderFileBuilder.java
```diff
@@ -89,6 +92,11 @@ public void addFunction(String javaName, String nativeName, FunctionInfo functio
             Constant mhConstant = constantBuilder.addMethodHandle(javaName, nativeName, functionInfo, false)
                     .emitGetter(this, MEMBER_MODS, Constant.QUALIFIED_NAME, nativeName);
             emitFunctionWrapper(mhConstant, javaName, nativeName, functionInfo);
+            if (functionInfo.methodType().returnType().equals(MemorySegment.class)) {
+                // emit scoped overload
+                emitFunctionWrapperNoAllocatorOverload(javaName, functionInfo);
+                emitFunctionWrapperScopedOverload(javaName, functionInfo);
+            }
```
JornVernee, line 99: With the emitFunctionWrapper call above this, don't we get a publicly accessible function wrapper that will always throw a WrongMethodTypeException because it's missing an allocator argument? Maybe it should be conditionally made private? (since the other 2 functions reference it, it's still needed).

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/OutputFactory.java
```diff
@@ -149,6 +150,13 @@ private String getRuntimeHelperSource() throws URISyntaxException, IOException {
                                 .replace("${C_LANG}", C_LANG_CONSTANTS_HOLDER);
     }
 
+    private String getNativeScopeSource() throws URISyntaxException, IOException {
+        URL runtimeHelper = OutputFactory.class.getResource("resources/NativeScope.java.template");
+        return (pkgName.isEmpty()? "" : "package " + pkgName + ";\n") +
+                String.join("\n", Files.readAllLines(Paths.get(runtimeHelper.toURI())))
+                        .replace("${C_LANG}", C_LANG_CONSTANTS_HOLDER);
+    }
+
```
JornVernee, line 159: This could be partially merged with `getRuntimeHelperSource`

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/TypedefBuilder.java
```diff
@@ -43,37 +41,6 @@ String superClass() {
...
-        append(MEMBER_MODS);
-        append(" MemorySegment allocatePointer(NativeScope scope) {\n");
-        incrAlign();
-        indent();
-        append("return scope.allocate(C_POINTER);\n");
-        decrAlign();
-        indent();
-        append("}\n");
-        decrAlign();
-    }
```
JornVernee, line 78: I don't get why this is removed. Was it not being used?

test/jdk/tools/jextract/test8244959/Test8244959.java
```diff
@@ -50,12 +50,11 @@
...
-                (byte) 1, 'b', -1.25f, 5.5d, -200L, Long.MAX_VALUE, (byte) -2, (short) 2, 3, (short) -4, 5L, 'a');
-            String str = toJavaString(s);
-            assertEquals(str, "1 b -1.25 5.50 -200 " + Long.MAX_VALUE + " -2 2 3 -4 5 a");
-        }
+        MemorySegment s = MemorySegment.allocateNative(1024);
+        my_sprintf(s,
+            toCString("%hhd %c %.2f %.2f %lld %lld %d %hd %d %d %lld %c"), 12,
+            (byte) 1, 'b', -1.25f, 5.5d, -200L, Long.MAX_VALUE, (byte) -2, (short) 2, 3, (short) -4, 5L, 'a');
+        String str = toJavaString(s);
+        assertEquals(str, "1 b -1.25 5.50 -200 " + Long.MAX_VALUE + " -2 2 3 -4 5 a");
```
JornVernee, line 58: I think using managed segments here could be problematic as well.

test/jdk/tools/jextract/test8246341/LibTest8246341Test.java
```diff
@@ -53,17 +52,16 @@
...
             assertEquals(argc, 4);
             assertEquals(toJavaStringRestricted(MemoryAccess.getAddressAtIndex(addr, 0)), "java");
             assertEquals(toJavaStringRestricted(MemoryAccess.getAddressAtIndex(addr, 1)), "python");
             assertEquals(toJavaStringRestricted(MemoryAccess.getAddressAtIndex(addr, 2)), "javascript");
             assertEquals(toJavaStringRestricted(MemoryAccess.getAddressAtIndex(addr, 3)), "c++");
-        })) {
-            func(callback);
-        }
+        });
+        func(callback);
```
JornVernee, line 64: Same here; managed segment but this is projecting to a MemoryAdress AFAIK.

test/jdk/tools/jextract/testFunctionPointer/LibFuncPtrTest.java
```diff
@@ -43,8 +43,7 @@
 public class LibFuncPtrTest {
     @Test
     public void test() {
-        try (var handle = func$f.allocate(x -> x*x)) {
-            assertEquals(func(handle, 35), 35 * 35 + 35);
-        } //deallocate
+        var handle = func$f.allocate(x -> x*x);
+        assertEquals(func(handle, 35), 35 * 35 + 35);
```
JornVernee, line 47: Again here, managed segment projected to address

JornVernee (approved):
I think there are some issues, particularly wrt to using managed memory segments that are projected down to addresses. I don't think managed segments should be used in those cases. But, we can get the branch building first and address later maybe.

src/jdk.incubator.jextract/share/classes/jdk/internal/clang/TranslationUnit.java
```diff
@@ -63,11 +65,10 @@ public Cursor getCursor() {
     }
 
     public final void save(Path path) throws TranslationUnitSaveException {
-        try (MemorySegment pathStr = CLinker.toCString(path.toAbsolutePath().toString())) {
-            SaveError res = SaveError.valueOf(Index_h.clang_saveTranslationUnit(tu, pathStr, 0));
-            if (res != SaveError.None) {
-                throw new TranslationUnitSaveException(path, res);
-            }
+        MemorySegment pathStr = CLinker.toCString(path.toAbsolutePath().toString());
+        SaveError res = SaveError.valueOf(Index_h.clang_saveTranslationUnit(tu, pathStr, 0));
```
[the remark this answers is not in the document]
mcimadamore: true

src/jdk.incubator.jextract/share/classes/jdk/internal/clang/Type.java
```diff
@@ -109,9 +111,8 @@ public long getNumberOfElements() {
 
     // Struct/RecordType
     private long getOffsetOf0(String fieldName) {
-        try (MemorySegment cfname = CLinker.toCString(fieldName)) {
-            return Index_h.clang_Type_getOffsetOf(type, cfname);
-        }
+        MemorySegment cfname = CLinker.toCString(fieldName);
+        return Index_h.clang_Type_getOffsetOf(type, cfname);
```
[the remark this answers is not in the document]
mcimadamore: true

src/jdk.incubator.jextract/share/classes/jdk/internal/clang/libclang/RuntimeHelper.java
```diff
@@ -78,17 +79,28 @@ private RuntimeHelper() {}
...
+    final static SegmentAllocator DEFAULT_ALLOCATOR = MemorySegment::allocateNative;
+
     static final MethodHandle downcallHandle(LibraryLookup[] LIBRARIES, String name, String desc, FunctionDescriptor fdesc, boolean variadic) {
         return lookup(LIBRARIES, name).map(
                 addr -> {
                     MethodType mt = MethodType.fromMethodDescriptorString(desc, LOADER);
                     return variadic ?
                         VarargsInvoker.make(addr, mt, fdesc) :
-                        LINKER.downcallHandle(addr, mt, fdesc);
+                        LINKER.downcallHandle(addr, DEFAULT_ALLOCATOR, mt, fdesc);
```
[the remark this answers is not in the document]
mcimadamore: Uhm.. in the new generated code this is not a problem - e.g. this is not what RuntimeHelper looks like in the new world. This is mostly a way to get libclang up and running. We could of course do what you suggest, I was unsure whether now is the time to do it (e.g. won't we regenerated libclang anyway?)

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/FunctionalInterfaceBuilder.java
```diff
@@ -103,16 +104,13 @@ private void emitFunctionalRestrictedFactory() {
...
-            indent();
-            append("public " + fiType.returnType().getName() + " apply(");
+            append("return (");
             String delim = "";
             for (int i = 0 ; i < fiType.parameterCount(); i++) {
                 append(delim + fiType.parameterType(i).getName() + " x" + i);
                 delim = ", ";
             }
-            append(") {\n");
+            append(") -> {\n");
```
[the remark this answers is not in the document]
mcimadamore: yes - I changed it since I was there

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/HeaderFileBuilder.java
mcimadamore: uhm, yes

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/HeaderFileBuilder.java
mcimadamore: I see what you mean.

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/TypedefBuilder.java
mcimadamore: It was a leftover from when we were using the @C annotation. These methods don't do anything special other than allocating a C_POINTER

src/jdk.incubator.jextract/share/classes/jdk/internal/clang/libclang/RuntimeHelper.java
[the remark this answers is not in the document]
JornVernee: Ah, ok. That makes sense.

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/TypedefBuilder.java
JornVernee: Ok. (It doesn't look all that interesting either :) )

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/FunctionalInterfaceBuilder.java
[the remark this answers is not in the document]
JornVernee: Ok, LGTM.

src/jdk.incubator.jextract/share/classes/jdk/internal/clang/libclang/RuntimeHelper.java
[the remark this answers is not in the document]
mcimadamore: This seems to be a list of "problematic" functions in a relatively simple jextract call:
```
at jdk.incubator.jextract/jdk.internal.clang.libclang.Index_h.clang_getTranslationUnitCursor(Index_h.java:3525)

at jdk.incubator.jextract/jdk.internal.clang.libclang.Index_h.clang_getCursorLocation(Index_h.java:4452)

at jdk.incubator.jextract/jdk.internal.clang.libclang.Index_h.clang_getNullLocation(Index_h.java:821)

at jdk.incubator.jextract/jdk.internal.clang.libclang.Index_h.clang_getFileName(Index_h.java:538)

at jdk.incubator.jextract/jdk.internal.clang.libclang.Index_h.clang_getCursorType(Index_h.java:4934)

at jdk.incubator.jextract/jdk.internal.clang.libclang.Index_h.clang_getTypeSpelling(Index_h.java:4945)

at jdk.incubator.jextract/jdk.internal.clang.libclang.Index_h.clang_getCursorSpelling(Index_h.java:7181)

at jdk.incubator.jextract/jdk.internal.clang.libclang.Index_h.clang_getTypeDeclaration(Index_h.java:5604)

at jdk.incubator.jextract/jdk.internal.clang.libclang.Index_h.clang_getCursorDefinition(Index_h.java:7559)

at jdk.incubator.jextract/jdk.internal.clang.libclang.Index_h.clang_Cursor_getArgument(Index_h.java:5122)

at jdk.incubator.jextract/jdk.internal.clang.libclang.Index_h.clang_getArgType(Index_h.java:5868)

at jdk.incubator.jextract/jdk.internal.clang.libclang.Index_h.clang_getResultType(Index_h.java:5751)

at jdk.incubator.jextract/jdk.internal.clang.libclang.Index_h.clang_getCanonicalType(Index_h.java:5310)
```

Now, all of these seem ok. When we create a new Cursor, or a new Type, we always wrap a segment, so the segment is kept alive. Same is true for SourceLocation. For CXString, we have a method that takes a CXString and gives us back a j.l.String, so again, the string segment will be alive at least until it's converted to a java String, after which we no longer care.

Honestly, in some (all?) of these cases it's not super clear to me that _deallocation_ was happening in the previous code?

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/HeaderFileBuilder.java
mcimadamore: Actually, I think the code is correct, as the generated "main" method will either have no allocator argument (in case return is not a struct) or will have allocator parameter (in case return is struct). The issue with the code is that, when no  allocator argument is present, we should pass DEFAULT_ALLOCATOR to the underlying MH.

mcimadamore: Scratch that - we don't need a new allocator parameter when struct is not returned as the underlying MH doesn't take an allocator.

I think the code is fine as is.

test/jdk/tools/jextract/test8244959/Test8244959.java
```diff
@@ -50,11 +51,13 @@
...
-        MemorySegment s = MemorySegment.allocateNative(1024);
-        my_sprintf(s,
-            toCString("%hhd %c %.2f %.2f %lld %lld %d %hd %d %d %lld %c"), 12,
-            (byte) 1, 'b', -1.25f, 5.5d, -200L, Long.MAX_VALUE, (byte) -2, (short) 2, 3, (short) -4, 5L, 'a');
-        String str = toJavaString(s);
-        assertEquals(str, "1 b -1.25 5.50 -200 " + Long.MAX_VALUE + " -2 2 3 -4 5 a");
+        try (ResourceScope scope = ResourceScope.ofConfined()) {
+            MemorySegment s = MemorySegment.allocateNative(1024, scope);
+            my_sprintf(s,
+                    toCString("%hhd %c %.2f %.2f %lld %lld %d %hd %d %d %lld %c"), 12,
```
JornVernee, line 57: `toCString` should also use the scope

JornVernee (approved):

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/HeaderFileBuilder.java
JornVernee: Ah, right. I see now we get:
```
MS func(SegmentAllocator, ...)
MS func(...) // uses default scope
MS func(ResourceScope, ...)
```

I misread this is the first overload not taking an allocator. What you have looks good.

src/jdk.incubator.jextract/share/classes/jdk/internal/clang/libclang/RuntimeHelper.java
```diff
@@ -78,17 +79,28 @@ private RuntimeHelper() {}
...
+                }).orElse(null);
+    }
+
+    static final MethodHandle downcallHandle(LibraryLookup[] LIBRARIES, String name, String desc, FunctionDescriptor fdesc, boolean variadic, NativeScope scope) {
```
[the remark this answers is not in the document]
JornVernee: What do you think of this? This method still seems unused in the latest version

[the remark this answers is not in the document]
mcimadamore: This is generated code, I don't think it's worth making changes on code that will be overwritten when we regenerate this. There are loads of methods that are unused in the libclang port (because jextract doesn't use all libclang functions).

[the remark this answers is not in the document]
JornVernee: Ah ok, I thought this was added by hand. Sorry

PR #475, opened 2021-03-24
Update memory access and linker documents

mcimadamore:
This patch updates the two documents describing usage of the foreign memory access and foreign linker API.

---------
### Progress

### Reviewers
 * [Jorn Vernee](https://openjdk.java.net/census#jvernee) (@JornVernee - Committer) ⚠️ Review applies to ae551e8aee7fa910cd83e5808f4d54e2405344df

### Download
To checkout this PR locally:
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/475/head:pull/475`
`$ git checkout pull/475`

`$ git checkout pull/475`
`$ git pull https://git.openjdk.java.net/panama-foreign pull/475/head`


doc/panama_ffi.md
```diff
@@ -47,6 +46,63 @@ Of course, since accessing the entire native heap is inherently *unsafe*, access
...
+foo(42, &size);
+```
+
+Here the function `foo` takes an output parameter, a pointer to an `int` variable. Unfortunately (and we have seen part of that pain in the `qsort` example above), implementing this idiom in Panama is not straightforward:
+
+```java
+MethodHandle foo = ...
+MemorySegment size = MemorySegment.allocateNative(C_INT);
+MemoryAccess.setInt(size, 5);
+foo.invokeExact(42, size);
```
JornVernee, line 66: I think this should call `size.address()` to be equivalent with the C code. Also, reachability of the segment is problematic in that case, so maybe this should switch to using an explicit resource scope.

doc/panama_ffi.md
```diff
@@ -47,6 +46,63 @@ Of course, since accessing the entire native heap is inherently *unsafe*, access
...
+There are a number of issues with the above code snippet:
+
+* compared to the C code, it is very verbose
+* allocation is very slow compared to C; allocating the `size` variable now takes a full `malloc`, while in C the variable was simply stack-allocated
+* we end up with a standalone segment with its own temporal bounds, which will have to be *closed* later, to avoid leaks
+
+To address these problems, Panama provides a `SegmentAllocator` abstraction, a functional interface which provides many useful operation to allocate commonly used values. For instance, the above code can be rewritten as follows:
+
+```java
+MemorySegment size = SegmentAllocator.ofDefault().allocate(C_INT, 5);
```
JornVernee, line 78: Same here

doc/panama_ffi.md
```diff
@@ -137,16 +195,30 @@ Note that, since the function `strlen` is part of the standard C library, which
 One we have obtained the downcall native method handle, we can just use it as any other method handle:
 
 ```java
-long len = strlen.invokeExact(CLinker.toCString("Hello").address()) // 5
+long len = strlen.invokeExact(CLinker.toCString("Hello").address()); // 5
```
JornVernee, line 198: Segment reachability here is problematic as well.

doc/panama_ffi.md
```diff
@@ -137,16 +195,30 @@ Note that, since the function `strlen` is part of the standard C library, which
...
+
+```java
+MethodHandle strlen_virtual = CLinker.getInstance().downcallHandle(
+		MethodType.methodType(long.class, MemoryAddress.class),
+        FunctionDescriptor.of(C_LONG, C_POINTER)
+);
+
+long len = strlen_virtual.invokeExact(
+	LibraryLookup.ofDefault().lookup("strlen").get()
+	CLinker.toCString("Hello").address()
```
JornVernee, line 213: Here as well

doc/panama_ffi.md
```diff
@@ -199,77 +271,17 @@ MemorySegment comparFunc = CLinker.getInstance().upcallStub(
...
 ```
 
+When no resource scope is specified (as in the above case), the upcall stub segment will be associated with the *default scope* - a non-closeable scope which does not support deterministic deallocation. This means that the upcall stub will be uninstalled when the upcall segment becomes *unreachable*. In cases where this is not desirable, the API also support associating a custom `ResourceScope` instance to the returned upcall segment.
+
 So, we finally have a memory segment — `comparFunc` — containing a stub that can be used to invoke our Java comparator function; this means we now have all we need to call the `qsort` downcall handle:
 
 ```java
-MemorySegment array = MemorySegment.allocateNative(4 * 10);
-array.copyFrom(MemorySegment.ofArray(new int[] { 0, 9, 3, 4, 6, 5, 1, 8, 2, 7 }));
+MemorySegment array = SegmentAllocator.ofDefault().allocateArray(new int[] { 0, 9, 3, 4, 6, 5, 1, 8, 2, 7 }));
```
JornVernee, line 279: Same here

doc/panama_ffi.md
```diff
@@ -72,7 +128,9 @@ At the core of Panama foreign function support we find the `CLinker` abstraction
 
 ```java
 interface CLinker {
+    MethodHandle downcallHandle(MethodType type, FunctionDescriptor function);
     MethodHandle downcallHandle(Addressable func, MethodType type, FunctionDescriptor function);
+    MemorySegment upcallStub(MethodHandle target, FunctionDescriptor function, ResourceScope scope);
     MemorySegment upcallStub(MethodHandle target, FunctionDescriptor function);
     ...
```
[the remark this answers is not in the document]
mcimadamore: yeah - since the examples never mention it, I didn't want to dive too much into that. I think I will simplify this further and just keep two methods (but then say that there are other overloads available).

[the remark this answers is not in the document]
JornVernee: Ok, sounds good

doc/panama_ffi.md
```diff
@@ -195,30 +195,34 @@ Note that, since the function `strlen` is part of the standard C library, which
...
 );
 
-long len = strlen_virtual.invokeExact(
-	LibraryLookup.ofDefault().lookup("strlen").get()
-	CLinker.toCString("Hello").address()
-); // 5
+try (ResourceScope scope = ResourceScope.ofConfined()) {
+    long len = strlen_virtual.invokeExact(
+        LibraryLookup.ofDefault().lookup("strlen").get() // address provided here!
+        CLinker.toCString("Hello").address()
```
JornVernee, line 217: The scope is not being used for the string here.

doc/panama_ffi.md
```diff
@@ -273,15 +279,24 @@ MemorySegment comparFunc = CLinker.getInstance().upcallStub(
...
+The above code creates  a memory segment — `comparFunc` — containing a stub that can be used to invoke our Java comparator function. The memory segment is associated with the provided resource scope instance; this means that the stub will be uninstalled when the resource scope is closed. It is also possible (not shown here) to create upcall stubs associated with the *default scope*, in which case the stub will be uninstalled when the upcall segment becomes *unreachable*.
+
+The snippet then creates an off-heap array from a Java array (using a `SegmemntAllocator`), which is then passed to the `qsort` handle, along with the comparator function we obtained from the foreign linker.  As a side-effect, after the call, the contents of the off-heap array will be sorted (as instructed by our comparator function, written in Java). We can than extract a new Java array from the segment, which contains the sorted elements. This is a more advanced example, but one that shows how powerful the native interop support provided by the foreign linker abstraction is, allowing full bidirectional interop support between Java and native.
```
JornVernee, line 300: s/SegmemntAllocator/SegmentAllocator

JornVernee (approved):

doc/panama_ffi.md
JornVernee: Looking at the full source code at the end, this also seems to mismatch with that.

doc/panama_ffi.md
mcimadamore: I've rewritten this section as I noted that it was referring to native method handles, which were not yet discussed. Instead I focussed on "creating a native array" as a way to demonstrate "clunkiness" of base memory access API.

mcimadamore: As for deterministic deallocation, I've changed all samples to just rely on it - and in the very first case (e.g. call to strlen) I explain that going from a segment to an address is a bit tricky.

PR #476, opened 2021-03-25
8264176: ResourceScope handles should keep parent scope reachable

mcimadamore:
While ResourceScope handles are handy to ensure that a scope is not closed too soon, they don't do much in order to help preventing the scope to become unreachable. This means that if a scope supports both deterministic and implicit deallocation, there might be cases where using a try-with-resources with a resource scope handle is not enough (as the scope might become unreachable inside the TWR block).

The solution is to add reachability fences in the various handle implementations, so that the parent scope is kept reachable.

I've refactored the implementation a bit, to remove the "closeable" flag - which was generally not used (only default scopes and global scope are non closeable). So, I have created a new class which <: SharedScope, `NonCloseableScope`. Both global scope and default scope are subclasses of that. This class recycles the handles it generates - e.g. only one is created, after which the same handle instance is returned.

An alternative would have been to make NonCloseableScope a subtype of both MemoryScope and ResourceScope.Handle - but that's difficult since both are AutoCloseable and feature a `close` method with different semantics.

---------
### Progress

### Issue
 * [JDK-8264176](https://bugs.openjdk.java.net/browse/JDK-8264176): ResourceScope handles should keep parent scope reachable

### Reviewers
 * [Jorn Vernee](https://openjdk.java.net/census#jvernee) (@JornVernee - Committer) ⚠️ Review applies to bc405e76d4f19086af34dea919b0d0d2239d34e9
 * [Paul Sandoz](https://openjdk.java.net/census#psandoz) (@PaulSandoz - Committer) ⚠️ Review applies to bc405e76d4f19086af34dea919b0d0d2239d34e9

### Download
To checkout this PR locally:
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/476/head:pull/476`
`$ git checkout pull/476`

`$ git checkout pull/476`
`$ git pull https://git.openjdk.java.net/panama-foreign pull/476/head`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/ConfinedScope.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/ConfinedScope.java
@@ -28,6 +28,7 @@ package jdk.internal.foreign;
 import jdk.internal.vm.annotation.ForceInline;
 
 import java.lang.ref.Cleaner;
+import java.lang.ref.Reference;
 
 /**
  * A confined scope, which features an owner thread. The liveness check features an additional
@@ -120,6 +121,7 @@ final class ConfinedScope extends MemoryScope {
             if (!released) {
                 released = true;
                 lockCount--;
+                Reference.reachabilityFence(ConfinedScope.this);
             }
         }
     }
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/ConfinedScope.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/ConfinedScope.java
@@ -41,8 +41,8 @@ final class ConfinedScope extends MemoryScope {
     private int lockCount = 0;
     private final Thread owner;
 
-    public ConfinedScope(Thread owner, Object ref, Cleaner cleaner, boolean closeable) {
-        super(ref, cleaner, closeable, new ConfinedResourceList());
+    public ConfinedScope(Thread owner, Object ref, Cleaner cleaner) {
+        super(ref, cleaner, new ConfinedResourceList());
         this.owner = owner;
     }
 
@@ -64,7 +64,6 @@ final class ConfinedScope extends MemoryScope {
     @Override
     public Handle acquire() {
         checkValidState();
-        if (!closeable) return DUMMY_LOCK;
         lockCount++;
         return new ConfinedHandle();
     }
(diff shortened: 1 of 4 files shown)


src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/ConfinedScope.java
```diff
@@ -121,6 +121,7 @@ public void close() {
             if (!released) {
                 released = true;
                 lockCount--;
+                Reference.reachabilityFence(ConfinedScope.this);
```
mcimadamore, line 124: Not sure this is needed - I added it as we were worried that possible optimization (e.g. scalarization) might break reachability (this class is an innner class and keeps a strong reference to its enclosing, after all).

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/SharedScope.java
```diff
@@ -184,6 +184,7 @@ public void close() {
                         throw new IllegalStateException("Already closed");
                     }
                 } while (!STATE.compareAndSet(jdk.internal.foreign.SharedScope.this, value, value - 1));
+                Reference.reachabilityFence(SharedScope.this);
```
mcimadamore, line 187: Not sure this is needed - I added it as we were worried that possible optimization (e.g. scalarization) might break reachability (this class is an innner class and keeps a strong reference to its enclosing, after all).

JornVernee (approved):
Looks good!

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryScope.java
```diff
@@ -191,15 +186,48 @@ public MemorySegment allocate(long bytesSize, long bytesAlignment) {
...
     }
 
-    public static MemoryScope GLOBAL = new SharedScope( null, null, false) {
+    /**
+     * A non-closeable, shared scope. Similar to a shared scope, but its {@link #close()} method throws unconditionally.
+     * In addition, non-closeable scopes feature a much simpler scheme for generating resource scope handles, where
+     * the same per-scope handle is shared across multiple calls to {@link #acquire()}. In fact, for non-closeable
+     * scopes, it is sufficient for resource scope handles to keep a strong reference to their scopes, to prevent closure.
+     */
+    static class NonCloseableScope extends SharedScope {
```
PaulSandoz, line 195:
```suggestion
    static class NonCloseableSharedScope extends SharedScope {
```

PaulSandoz (approved):

PR #477, opened 2021-03-26
8264187: Add a method for creating a slicing MethodHandle

JornVernee:
Hi,

This patch adds a `sliceHandle(PathElement...)` method to MemoryLayout, which can be used to create a method handle that, when given a segment, will return a slice of that segment for the layout element selected by the earlier given layout path.

This for instance makes it easier to iterate over an array of structs by first creating a slice for each element, and then accessing struct fields from there, without having to resolve to manual offset computation. For example:

```
SequenceLayout seq = MemoryLayout.ofSequence(5,
    MemoryLayout.ofStruct(
       MemoryLayout.ofValueBits(32, ByteOrder.nativeOrder()).withName("x"),
       MemoryLayout.ofValueBits(32, ByteOrder.nativeOrder()).withName("y")
    ));

MethodHandle sliceHandle = seq.sliceHandle(sequenceElement());

MemoryLayout structLayout = seq.select(sequenceElement());
VarHandle xHandle = structLayout.varHandle(int.class, groupElement("x"));
VarHandle yHandle = structLayout.varHandle(int.class, groupElement("y"));

try (ResourceScope scope = ResourceScope.ofConfined()) {
    MemorySegment segment = MemorySegment.allocateNative(seq, scope);
    long count = seq.elementCount().orElseThrow();
    for (int i = 0; i < count; i++) {
        // create view segment of single struct element
        MemorySegment struct = (MemorySegment) sliceHandle.invokeExact(segment, (long) i);
        xHandle.set(struct, 42);
        yHandle.set(struct, 84);
    }
}
```

Thanks,
Jorn

---------
### Progress

### Issue
 * [JDK-8264187](https://bugs.openjdk.java.net/browse/JDK-8264187): Add a method for creating a slicing MethodHandle

### Reviewers
 * [Maurizio Cimadamore](https://openjdk.java.net/census#mcimadamore) (@mcimadamore - Committer) ⚠️ Review applies to 8db12b617ad411ea70f8bf0b1b8977e6128e434f

### Download
To checkout this PR locally:
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/477/head:pull/477`
`$ git checkout pull/477`

`$ git checkout pull/477`
`$ git pull https://git.openjdk.java.net/panama-foreign pull/477/head`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryLayout.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryLayout.java
@@ -494,6 +494,49 @@ public interface MemoryLayout extends Constable {
     }
 
     /**
+     * Creates a method handle which can be used to create a slice of the layout selected by a given layout path,
+     * where the path is considered rooted in this layout.
+     *
+     * <p>The returned method handle has a return type of {@code MemorySegment}, features a {@code MemorySegment}
+     * parameter as leading parameter representing the segment to be sliced, and features as many trailing {@code long}
+     * parameter types as there are free dimensions in the provided layout path (see {@link PathElement#sequenceElement()},
+     * where the order of the parameters corresponds to the order of the path elements.
+     * The returned method handle can be used to create a slice similar to using {@link MemorySegment#asSlice(long, long)},
+     * but where the offset argument is dynamically compute based on indices specified when invoking the method handle.
+     *
+     * <p>The offset of the returned segment is computed as follows:
+     *
+     * <blockquote><pre>{@code
+    bitOffset = c_1 + c_2 + ... + c_m + (x_1 * s_1) + (x_2 * s_2) + ... + (x_n * s_n)
+    offset = bitOffset / 8
+     * }</pre></blockquote>
+     *
+     * where {@code x_1}, {@code x_2}, ... {@code x_n} are <em>dynamic</em> values provided as {@code long}
+     * arguments, whereas {@code c_1}, {@code c_2}, ... {@code c_m} and {@code s_0}, {@code s_1}, ... {@code s_n} are
+     * <em>static</em> stride constants which are derived from the layout path.
+     *
+     * <p>After the offset is computed, the returned segment is create as if by calling:
+     * <blockquote><pre>{@code
+    segment.asSlice(offset, layout.byteSize());
+     * }</pre></blockquote>
+     *
+     * where {@code segment} is the segment to be sliced, and where {@code layout} is the layout selected by the given
+     * layout path.
+     *
+     * <p>The method handle will throw an {@link UnsupportedOperationException} if the computed
+     * offset in bits is not a multiple of 8.
+     *
+     * @param elements the layout path elements.
+     * @return a method handle which can be used to create a slice of the selected layout element, given a segment.
+     * @throws UnsupportedOperationException if the size of the selected layout in bits is not a multiple of 8.
+     */
+    default MethodHandle sliceHandle(PathElement... elements) {
+        return computePathOp(LayoutPath.rootPath(this, MemoryLayout::bitSize), LayoutPath::sliceHandle,
+                Set.of(), elements);
+    }
+
+
+    /**
      * Selects the layout from a path rooted in this layout.
      *
      * @param elements the layout path elements.
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryLayout.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryLayout.java
@@ -382,8 +382,9 @@ public interface MemoryLayout extends Constable {
      * }</pre></blockquote>
      *
      * where {@code x_1}, {@code x_2}, ... {@code x_n} are <em>dynamic</em> values provided as {@code long}
-     * arguments, whereas {@code c_1}, {@code c_2}, ... {@code c_m} and {@code s_0}, {@code s_1}, ... {@code s_n} are
-     * <em>static</em> stride constants which are derived from the layout path.
+     * arguments, whereas {@code c_1}, {@code c_2}, ... {@code c_m} are <em>static</em> offset constants
+     * and {@code s_0}, {@code s_1}, ... {@code s_n} are <em>static</em> stride constants which are derived from
+     * the layout path.
      *
      * @param elements the layout path elements.
      * @return a method handle that can be used to compute the bit offset of the layout element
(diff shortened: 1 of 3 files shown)


src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryLayout.java
```diff
@@ -493,6 +493,49 @@ default VarHandle varHandle(Class<?> carrier, PathElement... elements) {
                 Set.of(), elements);
     }
 
+    /**
+     * Creates a method handle which can be used to create a slice of the layout selected by a given layout path,
```
mcimadamore, line 497: I'd say "which, given a memory segment, returns a slice (hyperlink to MemorySegment::asSlice) corresponding to the layout selected by a given layout path"

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryLayout.java
```diff
@@ -493,6 +493,49 @@ default VarHandle varHandle(Class<?> carrier, PathElement... elements) {
...
+     * <p>The offset of the returned segment is computed as follows:
+     *
+     * <blockquote><pre>{@code
+    bitOffset = c_1 + c_2 + ... + c_m + (x_1 * s_1) + (x_2 * s_2) + ... + (x_n * s_n)
+    offset = bitOffset / 8
+     * }</pre></blockquote>
+     *
+     * where {@code x_1}, {@code x_2}, ... {@code x_n} are <em>dynamic</em> values provided as {@code long}
+     * arguments, whereas {@code c_1}, {@code c_2}, ... {@code c_m} and {@code s_0}, {@code s_1}, ... {@code s_n} are
+     * <em>static</em> stride constants which are derived from the layout path.
```
mcimadamore, line 516: There is an issue here, which is probably not introduced by you, but also present in other methods: c_1 and s_0 are not all strides; c_0 are offsets, while s_0 are strides.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryLayout.java
```diff
@@ -493,6 +493,49 @@ default VarHandle varHandle(Class<?> carrier, PathElement... elements) {
...
+     *
+     * <p>After the offset is computed, the returned segment is create as if by calling:
+     * <blockquote><pre>{@code
+    segment.asSlice(offset, layout.byteSize());
+     * }</pre></blockquote>
+     *
+     * where {@code segment} is the segment to be sliced, and where {@code layout} is the layout selected by the given
```
mcimadamore, line 523: Maybe we can say "as per MemoryLayout::select(path)" ?

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/LayoutPath.java
```diff
@@ -72,6 +73,8 @@
                     MethodType.methodType(long.class, MemorySegment.class, long.class, long.class, long.class));
             MH_ADD_SCALED_OFFSET = lookup.findStatic(LayoutPath.class, "addScaledOffset",
                     MethodType.methodType(long.class, long.class, long.class, long.class));
+            MH_SLICE = lookup.findStatic(LayoutPath.class, "slice",
```
mcimadamore, line 76: Why not just pointing to MS::asSlice?

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/LayoutPath.java
```diff
@@ -200,6 +203,21 @@ public MethodHandle offsetHandle() {
         return mh;
     }
 
+    private static MemorySegment slice(MemorySegment base, long offset, long length) {
+        return base.asSlice(offset, length);
+    }
+
+    public MethodHandle sliceHandle() {
+        MethodHandle offsetHandle = offsetHandle(); // bit offset
```
mcimadamore, line 211: Nice composition of existing bits!

test/jdk/java/foreign/TestLayoutPaths.java
```diff
@@ -489,5 +493,53 @@ public void testOffsetHandle(MemoryLayout layout, PathElement[] pathElements, lo
...
+    @Test(expectedExceptions = UnsupportedOperationException.class)
+    public void testSliceHandleUOEInvalidOffset() throws Throwable {
+        MemoryLayout layout = MemoryLayout.ofStruct(
+            MemoryLayout.ofPaddingBits(5),
+            MemoryLayout.ofValueBits(32, ByteOrder.nativeOrder()).withName("y") // offset not a multiple of 8
+        );
+
+        MethodHandle sliceHandle;
+        try {
+            sliceHandle = layout.sliceHandle(groupElement("y")); // should work
```
mcimadamore, line 533: Question: why is the failure dynamic? We know exactly, at construction time that the offset is gonna be odd?

mcimadamore (commented):
Looks good - I've left some comments

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryLayout.java
JornVernee: Okay, will fix this on the other ones as well

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryLayout.java
JornVernee: Sure

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/LayoutPath.java
JornVernee: Yeah, I'll do that. I did it like this at first just in case we wanted to change the slice API in the future, but looking now it seems kinda YAGNI

test/jdk/java/foreign/TestLayoutPaths.java
JornVernee: That is true in this case, but not in general where the offset could depend on some dynamic indexes.

However, I could change it to eagerly do the offset check in case we don't have any free dimensions.

mcimadamore: ugh - you mean accessing something like 3bit value array in the 8th element? Yeah...

JornVernee: Yes, exactly. And e.g. if we always pick 0 as offset but the stride is not a multiple of 8 we would not catch that either.

I've added an eager check if there are no strides, and added a test case as well.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryLayout.java
```diff
@@ -493,6 +496,50 @@ default VarHandle varHandle(Class<?> carrier, PathElement... elements) {
                 Set.of(), elements);
     }
 
+    /**
+     * Creates a method handle which, given a memory segment, returns a {@link MemorySegment#asSlice(long,long)}
```
mcimadamore, line 500:
```suggestion
     * Creates a method handle which, given a memory segment, returns a {@link MemorySegment#asSlice(long,long) slice}
```

mcimadamore (approved):
Looks good - added minor javadoc suggestion

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryLayout.java
JornVernee: I tried this, but it doesn't seem to show up in the javadoc

JornVernee: Ah, no wait it does. I was looking in the wrong place :)

PR #478, opened 2021-03-26
8264280: Foreign linker should be more friendly with implicit scopes

mcimadamore:
Following the recent introduction of ResourceScope, it is now "easier" to create segments backed by an implicit scope, which will deallocate memory when the scope becomes unreachable.

Because of this, some idioms, especially in the linker API, have become unsound, consider this:

```
MethodHandle STRLEN = CLinker.downcallHandle(...);
STRLEN.invokeExact(CLinker.toCString("Hello!").address());
```

The above code contains a subtle issue: since the segment is converted into an address (and, by the linker, into a raw long), it is possible for the segment (and hence the scope associated with the segment) to become unreachable *before* the native call has returned.

When working with jextract code, this idiom is even sneakier, given that jextract makes abundant use of `Addressable` in places where an address or a segment can be provided. In those cases the conversion is not even explicit in the user code, it is instead made by jextract.

This seems like an inconsistency in the foreign API; while VarHandle keep segments and scope alive, native MethodHandle don't. Of course we can't solve _all_ possible issues - after all a native call could still hold onto an address for longer, and dereference it later on, when the original segment went unreachable. But there's not much we can do to protect against this (in fact, same issue is possible even with deterministic deallocation). But for simple (and very common!) cases we should try harder to keep the Addressable arguments passed to a native call alive for the duration of the call.

This includes, notably, the Addressable parameter passed to "virtual" method handle, which also can be collected prematurely (with pretty bad consequences).

This patch rectifies that. To do so, the linker machinery will keep alive any addressable argument coming its way (thanks Jorn for wrestling with the combinators). On the API side, I've refactored the memory address implementation into an abstract class, which contains no state (ht: Valhalla) and all the implementations we had before. Then we have two thin subclasses, one that is used for all unchecked addresses, which is just a wrapper around a long; another that is an offset into a segment.

For now, I didn't make any big claim in the direction that an address "has a scope" - the API doesn't say that. The only change API-wise is that MemorySegment::address now documents that the returned address will keep the segment alive.

I've added two tests in TestUpcall/Downcall to make sure that everything works when arguments and returns are created using the default scope; the test was also passing with the previous implementation, so I went a bit more drastic, and added explicit calls to System.gc() - which immediately revealed the problem in the old code. Since calling System.gc() on every test iteration slows down test execution too much, I added some sampling logic, where System.gc() is only called every 100 tests.

---------
### Progress

### Issue
 * [JDK-8264280](https://bugs.openjdk.java.net/browse/JDK-8264280): Foreign linker should be more friendly with implicit scopes

### Reviewers
 * [Paul Sandoz](https://openjdk.java.net/census#psandoz) (@PaulSandoz - Committer) ⚠️ Review applies to 4f8f0db9fa4e9380c8debe20c90c89b30ca20707
 * [Jorn Vernee](https://openjdk.java.net/census#jvernee) (@JornVernee - Committer) ⚠️ Review applies to 4f8f0db9fa4e9380c8debe20c90c89b30ca20707

### Download
To checkout this PR locally:
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/478/head:pull/478`
`$ git checkout pull/478`

`$ git checkout pull/478`
`$ git pull https://git.openjdk.java.net/panama-foreign pull/478/head`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/ResourceScope.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/ResourceScope.java
@@ -56,7 +56,7 @@ import java.util.Spliterator;
  *     (see {@link CLinker#upcallStub(MethodHandle, FunctionDescriptor, ResourceScope)}</li>
  * </ul>
  *
- * <h2>Implicit closure</h2>
+ * <h2><a id = "implicit-closure">Implicit closure</a></h2>
  *
  * Resource scopes can be associated with a {@link Cleaner} instance (see {@link #ofConfined(Cleaner)}) - we call these
  * resource scopes <em>managed</em> resource scopes. A managed resource scope is closed automatically once the scope instance
@@ -253,7 +253,8 @@ public interface ResourceScope extends AutoCloseable {
     }
 
     /**
-     * Create a new <em>default scope</em>, a shared, non-closeable scope which only features implicit closure.
+     * Create a new <em>default scope</em>, a shared, non-closeable scope which only features
+     * <a href="ResourceScope.html#implicit-closure"><em>implicit closure</em></a>.
      * This resource scope is used as a valid default where no resource scope is provided by the user. For instance, this code:
      * <blockquote><pre>{@code
     MemorySegment.allocateNative(10);
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/ResourceScope.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/ResourceScope.java
@@ -253,7 +253,7 @@ public interface ResourceScope extends AutoCloseable {
     }
 
     /**
-     * Create a new <em>default scope</em>, a shared, non-closeable scope which only features
+     * Create a new <em>default scope</em>. The default scope is a shared and non-closeable scope which only features
      * <a href="ResourceScope.html#implicit-closure"><em>implicit closure</em></a>.
      * This resource scope is used as a valid default where no resource scope is provided by the user. For instance, this code:
      * <blockquote><pre>{@code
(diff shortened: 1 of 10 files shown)


src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/ResourceScope.java
```diff
@@ -253,7 +253,8 @@ static ResourceScope ofShared(Object attachment, Cleaner cleaner) {
     }
 
     /**
-     * Create a new <em>default scope</em>, a shared, non-closeable scope which only features implicit closure.
+     * Create a new <em>default scope</em>, a shared, non-closeable scope which only features
```
PaulSandoz, line 256:
```suggestion
     * Create a new <em>default scope</em>. The default scope a shared and non-closeable scope, which only features
```

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryAddressImpl.java
```diff
@@ -35,62 +35,52 @@
  * This class provides an immutable implementation for the {@code MemoryAddress} interface. This class contains information
  * about the segment this address is associated with, as well as an offset into such segment.
  */
-public final class MemoryAddressImpl implements MemoryAddress {
+public abstract class MemoryAddressImpl implements MemoryAddress {
```
PaulSandoz, line 38: Add a comment we could mark this class as a sealed class? Dunno if it coveys any performance advantage but it does covey that the class cannot be extended outside of its nest.

PaulSandoz (approved):
Looks good. I suppose it might be possible to optimize by pushing the logic of memory address creation to the segment's scope, but perhaps that has other performance implications. e.g. i am wondering if unchecked and native can share a common type holding the long `addr` value, where as heap segments get something different.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryAddressImpl.java
mcimadamore: true

JornVernee (approved):
LGTM

PR #485, opened 2021-03-30
8264434: Remove ResourceScope-less overload from API

mcimadamore:
This patch does a bunch of API cleanups related to overloaded methods:

* Remove ResouceScope-less overloads (in MemorySegment, CLinker, MemoryAddress, VaList)
* Remove ResourceScope factory overloads which take Object attachment (in ResourceScope)
* Fixup javadoc

The changes are mostly mechanical - e.g. an extra ResourceScope.ofImplicit() is added wherever needed.
Similarly, for address to segment conversion, an extra ResourceScope.globalScope() is added.

The term *implicit* has been selected to denote the subset of scopes that are not only managed by a Cleaner, but that do not play into explicit deallocation.

So, ResourceScope.ofImplicit() returns a new implicit scope, whereas ResourceScope.globalScope, returns a constant implicit scope which is assumed to be alive for the entire duration of the application.

A new predicate has been added on ResourceScope, namely `isImplicit` which allows client to see whether a segment supports deterministic closure (we used to have a similar predicate, but was called `isCloseable` and that created confusion w.r.t. the semantics of close() - the new naming is more apt, in that it clearly reflects a permanent property of the resource scope).

---------
### Progress

### Issue
 * [JDK-8264434](https://bugs.openjdk.java.net/browse/JDK-8264434): Remove ResourceScope-less overload from API

### Reviewers
 * [Jorn Vernee](https://openjdk.java.net/census#jvernee) (@JornVernee - Committer) ⚠️ Review applies to 7d3573d64818e177e1f3914fb3cf9d70310d9f66

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/485/head:pull/485` \
`$ git checkout pull/485`

`$ git checkout pull/485` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/485/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 485`

`$ git pr show -t 485`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/485.diff">https://git.openjdk.java.net/panama-foreign/pull/485.diff</a>

</details>

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/ResourceScope.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/ResourceScope.java
@@ -146,7 +146,7 @@ public interface ResourceScope extends AutoCloseable {
 
     /**
      * Is this resource scope an <em>implicit scope</em>?
-     * @return true if this scope is an an <em>implicit scope</em>.
+     * @return true if this scope is an <em>implicit scope</em>.
      * @see #ofImplicit()
      * @see #globalScope()
      */
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/ResourceScope.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/ResourceScope.java
@@ -244,7 +244,7 @@ public interface ResourceScope extends AutoCloseable {
     }
 
     /**
-     * Create a new <em>implicit scope</em>. The implicit scope is a shared and non-closeable scope which only features
+     * Create a new <em>implicit scope</em>. The implicit scope is a managed, shared, and non-closeable scope which only features
      * <a href="ResourceScope.html#implicit-closure"><em>implicit closure</em></a>.
      * Since implicit scopes can only be closed implicitly by the garbage collector, it is recommended that implicit
      * scopes are only used in cases where deallocation performance is not a critical concern, to avoid unnecessary
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/ResourceScope.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/ResourceScope.java
@@ -67,7 +67,7 @@ import java.util.Spliterator;
  * scope is closed explicitly, no further action will be taken when the scope becomes unreachable; that is, cleanup actions
  * (see {@link #addOnClose(Runnable)}) associated with a resource scope, whether managed or not, are called <em>exactly once</em>.
  * <p>
- * Some resource scopes (see {@link #ofImplicit()}, {@link #globalScope()} are said to be <em>implicit scopes</em>.
+ * Some managed resource scopes are implicitly managed (see {@link #ofImplicit()}, {@link #globalScope()}, and are said to be <em>implicit scopes</em>.
  * An implicit resource scope only features implicit closure, and always throws an {@link UnsupportedOperationException}
  * when the {@link #close()} method is called directly.
  *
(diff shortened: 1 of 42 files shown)


src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/ResourceScope.java
```diff
@@ -66,6 +66,10 @@
  * deterministic resource deallocation, while still prevent accidental native memory leaks. In case a managed resource
  * scope is closed explicitly, no further action will be taken when the scope becomes unreachable; that is, cleanup actions
  * (see {@link #addOnClose(Runnable)}) associated with a resource scope, whether managed or not, are called <em>exactly once</em>.
+ * <p>
+ * Some resource scopes (see {@link #ofImplicit()}, {@link #globalScope()} are said to be <em>implicit scopes</em>.
```
PaulSandoz, line 70:
```suggestion
 * Some managed resource scopes are implicitly managed (see {@link #ofImplicit()}, {@link #globalScope()}, and are said to be <em>implicit scopes</em>.
```

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/ResourceScope.java
```diff
@@ -241,40 +240,24 @@ static ResourceScope ofShared() {
...
-     * @return a new shared scope, managed by {@code cleaner} (where provided).
-     */
-    static ResourceScope ofShared(Object attachment, Cleaner cleaner) {
-        return MemoryScope.createShared(attachment, cleaner);
+        return MemoryScope.createShared(cleaner);
     }
 
     /**
-     * Create a new <em>default scope</em>. The default scope is a shared and non-closeable scope which only features
+     * Create a new <em>implicit scope</em>. The implicit scope is a shared and non-closeable scope which only features
```
PaulSandoz, line 247:
```suggestion
     * Create a new <em>implicit scope</em>. The implicit scope is a managed, shared, and non-closeable scope which only features
```

PaulSandoz (commented):
Looking at the changes in the tests it occurred to me that `Resource.ofX` does not work so well with a static import, `Resource.XScope` would work better and is consistent with `globalScope`.

May be worth reviewing factory methods of other classes too, namely `SegmentAllocator`. 

For common cases i would imagine static imports may be useful, so the method names should to stand on their own, at the expense of a few more characters (3 for scope) when the method is qualified.

JornVernee (approved):
LGTM. I agree with Paul that it would be nice to rename `ofXYZ` factories to something that is more amenable to static importing (as also discussed offline), e.g. `ResourceScope.ofImplicit()` -> `ResourceScope.implicit()` to be consistent with `SegmentAllocator.implicit()`.

mcimadamore:
> Looking at the changes in the tests it occurred to me that `Resource.ofX` does not work so well with a static import, `Resource.XScope` would work better and is consistent with `globalScope`.

Ha! I had exactly this conversation earlier (offline).

Same applied to MemoryLayout.ofStruct and others.

I'm happy to rename all this, but (i) I'd like to separate from this change and (ii) I need some guidance from JDK API gurus - e.g. some JDK factories do use `ofXYZ` (e.g. List.of, ClassDesc.of) is there some rule about naming which will land the API in a more consistent place with the rest of the JDK?
> 
> May be worth reviewing factory methods of other classes too, namely `SegmentAllocator`.
> 
> For common cases i would imagine static imports may be useful, so the method names should stand on their own, at the expense of a few more characters (3 for scope) when the method is qualified.

mcimadamore:
> LGTM. I agree with Paul that it would be nice to rename `ofXYZ` factories to something that is more amenable to static importing (as also discussed offline), e.g. `ResourceScope.ofImplicit()` -> `ResourceScope.implicit()` to be consistent with `SegmentAllocator.implicit()`.

I think whatever naming we choose, we need to have some visual distinction between factories which are just "accessors" e.g. ResourceScope.globalScope and things which create a new scope e.g. ResourceScope.ofImplicit. But, as discussed offline, I don't want to go down the road of calling the latter `makeImplicitScope` or `newImplicitScope`.

PaulSandoz (approved):

PR #487, opened 2021-03-31
8264515: Rename static factory methods in the foreign API

mcimadamore:
It has been pointed out recently, that some of the static factory names are not very friendly with use of static imports. The latest case is `ResourceScope::ofImplicit`, which would lead to code like this:

```
MemorySegment.allocateNative(100, ofImplicit());
```

But there's also the old good layout API:

```
MemoryLayout layout = MemoryLayout.ofSequence(10,
                                  MemoryLayout.ofStruct(
                                             MemoryLayout.ofValueBits(4).withName("x"),
                                             MemoryLayout.ofPadding(8),
                                             MemoryLayout.ofValueBits(4).withName("y"),
                                  )); 
```

We've been examining precedents, and also looking at some of the existing comments on the topic [1], and we came up with the following principles, which we have tried to adhere to in our renaming effort:

* a static factory should have a name that stands on its own (think `MethodType.methodType`)
* in some cases, when converting between different things, or when creating something out of something else the prefix `of` is acceptable (`LocalDateTime.ofInstant`); this also means static imports cannot be used in these cases
* if returning a new instance is relevant semantic-wise, factory name should be prefixed with `new` (as in `Files.newInputStream`); otherwise `new` can be omitted

Armed with these principles, we have made the following changes to the API:

### MemoryLayouts

Factories have been renamed from `ofXYZ` to `XYZLayout` - so the above example becomes:

```
MemoryLayout layout = sequenceLayout(10,
                                  structLayout(
                                             valueLayout(4).withName("x"),
                                             paddingLayout(8),
                                             valueLayout(4).withName("y"),
                                  )); 
```

Which is much more readable.

### ResourceScope

Here we opted for the following mapping:

* ResourceScope::ofImplicit -> ResourceScope::newImplicitScope
* ResourceScope::ofConfined -> ResourceScope::newConfinedScope
* ResourceScope::ofShared -> ResourceScope::newSharedScope
* ResourceScope::globalScope ->  ResourceScope::globalScope (unchanged)

So, the above example becomes:

```
MemorySegment.allocateNative(100, newImplicitScope());
```

Again, more readable and explicit.

### SegmentAllocator

Here it's tricky, as we have a mix of allocator-sounding names (`arenaBounded`/`arenaUnbounded`/`malloc`) and some odd ones out (`scoped`, `prefix`, `implicit`).

So, in the case of `arenaBounded`/`arenaUnbounded`, we think that having a couple of `arenaAllocator` overloads would be preferrable.

As for `scoped` and `prefix`, we think it's more useful to think of these as *conversions*, so use `of`. Following the precedent on `LocalDateTime`, let's also add the type name aftre `of`, e.g. `ofScope` and `ofSegment`, respectively.

This leaves us with `malloc` and `implicit`. Now, I think while `malloc` is allocator sounding, it is also a misleading name. This is really an allocator that just ends up calling `MemorySegment::allocateNative` with fresh scopes. 

We also noted that, being `SegmentAllocator` a functional interface, the gain with using `malloc` is relatively small:

```
SegemtAllocator allocator = malloc(ResourceScope::newConfinedScope);
```

vs.

```
SegmentAllocator allocator = (size, align) -> MemorySegment.allocateNative(size, align, newConfinedScope())
```

So we opted to remove `malloc`; we also applied a similar argument for the implicit allocator (which, in fact, is really like `malloc(ResourceScope::newImplicitScope)`). So, to summarize:

* SegmentAllocator::arenaBounded -> ResourceScope::arenaAllocator
* SegmentAllocator::arenaUnbounded -> ResourceScope::arenaAllocator
* SegmentAllocator::prefix -> ResourceScope::ofSegment
* SegmentAllocator::scoped -> ResourceScope::ofScope
* SegmentAllocator::malloc -> // removed
* SegmentAllocator::implicit -> // removed

### Conclusions

While (many) other naming schemes are possible, what we did seems consistent with precedent set by other JDK API, and seems to land the API in a place which is more amenable to using static imports, which seems like an improvement (gievn the number of static methods in the API).

While most of the changes were done using the IDE, some of them have been done manually - specifically those fixing the currently broken micro benchmark (after removal of "restricted" suffix yesterday), and fixes to scope-less MemorySegment::allocateNative in various javadoc samples.

[1] - https://blog.joda.org/2011/08/common-java-method-names.html

---------
### Progress

### Issue
 * [JDK-8264515](https://bugs.openjdk.java.net/browse/JDK-8264515): Rename static factory methods in the foreign API

### Reviewers
 * [Jorn Vernee](https://openjdk.java.net/census#jvernee) (@JornVernee - Committer) ⚠️ Review applies to af76b2e0b5a0b9b0ce4b9be50596dd9d78c281f2

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/487/head:pull/487` \
`$ git checkout pull/487`

`$ git checkout pull/487` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/487/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 487`

`$ git pr show -t 487`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/487.diff">https://git.openjdk.java.net/panama-foreign/pull/487.diff</a>

</details>

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/AbstractLayout.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/AbstractLayout.java
@@ -214,22 +214,22 @@ abstract class AbstractLayout implements MemoryLayout {
 
     static final ConstantDesc LITTLE_ENDIAN = DynamicConstantDesc.ofNamed(BSM_GET_STATIC_FINAL, "LITTLE_ENDIAN", CD_BYTEORDER, CD_BYTEORDER);
 
-    static final MethodHandleDesc MH_PADDING = MethodHandleDesc.ofMethod(DirectMethodHandleDesc.Kind.INTERFACE_STATIC, CD_MEMORY_LAYOUT, "ofPaddingBits",
+    static final MethodHandleDesc MH_PADDING = MethodHandleDesc.ofMethod(DirectMethodHandleDesc.Kind.INTERFACE_STATIC, CD_MEMORY_LAYOUT, "paddingLayout",
                 MethodTypeDesc.of(CD_MEMORY_LAYOUT, CD_long));
 
-    static final MethodHandleDesc MH_VALUE = MethodHandleDesc.ofMethod(DirectMethodHandleDesc.Kind.INTERFACE_STATIC, CD_MEMORY_LAYOUT, "ofValueBits",
+    static final MethodHandleDesc MH_VALUE = MethodHandleDesc.ofMethod(DirectMethodHandleDesc.Kind.INTERFACE_STATIC, CD_MEMORY_LAYOUT, "valueLayout",
                 MethodTypeDesc.of(CD_VALUE_LAYOUT, CD_long, CD_BYTEORDER));
 
-    static final MethodHandleDesc MH_SIZED_SEQUENCE = MethodHandleDesc.ofMethod(DirectMethodHandleDesc.Kind.INTERFACE_STATIC, CD_MEMORY_LAYOUT, "ofSequence",
+    static final MethodHandleDesc MH_SIZED_SEQUENCE = MethodHandleDesc.ofMethod(DirectMethodHandleDesc.Kind.INTERFACE_STATIC, CD_MEMORY_LAYOUT, "sequenceLayout",
                 MethodTypeDesc.of(CD_SEQUENCE_LAYOUT, CD_long, CD_MEMORY_LAYOUT));
 
-    static final MethodHandleDesc MH_UNSIZED_SEQUENCE = MethodHandleDesc.ofMethod(DirectMethodHandleDesc.Kind.INTERFACE_STATIC, CD_MEMORY_LAYOUT, "ofSequence",
+    static final MethodHandleDesc MH_UNSIZED_SEQUENCE = MethodHandleDesc.ofMethod(DirectMethodHandleDesc.Kind.INTERFACE_STATIC, CD_MEMORY_LAYOUT, "sequenceLayout",
                 MethodTypeDesc.of(CD_SEQUENCE_LAYOUT, CD_MEMORY_LAYOUT));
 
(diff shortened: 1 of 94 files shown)


JornVernee (approved):
Looks good!

PaulSandoz (approved):
Very good.

Something to ponder after this PR. I have my suspicion that the distinction with the "of" naming convention will be lost on many as such methods get swept up in the `static import ...SegmentAllocator.*`. It may be better to be simpler (less nuanced) and consistent.

mcimadamore:
> Very good.
> 
> Something to ponder after this PR. I have my suspicion that the distinction with the "of" naming convention will be lost on many as such methods get swept up in the `static import ...SegmentAllocator.*`. It may be better to be simpler (less nuanced) and consistent.

True - that's why we conceded for ofXYZ - e.g.

```
toCString("Hello", ofScope(scope))
```
Which is not ideal, but not downright terrible.

The alternative would be to have `prefixAllocator` and `scopedAllocator` - we considered that and went against that (at least for the time being) as they suggest that these allocators are really more than what they really are. E.g. when I create an arena allocator, there's some work going on - whereas for scope/segment -> allocator is really just wiring up things. The distinction is subtle, yes, but I'd like to see how much mileage we can get :-)

mcimadamore:
One thing not addressed in this patch, which should:

`MemorySegment.ofNative()`

This used to be: `MemorySegment::ofNativeRestricted`

We can rename it to `allNative`, or `nativeHeap`, or ... 

any ideas/suggestions?

PaulSandoz:
> One thing not addressed in this patch, which should:
> 
> `MemorySegment.ofNative()`
> 
> This used to be: `MemorySegment::ofNativeRestricted`
> 
> We can rename it to `allNative`, or `nativeHeap`, or ...
> 
> any ideas/suggestions?

`nativeGlobalRegion` ?

mcimadamore:
> `nativeGlobalRegion` ?

Yeah - or, slight variations:

`globalNativeRegion` (that seems to match more with `globalScope` - e.g. put global first)
or
`globalNativeSegment` (use segment instead of region, as region is not a word in our API)

PR #494, opened 2021-04-08
8264933: Improve stream support in memory segments

mcimadamore:
This patch improves stream support for memory segments - it does so in two ways:

* first, tweaking the `MemorySegment::spliterator` method to take an *element layout*, instead of a full sequence layout whose size had to match that of the segment.
* secondly, by adding convenience methods to create streams directly, w/o needing to call `StreamSupport.stream`. As for `Collection`, two methods are provided: `MemorySegment::stream` and `MemorySegment::parallelStream`.

I've fixed up the documentation in places - I realized that, for example, the javadoc for `MemorySegment::spliterator` was missing a `@throws` clause for when there's a size mismatch between layout and segment.
Also, I've removed scope liveness check on the spliterator call - after all, a spliterator is going to make slices - a liveness check will occur when the slice is accessed.

---------
### Progress

### Issue
 * [JDK-8264933](https://bugs.openjdk.java.net/browse/JDK-8264933): Improve stream support in memory segments

### Reviewers
 * [Jorn Vernee](https://openjdk.java.net/census#jvernee) (@JornVernee - Committer) ⚠️ Review applies to 59033546afafbc7506249639cee037d6f0ec9cdb
 * [Paul Sandoz](https://openjdk.java.net/census#psandoz) (@PaulSandoz - Committer) ⚠️ Review applies to 59033546afafbc7506249639cee037d6f0ec9cdb

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/494/head:pull/494` \
`$ git checkout pull/494`

`$ git checkout pull/494` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/494/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 494`

`$ git pr show -t 494`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/494.diff">https://git.openjdk.java.net/panama-foreign/pull/494.diff</a>

</details>


forax:
I fail to see how a Spliterator that asks a stream to do has many recursive calls/creates as many sub-segments has the number of elements is a good idea ?

mcimadamore:
> I fail to see how a Spliterator that asks a stream to do has many recursive calls/creates as many sub-segments has the number of elements is a good idea ?

This is something that has been present since the second iteration of the API. And, it can be used quite effectively. We have benchmarks (see ParallelSum) which use fork join recursive actions to do a parallel sum of the contents of a segment. Provided that the "size of the element" is chosen appropriately, the speedup obtained is rather nice:

```
Benchmark                   Mode  Cnt   Score   Error  Units
ParallelSum.segment_serial  avgt   30  86.004 ? 0.941  ms/op
```

vs:

```
Benchmark                                 Mode  Cnt   Score   Error  Units
ParallelSum.segment_stream_parallel       avgt   30  45.211 ? 1.105  ms/op
ParallelSum.segment_stream_parallel_bulk  avgt   30  23.057 ? 0.353  ms/op
```

Here we compare a serial sum of all the elements in a segment (with a flat for loop) against a parallel sum using parallel streams; in the first parallel benchmark, we use a split size that is the same as the element size (e.g. 4 bytes); while this improves throughput, it is not ideal as too many intermediate segments are created. But that's why the spliterator/stream methods accept an element layout: you can also specify a "bulk element" (e.g. 1024 ints each) and then process these in parallel. As you can see the speed up increases another 2x by doing this.

Considering how little code is needed to write this:

```
   @Benchmark
    public int segment_stream_parallel_bulk() {
        return segment.parallelStream(ELEM_LAYOUT_BULK).mapToInt(SEGMENT_TO_INT_BULK).sum();
    }
```

I think this makes sense; of course it's not a silver bullet, and has to be handled with care, but here we assume the audience knows what they are doing.

Apart from parallel processing, turning a segment into a stream of slices is also useful to perform ad-hoc marshalling/unmarshalling - as written in the panama-dev email:

```
segment.stream(C_POINTER)
       .map(MemoryAccess::getAddress)
       .map(CLinker::toJavaString)
       .toArray(String[]::new);
``` 

In this case, we're not after performances - we just want to express more directly what would otherwise be expressed using a big for loop.

forax:
So not using the "bulk" version leads to less performance but because it's a Stream of MemorySegment, you have to know as a user if you use the bulk version or not.

I think i would prefer the stream to be a stream of the addresses,a LongStream instead of being a Stream of MemorySegment, so your last example can be written
```
  segment.addresses(C_POINTER)
       .map(CLinker::toJavaString)
       .toArray(String[]::new);
```
Here, the element is just there to get its size.

Having a stream of offsets/addresses will also work better with VarHandles, by example
```
  segment.addresses(AN_INT)
       .mapToInt(offset -> HANDLE.getInt(segment, offset))
       .sum();
```

mcimadamore:
> So not using the "bulk" version leads to less performance but because it's a Stream of MemorySegment, you have to know as a user if you use the bulk version or not.

Yes
> 
> I think i would prefer the stream to be a stream of the addresses,a LongStream instead of being a Stream of MemorySegment, so your last example can be written
> 
> ```
>   segment.addresses(C_POINTER)
>        .map(CLinker::toJavaString)
>        .toArray(String[]::new);
> ```
> 
> Here, the element is just there to get its size.
> 
> Having a stream of offsets/addresses will also work better with VarHandles, by example
> 
> ```
>   segment.addresses(AN_INT)
>        .mapToInt(offset -> HANDLE.getInt(segment, offset))
>        .sum();
> ```

I see what you mean: instead of returning slices, return offsets, and let the user deal with it.  Honestly, I'm not sure - it seems like we are, again, prematurely concluding that slicing is gonna be more expensive, while the benchmarks I have show that the performance we get are the same as with Unsafe.

Note that we want each returned segment to be disjoint - an offset only tells me the starting point - doesn't tell me how big it is - which means it is possible/easier for clients to make mistake and write on "somebody else's" slice.

JornVernee (approved):
LGTM!

forax:
> I see what you mean: instead of returning slices, return offsets, and let the user deal with it. Honestly, I'm not sure - it seems like we are, again, prematurely concluding that slicing is gonna be more expensive, while the benchmarks I have show that the performance we get are the same as with Unsafe.

Yes !
I see no issue with the Spliterator<MemorySegment>, it's the right abstraction and it doesn't seem to cost too much.

But i disagree that a Stream<MemorySegment> is the right abstraction.
The design of this API separates the Segment from the way to access to it, and the user links the two together with offsets.

> Note that we want each returned segment to be disjoint - an offset only tells me the starting point - doesn't tell me how big it is - which means it is possible/easier for clients to make mistake and write on "somebody else's" slice.

This is true when using VarHandle too, as a user you have to use the right segment.

mcimadamore:
> This is true when using VarHandle too, as a user you have to use the right segment.

Yes! Which is why having a memory segment slice that give you access only to the portion of data you want is the right way to approach this:

```
segment.addresses(AN_INT)
       .mapToInt(MemoryAccess::getInt)
       .sum();
```

I don't question that, in some cases, having a stream of offsets would be handy - but the goal of this PR is mostly to avoid what the API forces you to do today: `StreamSupport.stream(segment.spliterator(LAYOUT), true/false)`. Surely we can add other accessors in the future?

Is it possible we are quibbling over naming? E.g. perhaps instead of calling this new method `MemorySegment::stream` and suggesting that this is _the way_ to view a segment as a stream, perhaps calling it "slices" would be more honest? I'm open to that, although that needs some way to create a parallel stream too.

mcimadamore:
Proposal: instead of what we do in this PR, let's tweak things a little:

* for usability, let's add `MemorySegment::slices(MemoryLayout)`, which returns a _sequential_ stream of all the slices in a segment (with given element layout).
* for advanced use cases, including parallel streams, users will have to use spliterator + StreamSupport, as before.

This doesn't close the door to other stream-ified views to be added in the future. What do people think?

forax:
I don't see the point of slicing a segment into slice of one element.
Can we have a second argument with the maximum size of a sub-segment ?

mcimadamore:
> I don't see the point of slicing a segment into slice of one element.
> Can we have a second argument with the maximum size of a sub-segment ?

I have shown already real use cases for it - so there is a point :-)
Most of the times memory segments will be flat arrays of "stuff" - and it will be common, for user, to iterate over that "stuff" and element-wise convert it back to something Java-ish.

So, I'd suggest to stop second guessing use cases - as that's not helping the discussion too much; we already have had a spliterator of segment slices for the past 2 years - this change was a little usability boost to simplify what code was already doing. I think we are turning it into more than it is, really.

The definition of "element" here is fluid - it's whatever layout you pass - can be 4 bytes, or 1K. I also don't get the "maximum size of subsegment". The subsegment has a definite size, that of the layout provided by the user.

But really, at the bottom of it - assuming I have my array of string pointers, w/o this capability I have to do:

```
String[] strings = new String[n];
for (int i = 0; i < n ; i++) {
    MemoryAddress addr = MemoryAccess.getAddressAtIndex(segment, n);
    strings[i] = CLinker.toJavaString(addr);
}
```

If somebody wants to write higher level code, why not allowing users to do this:

```
segment.slices(C_POINTER)
    .map(MemoryAccess::getAddress)
    .map(CLinker::toJavaString)
    .toArray(String[]::new)
```

If you don't want to use it, because you are afraid of slice creation (or something else), then don't use it - but I'm really struggling here to see why such an API point is considered "bad"/"harmful" which your comments seem to imply.

Could we please be more explicit about what the real concern is here, rather than just expressing subjective opinions like "I don't see the point in doing XYZ" (since clearly if there was no point we wouldn't be filing this PR).

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -161,21 +163,46 @@
...
-     * thread owning this segment
+     * @throws IllegalArgumentException if this segment size is not a multiple of the size of {@code elementLayout}.
+     */
+    Spliterator<MemorySegment> spliterator(MemoryLayout elementLayout);
+
+    /**
+     * Returns a sequential {@code Stream} over disjoint slices (whose size matches that of the specified layout)
+     * in this segment. Calling this method is equivalent to the following code:
+     * <blockquote><pre>{@code
+    StreamSupport.stream(segment.spliterator(), false);
```
PaulSandoz, line 185:
```suggestion
    StreamSupport.stream(segment.spliterator(elementLayout), false);
```

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -161,21 +163,46 @@
...
+     * <blockquote><pre>{@code
+    StreamSupport.stream(segment.spliterator(), true);
+     * }</pre></blockquote>
+     *
+     * @param elementLayout the layout to be used for splitting.
+     * @return a parallel {@code Stream} over disjoint slices in this segment.
+     * @throws IllegalArgumentException if this segment size is not a multiple of the size of {@code elementLayout}.
      */
-    Spliterator<MemorySegment> spliterator(SequenceLayout layout);
+    Stream<MemorySegment> parallelStream(MemoryLayout elementLayout);
```
PaulSandoz, line 205: I don't think we need this method. Except for collections we generally don't use this pattern, and instead encourage the use of `Stream.parallel()` in such cases.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -161,21 +163,46 @@
...
+     * in this segment. Calling this method is equivalent to the following code:
+     * <blockquote><pre>{@code
+    StreamSupport.stream(segment.spliterator(), false);
+     * }</pre></blockquote>
+     *
+     * @param elementLayout the layout to be used for splitting.
+     * @return a sequential {@code Stream} over disjoint slices in this segment.
+     * @throws IllegalArgumentException if this segment size is not a multiple of the size of {@code elementLayout}.
+     */
+    Stream<MemorySegment> stream(MemoryLayout elementLayout);
```
PaulSandoz, line 192: Renaming to `elements` could also work.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -161,21 +163,46 @@
...
      * @return the element spliterator for this segment
-     * @throws IllegalStateException if the segment is not <em>alive</em>, or if access occurs from a thread other than the
-     * thread owning this segment
+     * @throws IllegalArgumentException if this segment size is not a multiple of the size of {@code elementLayout}.
+     */
+    Spliterator<MemorySegment> spliterator(MemoryLayout elementLayout);
+
+    /**
+     * Returns a sequential {@code Stream} over disjoint slices (whose size matches that of the specified layout)
+     * in this segment. Calling this method is equivalent to the following code:
```
PaulSandoz, line 183: The equivalent text could be converted to an API note.

PaulSandoz (approved):

PR #503, opened 2021-04-14
Add test for resource scope acquire with bulk operation

mcimadamore:
This patch adds a test for a bulk segment mismatch inside an acquire block. This benchmark measures the relative overhead of the acquire operation, relative to the cost of the bulk mismatch. Results for large segments are as follows:

```
Benchmark                                           (scopeKind)  Mode  Cnt      Score     Error  Units
BulkMismatchAcquire.mismatch_large_segment             CONFINED  avgt   30  37662.474 ? 579.201  ns/op
BulkMismatchAcquire.mismatch_large_segment               SHARED  avgt   30  37583.186 ? 427.663  ns/op
BulkMismatchAcquire.mismatch_large_segment             IMPLICIT  avgt   30  37577.151 ? 474.734  ns/op
BulkMismatchAcquire.mismatch_large_segment_acquire     CONFINED  avgt   30  37801.227 ? 515.459  ns/op
BulkMismatchAcquire.mismatch_large_segment_acquire       SHARED  avgt   30  37159.676 ? 369.788  ns/op
BulkMismatchAcquire.mismatch_large_segment_acquire     IMPLICIT  avgt   30  37854.949 ? 464.150  ns/op
```

Here, we can see that the cost of the acquire is, in all cases, negligible, compared with the cost of the underlying bulk operation. We expect this to be the use case where resource scope handles are used the most.

For completeness, here's the same benchmark, but with a smaller segment size:

```
Benchmark                                           (scopeKind)  Mode  Cnt   Score   Error  Units
BulkMismatchAcquire.mismatch_small_segment             CONFINED  avgt   30   5.826 ? 0.091  ns/op
BulkMismatchAcquire.mismatch_small_segment               SHARED  avgt   30   5.160 ? 0.072  ns/op
BulkMismatchAcquire.mismatch_small_segment             IMPLICIT  avgt   30   5.274 ? 0.090  ns/op
BulkMismatchAcquire.mismatch_small_segment_acquire     CONFINED  avgt   30   7.386 ? 0.035  ns/op
BulkMismatchAcquire.mismatch_small_segment_acquire       SHARED  avgt   30  15.130 ? 0.649  ns/op
BulkMismatchAcquire.mismatch_small_segment_acquire     IMPLICIT  avgt   30   6.877 ? 0.083  ns/op
```

In this case, we see that for implicit and confined scope the cost is low; for shared segments we have a spike (which is related to the CAS logic in the shared handle). We believe this to be an acceptable compromise, and an unavoidable consequence of having true, deterministic shared segments.

In addition to adding the benchmark - the patch also does two other things:

* Fixes a bug in the NonCloseableScope::acquire logic (the polarity of the null check is wrong, and this was causing NPEs on the benchmark - whoops)
* Fixes the unrelated test issue with SpliteratorTest - this was also fixed by PR/494, but seeing that it's not converging, I'd like to get the tests fixed sooner rather than later

---------
### Progress

### Reviewers
 * [Chris Hegarty](https://openjdk.java.net/census#chegar) (@ChrisHegarty - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/503/head:pull/503` \
`$ git checkout pull/503`

`$ git checkout pull/503` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/503/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 503`

`$ git pr show -t 503`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/503.diff">https://git.openjdk.java.net/panama-foreign/pull/503.diff</a>

</details>

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/ResourceScopeImpl.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/ResourceScopeImpl.java
@@ -204,7 +204,7 @@ public abstract class ResourceScopeImpl implements ResourceScope, ScopedMemoryAc
 
         @Override
         public Handle acquire() {
-            if (handle != null) {
+            if (handle == null) {
                 // capture 'this'
                 handle = () -> Reference.reachabilityFence(NonCloseableSharedScope.this);
             }
--- a/test/jdk/java/util/stream/test/org/openjdk/tests/java/util/stream/SpliteratorTest.java
+++ b/test/jdk/java/util/stream/test/org/openjdk/tests/java/util/stream/SpliteratorTest.java
@@ -65,7 +65,7 @@ public class SpliteratorTest {
 
     @Test(dataProvider = "SegmentSpliterator", dataProviderClass = SegmentTestDataProvider.class )
     public void testSegmentSpliterator(String name, SequenceLayout layout, SpliteratorTestHelper.ContentAsserter<MemorySegment> contentAsserter) {
-        try (ResourceScope scope = ResourceScope.ofConfined()) {
+        try (ResourceScope scope = ResourceScope.newConfinedScope()) {
             MemorySegment segment = MemorySegment.allocateNative(layout, scope);
             SegmentTestDataProvider.initSegment(segment);
             SpliteratorTestHelper.testSpliterator(() -> segment.spliterator(layout),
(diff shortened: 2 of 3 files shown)


ChrisHegarty (approved):
LGTM.  ( I happened to noticed the polarity of the null check was wrong too, but you got to it before me :-) )

PR #509, opened 2021-04-17
MemorySegmentPool + Allocator

rsmogura:
(Preview)

The MemorySegmentPool is a pool maintaining memory segments, optionally can expose allocator which can be bound to other scope, and which will return allocated segments back to pool.

However the best results has been achieved by using getSegmentEntry & putSegmentEntry methods.

The pool is intended to be used by long running applications (i.e. like global shared pool), where fast allocation and de-allocation of segments is critical (was designed during implementation of I/O subsystem with Panama, as a pool for temporary buffers between system I/O methods and Java byte arrays from InputStreams).

The pool uses hand-made SpinLockQueue as the Deque from JDK offers too much functionality and overhead.

---------
### Progress

### Integration blocker
&nbsp;⚠️ Whitespace errors (failed with updated jcheck configuration in pull request)

### Reviewers
 * @Aungkhant884 (no known openjdk.org user name / role)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign.git pull/509/head:pull/509` \
`$ git checkout pull/509`

`$ git checkout pull/509` \
`$ git pull https://git.openjdk.org/panama-foreign.git pull/509/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 509`

`$ git pr show -t 509`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/509.diff">https://git.openjdk.org/panama-foreign/pull/509.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/panama-foreign/pull/509#issuecomment-821886607)


mcimadamore:
I've fetched the branch, and tried to hook it up on a benchmark we have (`StrLenTest`) which allocates a native string and calls strlen on it. Here are the results:

```
StrLenTest.panama_strlen                      5  avgt   30  113.616 ? 3.525  ns/op
StrLenTest.panama_strlen                     20  avgt   30  118.923 ? 0.881  ns/op
StrLenTest.panama_strlen                    100  avgt   30  148.828 ? 1.055  ns/op
StrLenTest.panama_strlen_arena                5  avgt   30   37.072 ? 1.336  ns/op
StrLenTest.panama_strlen_arena               20  avgt   30   46.380 ? 1.600  ns/op
StrLenTest.panama_strlen_arena              100  avgt   27   87.728 ? 4.565  ns/op
StrLenTest.panama_strlen_pool                 5  avgt   30   66.258 ? 0.744  ns/op
StrLenTest.panama_strlen_pool                20  avgt   30   69.817 ? 0.364  ns/op
StrLenTest.panama_strlen_pool               100  avgt   30   85.723 ? 0.966  ns/op
StrLenTest.panama_strlen_prefix               5  avgt   30   27.365 ? 0.536  ns/op
StrLenTest.panama_strlen_prefix              20  avgt   30   32.963 ? 0.439  ns/op
StrLenTest.panama_strlen_prefix             100  avgt   30   59.558 ? 1.001  ns/op
StrLenTest.panama_strlen_unsafe               5  avgt   30   40.098 ? 0.319  ns/op
StrLenTest.panama_strlen_unsafe              20  avgt   30   43.820 ? 0.598  ns/op
StrLenTest.panama_strlen_unsafe             100  avgt   30   63.797 ? 4.497  ns/op
StrLenTest.panama_strlen_unsafe_trivial       5  avgt   30   26.341 ? 0.266  ns/op
StrLenTest.panama_strlen_unsafe_trivial      20  avgt   30   32.249 ? 0.553  ns/op
StrLenTest.panama_strlen_unsafe_trivial     100  avgt   30   59.589 ? 0.772  ns/op
```

The pool solution seems to be 2x faster than plain usage of `MemorySegment::allocateNative`. That said, pooled allocation is still significantly slower than calling malloc outside the `MemorySegment` API (see `unsafe` benchmarks), and also slower than just writing the native string onto the same segment over and over (see `prefix` benchmarks).

mcimadamore:
if I remove CAS from the queue (e.g. provide an optimized queue implementation for confined code), results are

```
Benchmark                      (size)  Mode  Cnt   Score   Error  Units
StrLenTest.panama_strlen_pool       5  avgt   30  52.930 ? 0.643  ns/op
StrLenTest.panama_strlen_pool      20  avgt   30  57.778 ? 0.789  ns/op
StrLenTest.panama_strlen_pool     100  avgt   30  74.229 ? 0.948  ns/op
```
Which are still not where I'd expected them to be.

mcimadamore:
And this is using low level API (this also optimizes the queue by removing CAS):

```
Benchmark                          (size)  Mode  Cnt   Score   Error  Units
StrLenTest.panama_strlen_pool_low       5  avgt   30  42.199 ? 0.475  ns/op
StrLenTest.panama_strlen_pool_low      20  avgt   30  44.181 ? 0.248  ns/op
StrLenTest.panama_strlen_pool_low     100  avgt   30  60.410 ? 1.280  ns/op
```

rsmogura:
I optimized code and CAS, but I'm not sure if CAS isn't too optimistic.

mcimadamore:
>  Direct API has competitive performance with unsafe version

Yeah - this is the thing I find suspicious: the unsafe version "cheats" by avoiding creation of a segment the "standard" way, which required memory to be zeroed, memory limits to be updated (there are a couple of CAS to do that on allocation and then on release). So it shaves off all the "redundant" cost. But, it still does "malloc" and "free"! So, if your memory pool, can perform like the unsafe version, it seems to mean that it's not faster than malloc/free - which is odd, and I don't have an explanation as to why it is the way it is.

I think your reference/target should be more `panama_strlen_prefix`, which performs no allocation whatsoever (I would imagine that in a benchmark where I keep requesting a segment of the same size, the pool will keep recycling the existing memory and not call malloc/free, so it should be similar?).

rsmogura:
Malloc / free actually use internally some good optimized algorithms.

(I briefly checked implementation of malloc and it sometimes uses arenas to allocate memory, I think that zeroing can happen due to system allocating zeroed page).

However, AFIK, malloc / free have a tendency to fragmentize memory, and generally C guys avoids it due to fragmentization of memory.

From all the checks, I noticed that CAS during spin lock is the biggest bottleneck, but I have no idea how to remove it.

Today evening I'll check how malloc works maybe I'll have some idea.

mcimadamore:
> Malloc / free actually use internally some good optimized algorithms.
> 
> (I briefly checked implementation of malloc and it sometimes uses arenas to allocate memory, I think that zeroing can happen due to system allocating zeroed page).

I don't think malloc concerns with zeroing - calloc does (and yes, calloc exploits behavior of mmap to avoid redundant zeroing when requesting large allocations).
> 
> However, AFIK, malloc / free have a tendency to fragmentize memory, and generally C guys avoids it due to fragmentization of memory.
Sure, I guess my point was more related on the need of a custom allocator in your specific case - there are many sources for overhead associated to MemorySegment::allocateNative:

* memory zeroing (cost grows linearly with allocated size)
* update memory limits (2 CAS on allocation, 2 CAS on free)
* some memory checks in Unsafe::allocateMemory
* alignment checks and slicing
* cost of malloc/free

If you just call "malloc" as is (CLinker::allocateMemory) - and then wrap the results in a segment (MemoryAddress::asSegment) the numbers you get are typically quite good (this is, essentially, what the `unsafe` benchmark does). So the question is: have you tried something like this with your IO implementation and see which numbers are coming out? In other words, are you sure that the reasons you are trying to avoid malloc/free have to do with malloc/free themselves, rather than to the extra overhead/bookkeeping added by `MemorySegment::allocateNative` ?

Of course, if you have experienced fragmentation issues with malloc, the ideas above are not going to help, and I agree that a better allocator is needed.

rsmogura:
I'll check one additional thing as analyzing assembly output I've found two entries for spinlock, not sure if this is expected.

mcimadamore:
New numbers:

```
Benchmark                                          (size)  Mode  Cnt   Score   Error  Units
StrLenTest.panama_strlen_memsegmentpool_allocator       5  avgt   30  56.377 ? 1.180  ns/op
StrLenTest.panama_strlen_memsegmentpool_allocator      20  avgt   30  59.849 ? 0.759  ns/op
StrLenTest.panama_strlen_memsegmentpool_allocator     100  avgt   30  83.250 ? 3.614  ns/op
StrLenTest.panama_strlen_memsegmentpool_direct          5  avgt   30  43.117 ? 0.525  ns/op
StrLenTest.panama_strlen_memsegmentpool_direct         20  avgt   30  46.241 ? 0.469  ns/op
StrLenTest.panama_strlen_memsegmentpool_direct        100  avgt   30  63.124 ? 0.708  ns/op
```

Yes - it seems to match the unsafe case (which does malloc/free).

rsmogura:
I did more optimizations and now results are 34 ns / op close to arena

Interesting is that biggest gain was due to this change, performance and in generated code
//    while ((int) LOCK.compareAndExchange(this, 0, 1) != 1) { }
    while (!LOCK.compareAndSet(this, 0, 1)) { }

I thought both should give sameresults

rsmogura:
I have following results (did I optimized something too much)
```
Benchmark                                          (size)  Mode  Cnt    Score   Error  Units
StrLenTest.jni_strlen                                   5  avgt   30  109.788 ? 6.553  ns/op
StrLenTest.jni_strlen                                  20  avgt   30  118.230 ? 2.807  ns/op
StrLenTest.jni_strlen                                 100  avgt   30  185.568 ? 1.925  ns/op
StrLenTest.panama_strlen                                5  avgt   30  114.209 ? 0.819  ns/op
StrLenTest.panama_strlen                               20  avgt   30  121.100 ? 3.347  ns/op
StrLenTest.panama_strlen                              100  avgt   30  152.947 ? 1.934  ns/op
StrLenTest.panama_strlen_arena                          5  avgt   30   36.248 ? 0.646  ns/op
StrLenTest.panama_strlen_arena                         20  avgt   30   48.894 ? 1.081  ns/op
StrLenTest.panama_strlen_arena                        100  avgt    9   94.346 ? 7.055  ns/op
StrLenTest.panama_strlen_memsegmentpool_allocator       5  avgt   30   59.867 ? 4.796  ns/op
StrLenTest.panama_strlen_memsegmentpool_allocator      20  avgt   30   57.990 ? 2.443  ns/op
StrLenTest.panama_strlen_memsegmentpool_allocator     100  avgt   30   72.129 ? 0.289  ns/op
StrLenTest.panama_strlen_memsegmentpool_direct          5  avgt   30   34.098 ? 0.128  ns/op
StrLenTest.panama_strlen_memsegmentpool_direct         20  avgt   30   37.481 ? 0.199  ns/op
StrLenTest.panama_strlen_memsegmentpool_direct        100  avgt   30   53.836 ? 0.267  ns/op
StrLenTest.panama_strlen_prefix                         5  avgt   30   26.680 ? 0.340  ns/op
StrLenTest.panama_strlen_prefix                        20  avgt   30   32.913 ? 0.099  ns/op
StrLenTest.panama_strlen_prefix                       100  avgt   30   61.064 ? 0.218  ns/op
StrLenTest.panama_strlen_unsafe                         5  avgt   30   44.110 ? 2.567  ns/op
StrLenTest.panama_strlen_unsafe                        20  avgt   30   46.377 ? 0.151  ns/op
StrLenTest.panama_strlen_unsafe                       100  avgt   30   65.233 ? 1.631  ns/op
StrLenTest.panama_strlen_unsafe_trivial                 5  avgt   30   27.910 ? 0.972  ns/op
StrLenTest.panama_strlen_unsafe_trivial                20  avgt   30   37.179 ? 3.002  ns/op
StrLenTest.panama_strlen_unsafe_trivial               100  avgt   30   62.084 ? 1.610  ns/op
```

mcimadamore:
I agree that now the benchmark seems to behave much better. I'm not aware as to why compareAndSet has such an advantage w.r.t. compareAndExchange.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegmentPool.java
```diff
@@ -0,0 +1,201 @@
...
+      final var bitBoundedSize = 1L << bitBound;
+      segment = allocateNewEntry(segmentDequeue, bitBoundedSize);
+    }
+
+    return segment;
+  }
+
+  private MemorySegment getSegmentForScope(ResourceScope resourceScope, long size, long alignment) {
+    final var segmentEntry = getSegmentEntryBySize(size, alignment);
+    resourceScope.addOnClose(() -> segmentEntry.close());
```
mcimadamore, line 154: There's a potential safety issue here - resourceScope doesn't promise that cleanup actions will be called in a certain order. But I think this should be ok, because the cleanup actions are invoked _after_ the user-provided scope is closed. Meaning that the segments associated with that scope will not be accessible anyway.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegmentPool.java
```diff
@@ -0,0 +1,201 @@
...
+
+  /**
+   * Gets segment from pool or allocates new one. Internally segments are cached.
+   * The size of segment can be larger than requested.
+   *
+   * @param size the size of segment.
+   *
+   * @return segment of size at least `size`
+   */
+  @ForceInline
```
mcimadamore, line 133: I've tried removing @ForceInline here and I did not notice any regression in the benchmark.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegmentPool.java
```diff
@@ -0,0 +1,201 @@
...
+
+  private static void validateMaxSizes(int maxSizes[]) {
+    Arrays.stream(maxSizes).filter(i -> i < 0).findAny()
+        .ifPresent(i -> {
+          throw new IllegalStateException("Invalid max size " + i);
+        });
+  }
+
+  public static class MemoryPoolSegment extends SpinLockQueue.Entry<MemoryPoolSegment> implements AutoCloseable {
+    private final NativeMemorySegmentImpl memorySegment;
```
mcimadamore, line 179: I think this should be MemoryAddress, and the scope should be "late bound" when a segment is retrieved from the queue and requested by a client. Otherwise clients get segments with a different scope than the one they are in - which in itself is ok, but I think it's counterintuitive for an allocator associated with a scope to return segments that last longer than that scope. 

In fact I'm not even 100% that you need a scope for the pool itself (which could mean that the acquire is not needed).

test/micro/org/openjdk/bench/jdk/incubator/foreign/StrLenTest.java
```diff
@@ -145,6 +151,27 @@ public int panama_strlen_unsafe_trivial() throws Throwable {
         return res;
     }
 
+    @Benchmark
+    public int panama_strlen_memsegmentpool_allocator() throws Throwable {
+        try(ResourceScope scope = ResourceScope.newConfinedScope()) {
+            final var allocator = memorySegmentPool.allocatorForScope(scope);
+            return (int)STRLEN.invokeExact(CLinker.toCString(str, allocator).address());
```
mcimadamore, line 158: This comparison is not 100% identical - here we're calling CLinker.toCString, while the low level benchmark just does a bulk copy. For some reason the latter seems to work much better, and I get better numbers if I do that in this benchmark too.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegmentPool.java
rsmogura: Good catch, in case when scope is getting closed in fact segment can be abandoned, and will be permanently removed from pool.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegmentPool.java
rsmogura: I made a performance test, and it looks like creating new segment adds around 8ns to the test results (grow from 32 to 40).

I wonder if this would be a must have for direct pool API. Direct pool does not have to be associated with new scope.

However, for allocator version I think this should be done, as it can prevent leaking shared segment.

mcimadamore: It seems like I misunderstood some of what was going on in the patch. I think I was confused by the fact that memory didn't seem to be ever reclaimed - this led me to the conclusion that having a scope for the pool was redundant. But then, when inspecting a running application closer, I noted that segments are left in the pool - even if the scope is closed (in your original patch). That's because you use CLinker::allocate memory to allocate the segment, which is ok, but then there's never a cleanup action that goes from the pool scope closure to the closure of the segments in the pool. This is, I think, a bug.

rsmogura:
I'll try to sit on this today, and fix remaining comments and other issues, plus add more docs / comments.

rsmogura:
Hi,

It's still bit messy, but I worked on managing releasing on scope(s) close - I hope all cases are handled.

I think the method addOrCleanUpIfFail should be a public API. I think it could provide some argument indicating reason of calling (clean up or fail in clean up), or the  clean up callback interface could be split into two methods cleanup and fail. Just an observation.

I'm adding new set of benchamarks
```
Benchmark                                          (size)  Mode  Cnt    Score   Error  Units
StrLenTest.jni_strlen                                   5  avgt   30  103.968 ? 0.901  ns/op
StrLenTest.jni_strlen                                  20  avgt   30  116.378 ? 2.838  ns/op
StrLenTest.jni_strlen                                 100  avgt   30  180.812 ? 4.355  ns/op
StrLenTest.panama_strlen                                5  avgt   30  118.095 ? 2.401  ns/op
StrLenTest.panama_strlen                               20  avgt   30  120.401 ? 0.515  ns/op
StrLenTest.panama_strlen                              100  avgt   30  162.540 ? 7.410  ns/op
StrLenTest.panama_strlen_arena                          5  avgt   30   35.753 ? 0.345  ns/op
StrLenTest.panama_strlen_arena                         20  avgt   30   48.421 ? 1.136  ns/op
StrLenTest.panama_strlen_arena                        100  avgt    9   94.138 ? 6.826  ns/op
StrLenTest.panama_strlen_memsegmentpool_allocator       5  avgt   30   61.217 ? 3.955  ns/op
StrLenTest.panama_strlen_memsegmentpool_allocator      20  avgt   30   65.178 ? 2.817  ns/op
StrLenTest.panama_strlen_memsegmentpool_allocator     100  avgt   30   78.562 ? 0.389  ns/op
StrLenTest.panama_strlen_memsegmentpool_direct          5  avgt   30   33.454 ? 0.508  ns/op
StrLenTest.panama_strlen_memsegmentpool_direct         20  avgt   30   36.430 ? 0.222  ns/op
StrLenTest.panama_strlen_memsegmentpool_direct        100  avgt   30   53.350 ? 0.349  ns/op
StrLenTest.panama_strlen_prefix                         5  avgt   30   26.351 ? 0.258  ns/op
StrLenTest.panama_strlen_prefix                        20  avgt   30   33.453 ? 0.639  ns/op
StrLenTest.panama_strlen_prefix                       100  avgt   30   61.366 ? 0.351  ns/op
StrLenTest.panama_strlen_unsafe                         5  avgt   30   41.748 ? 0.226  ns/op
StrLenTest.panama_strlen_unsafe                        20  avgt   30   46.536 ? 0.582  ns/op
StrLenTest.panama_strlen_unsafe                       100  avgt   30   66.818 ? 1.365  ns/op
StrLenTest.panama_strlen_unsafe_trivial                 5  avgt   30   27.182 ? 0.148  ns/op
StrLenTest.panama_strlen_unsafe_trivial                20  avgt   30   32.676 ? 0.203  ns/op
StrLenTest.panama_strlen_unsafe_trivial               100  avgt   30   60.266 ? 0.219  ns/op
```
AllocatorsForLongRun.arena                                 1  avgt   10    233.854 ?   0.825  ns/op
AllocatorsForLongRun.arena                                16  avgt   10    692.773 ?   8.502  ns/op
AllocatorsForLongRun.arena                               200  avgt   10   7192.740 ? 300.275  ns/op
AllocatorsForLongRun.malloc_free                           1  avgt   10     29.676 ?   0.655  ns/op
AllocatorsForLongRun.malloc_free                          16  avgt   10    600.062 ?  17.034  ns/op
AllocatorsForLongRun.malloc_free                         200  avgt   10   9672.720 ? 279.714  ns/op
AllocatorsForLongRun.pool_allocator                        1  avgt   10     34.617 ?   0.154  ns/op
AllocatorsForLongRun.pool_allocator                       16  avgt   10    363.034 ?   1.444  ns/op
AllocatorsForLongRun.pool_allocator                      200  avgt   10   4559.017 ?  14.477  ns/op
AllocatorsForLongRun.pool_allocator_exhausted              1  avgt   10     69.087 ?   0.449  ns/op
AllocatorsForLongRun.pool_allocator_exhausted             16  avgt   10   1006.549 ?  28.121  ns/op
AllocatorsForLongRun.pool_allocator_exhausted            200  avgt   10  16081.760 ? 282.773  ns/op
AllocatorsForLongRun.pool_direct                           1  avgt   10     17.413 ?   0.069  ns/op
AllocatorsForLongRun.pool_direct                          16  avgt   10    292.933 ?   1.274  ns/op
AllocatorsForLongRun.pool_direct                         200  avgt   10   5381.742 ?   8.380  ns/op
```
```

test/micro/org/openjdk/bench/jdk/incubator/foreign/AllocatorsForLongRun.java
```diff
@@ -0,0 +1,156 @@
...
+    }
+    allocatedAddresses.forEach(CLinker::freeMemory);
+  }
+
+  /**
+   * Do read to avoid situation allocator will allocate not mapped memory.
+   */
+  private void readSegment(MemorySegment s) {
+    final var size = s.byteSize();
+    for (long l = 0; l <  size; l += 256) {
```
mcimadamore, line 149: this is a loop on `long` - as thing stands - this is gonna defeat all optimizations. For the time being - replace with an `int` loop (and then cast coordinate back to `long`). Or just use `MemoryAccess.setByteAtIndex`.

mcimadamore:
> I think the method addOrCleanUpIfFail should be a public API. I think it could provide some argument indicating reason of calling (clean up or fail in clean up), or the clean up callback interface could be split into two methods cleanup and fail. Just an observation.

I get what you mean - for now I didn't want to give guaranteed call of cleanup actions added by user. That can be achieved using acquire.

As for distinguishing as to whether a cleanup as been called as a result of close() or fail, I don't think that matters: when you have 10 threads racing for the same segment - are you gonna able to distinguish between:

* I tried to add a cleanup action, but the "add" failed, and the cleanup was called
* the segment was closed immediately after I added the cleanup action

I don't think this is a material difference. In both cases, misbehavior signals a lack of adequate countermeasures in user code, IMHO.

test/micro/org/openjdk/bench/jdk/incubator/foreign/AllocatorsForLongRun.java
mcimadamore: Tweaked the loop and numbers didn't change much from yours:

```
Benchmark                                      (allocations)  Mode  Cnt      Score     Error  Units
AllocatorsForLongRun.arena                                 1  avgt   30    185.742 ?   4.349  ns/op
AllocatorsForLongRun.arena                                16  avgt   30    616.261 ?  14.974  ns/op
AllocatorsForLongRun.arena                               200  avgt   30   6574.496 ?  55.602  ns/op
AllocatorsForLongRun.malloc_free                           1  avgt   30     25.888 ?   0.272  ns/op
AllocatorsForLongRun.malloc_free                          16  avgt   30    602.258 ?  11.776  ns/op
AllocatorsForLongRun.malloc_free                         200  avgt   30  10126.972 ? 151.182  ns/op
AllocatorsForLongRun.pool_allocator                        1  avgt   30     35.907 ?   0.474  ns/op
AllocatorsForLongRun.pool_allocator                       16  avgt   30    378.874 ?   8.533  ns/op
AllocatorsForLongRun.pool_allocator                      200  avgt   30   4489.656 ?  40.615  ns/op
AllocatorsForLongRun.pool_allocator_exhausted              1  avgt   30     65.074 ?   3.399  ns/op
AllocatorsForLongRun.pool_allocator_exhausted             16  avgt   30    994.809 ?  22.971  ns/op
AllocatorsForLongRun.pool_allocator_exhausted            200  avgt   30  16247.051 ? 223.768  ns/op
AllocatorsForLongRun.pool_direct                           1  avgt   30     15.827 ?   0.398  ns/op
AllocatorsForLongRun.pool_direct                          16  avgt   30    269.499 ?   3.384  ns/op
AllocatorsForLongRun.pool_direct                         200  avgt   30   3491.204 ?  35.959  ns/op
```

Seems like 16 allocations is the break even for arena - after which (on 200) arena is better than malloc (I can only imagine that advantage of arena will keep growing with number of allocations). Malloc/free is still surprisingly good, all things considered, especially hard to beat on single shot allocations.

The problem I see with the pool strategy is that it's faster than malloc - but not in a radical way (there's no 10x here). And you have to consider best case, and worst case (the best case is better than malloc, the worst case, exhausted, is worse). So it looks like something that can be a great thing, if that's what a program needs, and provided it's used as intended - but it doesn't seem (yet?) to deliver that kind of horizontal, across the board, scaling that would justify its inclusion in the API (although it's great to see that such an allocator can be written on top of the API).

What I like about the pool though, is the approach you had for the API - I think that when we will look at allocators again (as I said, we did have some other allocators we were looking at, not too different from what you are trying to do here), I think the API that will be offered will probably be very similar to what you have in here - as I think it's spot on, and plays to the advantages of the new memory API.

mcimadamore: btw, things improve considerably if locking code is removed:

```
Benchmark                                      (allocations)  Mode  Cnt      Score     Error  Units
AllocatorsForLongRun.pool_allocator                        1  avgt   30     32.553 ?   0.583  ns/op
AllocatorsForLongRun.pool_allocator                       16  avgt   30    333.564 ?   3.942  ns/op
AllocatorsForLongRun.pool_allocator                      200  avgt   30   3552.176 ?  43.852  ns/op
AllocatorsForLongRun.pool_allocator_exhausted              1  avgt   30     62.226 ?   1.408  ns/op
AllocatorsForLongRun.pool_allocator_exhausted             16  avgt   30    890.346 ?  15.812  ns/op
AllocatorsForLongRun.pool_allocator_exhausted            200  avgt   30  16232.458 ? 562.777  ns/op
AllocatorsForLongRun.pool_direct                           1  avgt   30     14.983 ?   0.134  ns/op
AllocatorsForLongRun.pool_direct                          16  avgt   30    246.688 ?   3.411  ns/op
AllocatorsForLongRun.pool_direct                         200  avgt   30   2990.063 ?  34.049  ns/op
```

So I think providing an alternate queue implementation to use if the pool scope is confined could make sense.

mcimadamore: Did some other experiments comparing direct_pool vs. arena with different allocation sizes:

```
Benchmark                         (allocations)  Mode  Cnt      Score     Error  Units
AllocatorsForLongRun.pool_direct              1  avgt   30     17.313 ?   0.413  ns/op
AllocatorsForLongRun.pool_direct             10  avgt   30    181.496 ?   2.163  ns/op
AllocatorsForLongRun.pool_direct            100  avgt   30   1759.230 ?  24.107  ns/op
AllocatorsForLongRun.pool_direct           1000  avgt   30  32862.773 ? 235.450  ns/op

Benchmark                   (allocations)  Mode  Cnt      Score     Error  Units
AllocatorsForLongRun.arena              1  avgt   30    183.450 ?   4.108  ns/op
AllocatorsForLongRun.arena             10  avgt   30    462.525 ?   7.531  ns/op
AllocatorsForLongRun.arena            100  avgt   30   3208.686 ?  41.565  ns/op
AllocatorsForLongRun.arena           1000  avgt   30  40230.111 ? 678.682  ns/op

Benchmark                         (allocations)  Mode  Cnt      Score     Error  Units
AllocatorsForLongRun.malloc_free              1  avgt   30     26.440 ?   0.286  ns/op
AllocatorsForLongRun.malloc_free             10  avgt   30    349.220 ?   6.760  ns/op
AllocatorsForLongRun.malloc_free            100  avgt   30   4755.480 ? 124.909  ns/op
AllocatorsForLongRun.malloc_free           1000  avgt   30  55986.975 ? 764.037  ns/op
```

Admittedly, this doesn't look too good for arena allocation. And I couldn't at first understand why. But then I realized that arena allocator is based on MemorySegment::allocateNative, not the fast/dirty CLinker::allocate/free. For fun I tried to tweak the arena allocator to use those instead:

```
Benchmark                   (allocations)  Mode  Cnt      Score     Error  Units
AllocatorsForLongRun.arena              1  avgt   30     73.263 ?   6.777  ns/op
AllocatorsForLongRun.arena             10  avgt   30    236.181 ?  23.822  ns/op
AllocatorsForLongRun.arena            100  avgt   30   1902.569 ?  70.081  ns/op
AllocatorsForLongRun.arena           1000  avgt   30  22995.554 ? 593.434  ns/op
```

Now, apart for the one shot allocation, arena performs similar to the (direct) pool allocator, and quite better than malloc/free. It is a bit sad that arena loses so much due to "secondary" stuff like memory zeroing, reserving memory etc.
Perhaps this open the door to an additional arena allocator factory which takes a segment allocator, instead of a ResourceScope (so that if users are ok with loss of initialization, they have a way to opt out).

rsmogura:
I made a change to test, and results look strange.

Looks like using pooled buffer is very slow. I think this is something what was pointed out during test with StrLen, but now this phenomena can be better observed when manipulating number of read bytes.

I can't find right now what causes this.

test/micro/org/openjdk/bench/jdk/incubator/foreign/AllocatorsForLongRun.java
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  /**
+   * Do read to avoid situation allocator will allocate not mapped memory.
+   */
+  private void readSegment(MemorySegment s) {
+    final var size = (int) s.byteSize();
+    for (int idx = 0; idx <  size; idx += 1024) {
+//      MemoryAccess.setByteAtOffset(s, l, (byte)0);
+      BYTE.set(s, 0, (byte) 1);
```
mcimadamore, line 162: You need to cast the offset (idx?) back to a long when you call BYTE.set

mcimadamore: That said, IMHO, in the new benchmark, the cost of reading dominates everything else. It's not an issue with the memory access API either - replacing read with Unsafe::putInt it still slow. Honestly, since you are after measuring allocation performances, I'm not sure if `readSegment` even belongs here. The more allocations you have, with bigger sized segments, the more read will dominate anything else, to the point of rendering the benchmark not very useful to measure what you want to measure. In fact, even the `strlen` call was an issue in the previous benchmark - as the size of string grows, the time spent in copying data, and strlen grows as well, which then masks the allocation costs. Here you have a similar situation, I think.

rsmogura: That's a pity JMH does not allow sub measurements

rsmogura: I think that issue is related how JVM shapes code. If I'll write something like this 
readSegment(s.memoryAddress.asSegment(size, scope));
so I'll allocate a new segment based on address than the read performance goes up, and it's same as malloc version

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegmentPool.java
```diff
@@ -0,0 +1,298 @@
...
+    final var segmentEntry = getMemoryPoolSegment(bitBound);
+
+    ((ResourceScopeImpl) resourceScope).addOrCleanupIfFail(new ResourceCleanup() {
+      @Override
+      public void cleanup() {
+        segmentEntry.close();
+      }
+    });
+
+    return segmentEntry.memoryAddress.asSegment(1L << bitBound, resourceScope);
```
mcimadamore, line 199: I wouldn't be surprised if this was the issue - the segment is created with a size that is determined by a long shift. If this segment gets scalarized, I wonder if the VM is able to see that the segment size is actually an int and not a long. Worth a shot.

test/micro/org/openjdk/bench/jdk/incubator/foreign/AllocatorsForLongRun.java
mcimadamore: See my other code comment - I think the issue here might be that the VM can't clearly see the segment size, so counted loop optimizations are not kicking in.

mcimadamore: Yeah - I confirm that if you add another getter for `MemoryPoolSegment::memorySegment`, like this:

```
@ForceInline
    public MemorySegment memorySegment(long size) {
      return memoryAddress.asSegment(size, scope);
    }
```
    
Then performance improve considerably. This smells like an inlining issue - e.g. the `size` value that is passed by the code to `getSegmentEntryBySize` is not recognized as the size of the segment - C2 doesn't seem to be able to see through it. There is too much control flow inside `getSegmentEntryBySize` - the trick is to create a "fresh" segment as close as possible to the application code.

The safe version of the API already does this, in fact suffers less from this issue (it's slower than malloc for smaller allocation sizes, but that has always been the case - but it's faster than the direct version!).

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegmentPool.java
rsmogura: I checked it, and I get same results. I'll shortly send an updated version

rsmogura: Catched! It's not int, I use calculate leading zeros, I guess this a barrier for VM. Replacing size by size with align makes it fast.

But yes it's about size.

rsmogura:
New set of benchmarks

```
Benchmark                                      (allocations)  Mode  Cnt         Score       Error  Units
AllocatorsForLongRun.arena                                 1  avgt   10      8578.213 ?   119.631  ns/op
AllocatorsForLongRun.arena                                16  avgt   10    190038.468 ?  1879.850  ns/op
AllocatorsForLongRun.arena                               200  avgt   10  12738712.433 ? 92041.354  ns/op
AllocatorsForLongRun.malloc_free                           1  avgt   10        43.774 ?     0.668  ns/op
AllocatorsForLongRun.malloc_free                          16  avgt   10      2639.027 ?   433.626  ns/op
AllocatorsForLongRun.malloc_free                         200  avgt   10     26356.097 ?   191.074  ns/op
AllocatorsForLongRun.pool_allocator                        1  avgt   10        30.470 ?     0.062  ns/op
AllocatorsForLongRun.pool_allocator                       16  avgt   10       323.179 ?     0.611  ns/op
AllocatorsForLongRun.pool_allocator                      200  avgt   10      3884.727 ?    10.954  ns/op
AllocatorsForLongRun.pool_allocator_exhausted              1  avgt   10        80.138 ?     1.547  ns/op
AllocatorsForLongRun.pool_allocator_exhausted             16  avgt   10      3453.099 ?   696.754  ns/op
AllocatorsForLongRun.pool_allocator_exhausted            200  avgt   10     31563.155 ?   153.992  ns/op
AllocatorsForLongRun.pool_direct                           1  avgt   10        18.316 ?     0.092  ns/op
AllocatorsForLongRun.pool_direct                          16  avgt   10       261.271 ?     3.951  ns/op
AllocatorsForLongRun.pool_direct                         200  avgt   10      3397.057 ?    35.783  ns/op
```
With write
```
Benchmark                                      (allocations)  Mode  Cnt         Score        Error  Units
AllocatorsForLongRun.arena                                 1  avgt   10     12101.342 ?   2231.185  ns/op
AllocatorsForLongRun.arena                                16  avgt   10    243313.864 ?   5256.439  ns/op
AllocatorsForLongRun.arena                               200  avgt   10  13135973.454 ?  37170.906  ns/op
AllocatorsForLongRun.malloc_free                           1  avgt   10      1571.583 ?     22.135  ns/op
AllocatorsForLongRun.malloc_free                          16  avgt   10     65492.794 ?  22068.908  ns/op
AllocatorsForLongRun.malloc_free                         200  avgt   10   6800966.660 ?  39904.879  ns/op
AllocatorsForLongRun.pool_allocator                        1  avgt   10      1420.386 ?      5.112  ns/op
AllocatorsForLongRun.pool_allocator                       16  avgt   10     29433.570 ?    160.868  ns/op
AllocatorsForLongRun.pool_allocator                      200  avgt   10    640035.597 ?  38522.118  ns/op
AllocatorsForLongRun.pool_allocator_exhausted              1  avgt   10      1526.161 ?     18.930  ns/op
AllocatorsForLongRun.pool_allocator_exhausted             16  avgt   10    397319.239 ?   6746.863  ns/op
AllocatorsForLongRun.pool_allocator_exhausted            200  avgt   10  10774953.962 ? 761368.809  ns/op
AllocatorsForLongRun.pool_direct                           1  avgt   10      2242.986 ?     15.156  ns/op
AllocatorsForLongRun.pool_direct                          16  avgt   10     46143.779 ?    442.033  ns/op
AllocatorsForLongRun.pool_direct                         200  avgt   10    831948.538 ?   6819.482  ns/op
```

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegmentPool.java
mcimadamore: Good catch - I'm not sure I follow the various @DontInline and such - I hope the code is not too tailored to the specific benchmark. But the numbers are rather good - now even the clean API gives numbers that beat malloc considerably (I honestly would leave the direct API alone). This is indeed a good direction - while I'm not planning to add something like this at this stage, we will, at some point after 17, ramp up some discussion around allocation strategy, and it seems like we got a good contender here ;-)

rsmogura:
That's depressing to admit, but the slow size reading is not caused by external circumstances.

I was returning cached segment to the client using direct API.

This segment can have bigger size than requested (bucketing), and one allocated with malloc. So if upper bound of read was segment size more data could be read. So I have to admit it's a bug in test.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegmentPool.java
rsmogura: I should thank you for help and motivation.

Now I've got highly tuned allocator to implement native I/O.

That's will be critical to provide buffers between native and heap memory.

Thanks again!

mcimadamore:
> That's depressing to admit, but the slow size reading is not caused by external circumstances.
> 
> I was returning cached segment to the client using direct API.
> 
> This segment can have bigger size than requested (bucketing), and one allocated with malloc. So if upper bound of read was segment size more data could be read. So I have to admit it's a bug in test.

I see - but shouldn't the high level API resize the segment?

I think a more robust approach would be:

* drop the low level API
* implement pool entries just in terms of a MemoryAddress (what has been returned by malloc)
* when the pool is created, add a cleanup action, so that all the entries in the pool are "freed"
* when clients request a segment, and an entry exists of matching size, take the address and create a sized segment from it with the right scope
* as now, register a cleanup action on the requesting scope, so that the segments allocated are returned to the pool when closed

rsmogura:
Hi @mcimadamore ,

I hope it looks more better. I wonder if you see some obvious issues here (I understand it's probably still not this stage to handle more complicated memory operations in Panama).

After finding issue with tests, a lot of things get simplified.

To summarize changes:
Dropped internal API

Result segment is created based on address (I miss a bit a fast way to create segment without checking scope - something like plain constructor call)

Added separated cleaner to clean pool on pool's scope closure - in case of implicate scope the hard reference to scope was kept via cleaner ref -> lambda to free method -> this -> scope. That's somehow quite fragile part for cleaner as it's easy to create memory leak.

I think I've found possible race in shared resource list (but put it in other thread) - if two add threads t1, t2 will start to race with close thread. They can leave list in following state cleaner2 -> cleaner 1 -> close marker.

While outcome will be t2 -> ok, t1 -> exception, cleaner -> cleans what was before.

Thanks for previews reviews & help,
Rado

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegmentPool.java
```diff
@@ -0,0 +1,375 @@
...
+      }
+    }
+
+    @ForceInline
+    private MemoryPoolItem(SpinLockQueue<MemoryPoolItem> queue, MemoryAddress sourceAddress, long size, ResourceScope scope) {
+      super();
+      this.owner = queue;
+      this.sourceAddress = sourceAddress;
+
+      ((ResourceScopeImpl) scope).addOrCleanupIfFail(new ResourceCleanup() {
```
mcimadamore, line 316: Not 100% sure you need this. You already have added a cleanup action which calls `ClearRunnable`. Why not just deallocating stuff from there? In fact, I think the scope should be also moved to the spin lock queue - each queue has a scope, and when the scope is closed, the queue is locked, and all entries inside it are released. This will save a cleanup action addition on every new memory pool item (that's how I'd do it at least).

mcimadamore:
I agree the latest version looks much cleaner - I think you can simplify the MemoryPoolItem further, by just storing an address and move all cleanup logic in CleanerRunnable (which would be run when the pool scope is closed).

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegmentPool.java
rsmogura: That's very fair point!!! Thanks.

After calling cleaner queue max size is set to 0, so any pending entry will have to be released on close.

Outstanding entires can be closed using iterator.

rsmogura:
With the latest changes I could go down from 61ns to 54ns per operation.

I hope I didn't mess with dropping VarHandles but it gave 1ns.

Clearing MemoryPoolItem helped a bit, too.

Most important, replacing cleaner from public API by ResourceCleaner gave big win.

I still, think having a good indicator if cleaner was added or not is a good idea (it can be a Boolean return or concrete exception). For clients wanting to use public API, IllegalStateException is not 100% sure indicator. 

Once again I would like to thank you @mcimadamore for a good review! You helped a lot!

mcimadamore:
Ping: does this PR needs to be open? (Perhaps it might be useful to update and make it a draft JEP, as an example of how the allocator API could be used?)

Aungkhant884 (approved):
Hi @Aungkhant884, thanks for making a comment in an OpenJDK project!

All comments and discussions in the OpenJDK Community must be made available under the OpenJDK [Terms of Use](https://openjdk.java.net/legal/tou/). If you already are an OpenJDK [Author](https://openjdk.java.net/bylaws#author), [Committer](https://openjdk.java.net/bylaws#committer) or [Reviewer](https://openjdk.java.net/bylaws#reviewer), please click [here](https://bugs.openjdk.java.net/secure/CreateIssue.jspa?pid=11300&issuetype=1) to open a new issue so that we can record that fact. Please Use "Add GitHub user Aungkhant884 for the summary.

If you are not an OpenJDK Author, Committer or Reviewer, simply check the box below to accept the OpenJDK Terms of Use for your comments.

 - [ ]  I agree to the [OpenJDK Terms of Use](https://openjdk.java.net/legal/tou/) for all comments I make in a project in the [OpenJDK GitHub organization](https://github.com/openjdk).

Your comment will be automatically restored once you have accepted the OpenJDK [Terms of Use](https://openjdk.java.net/legal/tou/).

Aungkhant884 (commented):
Hi @Aungkhant884, thanks for making a comment in an OpenJDK project!

All comments and discussions in the OpenJDK Community must be made available under the OpenJDK [Terms of Use](https://openjdk.java.net/legal/tou/). If you already are an OpenJDK [Author](https://openjdk.java.net/bylaws#author), [Committer](https://openjdk.java.net/bylaws#committer) or [Reviewer](https://openjdk.java.net/bylaws#reviewer), please click [here](https://bugs.openjdk.java.net/secure/CreateIssue.jspa?pid=11300&issuetype=1) to open a new issue so that we can record that fact. Please Use "Add GitHub user Aungkhant884 for the summary.

If you are not an OpenJDK Author, Committer or Reviewer, simply check the box below to accept the OpenJDK Terms of Use for your comments.

 - [ ]  I agree to the [OpenJDK Terms of Use](https://openjdk.java.net/legal/tou/) for all comments I make in a project in the [OpenJDK GitHub organization](https://github.com/openjdk).

Your comment will be automatically restored once you have accepted the OpenJDK [Terms of Use](https://openjdk.java.net/legal/tou/).

rsmogura:
Hi @mcimadamore - I hope you have good day. I wonder if this PR should be kept or closed. I remember that plan was to introduce it after stabilisation of API, but not sure if this plan have not changed?

mcimadamore:
> Hi @mcimadamore - I hope you have good day. I wonder if this PR should be kept or closed. I remember that plan was to introduce it after stabilisation of API, but not sure if this plan have not changed?

Ho @rsmogura - while we do have plans to include more allocators/arenas out of the box, I'm not too sure a PR would be the best way to keep this initiative alive. I'd say let's close the PR (otherwise folks might thing this is going to get integrated soon, which might be misleading), and maybe drop the code in some branch, where it can be kept in sync with mainline?

PR #510, opened 2021-04-19
Improve ResourceScope javadoc

mcimadamore:
Chris and I had some discussions around the new `ResourceScope` API, and realized that the toplevel javadoc doesn't make it easy for developers to grasp all the various dimensions of resource scopes.
This is mostly caused by the fact that the concept of implicit closure is used to refer to _all_ scopes that support some form of GC-backed closure - and this brings confusion, as we have added the concept of "implicit" scopes to denote those scopes that *cannot* be closed explicitly.

I have rewritten the javadoc to talk about explicit and implicit scopes; we now say, in the section on explicit scopes, that explicit scopes can sometimes be associated with a Cleaner, but it is now clearer that that's mostly a fallback option.

When working on this, I think that we should probably make `ResourceScope::acquire` fail for implicit scope - either return null or throw. In other words, while acquiring handles makes sense to prevent deterministic deallocation, it seems like it doesn't make a lot of sense to prevent implicit deallocation (we already have reachability fences for this). What the implementation does right now permanently capture the scope instance in a handle instance, so the original implicit scope will never be closed as long as handles are reachable, even after close. In other words, having an handle that can be closed doesn't make sense for implicit scopes, since `Handle::close` doesn't do anything intersting in that case (or is not sufficient).

Any thoughts?

---------
### Progress

### Reviewers
 * [Chris Hegarty](https://openjdk.java.net/census#chegar) (@ChrisHegarty - Committer) ⚠️ Review applies to ed498870936a85ff6dac0809a932dac63c48563f

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/510/head:pull/510` \
`$ git checkout pull/510`

`$ git checkout pull/510` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/510/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 510`

`$ git pr show -t 510`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/510.diff">https://git.openjdk.java.net/panama-foreign/pull/510.diff</a>

</details>

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/ResourceScope.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/ResourceScope.java
@@ -43,7 +43,7 @@ import java.util.Spliterator;
  *
  * Resource scopes obtained from {@link #newConfinedScope()}, {@link #newSharedScope()} support <em>deterministic deallocation</em>;
  * We call these resource scopes <em>explicit scopes</em>. Explicit resource scopes can be closed explicitly (see {@link ResourceScope#close()}).
- * When a resource scope is closed, it is no longer <em>alive</em> (see {@link #isAlive()}, and subsequent operation on
+ * When a resource scope is closed, it is no longer <em>alive</em> (see {@link #isAlive()}, and subsequent operations on
  * resources associated with that scope (e.g. attempting to access a {@link MemorySegment} instance) will fail with {@link IllegalStateException}.
  * <p>
  * Closing a resource scope will cause all the cleanup actions associated with that scope (see {@link #addOnClose(Runnable)}) to be called.
@@ -75,7 +75,8 @@ import java.util.Spliterator;
  * <p>
  * An important implicit resource scope is the so called {@link #globalScope() global scope}; the global scope is
  * an implicit scope that is guaranteed to never become <a href="../../../java/lang/ref/package.html#reachability">unreachable</a>.
- * As a results, resources associated to the global scope are never cleaned up.
+ * As a results, the global scope will never attempt to release resources associated with it. Such resources must, where
+ * needed, be managed independently by clients.
  *
  * <h2><a id = "thread-confinement">Thread confinement</a></h2>
  *
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/ResourceScope.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/ResourceScope.java
@@ -39,11 +39,12 @@ import java.util.Spliterator;
  * with a resource scope can only be accessed while the resource scope is <em>alive</em> (see {@link #isAlive()}),
  * and by the thread associated with the resource scope (if any).
  *
- * <h2>Explicit closure</h2>
+ * <h2>Explicit resource scopes</h2>
  *
- * Resource scopes created using one of the factories in this class can be closed explicitly (see {@link ResourceScope#close()}).
+ * Resource scopes obtained from {@link #newConfinedScope()}, {@link #newSharedScope()} support <em>deterministic deallocation</em>;
+ * We call these resource scopes <em>explicit scopes</em>. Explicit resource scopes can be closed explicitly (see {@link ResourceScope#close()}).
  * When a resource scope is closed, it is no longer <em>alive</em> (see {@link #isAlive()}, and subsequent operation on
- * resources derived from that scope (e.g. attempting to access a {@link MemorySegment} instance) will fail with {@link IllegalStateException}.
+ * resources associated with that scope (e.g. attempting to access a {@link MemorySegment} instance) will fail with {@link IllegalStateException}.
  * <p>
  * Closing a resource scope will cause all the cleanup actions associated with that scope (see {@link #addOnClose(Runnable)}) to be called.
  * Moreover, closing a resource scope might trigger the releasing of the underlying memory resources associated with said scope; for instance:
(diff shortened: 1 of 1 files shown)


src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/ResourceScope.java
```diff
@@ -55,21 +56,26 @@
...
+ * <em>implicit scopes</em>. Calling {@link #close()} on an implicit resource scope always results in an exception.
+ * Resources associated with implicit scopes are released once the scope instance becomes
+ * <a href="../../../java/lang/ref/package.html#reachability">unreachable</a>.
  * <p>
- * Some managed resource scopes are implicitly managed (see {@link #newImplicitScope()}, {@link #globalScope()}, and are said to be <em>implicit scopes</em>.
- * An implicit resource scope only features implicit closure, and always throws an {@link UnsupportedOperationException}
- * when the {@link #close()} method is called directly.
+ * An important implicit resource scope is the so called {@link #globalScope() global scope}; the global scope is
+ * an implicit scope that is guaranteed to never become <a href="../../../java/lang/ref/package.html#reachability">unreachable</a>.
+ * As a results, resources associated to the global scope are never cleaned up.
```
PaulSandoz, line 78: By "cleaned up" do you mean that no custom clean up actions added to the scope take effect (they are ignored)? A heap segment is associated with the global scope, but it does not imply the heap segment's backing memory (an array) is always kept alive.  We could say that any resource associated with the global scope is "external" and managed independently (by other means), or not at all (such as the native everything segment)

PaulSandoz (commented):
Regarding `GLOBAL_SCOPE.acquire`, it seems harmless. Writing uniform scope logic might get harder if the global scope behaves differently, but i don't have a strong sense of how libraries might operate on scope.s It might well be the case if something accepts and wraps a segment e.g. a matrix whose element buffer is allocated independently/externally and it needs to be kept alive for the life time of the matrix.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/ResourceScope.java
mcimadamore: So, the global scope is a singleton scope that is always alive. Since it's always alive, it's never closed, which means cleanup actions never get a chance to get called, yes. I'm not sure talking about resources being "external" makes the point clearer? The thing I'm trying to get across is that global scope is good for things that do **not** require cleanup (e.g. are assumed to be alive for the entire duration of the application).

mcimadamore:
> Regarding `GLOBAL_SCOPE.acquire`, it seems harmless. Writing uniform scope logic might get harder if the global scope behaves differently, but i don't have a strong sense of how libraries might operate on scope.s It might well be the case if something accepts and wraps a segment e.g. a matrix whose element buffer is allocated independently/externally and it needs to be kept alive for the life time of the matrix.

Not just about GLOBAL_SCOPE - but also about newImplicitScope. Yes, there is a tension between uniform usage and how meaningful these operations are. My intuition was to exploit a trick of try with resources which is tolerant to `null`s.

So, if you do

```
try (var handle = scope.acquire()) {
   ...
}
```
And, sometimes, the handle is null, the TWR is still valid (javac inserts a null check in the finalizer), but at least you don't have an odd handle on your hands which doesn't really do what it says on the tin (e.g. calling `Handle::close` does nothing).

mcimadamore:
> > Regarding `GLOBAL_SCOPE.acquire`, it seems harmless. Writing uniform scope logic might get harder if the global scope behaves differently, but i don't have a strong sense of how libraries might operate on scope.s It might well be the case if something accepts and wraps a segment e.g. a matrix whose element buffer is allocated independently/externally and it needs to be kept alive for the life time of the matrix.
> 
> Not just about GLOBAL_SCOPE - but also about newImplicitScope. Yes, there is a tension between uniform usage and how meaningful these operations are. My intuition was to exploit a trick of try with resources which is tolerant to `null`s.
> 
> So, if you do
> 
> ```
> try (var handle = scope.acquire()) {
>    ...
> }
> ```
> 
> And, sometimes, the handle is null, the TWR is still valid (javac inserts a null check in the finalizer), but at least you don't have an odd handle on your hands which doesn't really do what it says on the tin (e.g. calling `Handle::close` does nothing).

On this topic, I think an API change that I've been considering recently might help things a bit:

```
var handle = scope.acquire();
try {
   // critical region
} finally {
   scope.release(handle);
}
```

This is lower level, and doesn't expose try with resources. It's not a big deal, I think, as using try with resources in the way the current API suggests generates annoying warnings (as the handle is never used inside the try block). The nice thing about this approach is that we can:

* return a dummy handle (e.g. a constant, well known to the system) for all implicit scopes
* in idiomatic code, the scope is always reachable at the end of the try/finally

In other words, writing code this way will make it work, both when deterministic deallocation is used (because acquire will return a "real" handle) and when implicit deallocation is used (because the scope is still reachable, so its resources cannot be freed by the GC in the middle of the try block).

I think removing the `close` method from `Handle` mitigates a lot of the concerns I have on the current acquire API.

ChrisHegarty:
>>> ...
> > 
> > And, sometimes, the handle is null, the TWR is still valid (javac inserts a null check in the finalizer), but at least you don't have an odd handle on your hands which doesn't really do what it says on the tin (e.g. calling `Handle::close` does nothing).

In my experiments, I found that most usage of a resource handle in a TWR construct raised a warning about the AutoCloseable not being used, which I then had to suppress. (which is not a concern with the proposal below)

> On this topic, I think an API change that I've been considering recently might help things a bit:
> 
> ```
> var handle = scope.acquire();
> try {
>    // critical region
> } finally {
>    scope.release(handle);
> }
> ```

This is a familiar idiom, say when using explicit j.u.c locks. Nice.

> This is lower level, and doesn't expose try with resources. It's not a big deal, I think, as using try with resources in the way the current API suggests generates annoying warnings (as the handle is never used inside the try block).

Agree.

> The nice thing about this approach is that we can:
> 
> * return a dummy handle (e.g. a constant, well known to the system) for all implicit scopes
> * in idiomatic code, the scope is always reachable at the end of the try/finally

Yes, this is nice.

> In other words, writing code this way will make it work, both when deterministic deallocation is used (because acquire will return a "real" handle) and when implicit deallocation is used (because the scope is still reachable, so its resources cannot be freed by the GC in the middle of the try block).
> 
> I think removing the `close` method from `Handle` mitigates a lot of the concerns I have on the current acquire API. If this looks good I can pursue in a separate patch.

Yes, please.

> For the records, if not using try-with-resources ends up being too annoying, we can always add an high-order function which does acquire/release on top. But current experiments with async IO don't seem to suggest that this is an issue.

In my experiments, I want to restrict a scattering/gathering I/O operation to buffer views over at most one explicit scope. Doing this requires me to be able to determine scope equality either by carrying the scope explicitly or possibly retrieving it from the handle (which is not currently possible). To be kind to the GC, I think I can keep the former solution, but the latter would be convenient.

mcimadamore:
> In my experiments, I want to restrict a scattering/gathering I/O operation to buffer views over at most one explicit scope. Doing this requires me to be able to determine scope equality either by carrying the scope explicitly or possibly retrieving it from the handle (which is not currently possible). To be kind to the GC, I think I can keep the former solution, but the latter would be convenient.

I think we want to land in a place where acquiring an implicit scope is a no-op. I'm fine with having a scope() accessor on the Handle class - but note that if we require this accessor to be always precise, even in the case of implicit scopes, that would mean that when acquiring an implicit scope we have to create a _new_ handle, just so that we can return the right scope. Worse, the handle will interact with Cleaners (as the handle will keep a strong reference to the original scope). I think mixing reachability with handles is wrong - acquiring an handle on an implicit should **not** make a scope always reachable (until the handle itself becomes unreachable), as that would make it even harder to reason about reachability.

So, the thing I can offer here is a scope() accessor which only works for explicit scopes. That is, the javadoc for ResourceScope::acquire can say that, if the acquired scope is an implicit scope a "global" handle will be returned, owned by the global scope - and that releasing a global handle is idempotent.

This way, the accessor will give you the info you want, in case where acquire does something useful - but it won't give you any info in case acquire is  a no-op, which should be ok?

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/ResourceScope.java
```diff
@@ -39,11 +39,12 @@
  * with a resource scope can only be accessed while the resource scope is <em>alive</em> (see {@link #isAlive()}),
  * and by the thread associated with the resource scope (if any).
  *
- * <h2>Explicit closure</h2>
+ * <h2>Explicit resource scopes</h2>
  *
- * Resource scopes created using one of the factories in this class can be closed explicitly (see {@link ResourceScope#close()}).
+ * Resource scopes obtained from {@link #newConfinedScope()}, {@link #newSharedScope()} support <em>deterministic deallocation</em>;
+ * We call these resource scopes <em>explicit scopes</em>. Explicit resource scopes can be closed explicitly (see {@link ResourceScope#close()}).
  * When a resource scope is closed, it is no longer <em>alive</em> (see {@link #isAlive()}, and subsequent operation on
```
ChrisHegarty, line 46: possibly a typo here "operationS on"

ChrisHegarty (approved):

mcimadamore:
> > In my experiments, I want to restrict a scattering/gathering I/O operation to buffer views over at most one explicit scope. Doing this requires me to be able to determine scope equality either by carrying the scope explicitly or possibly retrieving it from the handle (which is not currently possible). To be kind to the GC, I think I can keep the former solution, but the latter would be convenient.
> 
> I think we want to land in a place where acquiring an implicit scope is a no-op. I'm fine with having a scope() accessor on the Handle class - but note that if we require this accessor to be always precise, even in the case of implicit scopes, that would mean that when acquiring an implicit scope we have to create a _new_ handle, just so that we can return the right scope. Worse, the handle will interact with Cleaners (as the handle will keep a strong reference to the original scope). I think mixing reachability with handles is wrong - acquiring an handle on an implicit should **not** make a scope always reachable (until the handle itself becomes unreachable), as that would make it even harder to reason about reachability.
> 
> So, the thing I can offer here is a scope() accessor which only works for explicit scopes. That is, the javadoc for ResourceScope::acquire can say that, if the acquired scope is an implicit scope a "global" handle will be returned, owned by the global scope - and that releasing a global handle is idempotent.
> 
> This way, the accessor will give you the info you want, in case where acquire does something useful - but it won't give you any info in case acquire is a no-op, which should be ok?

PaulSandoz (approved):

PR #511, opened 2021-04-20
Investigate alternate strategy to acquire resource scopes

mcimadamore:
The code in this PR demonstrates a possible, alternate approach to resource scope acquire.

Instead of having `ResourceScope.Handle` be `AutoCloseable` (which is problematic, given that `javac` issues warning when an auto closeable is not used inside the body of the try block) this patch adds a new method on resource scopes, namely `ResourceScope::release(Handle)`.

The semantics of this method is similar to the previous `Handle::close` - one thing that is different is how implicit scopes are handled - instead of trying to keep the implicit scope alive through the handle instance, here we simply return a "dummy" handle, which does nothing.

The idea is that if you write idiomatic code like:

```java
var handle = scope.acquire();
try {
   // critical region
} finally {
   scope.release(handle); 
} 
```

This will work for both explicit scopes (since acquiring actively prevents calls to `ResourceScope::close`) - but also for implicit scopes, since the `release` call inside the finally block will effectively require the scope to be reachable at this stage (see reachability fence inside `ResourceScopeImpl::release`).

---------
### Progress

### Reviewers
 * [Chris Hegarty](https://openjdk.java.net/census#chegar) (@ChrisHegarty - Committer) ⚠️ Review applies to b74bba6c5dae89d5b0be947a1df03080be06f5a2

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/511/head:pull/511` \
`$ git checkout pull/511`

`$ git checkout pull/511` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/511/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 511`

`$ git pr show -t 511`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/511.diff">https://git.openjdk.java.net/panama-foreign/pull/511.diff</a>

</details>

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/ResourceScopeImpl.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/ResourceScopeImpl.java
@@ -126,11 +126,15 @@ public abstract class ResourceScopeImpl implements ResourceScope, ScopedMemoryAc
 
     @Override
     public void release(Handle handle) {
-        Objects.requireNonNull(handle);
-        if (!isImplicit() && handle.scope() != this) {
-            throw new IllegalArgumentException("Cannot release an handle acquired from another scope");
+        try {
+            Objects.requireNonNull(handle);
+            if (!isImplicit() && handle.scope() != this) {
+                throw new IllegalArgumentException("Cannot release an handle acquired from another scope");
+            }
+            ((HandleImpl) handle).close();
+        } finally {
+            Reference.reachabilityFence(this);
         }
-        ((HandleImpl)handle).close();
     }
 
     /**
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/ResourceScopeImpl.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/ResourceScopeImpl.java
@@ -31,7 +31,6 @@ import jdk.incubator.foreign.ResourceScope;
 import jdk.incubator.foreign.SegmentAllocator;
 import jdk.internal.misc.ScopedMemoryAccess;
 import jdk.internal.ref.CleanerFactory;
-import jdk.internal.vm.annotation.Stable;
 
 import java.lang.ref.Cleaner;
 import java.lang.ref.Reference;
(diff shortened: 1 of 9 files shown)


ChrisHegarty:
I like this change, and the idiom that it promotes.  I use similar in experiments that I've been doing, and it works well.

There is one small wart that we can probably live it, but I'd like to ensure that we're comfortable with it - Handle::close returning the global scope for "The implicit resource scope handle" (I like this term). It seems odd at first that the scope method would not return the actual implicit scope, but I understand the reasoning for it - and I accept that the handle is kinda logically meaningless, but exists to help promote good usage patterns.

mcimadamore:
> There is one small wart that we can probably live it, but I'd like to ensure that we're comfortable with it - Handle::close returning the global scope for "The implicit resource scope handle" (I like this term). It seems odd at first that the scope method would not return the actual implicit scope, but I understand the reasoning for it - and I accept that the handle is kinda logically meaningless, but exists to help promote good usage patterns.

Yep - that is the compromise - if we're not comfortable we can remove the accessor - but then it become harder to explain why some calls to ResourceScope::release fail.

ChrisHegarty (approved):

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/ResourceScopeImpl.java
```diff
@@ -124,6 +124,27 @@ public static ResourceScopeImpl createShared(Cleaner cleaner) {
         return new SharedScope(cleaner);
     }
 
+    @Override
+    public void release(Handle handle) {
+        try {
+            Objects.requireNonNull(handle);
+            if (!isImplicit() && handle.scope() != this) {
```
PaulSandoz, line 131: Is this check correct, since an implicit scope could accept a handle from a non-implicit scope?

mcimadamore: yeah - something is off here

PaulSandoz (approved):

PR #513, opened 2021-04-22
8265751: MemoryAddress should have a scope accessor

mcimadamore:
Following the recent API changes, I think `MemoryAddress` has been left in an inconsistent state. On the one hand, we clearly want `MemoryAddress` to have a link to a `ResourceScope`; we clearly state so in the javadoc for:

* `CLinker::upcallStub`
* `MemorySegment::address`

(and we don't say, but we probably should in `VaList::address`).

This link is crucial, as it prevents scope closure, if an address is passed to a native function and the scope is implicit.

However, this dependency is, currently, ad-hoc, and not cleanly stated in the API.

This patch rectifies that; now all memory address have a scope (which can be the "global scope" for addresses obtained from native, or from `long` values).

There are not many changes in terms of implementation, but by putting the association between addresses and scopes front and center we can make the javadoc clearer.

Also, this allows us to move the scope check from `MemorySegment::address` (that has always been a bit odd) to `MemoryAddress::toRawLongAddress` - that is, closer to where the address value is actually used.

This allows us to, effectively, make calls to downcall method handles *safer*, since now a scope check will take place before the raw long value is extracted.

Note that this doesn't negatively affect performance: the check was occurring even before, albeit at a different level (closer to the user code). I've run all the downcall benchmarks we have and found no regressions.

I've added a test which makes sure that passing "closed" segments/addresses to native function is no longer possible.

---------
### Progress

### Issue
 * [JDK-8265751](https://bugs.openjdk.java.net/browse/JDK-8265751): MemoryAddress should have a scope accessor

### Reviewers
 * [Paul Sandoz](https://openjdk.java.net/census#psandoz) (@PaulSandoz - Committer) ⚠️ Review applies to 1520286b602842604c87b80c2e84846ba13eed92

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/513/head:pull/513` \
`$ git checkout pull/513`

`$ git checkout pull/513` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/513/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 513`

`$ git pr show -t 513`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/513.diff">https://git.openjdk.java.net/panama-foreign/pull/513.diff</a>

</details>


mcimadamore:
I have also fixed several references to "this segment is not alive" which were still present in the MemorySegment javadoc

PaulSandoz (approved):

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryAddress.java
```diff
@@ -117,8 +132,9 @@ default MemoryAddress address() {
     MemorySegment asSegment(long bytesSize, ResourceScope scope);
```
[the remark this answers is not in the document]
mcimadamore: very good point - yes, this will fail - and not only because of the address scope - but also because of the new scope.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryAddressImpl.java
```diff
@@ -68,16 +73,20 @@ public MemoryAddress addOffset(long offset) {
...
         return offset() - segmentImpl.min();
     }
 
     @Override
     public long toRawLongValue() {
-        if (base() != null) {
-            throw new UnsupportedOperationException("Not a native address");
+        if (segment != null) {
+            if (segment.base() != null) {
+                throw new UnsupportedOperationException("Not a native address");
```
rsmogura, line 87: Maybe I'm wrong, but if I would like to deregister resource associated with this address here (I.e by calling unmap) during cleanup, will it cause exception?

PaulSandoz (approved):

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MemoryAddressImpl.java
mcimadamore: In general yes. Although I argue that if you get such an exception, the cleanup action is doing something suspicious. If you are referring to your allocator work, I think the trick there is to create entries that are backed by plain addresses (and where the addresses are the ones you obtain from malloc - so they have the global scope, no issue there).

But - if you find yourself in a tight spot during cleanup, you could always do:

```
long rawAddress = address.segmentOffset(MemorySegment.globalNativeScope);
MemoryAddress addr = MemoryAddress.ofLong(rawAddress);
```

rsmogura: Yes, I was bit scared about it and one other thing I got in mind, but I think it will work. Thank you.

mcimadamore:
For the records, the latest iteration also fixes some test failures which crept in because some of the previous changes have not been merged against latest code.

PR #519, opened 2021-04-24
Minor change for possible race in shared resource list

rsmogura:
(1) One or more threads can see "prev" as CLOSED_LIST. Even those CAS will be performed
and list head will be set to new cleanup. Than exception is going to be thrown.

(2) Than cleanup will continue (maybe skipping just added cleanup).

If in addition, other thread will join in parallel to (1) with cleanup2, than
queue could look like this HEAD = cleanup2 -> list end marker.

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/519/head:pull/519` \
`$ git checkout pull/519`

`$ git checkout pull/519` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/519/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 519`

`$ git pr show -t 519`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/519.diff">https://git.openjdk.java.net/panama-foreign/pull/519.diff</a>

</details>

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/SharedScope.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/SharedScope.java
@@ -134,14 +134,15 @@ class SharedScope extends ResourceScopeImpl {
         void add(ResourceCleanup cleanup) {
             while (true) {
                 ResourceCleanup prev = (ResourceCleanup) FST.getAcquire(this);
-                cleanup.next = prev;
-                ResourceCleanup newSegment = (ResourceCleanup) FST.compareAndExchangeRelease(this, prev, cleanup);
-                if (newSegment == ResourceCleanup.CLOSED_LIST) {
+                if (prev == ResourceCleanup.CLOSED_LIST) {
+                    cleanup.next = null; // Release, as long as two threads can't add same cleanup it's ok
                     // too late
                     throw new IllegalStateException("Already closed");
-                } else if (newSegment == prev) {
-                    return; //victory
                 }
+                cleanup.next = prev;
+                if (FST.compareAndSet(this, prev, cleanup)) {
+                    return; //victory
+                };
                 // keep trying
             }
         }
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/SharedScope.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/SharedScope.java
@@ -135,7 +135,7 @@ class SharedScope extends ResourceScopeImpl {
             while (true) {
                 ResourceCleanup prev = (ResourceCleanup) FST.getAcquire(this);
                 if (prev == ResourceCleanup.CLOSED_LIST) {
-                    cleanup.next = null; // Release, as long as two threads can't add same cleanup it's ok
+                    cleanup.next = null; // Sanity, as long as two threads can't add same `cleanup` it's ok to release next
                     // too late
                     throw new IllegalStateException("Already closed");
                 }
@@ -152,15 +152,18 @@ class SharedScope extends ResourceScopeImpl {
             // (because MemoryScope::justClose ensured that this thread won the race to close the scope).
             // So, the only "bad" thing that could happen is that some other thread adds to this list
             // while we're closing it.
-            if (FST.getAcquire(this) != ResourceCleanup.CLOSED_LIST) {
+            //
+            // That's a speculative check
+            ResourceCleanup prev = (ResourceCleanup) FST.getAcquire(this);
+            if (prev != ResourceCleanup.CLOSED_LIST) {
                 //ok now we're really closing down
-                ResourceCleanup prev = null;
                 while (true) {
-                    prev = (ResourceCleanup) FST.getAcquire(this);
-                    // no need to check for DUMMY, since only one thread can get here!
-                    if (FST.weakCompareAndSetRelease(this, prev, ResourceCleanup.CLOSED_LIST)) {
+                    // do atomic CAS (with full fence), to prevent races with add method
+                    if (FST.compareAndSet(this, prev, ResourceCleanup.CLOSED_LIST)) {
                         break;
                     }
+                    // Refresh prev
+                    prev = (ResourceCleanup) FST.getAcquire(this);
                 }
                 cleanup(prev);
             } else {


mcimadamore:
I'm not sure there's a problem. Yes, we set "next" on the provided resource cleanup object - and if we throw an exception, this value will be CLOSED_LIST. However, in the code logic, it seems to me that we are always creating a "new" ResourceCleanup before adding it to the ResourceList - so if adding throws, the resource has not been added, so this cannot cause issues when cleaning up the resource scope.

The only way I can see this being a problem is the "same" ResourceScope instance is repeatedly added to a resource scope, possibly from multiple threads - in which case there might be a race in setting the next field - but this is not how the API implementation operates internally. Is there a test which demonstrates the issues, e.g. where wrong behavior might occur? Note that even the new patch races to set the next field to NULL... it seems to me that correctness of this code is enforced holistically, by looking at the places where the code is used.

mcimadamore:
To be clear, what the logic behind the code is supposed to do is this:

* read the first element of the resource list
* set that element as the next in the cleanup to be added
* keep trying replacing the first element with the new one; there are two cases here:
  - the returned value from compareAndExchange will be CLOSED_LIST; in this case it means the list was already closed; no add happened (as the compareAndExchangeFailed) and an exception is thrown. We could, for completeness reset the `next` field in the cleanup action to be added before throwing.
  
  - the returned value is the new cleanup action; this means the compare and exchange succeeded - the cleanup is added to the list.
  
  - the returned value is something else - means that some thread has sneaked in, and added some other cleanup to the list; keep trying from the beginning.
  
Apart from the fact that cleanup.next is left to the `prev` value when an exception is thrown, I see no other issue - and even then, it's not clear to me what the impact of that issue is?

mcimadamore:
OK, I now get what you mean. The scope is already close. And the first element is already CLOSED_LIST. This means you can happily replace CLOSED_LIST with something else in the CAS.

mcimadamore:
To be clear, compareAndExchange returns the value _previously_ set on memory. So, if the actual value happens to be CLOSED_LIST, compareAndExchange would return that. So here:

```
ResourceCleanup newSegment = (ResourceCleanup) FST.compareAndExchangeRelease(this, prev, cleanup);
                if (newSegment == ResourceCleanup.CLOSED_LIST) {
                    // too late
                    throw new IllegalStateException("Already closed");  
                }
```

we will execute the code in the if branch and throw - but, sadly, there is a chance that we would also already have replaced the first element of the list (CLOSED_LIST) with something else.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/SharedScope.java
```diff
@@ -134,14 +134,15 @@ public boolean isAlive() {
...
-                ResourceCleanup newSegment = (ResourceCleanup) FST.compareAndExchangeRelease(this, prev, cleanup);
-                if (newSegment == ResourceCleanup.CLOSED_LIST) {
+                if (prev == ResourceCleanup.CLOSED_LIST) {
+                    cleanup.next = null; // Release, as long as two threads can't add same cleanup it's ok
                     // too late
                     throw new IllegalStateException("Already closed");
-                } else if (newSegment == prev) {
-                    return; //victory
                 }
+                cleanup.next = prev;
```
mcimadamore, line 142: Why not moving this assigment inside the `if` block (e.g. when add succeeded) and then remove the `cleanup.next = null` ?

rsmogura:
Yes, and I think this requires at least three participants.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/SharedScope.java
rsmogura: Do you mean something like this?
```
if (FST.compareAndSet(this, prev, cleanup)) {
cleanup.next = prev;
}
```
It could create race with clean block, as clean block could read and process new head, before the pointer to next has been assigned.

(I actually bit wonder, if this cleaning is required, I thought about case that this class will be kept by someone, and it can create some memory leak, but I think there's no many places where it can happen)

mcimadamore: I guess what I was suggesting is this:

```
void add(ResourceCleanup cleanup) {
            while (true) {
                ResourceCleanup prev = (ResourceCleanup) FST.getAcquire(this);
                if (prev == ResourceCleanup.CLOSED_LIST) {
                    // too late
                    throw new IllegalStateException("Already closed");
                }
                if (FST.compareAndSet(this, prev, cleanup)) {
                    cleanup.next = prev;
                    return; //victory
                };
                // keep trying
            }
        }
 ```

E.g. setup `prev` only if we are sure that the element has been added to the list - but that probably won't work - as it leaves the added element in a partially invalid state.

mcimadamore (commented):
I'm starting to get a bit skeptical of this code. In general we try to be lock free where possible, which is fine. We certainly need to be lock free for things like memory access, as the cost of locking is way larger than that associated with memory access.

We also need a lock-free acquire scheme, as accessor threads might want to do acquire/access/release (actually, in this case contention is also an important factor and I'm not sure CAS will deliver best results, but, fine for now, we can uber optimize later).

In the case of close vs. add I'm less sure that the performance component is critical. One thread wants to shut the resource scope down for good. The other is attempting to perform some kind of setup operation on the scope. It seems to me that these two operation (close/add) generally belong to different "stages" of a resource scope life cycle.

So, while it is possible to come up (with some complication) with a lock free scheme which allows adds and close to occur safely between threads, I'm less sure here that the extra complication justifies the cost.

Since I don't think we care of being able to "race" between close/add, I think the question boils down to: how much do we care that two different threads are not able to add a cleanup action _on the same scope_ without a lock?

In other words, let's suppose, for a moment, that shared ResourceScope had its "add" and "close" method synchronized (either using the `synchronized` keyword, or with other means). Would that result in a loss of usability or performance for clients? In other words, can we make resource list always based on a single-thread assumption?

rsmogura:
It's more theoretical, then practical, as code paths, rather makes CAS in cleanup not concurrent with add and two threads can't practically add cleanup action on the same scope, as cleanup action is triggered after marking scope as CLOSED.

rsmogura:
Let me close it.

PR #527, opened 2021-05-06
8266627: CLinker allocateMemory, freeMemory implementation should not use default lookup

mcimadamore:
This patch removes the workaround on Windows which requires loading msvcrt.dll.

The solution is to create (Java side) a memory segment and fill it with function pointers on the VM side.
This way, Java code can look up VM functions easily. I've moved this support onto a different class, as the approach might, in the future, be required for other VM functions.

---------
### Progress

### Issue
 * [JDK-8266627](https://bugs.openjdk.java.net/browse/JDK-8266627): CLinker allocateMemory, freeMemory implementation should not use default lookup

### Reviewers
 * [Jorn Vernee](https://openjdk.java.net/census#jvernee) (@JornVernee - Committer) ⚠️ Review applies to 20d183b669fc9afc735d68c050321880541d84cb

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/527/head:pull/527` \
`$ git checkout pull/527`

`$ git checkout pull/527` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/527/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 527`

`$ git pr show -t 527`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/527.diff">https://git.openjdk.java.net/panama-foreign/pull/527.diff</a>

</details>

### DIFF
--- a/src/java.base/share/native/libjava/VMFunctions.c
+++ b/src/java.base/share/native/libjava/VMFunctions.c
@@ -1,0 +1,41 @@
+/*
+ * Copyright (c) 2021, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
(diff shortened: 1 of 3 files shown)


src/java.base/share/native/libjava/VMFunctions.c
```diff
@@ -0,0 +1,41 @@
...
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+#include <stdlib.h>
+#include <assert.h>
```
JornVernee, line 27: Curious; why is this assert.h include needed? (usually assert.h conflicts with hotspot's own assert)

JornVernee (approved):
LGTM

src/java.base/share/native/libjava/VMFunctions.c
mcimadamore: No reason, I realized that I had it at some point but then cleaned up code... it's a leftover

PR #529, opened 2021-05-07
8266682: sync jextract for LibraryLookup removal

sundararajana:
Using System.load/.loadLibrary in generated code and jextract code. Class initialisation dummy method to take sure constants classes initialise header-class before looking up names.

### DIFF
--- a/src/jdk.incubator.jextract/share/classes/jdk/internal/clang/LibClang.java
+++ b/src/jdk.incubator.jextract/share/classes/jdk/internal/clang/LibClang.java
@@ -27,7 +27,6 @@ package jdk.internal.clang;
 
 import jdk.incubator.foreign.CLinker;
 import jdk.incubator.foreign.FunctionDescriptor;
-import jdk.incubator.foreign.LibraryLookup;
 import jdk.incubator.foreign.MemoryAddress;
 import jdk.incubator.foreign.MemorySegment;
 import jdk.incubator.foreign.ResourceScope;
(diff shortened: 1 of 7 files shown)


src/jdk.incubator.jextract/share/classes/jdk/internal/clang/LibClang.java
```diff
@@ -49,12 +48,14 @@
             //this is an hack - needed because clang_toggleCrashRecovery only takes effect _after_ the
             //first call to createIndex.
```
mcimadamore, line 49: Maybe add a link to:
https://reviews.llvm.org/D23662

This should really get fixed upstream.

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/ToplevelBuilder.java
```diff
@@ -208,7 +213,17 @@ private void emitLibraries(String[] libraryNames) {
...
             indent();
-            append("});\n\n");
+            append("});\n");
+            decrAlign();
+            indent();
+            append("}\n\n");
+            decrAlign();
+        }
+
+        private void emitInit() {
```
mcimadamore, line 223: Following offline discussion, I think we can simplify this by removing library loading logic from header classes and sticking it in runtime helper.

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/resources/RuntimeHelper.java.template
```diff
@@ -36,28 +39,22 @@ final class RuntimeHelper {
...
-            return Arrays.stream(libNames)
-                 .map(libName -> {
-                      if (libName.indexOf(File.separatorChar) != -1) {
-                          return LibraryLookup.ofPath(Path.of(libName));
-                      } else {
-                          return LibraryLookup.ofLibrary(libName);
-                      }
-                 })
-                .toArray(LibraryLookup[]::new);
+    static final void loadLibraries(String... libNames) {
```
mcimadamore, line 42: We should be able to remove this

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/OutputFactory.java
```diff
@@ -148,7 +149,23 @@ private String getRuntimeHelperSource() throws URISyntaxException, IOException {
         URL runtimeHelper = OutputFactory.class.getResource("resources/RuntimeHelper.java.template");
         return (pkgName.isEmpty()? "" : "package " + pkgName + ";\n") +
                         String.join("\n", Files.readAllLines(Paths.get(runtimeHelper.toURI())))
-                                .replace("${C_LANG}", C_LANG_CONSTANTS_HOLDER);
+                                .replace("${LOADLIBRARIES}", generateLoadLibraries());
+    }
+
+    private String generateLoadLibraries() {
+        StringBuilder buf = new StringBuilder();
+        buf.append("RuntimeHelper.loadLibraries(new String[] {");
```
mcimadamore, line 157: Instead of calling load libraries, we can just emit calls to System.load/loadLibraries here

mcimadamore (approved):
Looks good - I've tested it on a couple of samples on linux and performance numbers seems good. Tests all pass here.

JornVernee (approved):
Works on Windows as well.

PR #531, opened 2021-05-10
8266814: Improve library loading with SymbolLookup abstraction

mcimadamore:
This patch implements the library loading abstraction described in:

https://mail.openjdk.java.net/pipermail/panama-dev/2021-May/013684.html

That is, a functional interface called `SymbolLookup`, and a couple of factories to get a lookup for a given classloader, and to get a *system* lookup, useful to lookup C symbols.

To implement the system lookup, we load `msvcrt.dll` on Windows, while we build and load an empty library (which depends on libc) on Mac/Linux. This approach is better than relying on RTLD_DEFAULT (which can sometimes leak symbols from libraries loaded independently). Also, doing this bypasses the problem of figuring out the location of libc, which, on Linux system is particularly gnarly, because of the multi-arch support.

---------
### Progress

### Issue
 * [JDK-8266814](https://bugs.openjdk.java.net/browse/JDK-8266814): Improve library loading with SymbolLookup abstraction

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/531/head:pull/531` \
`$ git checkout pull/531`

`$ git checkout pull/531` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/531/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 531`

`$ git pr show -t 531`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/531.diff">https://git.openjdk.java.net/panama-foreign/pull/531.diff</a>

</details>


src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/SymbolLookup.java
```diff
@@ -0,0 +1,97 @@
...
+    /**
+     * Obtains a symbol lookup suitable to find symbols in native libraries associated with the caller's classloader
+     * (that is, libraries loaded using {@link System#loadLibrary} or {@link System#load}).
+     * <p>
+     * This method is <a href="package-summary.html#restricted"><em>restricted</em></a>.
+     * Restricted method are unsafe, and, if used incorrectly, their use might crash
+     * the JVM or, worse, silently result in memory corruption. Thus, clients should refrain from depending on
+     * restricted methods, and use safe and supported functionalities, where possible.
+     *
+     * @return a symbol lookup suitable to find symbols in libraries loaded by given classloader.
```
sundararajana, line 60: "by given classloader" should be "by the caller's class loader"

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/SymbolLookup.java
```diff
@@ -0,0 +1,97 @@
...
+     * the JVM or, worse, silently result in memory corruption. Thus, clients should refrain from depending on
+     * restricted methods, and use safe and supported functionalities, where possible.
+     *
+     * @param loader the classloader whose symbol lookup is to be retrieved.
+     * @return a symbol lookup suitable to find symbols in libraries loaded by given classloader.
+     */
+    @CallerSensitive
+    static SymbolLookup loaderLookup(ClassLoader loader) {
+        Reflection.ensureNativeAccess(Reflection.getCallerClass());
+        Objects.requireNonNull(loader);
```
sundararajana, line 85: why do we filter null (bootstrap) loader?

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/package-info.java
```diff
@@ -96,8 +96,9 @@
  * operation either succeeds - and accesses a valid memory location - or fails.
  *
  * <h2>Foreign function access</h2>
- * The key abstractions introduced to support foreign function access is {@link jdk.incubator.foreign.CLinker}.
- * CLinker provides linking capabilities which allow to model foreign functions as {@link java.lang.invoke.MethodHandle} instances,
+ * The key abstractions introduced to support foreign function access are {@link jdk.incubator.foreign.SymbolLookup} and {@link jdk.incubator.foreign.CLinker}.
+ * The former is used to load foreign libraries, as well as to lookup symbols inside said libraries; the latter
```
sundararajana, line 100: "is used to load foreign libraries" should be removed?

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/SymbolLookup.java
mcimadamore: Because boot loader is under control of the JDK, not the user. I don't think it's desirable to introduce coupling between user code and set of libraries which happen to be loaded by boot loader.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/SymbolLookup.java
```diff
@@ -0,0 +1,97 @@
...
+ * elements to a method in this class causes a {@link NullPointerException NullPointerException} to be thrown. </p>
+ */
+@FunctionalInterface
+public interface SymbolLookup {
+
+    /**
+     * Looks up a symbol with given name in this lookup.
+     *
+     * @param name the symbol name.
+     * @return the memory address associated with the library symbol (if any).
```
JornVernee, line 47: I think we want to remove library-centric text here?
```suggestion
     * @return the memory address associated with the symbol (if any).
```

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/SymbolLookup.java
```diff
@@ -0,0 +1,97 @@
...
+    /**
+     * Obtains a symbol lookup suitable to find symbols in native libraries associated with the caller's classloader
+     * (that is, libraries loaded using {@link System#loadLibrary} or {@link System#load}).
+     * <p>
+     * This method is <a href="package-summary.html#restricted"><em>restricted</em></a>.
+     * Restricted method are unsafe, and, if used incorrectly, their use might crash
+     * the JVM or, worse, silently result in memory corruption. Thus, clients should refrain from depending on
+     * restricted methods, and use safe and supported functionalities, where possible.
+     *
+     * @return a symbol lookup suitable to find symbols in libraries loaded by the caller's classloader.
```
JornVernee, line 60: Should this include an `@throws` for the ICE that can be thrown the native access check?

make/modules/jdk.incubator.foreign/Lib.gmk
```diff
@@ -0,0 +1,59 @@
...
+    OPTIMIZATION := HIGH, \
+    DISABLED_WARNINGS_gcc := sign-compare pointer-arith, \
+    DISABLED_WARNINGS_clang := sign-compare pointer-arith format-nonliteral, \
+    CFLAGS := $(CFLAGS_JDKLIB), \
+    CXXFLAGS := $(CXXFLAGS_JDKLIB), \
+    LDFLAGS := -Wl$(COMMA)--no-as-needed -lc -lm $(LDFLAGS_JDKLIB) $(call SET_SHARED_LIBRARY_ORIGIN), \
+    LIBS := $(LIBCXX), \
+))
+
+else ifeq ($(call isTargetOs, windows), false)
```
JornVernee, line 41: I feel like this code, and maybe also the associated code in SystemLookup that loads msvcrt.dll could do with a comment that explains why the same strategy is not used on Windows; namely that symbol lookup on Windows does not search a library's dependencies, as opposed to dlsym, so it's not as easy to re-export the symbols in msvcrt by creating a shim library.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/SystemLookup.java
```diff
@@ -0,0 +1,60 @@
...
+
+public class SystemLookup implements SymbolLookup {
+
+    private SystemLookup() { }
+
+    final static SystemLookup INSTANCE = new SystemLookup();
+
+    final NativeLibrary syslookup = switch (CABI.current()) {
+        case SysV, AArch64 -> NativeLibraries.rawNativeLibraries(SystemLookup.class, false).loadLibrary("syslookup");
+        case Win64 -> NativeLibraries.rawNativeLibraries(SystemLookup.class, false)
```
JornVernee, line 45: e.g. a short comment here that explains why Windows is different.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/SystemLookup.java
```diff
@@ -0,0 +1,60 @@
...
+                .loadLibrary(Path.of(System.getenv("SystemRoot"), "System32", "msvcrt.dll").toString());
```
JornVernee, line 46: Looks like this loadLibrary call does not except paths:

```
assert name.indexOf(File.separatorChar) < 0;
```

So this is causing an assertion error on Windows. I'll try to find a fix.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/SystemLookup.java
mcimadamore: I'll add a comment here

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/SystemLookup.java
JornVernee: This patch fixes the problem:

```suggestion
                .loadLibrary(null, Path.of(System.getenv("SystemRoot"), "System32", "msvcrt.dll").toFile());
```

sundararajana: was it working?!

mcimadamore: > As far as I can tell there's no `load` on NativeLibraries, but this method loads from the supplied file.

Right

JornVernee (approved):

PR #534, opened 2021-05-11
8266911: On Windows the system lookup should load ucrtbase if possible

JornVernee:
The current system lookup on Windows is always backed by msvcrt.dll, but since Windows 10 the core runtime is located in ucrtbase.dll (though the legacy msvcrt.dll is still available).

We should try to load ucrtbase first if it's available, as it's more up to date.


mcimadamore (approved):
Very good

PR #536, opened 2021-05-14
Remove printf and vprintf tests

JornVernee:
This patch removes the printf and vprintf tests, since the symbols are not available on Windows in the UCRT.

gmtime is similarly an inline function that maps to something else, but in that case there is a drop-in replacement symbol name _gmtime64 we can call instead.

This fixes the missing symbol errors in StdLibTest.

---------
### Progress

### Reviewers
 * [Maurizio Cimadamore](https://openjdk.java.net/census#mcimadamore) (@mcimadamore - Committer) ⚠️ Review applies to a0cf834b32f48e5024c6e9d1d9c9175d000a1e5e

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/536/head:pull/536` \
`$ git checkout pull/536`

`$ git checkout pull/536` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/536/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 536`

`$ git pr show -t 536`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/536.diff">https://git.openjdk.java.net/panama-foreign/pull/536.diff</a>

</details>

### DIFF
--- a/test/jdk/java/foreign/NativeTestHelper.java
+++ b/test/jdk/java/foreign/NativeTestHelper.java
@@ -30,8 +30,6 @@ import jdk.incubator.foreign.SegmentAllocator;
 
 public class NativeTestHelper {
 
-    public static final boolean IS_WINDOWS = System.getProperty("os.name").startsWith("Windows");
-
     static CLinker.TypeKind kind(MemoryLayout layout) {
         return (CLinker.TypeKind)layout.attribute(CLinker.TypeKind.ATTR_NAME).orElseThrow(
                 () -> new IllegalStateException("Unexpected value layout: could not determine ABI class"));
--- a/test/jdk/java/foreign/NativeTestHelper.java
+++ b/test/jdk/java/foreign/NativeTestHelper.java
@@ -30,6 +30,8 @@ import jdk.incubator.foreign.SegmentAllocator;
 
 public class NativeTestHelper {
 
+    public static final boolean IS_WINDOWS = System.getProperty("os.name").startsWith("Windows");
+
     static CLinker.TypeKind kind(MemoryLayout layout) {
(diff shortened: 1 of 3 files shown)


mcimadamore (commented):
Rather than removing could we disable them on windows? On other platforms they give valuable feedback.

JornVernee:
As discussed offline, rather than removing the tests, or disabling them on Windows, I've included a helper library that has a function that can give us pointers to generated inline code for the missing functions on Windows. This allows keeping the tests on on all platforms.

mcimadamore (approved):
Looks good. One possible simplification would be, since you already have a supporting library for windows, to just define printf/vprintf/gmtime there, and then define a lookup that first looks in systemLookup and then falls back in the loader lookup. In fact, you could define the test lookup always in this way, and just do a System.loadLibrary in the windows case.

test/jdk/java/foreign/StdLibTest.java
```diff
@@ -49,6 +50,7 @@
 
 import static jdk.incubator.foreign.MemoryAccess.*;
 
+import org.testng.SkipException;
```
mcimadamore, line 53: Still needed?

test/jdk/java/foreign/NativeTestHelper.java
```diff
@@ -30,6 +30,8 @@
 
 public class NativeTestHelper {
 
+    public static final boolean IS_WINDOWS = System.getProperty("os.name").startsWith("Windows");
```
mcimadamore, line 33: This is never used

mcimadamore (approved):

PR #537, opened 2021-05-17
8267240: Bounded arena allocator doesn't work if bounded size > BLOCK_SIZE

mcimadamore:
This patch follows the discussion in [1], and splits the unbounded allocator from the bounded one, as attempts to reusing logic were causing issues when bounded arena allocators were created with bounded size > BLOCK_SIZE.

I have looked for simplifications of the `trySlice` logic (also mentioned in the emails) - but in my opinion it cannot be simplified much. When an allocator is unbounded, if we receive a request that is more than half the size of the allocator block, we cannot guarantee that we can align it correctly in a segment whose size is BLOCK_SIZE. So, I think the limit of MAX_ALLOC_SIZE = BLOCK_SIZE / 2 is correct, in the sense that anything above that limit needs a standalone allocation (or we won't be able to align it in certain edge cases).

Of course, if we didn't care about alignment, the logic could be simplified, but since this is a native allocator, I think respecting alignment of the allocation request is important.

[1] - https://mail.openjdk.java.net/pipermail/panama-dev/2021-May/013796.html

---------
### Progress

### Issue
 * [JDK-8267240](https://bugs.openjdk.java.net/browse/JDK-8267240): Bounded arena allocator doesn't work if bounded size > BLOCK_SIZE

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/537/head:pull/537` \
`$ git checkout pull/537`

`$ git checkout pull/537` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/537/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 537`

`$ git pr show -t 537`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/537.diff">https://git.openjdk.java.net/panama-foreign/pull/537.diff</a>

</details>

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/ArenaAllocator.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/ArenaAllocator.java
@@ -40,7 +40,6 @@ public abstract class ArenaAllocator implements SegmentAllocator {
     public static class UnboundedArenaAllocator extends ArenaAllocator {
 
         private static final long DEFAULT_BLOCK_SIZE = 4 * 1024;
-        private static final long DEFAULT_MAX_ALLOC_SIZE = DEFAULT_BLOCK_SIZE / 2;
 
         public UnboundedArenaAllocator(ResourceScope scope) {
             super(MemorySegment.allocateNative(DEFAULT_BLOCK_SIZE, 1, scope));
@@ -58,7 +57,8 @@ public abstract class ArenaAllocator implements SegmentAllocator {
             if (slice != null) {
                 return slice;
             } else {
-                if (Utils.alignUp(bytesSize, bytesAlignment) > DEFAULT_MAX_ALLOC_SIZE) {
+                long maxPossibleAllocationSize = bytesSize + bytesAlignment - 1;
+                if (maxPossibleAllocationSize > DEFAULT_BLOCK_SIZE) {
                     // too big
                     return newSegment(bytesSize, bytesAlignment);
                 } else {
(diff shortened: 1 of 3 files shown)


src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/ArenaAllocator.java
```diff
@@ -63,22 +26,67 @@ private MemorySegment trySlice(long bytesSize, long bytesAlignment) {
...
+
+        @Override
+        public MemorySegment allocate(long bytesSize, long bytesAlignment) {
+            checkConfinementIfNeeded();
+            // try to slice from current segment first...
+            MemorySegment slice = trySlice(bytesSize, bytesAlignment);
+            if (slice != null) {
+                return slice;
+            } else {
+                if (Utils.alignUp(bytesSize, bytesAlignment) > DEFAULT_MAX_ALLOC_SIZE) {
```
JornVernee, line 61: This seems to be the subject of some of the emails. I think maybe it might be worth it to break this out into an intermediate variable with a good name?

I also think the DEFAULT_MAX_ALLOC_SIZE could just be DEFAULT_BLOCK_SIZE, if we change this check to check for the maximum allocation size computed by `byteSize + bytesAlignment - 1` instead of using `alignUp`:

```suggestion
                long maxPossibleAllocationSize = bytesSize + bytesAlignment - 1;
                if (maxPossibleAllocationSize > DEFAULT_BLOCK_SIZE) {
```

e.g. if we have an allocation aligned to 8 bytes, we can have a maximum of 7 bytes of alignment padding (in case the address is off by just 1 byte).

mcimadamore: I've accepted this comment

JornVernee (approved):
LGTM

PR #543, opened 2021-05-31
8267989: Exceptions thrown during upcalls should be handled

JornVernee:
Hi,

This patch regularizes exception handling for exceptions thrown during upcalls. Exceptions thrown during upcalls are now always handled by printing out the stack trace and then calling `System::exit` (see the JBS issue for some motivation).

I've added some documentation for the exception handling to `CLinker::upcallStub`, as well as a new public `int` constant in `CLinker` which is the error code that is passed to `System::exit`. The returned error code can also be configured by a system property, which for now is mostly useful for testing purposes to make sure we don't get a consistent false positive.

Thanks,
Jorn

---------
### Progress

### Issues
 * [JDK-8267989](https://bugs.openjdk.java.net/browse/JDK-8267989): Exceptions thrown during upcalls should be handled
 * [JDK-8268031](https://bugs.openjdk.java.net/browse/JDK-8268031): VarHandle combinator check for exceptions is too strict

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/543/head:pull/543` \
`$ git checkout pull/543`

`$ git checkout pull/543` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/543/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 543`

`$ git pr show -t 543`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/543.diff">https://git.openjdk.java.net/panama-foreign/pull/543.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/java/lang/invoke/VarHandles.java
+++ b/src/java.base/share/classes/java/lang/invoke/VarHandles.java
@@ -646,7 +646,7 @@ final class VarHandles {
             return new Class<?>[0];
         }
 
-        assert handle instanceof BoundMethodHandle : "Unecpexted handle type: " + handle;
+        assert handle instanceof BoundMethodHandle : "Unexpected handle type: " + handle;
         // unknown
         return null;
     }
--- a/src/java.base/share/classes/java/lang/invoke/VarHandles.java
+++ b/src/java.base/share/classes/java/lang/invoke/VarHandles.java
@@ -31,12 +31,9 @@ import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
-import java.lang.reflect.Parameter;
 import java.nio.ByteOrder;
 import java.util.ArrayList;
-import java.util.LinkedHashMap;
 import java.util.List;
-import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
@@ -45,8 +42,6 @@ import java.util.stream.Stream;
 import static java.lang.invoke.MethodHandleStatics.UNSAFE;
 import static java.lang.invoke.MethodHandleStatics.VAR_HANDLE_IDENTITY_ADAPT;
 import static java.lang.invoke.MethodHandleStatics.newIllegalArgumentException;
-import static java.util.stream.Collectors.joining;
-import static java.util.stream.Collectors.toList;
 
 final class VarHandles {
 
@@ -359,13 +354,13 @@ final class VarHandles {
         return target;
     }
 
-    public static VarHandle filterValue(VarHandle target, MethodHandle filterToTarget, MethodHandle filterFromTarget) {
+    public static VarHandle filterValue(VarHandle target, MethodHandle pFilterToTarget, MethodHandle pFilterFromTarget) {
         Objects.requireNonNull(target);
-        Objects.requireNonNull(filterToTarget);
-        Objects.requireNonNull(filterFromTarget);
+        Objects.requireNonNull(pFilterToTarget);
+        Objects.requireNonNull(pFilterFromTarget);
         //check that from/to filters do not throw checked exceptions
-        noCheckedExceptions(filterToTarget);
-        noCheckedExceptions(filterFromTarget);
+        MethodHandle filterToTarget = adaptForCheckedExceptions(pFilterToTarget);
+        MethodHandle filterFromTarget = adaptForCheckedExceptions(pFilterFromTarget);
 
         List<Class<?>> newCoordinates = new ArrayList<>();
         List<Class<?>> additionalCoordinates = new ArrayList<>();
@@ -473,8 +468,9 @@ final class VarHandles {
 
         List<Class<?>> newCoordinates = new ArrayList<>(targetCoordinates);
         for (int i = 0 ; i < filters.length ; i++) {
-            noCheckedExceptions(filters[i]);
-            MethodType filterType = filters[i].type();
+            MethodHandle filter = Objects.requireNonNull(filters[i]);
+            filter = adaptForCheckedExceptions(filter);
+            MethodType filterType = filter.type();
             if (filterType.parameterCount() != 1) {
                 throw newIllegalArgumentException("Invalid filter type " + filterType);
             } else if (newCoordinates.get(pos + i) != filterType.returnType()) {
@@ -564,10 +560,10 @@ final class VarHandles {
         return adjustedType;
     }
 
-    public static VarHandle collectCoordinates(VarHandle target, int pos, MethodHandle filter) {
+    public static VarHandle collectCoordinates(VarHandle target, int pos, MethodHandle pFilter) {
         Objects.requireNonNull(target);
-        Objects.requireNonNull(filter);
-        noCheckedExceptions(filter);
+        Objects.requireNonNull(pFilter);
+        MethodHandle filter = adaptForCheckedExceptions(pFilter);
 
         List<Class<?>> targetCoordinates = target.coordinateTypes();
         if (pos < 0 || pos >= targetCoordinates.size()) {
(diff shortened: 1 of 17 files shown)


src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
```diff
@@ -224,9 +232,14 @@ static SymbolLookup systemLookup() {
...
      * calling such a function pointer from native code will result in the execution of the provided method handle.
      *
-     * <p>The returned memory address is associated with the provided scope. When such scope is closed,
+     * <p>
+     * The returned memory address is associated with the provided scope. When such scope is closed,
      * the corresponding native stub will be deallocated.
      * <p>
+     * Any exceptions that occur during an upcall should be handled during the upcall. The target method handle
+     * should not throw any exceptions. If the target method handle does throw an exception, it will be handle by
+     * calling {@link System#exit System.exit(ERR_UNCAUGHT_EXCEPTION)}. (See {@link #ERR_UNCAUGHT_EXCEPTION})
```
JornVernee, line 241: There might be better wording for this, esp. the first sentence. Any suggestions?

test/jdk/java/foreign/TestUpcall.java
```diff
@@ -28,7 +28,7 @@
  * @modules jdk.incubator.foreign/jdk.internal.foreign
  * @build NativeTestHelper CallGeneratorHelper TestUpcall
  *
- * @run testng/othervm
+ * @run testng/othervm/timeout=720
```
JornVernee, line 31: I've increased the timeout of TestUpcall here as well, as I was seeing timeouts pretty frequently

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
```diff
@@ -114,6 +115,13 @@
  */
 public interface CLinker {
 
+    /**
+     * The value returned by the process when the target of an upcall throws an exception.
+     *
+     * @see CLinker#upcallStub(MethodHandle, FunctionDescriptor, ResourceScope)
+     */
+    int ERR_UNCAUGHT_EXCEPTION = privilegedGetProperty("jdk.incubator.foreign.uncaught_exception_code", 1);
```
mcimadamore, line 123: Do we need a system property? E.g. can it be a constant value?

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
mcimadamore: I think it is best to drop the first sentence, and go straight to the point, which is "The target method handle shoud not...".

Which reminds me, do we want to apply same logic we do for var handle combinators to detect if a method handle contains exceptions in its guts (by looking at signature of direct MH, or by looking for certain combinators).

test/jdk/java/foreign/TestUpcallException.java
```diff
@@ -0,0 +1,83 @@
...
+                "--enable-native-access=ALL-UNNAMED",
+                "-Djava.library.path=" + System.getProperty("java.library.path"),
+                "-Djdk.internal.foreign.ProgrammableUpcallHandler.USE_SPEC=" + useSpec,
+                "-Djdk.incubator.foreign.uncaught_exception_code=" + exitCode,
+                "-cp", Utils.TEST_CLASS_PATH,
+                "ThrowingUpcall")
+            .start();
+
+        int result = process.waitFor();
+        assertEquals(result, exitCode);
```
mcimadamore, line 80: Should we check for stacktrace output - e.g. the name of ThrowingUpcall" ?

mcimadamore (approved):
Looks good!

test/jdk/java/foreign/TestUpcallException.java
JornVernee: That's probably a good idea. Though, I'll note that it still allows for false positives if the tests throws an exception from ThrowingUpcall another way. Comparing the exit codes should catch that though.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
JornVernee: We could just use `1` here as well, but during testing I've found that the default uncaught exception handler also return `1`, so I was getting a false positive.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
```diff
@@ -224,9 +232,14 @@ static SymbolLookup systemLookup() {
      * Allocates a native stub with given scope which can be passed to other foreign functions (as a function pointer);
      * calling such a function pointer from native code will result in the execution of the provided method handle.
      *
-     * <p>The returned memory address is associated with the provided scope. When such scope is closed,
+     * <p>
+     * The returned memory address is associated with the provided scope. When such scope is closed,
      * the corresponding native stub will be deallocated.
      * <p>
+     * Any exceptions that occur during an upcall should be handled during the upcall. The target method handle
```
mcimadamore, line 239: I think it is best to drop the first sentence, and go straight to the point, which is "The target method handle shoud not...".

Which reminds me, do we want to apply same logic we do for var handle combinators to detect if a method handle contains exceptions in its guts (by looking at signature of direct MH, or by looking for certain combinators).

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/SharedUtils.java
```diff
@@ -359,14 +362,25 @@ private static MethodHandle reachabilityFenceHandle(Class<?> type) {
...
     static MethodHandle wrapWithAllocator(MethodHandle specializedHandle,
                                           int allocatorPos, long bufferCopySize,
                                           boolean upcall) {
         // insert try-finally to close the NativeScope used for Binding.Copy
         MethodHandle closer;
         int insertPos;
         if (specializedHandle.type().returnType() == void.class) {
-            closer = empty(methodType(void.class, Throwable.class)); // (Throwable) -> void
+            if (!upcall) {
+                closer = empty(methodType(void.class, Throwable.class)); // (Throwable) -> void
```
mcimadamore, line 380: Shouldn't ProgrammableUpcallHandler do the same thing?

JornVernee: It does? Both call SharedUtils.handelUncaughtException

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
JornVernee: Thinking some more. Having a system property is probably overkill, and checking stderr for the exception message in the test is good enough

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
JornVernee: I'll see about adding a check for exceptions.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/ProgrammableUpcallHandler.java
```diff
@@ -295,6 +295,9 @@ private static Object invokeInterpBindings(Object[] moves, MethodHandle leaf,
             } else {
                 return returnMoves;
             }
+        } catch(Throwable t) {
+            SharedUtils.handleUncaughtException(t);
```
mcimadamore, line 299: Ah - missed that :-)

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
```diff
@@ -114,6 +111,13 @@
  */
 public interface CLinker {
 
+    /**
+     * The value returned by the process when the target of an upcall throws an exception.
+     *
+     * @see CLinker#upcallStub(MethodHandle, FunctionDescriptor, ResourceScope)
+     */
+    int ERR_UNCAUGHT_EXCEPTION = 1;
```
mcimadamore, line 119: On a second thought, I'm not super sure of the value of documenting what the exit code should be? Maybe we can leave that unspecified?

test/jdk/java/foreign/TestUpcallException.java
```diff
@@ -0,0 +1,95 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.nio.file.Paths;
+
+import static jdk.incubator.foreign.CLinker.ERR_UNCAUGHT_EXCEPTION;
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.fail;
+
+public class TestUpcallException {
```
mcimadamore, line 49: I wonder if we should also test with a security manager (even though that's getting deprecated) - and see what happens when you call System.exit and SM is enabled with policy to block off calls to System.exit.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
JornVernee: I agree. Without a system property, having a constant doesn't really make sense. I looked around in the documentation of exit codes in the JDK, and most  places seem to at least specify the exit code  to be non-zero in case of error. I'll add a note like that to the documentation of upcallStub.

JornVernee:
I've switched to using `Shutdown.exit`, and added a `SecurityManager` that blocks the normal `System.exit` call into the test. I've also switched `CABI` from using `System.getProperty` to determine the platform, to using `priviledgedGetProperty`, to avoid any security exceptions (found during testing).

I've removed the public `ERR_UNCAUGHT_EXCEPTION` constant, and changed the documentation to say that the VM will exit with a non-zero exit status in the case of an uncaught exception.

Finally, I've changed the test to always print the output as well, besides checking the contents.

JornVernee:
I've now added the eager exception checking to upcallStub.

As discussed offline, this reuses the exception check done for the VarHandle adapters, but makes the check more lenient towards BoundMetrhodHandles that might or might not throw an exception. The current check was too conservative toward rejecting such handles, because it rejects a BoundMehtodHandle that contained any MethodHandle field that could throw an exception. But, such an exception might be caught be an exclosing method handle. (see also https://bugs.openjdk.java.net/browse/JDK-8268031)

Instead, when a filter used for a var handle combinator throws a checked a exception when invoked, it is caught, and an IllegalStateException is thrown instead (with the original exception as the cause).

src/java.base/share/classes/java/lang/invoke/VarHandles.java
```diff
@@ -604,42 +600,55 @@ public static VarHandle dropCoordinates(VarHandle target, int pos, Class<?>... v
...
-                if (arg instanceof MethodHandle){
-                    noCheckedExceptions((MethodHandle) arg);
-                }
-            }
+            return exceptionTypes(((DelegatingMethodHandle)handle).getTarget());
+        } else if (handle instanceof NativeMethodHandle) {
+            return new Class<?>[0];
         }
+
+        assert handle instanceof BoundMethodHandle : "Unecpexted handle type: " + handle;
```
JornVernee, line 649:
```suggestion
        assert handle instanceof BoundMethodHandle : "Unexpected handle type: " + handle;
```

mcimadamore (approved):
Looks very good - thanks for fixing the VH adaptation code as well!

PR #553, opened 2021-06-10
8268266: Investigate way to lazily customize upcall lambda forms

JornVernee:
Hi,

This patch makes upcall method handle customization lazy, which helps a lot in reducing the number of lambda form classes generated during for intstance the TestUpcall test, making it complete a lot faster.

This adds a very slight overhead per call, since there's another indirection, but I think the tradeoff is acceptable.

Since the patch was otherwise really small, this also adds a qsort benchmark that is also used the measure the cost of the extra indirection. Results are as follows:

```
QSort without lazy customization:

Benchmark                         Mode  Cnt      Score     Error  Units
QSort.jni_upcall_qsort_naive      avgt   30  21096.871 � 299.669  ns/op
QSort.jni_upcall_qsort_optimized  avgt   30   3603.495 �  16.464  ns/op
QSort.native_qsort                avgt   30     96.236 �   0.713  ns/op
QSort.panama_upcall_qsort         avgt   30    937.675 �  26.160  ns/op

QSort with lazy customization:

Benchmark                         Mode  Cnt      Score     Error  Units
QSort.jni_upcall_qsort_naive      avgt   30  21454.476 � 306.135  ns/op
QSort.jni_upcall_qsort_optimized  avgt   30   3691.395 �  35.671  ns/op
QSort.native_qsort                avgt   30     97.440 �   0.894  ns/op
QSort.panama_upcall_qsort         avgt   30   1072.929 �  14.948  ns/op
```

So, while there is some regression from this change, it's not super bad (not integer factor), and I think the tradeoff is worth it to generate fewer lambda form classes, which of course has it's own overhead not visible in this benchmark.

Thanks,
Jorn

---------
### Progress

### Issue
 * [JDK-8268266](https://bugs.openjdk.java.net/browse/JDK-8268266): Investigate way to lazily customize upcall lambda forms

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/553/head:pull/553` \
`$ git checkout pull/553`

`$ git checkout pull/553` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/553/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 553`

`$ git pr show -t 553`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/553.diff">https://git.openjdk.java.net/panama-foreign/pull/553.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/java/lang/invoke/MethodHandleImpl.java
+++ b/src/java.base/share/classes/java/lang/invoke/MethodHandleImpl.java
@@ -1468,11 +1468,6 @@ abstract class MethodHandleImpl {
             }
 
             @Override
-            public void ensureCustomized(MethodHandle mh) {
-                mh.customize();
-            }
-
-            @Override
             public VarHandle memoryAccessVarHandle(Class<?> carrier, boolean skipAlignmentMaskCheck, long alignmentMask,
                                                    ByteOrder order) {
                 return VarHandles.makeMemoryAddressViewHandle(carrier, skipAlignmentMaskCheck, alignmentMask, order);
--- a/src/java.base/share/classes/jdk/internal/access/JavaLangInvokeAccess.java
+++ b/src/java.base/share/classes/jdk/internal/access/JavaLangInvokeAccess.java
@@ -134,13 +134,6 @@ public interface JavaLangInvokeAccess {
     MethodHandle nativeMethodHandle(NativeEntryPoint nep, MethodHandle fallback);
 
     /**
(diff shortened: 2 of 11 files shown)


mcimadamore (approved):
Looks good - what is the cost for downcalls?

JornVernee:
> Looks good - what is the cost for downcalls?

Nothing, only upcalls were eagerly customized. So, there's no change for downcalls

PR #554, opened 2021-06-10
8268230: Foreign Linker API & Windows user32/kernel32: String conversion seems broken

JornVernee:
The problem is that we only add a single 0 byte as a null terminator, regardless of the charset used. For wider char sets, more 0 bytes need to be added. For instance, for UTF_16LE two 0 bytes need to be added.

This patch fixes the issue by adding the null terminator to the Java string, and only then encoding it as a `byte[]`.

---------
### Progress

### Issue
 * [JDK-8268230](https://bugs.openjdk.java.net/browse/JDK-8268230): Foreign Linker API & Windows user32/kernel32: String conversion seems broken

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/554/head:pull/554` \
`$ git checkout pull/554`

`$ git checkout pull/554` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/554/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 554`

`$ git pr show -t 554`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/554.diff">https://git.openjdk.java.net/panama-foreign/pull/554.diff</a>

</details>


src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
```diff
@@ -296,7 +296,14 @@ static SymbolLookup systemLookup() {
     static MemorySegment toCString(String str, SegmentAllocator allocator) {
         Objects.requireNonNull(str);
         Objects.requireNonNull(allocator);
-        return toCString(str.getBytes(), allocator);
+        return toCString(addNullTerminator(str).getBytes(), allocator);
+    }
+
+    private static String addNullTerminator(String str) {
```
mcimadamore, line 302: This is gonna allocate another string in an already allocation-heavy code. Wouldn't it be better to just add the correct termination sequence in the segment? I suggest running StrLen benchmark before/after to make sure string conversion performance isn't negatively impacted.

mcimadamore (commented):
Good catch - I think the fix can be made slightly more efficient

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
JornVernee: I'll check the benchmark.

I looked at the CharSet API, and AFAICS the only way to get the number of bytes for the null terminator would be to create a String with the value `"\0"` then encode that and check the resulting `byte[]`. I thought going the string concat route would have better chances of being optimized.

JornVernee: The benchmark results are as follows:

Before:

```
Benchmark                        (size)  Mode  Cnt    Score    Error  Units
StrLenTest.panama_strlen_prefix       5  avgt   30  124.874 � 15.751  ns/op
StrLenTest.panama_strlen_prefix      20  avgt   30  131.683 �  6.011  ns/op
StrLenTest.panama_strlen_prefix     100  avgt   30  161.046 � 10.580  ns/op
```

After:
```
Benchmark                        (size)  Mode  Cnt    Score   Error  Units
StrLenTest.panama_strlen_prefix       5  avgt   30  130.758 � 5.691  ns/op
StrLenTest.panama_strlen_prefix      20  avgt   30  145.012 � 6.804  ns/op
StrLenTest.panama_strlen_prefix     100  avgt   30  179.992 � 6.457  ns/op
```

I think C2 should be able to eliminate the intermediate string, but there's still a slight regression.

> Wouldn't it be better to just add the correct termination sequence in the segment?

The problem is finding out how many bytes to add. Looking at Charset again, there really doesn't seem to be a way to get the number of bytes per character. The closest seems to be `charset.newEncoder().averageBytesPerChar()`. I'll ask around as well.

JornVernee:
> I don't think reverse conversion works using Clinker.toJavaString.

Good catch, the `toJavaString` version also looks broken, since it looks for the first 0 byte, but that might just be the high-order byte of a character, which just happens to be 0, for wider char sets (for some reason I had assumed this was okay when thinking about it before).

JornVernee:
When looking into toJavaString, the problems turned out to be much greater. The problem essentially boils down to not having a `strlen` function for any arbitrary `Charset`, so we can only support a certain subset of `Charsets` for which we know our current way of determining the native string's length works.

We discussed this at length among the team members, and arrived at the intermediate conclusion to only support the 'platform native' Charset. Though, this turns out to be tricky as well, as this charset, which is also called the 'execution character set' in C lingo, is determined based on a compiler setting at build time of the native code. With GCC and Clang the default is UTF-8, while on Windows it depends on the current code page. While there is a way to get the current code page of the runtime system and determine the character set from that, we would not be able to avoid issues with code page mismatches between the build environment and runtime environment on Windows,

While it would still technically be possible to support different character sets as long as they work with `strlen`, at present there is no way to detect this for an arbitrary character set. So, if we kept the `Charset` parameter, we would not be able to sanity check it, which doesn't seem great either.

As a result of all this, for now we have arrived at the decision to only support the UTF-8 Charset for the toCString and toJavaString methods, and to leave encoding and decoding using other character sets (including determining the length of a native string) to be implemented manually.

I've updated this PR to remove the overloads that accept a `Charset`, and updated the implementation to always use UTF-8. I've added several test cases as well that test Unicode characters that get encoded with different amounts of bytes in UTF-8.

test/jdk/java/foreign/TestStringEncoding.java
```diff
@@ -0,0 +1,63 @@
...
+    @DataProvider
+    public static Object[][] strings() {
+        return new Object[][] {
+            { "testing",               8 },
+            { "",                      1 },
+            { "X",                     2 },
+            { "12345",                 6 },
+            { "yen \u00A5",            7 }, // in UTF-8 2 bytes: 0xC2 0xA5
+            { "snowman \u26C4",       12 }, // in UTF-8 three bytes: 0xE2 0x9B 0x84
+            { "rainbow \uD83C\uDF08", 13 }  // int UTF-8 four bytes: 0xF0 0x9F 0x8C 0x88
```
JornVernee, line 60:
```suggestion
            { "rainbow \uD83C\uDF08", 13 }  // in UTF-8 four bytes: 0xF0 0x9F 0x8C 0x88
```

mcimadamore (approved):
Looks sensible!

overheadhunter:
> leave encoding and decoding using other character sets (including determining the length of a native string) to be implemented manually.

Can you give us a hint, how such manual re-encoding could look like, if an C lib wants to consume a different charset?

If I call `CLinker.toCString(new String(bytes, UTF-16BE))`, it'll still be converted to UTF-8 internally.

JornVernee:
> If I call `CLinker.toCString(new String(bytes, UTF-16BE))`, it'll still be converted to UTF-8 internally.

Note that the char set that is passed to the `String` constructor only indicates the encoding of the `bytes` that are being passed. These bytes are then converted to one of the internal encodings used by Java strings to produce the new `String`, and then `toCString` converts from that `String` back to a UTF-8 encoded memory region.

---

For encoding to a native representation, here is an example that assumes the native representation of the string should be terminated with 2 null bytes (the exact format the native string should have depends ultimately on the domain, and what the library expects):

```java
String input = foo(); // get a Java string from somewhere
// Using a standard null terminated format here, but the library could also expect a different format
byte[] bytes = (input + '\0').getBytes(UTF_16BE);
MemorySegment text = MemorySegment.allocateNative(bytes.length);
text.copyFrom(MemorySegment.ofArray(bytes));
// use 'text'
```

For converting from a native string to a Java string a way to determine the length of a string is needed. Examples are `strlen` and `wcslen`, but it depends on the encoding the string is in, and again the domain (e.g. the library might expose a function to determine the length of the strings it returns).

Then something like this could be used:

```java
MemoryAddress input = foo(); // get a native string from somewhere
int length = utf16be_string_length(input);
byte[] bytes = new byte[length];
MemorySegment inputSegment = input.asSegment(length, ResourceScope.newImplicitScope());
MemorySegment.ofArray(bytes).copyFrom(inputSegment);
String text = new String(bytes, UTF_16BE);
// use 'text'
```

PR #555, opened 2021-06-10
Add support for high-level functions to copy to and from Java arrays

mcimadamore:
This patch includes some of the changes from Lee to support a set of static functions to allow bulk copy from Java arrays to segments (and viceversa) in a more succint fashion (so that the user doesn't have to create a temporary heap segment to do the copy).

I've added a new public method to `MemorySegment` which performs an *element-wise* bulk copy; it takes a source segment and a couple of element layouts: the source element layout and the destination element layout. The two layouts must have same size, but can have different alignments (which will be checked against the corresponding segments) and byte orders. If the byte order differs, a bulk copy with swap will be performed. As such, this method generalizes the previous `copyFrom` - as follows:

```
copyFrom(srcSegment) -> copyFrom(JAVA_BYTE, srcSegment, JAVA_BYTE)
```
I've added support for argument type profiling for MemoryCopy static methods to avoid type pollution in cases where same metod is called with different memory segment types.

I've done a pass over the javadoc, and make it more consistent with the rest of the API. I've also reworked the test a bit to use the data provider functionality of TestNG, since all the test cases were similar, except for the carrier type.

There are other cosmetic changes as well, compared to original code from Lee, such as naming of static fields which is now capitalized. Everything else is the same.

---------
### Progress

### Reviewers
 * [Jorn Vernee](https://openjdk.java.net/census#jvernee) (@JornVernee - Committer) ⚠️ Review applies to e66008dc8dd6b93fb90a866b5014200a8ef65896
 * [Uwe Schindler](https://openjdk.java.net/census#uschindler) (@uschindler - no project role)
 * @leerho (no known github.com user name / role) ⚠️ Review applies to 85a9786aa994cba0b62edf645960a84ee4e1823d

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/555/head:pull/555` \
`$ git checkout pull/555`

`$ git checkout pull/555` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/555/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 555`

`$ git pr show -t 555`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/555.diff">https://git.openjdk.java.net/panama-foreign/pull/555.diff</a>

</details>


uschindler:
Hi, the byte-swap code is still commented out in MemoryCopy class, although the method was implemented in MemorySegment.

uschindler:
Hi,
when testing this with Lucene, Irecognized the following:

```
org.apache.lucene.store.TestMultiMMap > test suite's output saved to /home/thetaphi/benchmark/lucene_candidate/lucene/core/build/test-results/test/outputs/OUTPUT-org.apache.lucene.store.TestMultiMMap.txt, copied below:
   >     java.lang.IllegalArgumentException: Source segment incompatible with alignment constraints
   >         at __randomizedtesting.SeedInfo.seed([90A99875BABA30A6:B162CD10DB95DFC]:0)
   >         at jdk.incubator.foreign/jdk.internal.foreign.AbstractMemorySegmentImpl.copyFrom(AbstractMemorySegmentImpl.java:158)
   >         at jdk.incubator.foreign/jdk.incubator.foreign.MemoryCopy.copyToArray(MemoryCopy.java:493)
   >         at org.apache.lucene.store.MemorySegmentIndexInput.readLongs(MemorySegmentIndexInput.java:204)
```

So it is impossible to copy stuff if the source segment is not at an aligned position. Unfortunately, with data inside a memory mapped file it may often happen that data is not properly aligned, but you still need to copy them (including swapping).

mcimadamore:
> Hi,
> when testing this with Lucene, Irecognized the following:
> 
> ```
> org.apache.lucene.store.TestMultiMMap > test suite's output saved to /home/thetaphi/benchmark/lucene_candidate/lucene/core/build/test-results/test/outputs/OUTPUT-org.apache.lucene.store.TestMultiMMap.txt, copied below:
>    >     java.lang.IllegalArgumentException: Source segment incompatible with alignment constraints
>    >         at __randomizedtesting.SeedInfo.seed([90A99875BABA30A6:B162CD10DB95DFC]:0)
>    >         at jdk.incubator.foreign/jdk.internal.foreign.AbstractMemorySegmentImpl.copyFrom(AbstractMemorySegmentImpl.java:158)
>    >         at jdk.incubator.foreign/jdk.incubator.foreign.MemoryCopy.copyToArray(MemoryCopy.java:493)
>    >         at org.apache.lucene.store.MemorySegmentIndexInput.readLongs(MemorySegmentIndexInput.java:204)
> ```
> 
> So it is impossible to copy stuff if the source segment is not at an aligned position. Unfortunately, with data inside a memory mapped file it may often happen that data is not properly aligned, but you still need to copy them (including swapping).

This is an issue in the layouts used by the MemoryCopy class. I meant to drop all alignment constraints from them when doing another pass on the code, but it seems like I forgot.

uschindler:
> > Hi,
> > when testing this with Lucene, Irecognized the following:
> > ```
> > org.apache.lucene.store.TestMultiMMap > test suite's output saved to /home/thetaphi/benchmark/lucene_candidate/lucene/core/build/test-results/test/outputs/OUTPUT-org.apache.lucene.store.TestMultiMMap.txt, copied below:
> >    >     java.lang.IllegalArgumentException: Source segment incompatible with alignment constraints
> >    >         at __randomizedtesting.SeedInfo.seed([90A99875BABA30A6:B162CD10DB95DFC]:0)
> >    >         at jdk.incubator.foreign/jdk.internal.foreign.AbstractMemorySegmentImpl.copyFrom(AbstractMemorySegmentImpl.java:158)
> >    >         at jdk.incubator.foreign/jdk.incubator.foreign.MemoryCopy.copyToArray(MemoryCopy.java:493)
> >    >         at org.apache.lucene.store.MemorySegmentIndexInput.readLongs(MemorySegmentIndexInput.java:204)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So it is impossible to copy stuff if the source segment is not at an aligned position. Unfortunately, with data inside a memory mapped file it may often happen that data is not properly aligned, but you still need to copy them (including swapping).
> 
> This is an issue in the layouts used by the MemoryCopy class. I meant to drop all alignment constraints from them when doing another pass on the code, but it seems like I forgot.

No problem, I just commented out the long/float methods and the tests were passing. I am now benchmarking, but again it looks not so well with the garbage created... (see #560). But I have to wait for the benchmark to finish.

mcimadamore:
I've fixed the alignment issue and uploaded a new revision.

uschindler:
If you're interested, here's my branch and modification to use this PR in Lucene - as you see the new `MemoryCopy` fits our needs perfectly: https://github.com/uschindler/lucene/blob/048434b9e3cb8aa515d9c1d7f57b44182704ab09/lucene/core/src/java/org/apache/lucene/store/MemorySegmentIndexInput.java#L164-L176

uschindler:
I don't want to post too much in this issue as it is not fully related, but it looks this is also affected by the same garbage problem described in https://bugs.openjdk.java.net/browse/JDK-8268743:

```
PROFILE SUMMARY from 14119950 events (total: 4897631M)
  tests.profile.mode=heap
  tests.profile.count=30
  tests.profile.stacksize=1
  tests.profile.linenumbers=false
PERCENT       HEAP SAMPLES  STACK
49.63%        2430891M      jdk.internal.foreign.MappedMemorySegmentImpl#dup()
28.37%        1389691M      jdk.internal.foreign.HeapMemorySegmentImpl$OfByte#fromArray()
21.43%        1049362M      jdk.internal.foreign.HeapMemorySegmentImpl$OfByte#dup()
0.09%         4624M         org.apache.lucene.util.FixedBitSet#<init>()
```

See the number of heap samples, this cannot be! The only difference to the JFR heap profile in https://bugs.openjdk.java.net/browse/JDK-8268743 is the order (here we have more dups of the MappedMemorySegment (source from where it is copied).

uschindler:
Good news: I can confirm after a quick test that memory copy now works with unaligned offsets! 👍

mcimadamore:
> I don't want to post too much in this issue as it is not fully related, but it looks this is also affected by the same garbage problem described in https://bugs.openjdk.java.net/browse/JDK-8268743:
> 
> ```
> PROFILE SUMMARY from 14119950 events (total: 4897631M)
>   tests.profile.mode=heap
>   tests.profile.count=30
>   tests.profile.stacksize=1
>   tests.profile.linenumbers=false
> PERCENT       HEAP SAMPLES  STACK
> 49.63%        2430891M      jdk.internal.foreign.MappedMemorySegmentImpl#dup()
> 28.37%        1389691M      jdk.internal.foreign.HeapMemorySegmentImpl$OfByte#fromArray()
> 21.43%        1049362M      jdk.internal.foreign.HeapMemorySegmentImpl$OfByte#dup()
> 0.09%         4624M         org.apache.lucene.util.FixedBitSet#<init>()
> ```
> 
> See the number of heap samples, this cannot be! The only difference to the JFR heap profile in https://bugs.openjdk.java.net/browse/JDK-8268743 is the order (here we have more dups of the MappedMemorySegment (source from where it is copied).

This is strange. All the copy methods are now forced inlined, so I can't understand why you would observe allocation coming from there (I would expect C2 to just optimize that method regardless of its context). The only thing left for you to try is this:

https://github.com/mcimadamore/panama-foreign/tree/small_copy_benchmark/

Which contains a static method `copy` in MemorySegment which takes offsets. This method avoids the slices (but you would still need to wrap arrays) - it would be interesting to see if the allocation profile on Lucene changes at all.

mcimadamore:
> > I don't want to post too much in this issue as it is not fully related, but it looks this is also affected by the same garbage problem described in https://bugs.openjdk.java.net/browse/JDK-8268743:
> > ```
> > PROFILE SUMMARY from 14119950 events (total: 4897631M)
> >   tests.profile.mode=heap
> >   tests.profile.count=30
> >   tests.profile.stacksize=1
> >   tests.profile.linenumbers=false
> > PERCENT       HEAP SAMPLES  STACK
> > 49.63%        2430891M      jdk.internal.foreign.MappedMemorySegmentImpl#dup()
> > 28.37%        1389691M      jdk.internal.foreign.HeapMemorySegmentImpl$OfByte#fromArray()
> > 21.43%        1049362M      jdk.internal.foreign.HeapMemorySegmentImpl$OfByte#dup()
> > 0.09%         4624M         org.apache.lucene.util.FixedBitSet#<init>()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > See the number of heap samples, this cannot be! The only difference to the JFR heap profile in https://bugs.openjdk.java.net/browse/JDK-8268743 is the order (here we have more dups of the MappedMemorySegment (source from where it is copied).
> 
> This is strange. All the copy methods are now forced inlined, so I can't understand why you would observe allocation coming from there (I would expect C2 to just optimize that method regardless of its context). The only thing left for you to try is this:
> 
> https://github.com/mcimadamore/panama-foreign/tree/small_copy_benchmark/
> 
> Which contains a static method `copy` in MemorySegment which takes offsets. This method avoids the slices (but you would still need to wrap arrays) - it would be interesting to see if the allocation profile on Lucene changes at all.

Also, would it be possible for you to check if the methods in MemoryCopy are compiled at all? @ForceInline won't do much if the method that caller is not compiled, it seems.

uschindler:
> Also, would it be possible for you to check if the methods in MemoryCopy are compiled at all? @ForceInline won't do much if the method that caller is not compiled, it seems.

Will do once I figured out how to change this in our benchmark tool. But I think I can just pass the correct command line options to print compilations inside the class.

Let me check your other branch first!

uschindler:
I am testing now this change with your branch: https://github.com/apache/lucene/commit/02738d8140bde23dfcd22eeca1ca1acb16e8b15e

uschindler:
With this variant, profile looks like this:
```
PROFILE SUMMARY from 316771 events (total: 112007M)
  tests.profile.mode=heap
  tests.profile.count=30
  tests.profile.stacksize=1
  tests.profile.linenumbers=false
PERCENT       HEAP SAMPLES  STACK
72.60%        81319M        jdk.internal.foreign.HeapMemorySegmentImpl$OfByte#fromArray()
4.23%         4738M         org.apache.lucene.util.FixedBitSet#<init>()
3.14%         3516M         org.apache.lucene.search.ExactPhraseMatcher$1$1#getImpacts()
2.77%         3105M         org.apache.lucene.search.ExactPhraseMatcher$1#getImpacts()
```

Performance is much better, but not yet ideal!

Maybe you should really think about implementing all those copy methods between arrays and segments without slicing and ideally also without 

Wouldn't it be so complicated to just implement all the MemoryCopy methods like that?

test/jdk/java/foreign/TestMemoryCopy.java
```diff
@@ -16,106 +16,390 @@
...
+import jdk.incubator.foreign.MemorySegment;
 
+import jdk.incubator.foreign.ValueLayout;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+/**
+ * These tests exercise the MemoryCopy copyFromArray(...) and copyToArray(...).
+ * To make these tests more challenging the segment is a view of the given array,
+ * which makes the copy operations overlapping self-copies.  Thus, this checks the claim:
```
JornVernee, line 49: I'm not sure I understand this claim; it seems to me that the test sets up an array view segment, but then copies the contents into a new array (using toArray) as src/dst to do the actual copying? i.e. the source and destination seem disjoint.

test/jdk/java/foreign/TestMemoryCopy.java
```diff
@@ -16,106 +16,390 @@
...
+
+            @Override
+            int length(double[] arr) {
+                return arr.length;
+            }
+        };
+    }
+
+    @DataProvider
+    Object[][] copyModes() {
```
JornVernee, line 387: Seems unused.

JornVernee (commented):
I guess this is still ongoing, but for now some comments inline.

mcimadamore:
> Performance is much better, but not yet ideal!
> 
> Maybe you should really think about implementing all those copy methods between arrays and segments without slicing and ideally also without wrapped segments.
> 
> Wouldn't it be so complicated to just implement all the MemoryCopy methods like that?

As you say, it is fairly easy to re-implement the various methods using unsafe directly. It does create a bit of a problem in terms of API stacking, in the sense that the `MemorySegment::copyFrom` method, while in principle is a "primitive" in reality cannot act like one because of performance-related concerns. But perhaps this is just a transient concern: eventually we'll get primitive classes and we'll be able to avoid allocation w/o having to rely on escape analysis.

I'll work to re-implement the various methods tomorrow.

src/hotspot/share/oops/methodData.cpp
```diff
@@ -1591,6 +1591,11 @@ bool MethodData::profile_memory_access(const methodHandle& m, int bci) {
         return true;
       }
     }
+    if (inv.klass() == vmSymbols::jdk_incubator_foreign_MemoryCopy()) {
+      if (inv.name()->starts_with("copy")) {
+        return true;
+      }
+    }
```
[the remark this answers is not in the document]
mcimadamore: Ok, I can revert that - IMHO, the fact that a memory barrier is always added is a peformance issue that should be addressed, as that would impact things like liveness checks - but we can get there when we get there.

test/jdk/java/foreign/TestMemoryCopy.java
mcimadamore: This comment was in the original patch - I'll fix or drop it.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -373,6 +374,46 @@ default MemorySegment asSlice(MemoryAddress newBase) {
...
+     * For example, this may occur if the same file is {@linkplain MemorySegment#mapFile mapped} to two segments.
+     *
+     * @implSpec
+     * If {@code srcElementLayout.byteSize() == 1}, or if {@code srcElementLayout.order() == dstElementLayout.order()}, calling
+     * this method must be equivalent to calling {@link #copyFrom(MemorySegment)} with the same source segment.
+     *
+     * @param dstElementLayout the element layout associated with this segment.
+     * @param src the source segment.
+     * @param srcElementLayout the element layout associated with the source segment.
+     * @throws IndexOutOfBoundsException if {@code src.byteSize() > this.byteSize()}.
```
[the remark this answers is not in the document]
mcimadamore: For now, that is consistent with what the other `MemorySegment::copyFrom` does.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryCopy.java
```diff
@@ -0,0 +1,574 @@
...
+ * This class provides convenient methods for copying data between primitive arrays and memory segments.
+ *
+ * <p>If the source (destination) segment is actually a view of the destination (source) array,
+ * and if the copy region of the source overlaps with the copy region of the destination,
+ * the copy of the overlapping region is performed as if the data in the overlapping region
+ * were first copied into a temporary segment before being copied to the destination.</p>
+ *
+ * <p>
+ * Copy operations defined in this class accept a <em>byte order</em> parameter. If the specified byte order is different
+ * from the <em>native</em> byte order, a byte swap operation is performed on each array elements
```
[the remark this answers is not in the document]
mcimadamore: That seems consistent with MemoryAccess, to be honest.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryCopy.java
```diff
@@ -0,0 +1,574 @@
...
+     */
+    @ForceInline
+    public static void copyFromArray(
+            char[] srcArray, int srcIndexChars, int srcCopyLengthChars,
+            MemorySegment dstSegment, long dstOffsetBytes,
+            ByteOrder order) {
+        Objects.requireNonNull(dstSegment);
+        MemorySegment srcSegmentSlice =
+                MemorySegment.ofArray(srcArray).asSlice(srcIndexChars * 2L, srcCopyLengthChars * 2L);
+        MemorySegment dstSegmentSlice = dstSegment.asSlice(dstOffsetBytes, srcCopyLengthChars * 2L);
```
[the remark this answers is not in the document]
leerho: The variable name `srcCopyLengthChars` is for clarity to the user of the API, who does not see the implementation code. Because there are two different types of units in the method call, one in Chars and one in Bytes, it is important to clarify the units of the measure as well as which of source or destination it is associated with.

test/jdk/java/foreign/TestMemoryCopy.java
leerho: I think what you are seeing the extra copy is in the creation of the `truthSegment`, where an extra copy had to be used because there is no API (yet), to help the code determine which direction the copy should be performed in a self-copy scenario.  If the call to the MemoryCopy API does not perform the self-copy correctly, comparing its result with the `truthSegment` will definitely reveal an error.

uschindler:
I ran the benchmark on this pull request again (non-tiered batch compilation). Here is the output of PrintCompilation, grepped on `MemorySegmentIndexInput`, our impl. As you see our `readBytes()` and `readLongs()` (both calling `MemoryCopy#copyToArray()`) appears:
```
thetaphi@serv1:~/benchmark/util$ fgrep MemorySegmentIndexInput ../logs/baseline_vs_patch.my_modified_version.0.stdout
   3010  824   !b        org.apache.lucene.store.MemorySegmentIndexInput::readByte (130 bytes)
   3035  845    b        org.apache.lucene.store.MemorySegmentIndexInput::length (5 bytes)
   6117 1414   !b        org.apache.lucene.store.MemorySegmentIndexInput$SingleSegmentImpl::readByte (28 bytes)
   6182 1453   !b        org.apache.lucene.store.MemorySegmentIndexInput::readBytes (50 bytes)
   6281 1489   !b        org.apache.lucene.store.MemorySegmentIndexInput::readLongs (57 bytes)
   6596 1570   !b        org.apache.lucene.store.MemorySegmentIndexInput$SingleSegmentImpl::readShort (28 bytes)
   6694 1588    b        org.apache.lucene.store.MemorySegmentIndexInput::ensureOpen (13 bytes)
   6966 1622   !b        org.apache.lucene.store.MemorySegmentIndexInput$SingleSegmentImpl::seek (31 bytes)
   6975 1627    b        org.apache.lucene.store.MemorySegmentIndexInput$SingleSegmentImpl::getFilePointer (9 bytes)
  11819 2091    b        org.apache.lucene.store.MemorySegmentIndexInput$SingleSegmentImpl::<init> (26 bytes)
  11823 2092    b        org.apache.lucene.store.MemorySegmentIndexInput::<init> (103 bytes)
  11883 2095    b        org.apache.lucene.store.MemorySegmentIndexInput::buildSlice (154 bytes)
  12122 2120   !b        org.apache.lucene.store.MemorySegmentIndexInput::readLong (42 bytes)
  15710 2204    b        org.apache.lucene.store.MemorySegmentIndexInput$SingleSegmentImpl::slice (9 bytes)
  15711 2205    b        org.apache.lucene.store.MemorySegmentIndexInput::slice (56 bytes)
  15731 2219   !b        org.apache.lucene.store.MemorySegmentIndexInput::clone (37 bytes)
  16198 2242    b        org.apache.lucene.store.MemorySegmentIndexInput$SingleSegmentImpl::clone (5 bytes)
```

MemoryCopy is also compiled:

```
thetaphi@serv1:~/benchmark/util$ fgrep MemoryCopy ../logs/baseline_vs_patch.my_modified_version.0.stdout
   6187 1454    b        jdk.incubator.foreign.MemoryCopy::copyToArray (44 bytes)
   6286 1490    b        jdk.incubator.foreign.MemoryCopy::copyToArray (76 bytes)
```

Not sure why escape analysis does not work. Seems to be a bug or it is again the SMALL segments.

uschindler:
With tiered compilation (all defaults):

```
thetaphi@serv1:~/benchmark/util$ egrep '(MemorySegmentIndexInput::readBytes|MemorySegmentIndexInput::readLongs|MemoryCopy)' ../logs/
baseline_vs_patch.my_modified_version.0.stdout
    575 2580   !   3       org.apache.lucene.store.MemorySegmentIndexInput::readBytes (50 bytes)
    575 2581       3       jdk.incubator.foreign.MemoryCopy::copyToArray (44 bytes)
   1132 4188   !   3       org.apache.lucene.store.MemorySegmentIndexInput::readLongs (57 bytes)
   1132 4193       3       jdk.incubator.foreign.MemoryCopy::copyToArray (76 bytes)
   1200 4316   !   4       org.apache.lucene.store.MemorySegmentIndexInput::readBytes (50 bytes)
   1209 2580   !   3       org.apache.lucene.store.MemorySegmentIndexInput::readBytes (50 bytes)   made not entrant
   1209 4343   !   4       org.apache.lucene.store.MemorySegmentIndexInput::readLongs (57 bytes)
   1220 4188   !   3       org.apache.lucene.store.MemorySegmentIndexInput::readLongs (57 bytes)   made not entrant
   3324 2580   !   3       org.apache.lucene.store.MemorySegmentIndexInput::readBytes (50 bytes)   made zombie
   3328 4188   !   3       org.apache.lucene.store.MemorySegmentIndexInput::readLongs (57 bytes)   made zombie
```

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryCopy.java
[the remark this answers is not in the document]
JornVernee: True. Ok, that actually makes sense now that I compare it with what memaccess var handles do. i.e. the byte order always corresponds to the memory segment, since the array is of course in native order. So, in practice this means a swap occurs when the specified or is non-native. Thanks.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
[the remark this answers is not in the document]
JornVernee: Ok, makes sense.

test/jdk/java/foreign/TestMemoryCopy.java
JornVernee: Actually, I'm talking about the call to toArray here (as well as for CopyTo): 

```
ByteOrder bo = mode.swap ? NON_NATIVE_ORDER : NATIVE_ORDER;
//CopyFrom
Object srcArr = helper.toArray(base); // <---------------------
int srcIndex = mode.direction ? 0 : indexShifts;
int srcCopyLen = helper.length(srcArr) - indexShifts;
```

This call to `toArray` ends of calling one of the `MemorySegment::to*Array` methods, which always copy the contents of the segment.

uschindler:
Hi @mcimadamore, 

I was thinking about that last night ant I think I know what the issue and why it affects Lucene: As you remember, I mentioned that without tiered compilation and batch compilation enabled (the default for our benchmark to get more predictable result), the garbage produced and its collection is horrible. These are the first results above. As you see the copy methods are optimized only after 6 seconds. The total runtime of the benchmark was 73 seconds (50 seconds with our old code). So it looks like for the first 6 seconds the benchmark is hammering with many parallel queries the lucene index, creating incredible amounts of garbage (3 object instances for each memory copy), until the compilation kicks in. This is completely new to lucene, because the very low level code is carefully written to not produce any object instances, because from our experience, escape analysis kicks in way too late. The readBytes() method is not called in a loop, but it is still called quite often in a very complex code path making it hard for the optimizer to kick in. What happens now: In our benchmark, we use many threads to execute fulltext queries and without the optimization kicking in, it creates millions of objects and those have to be garbage collected. Due to the high load (many threads and expensive low level query execution), GC has less CPU resources and then fails to kick in early enough and wastes at end like 23 seconds extra in cleaning up the useless garbage, possibly with stop-the-world phases.

With tiered compilation enabled, the compilation kicks in much earlier (after half a second already for C1) and after 1.2 seconds for C2.  This avoids a lot of garbage, (about 1/6s also in the output at end). The total runtime of the benchmark goes down from 73 seconds to approx 55 seconds). So garbage collector does not go crazy, but we still see a slowdown.

IMHO, inside Panama internals you should not rely on escape analysis to kick in early enough and implement stuff like copy methods without producing too much garbage by default.

> It does create a bit of a problem in terms of API stacking, in the sense that the MemorySegment::copyFrom method, while in principle is a "primitive" in reality cannot act like one because of performance-related concerns. 

Maybe document in the API of MemoryCopy, that those methods behave like the slicing code, but are much more efficient.

But I agree you should really work on letting the escape analysis kick in earlier, especially if code is not used in loops (like memory copy).

Thanks,
Uwe

mcimadamore:
> Hi @mcimadamore,
> 
> I was thinking about that last night ant I think I know what the issue and why it affects Lucene: As you remember, I mentioned that without tiered compilation and batch compilation enabled (the default for our benchmark to get more predictable result), the garbage produced and its collection is horrible. These are the first results above. As you see the copy methods are optimized only after 6 seconds. The total runtime of the benchmark was 73 seconds (50 seconds with our old code). So it looks like for the first 6 seconds the benchmark is hammering with many parallel queries the lucene index, creating incredible amounts of garbage (3 object instances for each memory copy), until the compilation kicks in. This is completely new to lucene, because the very low level code is carefully written to not produce any object instances, because from our experience, escape analysis kicks in way too late. The readBytes() method is not called in a loop, but it is still called quite often in a very complex code path making it hard for the optimizer to kick in. What happens now: In our benchmark, we use many threads to execute fulltext queries and without the optimization kicking in, it creates millions of objects and those have to be garbage collected. Due to the high load (many threads and expensive low level query execution), GC has less CPU resources and then fails to kick in early enough and wastes at end like 23 seconds extra in cleaning up the useless garbage, possibly with stop-the-world phases.
> 
> With tiered compilation enabled, the compilation kicks in much earlier (after half a second already for C1) and after 1.2 seconds for C2. This avoids a lot of garbage, (about 1/6s also in the output at end). The total runtime of the benchmark goes down from 73 seconds to approx 55 seconds). So garbage collector does not go crazy, but we still see a slowdown of 10% for some queries, but not 40% slowdown for queries like in the non-tiered case.
> 
> IMHO, inside Panama internals you should not rely on escape analysis to kick in early enough and implement stuff like copy methods without producing too much garbage by default.
> 
> > It does create a bit of a problem in terms of API stacking, in the sense that the MemorySegment::copyFrom method, while in principle is a "primitive" in reality cannot act like one because of performance-related concerns.
> 
> Maybe document in the API of MemoryCopy, that those methods behave like the slicing code, but are much more efficient.
> 
> But I agree you should really work on letting the escape analysis kick in earlier, especially if code is not used in loops (like memory copy).
> 
> Thanks,
> Uwe

I think I agree with the conclusion - code that is doing memory copy will typically not sit inside a loop (although sometimes it might) and will get optimized late. If we rely on compilation/escape analysis to get rid of slicing in such cases we might never get there, or get there after the horses have left the proverbial barn. While I think Lucene is especially sensitive in this direction (I'm not aware of other similar reports), it is a real world scenario, which means we have to readjust our strategy here.

Your results (thanks for spending time in validating the various patch) are pretty clear: there is a direct correlation between the amount of garbage that is generated and the overall throughput of the benchmark (I was mildly afraid we were dealing with _two_ issues here). The solution is relatively straightforward - and maybe at some point in the future Valhalla will help allowing us to express the code in terms of higher-level APIs, w/o sacrificing on the allocation cost.

mcimadamore:
@uschindler I've just pushed a new iteration which uses unsafe directly when copying to/from arrays. Please give that a try - if numbers look good we will finalize this version.

uschindler:
> @uschindler I've just pushed a new iteration which uses unsafe directly when copying to/from arrays. Please give that a try - if numbers look good we will finalize this version.

There's something fishy:

```
Exception in thread "main" java.lang.UnsupportedOperationException: Attempt to write a read-only segment
        at jdk.incubator.foreign/jdk.internal.foreign.AbstractMemorySegmentImpl.checkAccess(AbstractMemorySegmentImpl.java:359)
        at jdk.incubator.foreign/jdk.incubator.foreign.MemoryCopy.copyToArray(MemoryCopy.java:108)
        at org.apache.lucene.store.MemorySegmentIndexInput.readBytes(MemorySegmentIndexInput.java:151)
```

uschindler:
Looks like a copy-paste error:

```java
srcImpl.checkAccess(srcOffsetBytes, dstCopyLengthBytes, false);
```

this should pass `true`, because the source check is readonly.

Maybe change the test to make the source segment read only. We catched it as we use a memory mapped file mapped read only.

mcimadamore:
> Looks like a copy-paste error:
> 
> ```java
> srcImpl.checkAccess(srcOffsetBytes, dstCopyLengthBytes, false);
> ```
> 
> this should pass `true`, because the source check is readonly.
> 
> Maybe change the test to make the source segment read only. We catched it as we use a memory mapped file mapped read only.

Fixed now, apologizes for the inconvenience!

uschindler:
Hi,

I can confirm. The garbage collection issues are gone. The heap profile of Lucene looks identical to our old ByteBuffer-based implementation:

```
PERCENT       HEAP SAMPLES  STACK
16.78%        4781M         org.apache.lucene.util.FixedBitSet#<init>()
10.17%        2897M         org.apache.lucene.search.ExactPhraseMatcher$1$1#getImpacts()
8.51%         2424M         java.util.AbstractList#iterator()
7.67%         2184M         org.apache.lucene.search.ExactPhraseMatcher$1#getImpacts()
4.37%         1245M         org.apache.lucene.codecs.lucene90.blocktree.SegmentTermsEnumFrame#<init>()
3.46%         985M          org.apache.lucene.queryparser.charstream.FastCharStream#refill()
3.11%         885M          org.apache.lucene.util.ArrayUtil#growExact()
2.93%         834M          java.util.ArrayList#grow()
2.37%         675M          org.apache.lucene.codecs.lucene90.Lucene90PostingsReader$BlockDocsEnum#<init>()
2.14%         609M          org.apache.lucene.util.BytesRef#<init>()
1.76%         502M          java.util.ArrayList#iterator()
1.73%         491M          org.apache.lucene.util.fst.ByteSequenceOutputs#read()
1.52%         432M          org.apache.lucene.util.PriorityQueue#<init>()
1.49%         425M          jdk.internal.misc.Unsafe#allocateUninitializedArray()
1.30%         369M          org.apache.lucene.codecs.lucene90.blocktree.IntersectTermsEnumFrame#load()
1.27%         362M          org.apache.lucene.codecs.lucene90.Lucene90PostingsReader#newTermState()
1.26%         359M          org.apache.lucene.codecs.lucene90.blocktree.SegmentTermsEnum#getFrame()
1.20%         342M          org.apache.lucene.codecs.lucene90.Lucene90PostingsReader$BlockImpactsDocsEnum#<init>()
1.10%         312M          org.apache.lucene.codecs.lucene90.blocktree.SegmentTermsEnum#<init>()
1.03%         292M          org.apache.lucene.store.MemorySegmentIndexInput#buildSlice()
0.99%         283M          org.apache.lucene.queryparser.charstream.FastCharStream#GetImage()
0.98%         279M          org.apache.lucene.codecs.lucene90.Lucene90PostingsReader$BlockImpactsPostingsEnum#<init>()
0.98%         279M          org.apache.lucene.codecs.lucene90.ForUtil#<init>()
0.94%         267M          org.apache.lucene.util.DocIdSetBuilder$Buffer#<init>()
0.90%         257M          java.util.AbstractList#listIterator()
0.88%         250M          org.apache.lucene.codecs.lucene90.blocktree.IntersectTermsEnumFrame#<init>()
0.81%         231M          jdk.internal.foreign.MappedMemorySegmentImpl#dup()
0.80%         229M          org.apache.lucene.codecs.lucene90.compressing.Lucene90CompressingStoredFieldsReader$BlockState#document()
0.77%         219M          java.util.Arrays#copyOfRange()
0.69%         198M          java.util.Arrays#asList()
```

ByteBuffer profile:

```
PERCENT       HEAP SAMPLES  STACK
16.44%        4774M         org.apache.lucene.util.FixedBitSet#<init>()
9.88%         2870M         org.apache.lucene.search.ExactPhraseMatcher$1$1#getImpacts()
8.34%         2421M         java.util.AbstractList#iterator()
7.42%         2155M         org.apache.lucene.search.ExactPhraseMatcher$1#getImpacts()
4.40%         1278M         org.apache.lucene.codecs.lucene90.blocktree.SegmentTermsEnumFrame#<init>()
3.30%         959M          org.apache.lucene.util.ArrayUtil#growExact()
3.09%         896M          java.util.ArrayList#grow()
2.89%         839M          org.apache.lucene.queryparser.charstream.FastCharStream#refill()
2.26%         655M          org.apache.lucene.codecs.lucene90.Lucene90PostingsReader$BlockDocsEnum#<init>()
1.98%         573M          org.apache.lucene.util.BytesRef#<init>()
1.91%         553M          java.util.ArrayList#iterator()
1.69%         490M          org.apache.lucene.util.fst.ByteSequenceOutputs#read()
1.54%         448M          jdk.internal.misc.Unsafe#allocateUninitializedArray()
1.53%         444M          java.nio.DirectByteBufferR#duplicate()
1.52%         440M          org.apache.lucene.util.PriorityQueue#<init>()
1.42%         411M          org.apache.lucene.codecs.lucene90.blocktree.IntersectTermsEnumFrame#load()
1.28%         371M          org.apache.lucene.codecs.lucene90.blocktree.SegmentTermsEnum#getFrame()
1.25%         362M          org.apache.lucene.codecs.lucene90.Lucene90PostingsReader#newTermState()
1.23%         356M          org.apache.lucene.queryparser.charstream.FastCharStream#GetImage()
1.20%         348M          org.apache.lucene.codecs.lucene90.Lucene90PostingsReader$BlockImpactsDocsEnum#<init>()
1.00%         289M          org.apache.lucene.codecs.lucene90.blocktree.SegmentTermsEnum#<init>()
0.92%         268M          java.util.AbstractList#listIterator()
0.91%         265M          org.apache.lucene.codecs.lucene90.blocktree.IntersectTermsEnumFrame#<init>()
0.91%         264M          org.apache.lucene.codecs.lucene90.ForUtil#<init>()
0.89%         257M          java.util.Arrays#copyOf()
0.82%         238M          org.apache.lucene.util.DocIdSetBuilder$Buffer#<init>()
0.82%         237M          org.apache.lucene.codecs.lucene90.Lucene90PostingsReader$BlockImpactsPostingsEnum#<init>()
0.79%         230M          org.apache.lucene.codecs.lucene90.compressing.Lucene90CompressingStoredFieldsReader$BlockState#document()
0.77%         224M          java.nio.DirectByteBufferR#slice()
0.70%         202M          java.util.Arrays#asList()
```

As you see we have a few segment slices/dups, but similar for MappedByteBuffers (this comes from the fact that Lucene sometimes opens views on compound files.

The performance is still a little bit slower for some query types (mainly facets), but I have the feeling this comes from the overhead when coyping small arrays, as well as scoping checks.

@mcimadamore: You mentioned that for small arrays, copy-loops may fit better. Do you have any suggestions what loop sizes we are talking about. Lucene's `long[]` are safely sizes <= 64 and `float[]` of sizes <= 1024 elements, so maybe it's a good idea to just use the default read loop and not specialize it to do bulk copy.

Here are Lucene's results:

```
                    TaskQPS baseline      StdDevQPS my_modified_version      StdDev                Pct diff p-value
   BrowseMonthTaxoFacets        1.17      (5.7%)        1.07      (6.4%)   -8.6% ( -19% -    3%) 0.000
BrowseDayOfYearTaxoFacets        1.13      (7.1%)        1.04      (7.8%)   -7.9% ( -21% -    7%) 0.001
    BrowseDateTaxoFacets        1.13      (7.1%)        1.04      (8.0%)   -7.7% ( -21% -    7%) 0.001
    HighTermTitleBDVSort       54.84     (20.1%)       50.84     (15.8%)   -7.3% ( -35% -   35%) 0.202
                PKLookup      188.75      (2.2%)      181.26      (1.4%)   -4.0% (  -7% -    0%) 0.000
             AndHighHigh       61.16      (6.1%)       59.56      (5.6%)   -2.6% ( -13% -    9%) 0.158
   HighTermDayOfYearSort       63.43     (10.8%)       62.28     (14.2%)   -1.8% ( -24% -   26%) 0.650
                  Fuzzy2       57.80      (5.1%)       56.81      (5.1%)   -1.7% ( -11% -    8%) 0.288
       HighTermMonthSort       66.31     (11.8%)       65.22     (17.0%)   -1.7% ( -27% -   30%) 0.720
                 Respell       68.71      (2.4%)       67.71      (2.1%)   -1.5% (  -5% -    3%) 0.041
              HighPhrase       10.97      (4.8%)       10.83      (6.2%)   -1.3% ( -11% -   10%) 0.468
              AndHighMed       44.63      (4.4%)       44.09      (4.2%)   -1.2% (  -9% -    7%) 0.370
              TermDTSort      105.50     (17.0%)      104.58     (19.8%)   -0.9% ( -32% -   43%) 0.880
               OrHighMed       34.38      (3.5%)       34.14      (3.9%)   -0.7% (  -7% -    6%) 0.556
               OrHighLow      337.44      (3.1%)      335.46      (4.6%)   -0.6% (  -8% -    7%) 0.634
               MedPhrase      179.30      (2.7%)      178.40      (4.1%)   -0.5% (  -7% -    6%) 0.646
              AndHighLow      369.54      (2.9%)      367.78      (3.3%)   -0.5% (  -6% -    5%) 0.628
                  Fuzzy1       52.74      (9.8%)       52.49      (9.2%)   -0.5% ( -17% -   20%) 0.874
              OrHighHigh       12.03      (3.8%)       11.97      (3.7%)   -0.4% (  -7% -    7%) 0.717
                  IntNRQ       73.05      (1.5%)       72.78      (2.5%)   -0.4% (  -4% -    3%) 0.569
         MedSloppyPhrase       25.01      (1.9%)       24.98      (1.7%)   -0.1% (  -3% -    3%) 0.856
                Wildcard       76.65     (21.9%)       76.60     (21.5%)   -0.1% ( -35% -   55%) 0.993
    HighIntervalsOrdered       10.04      (4.3%)       10.03      (5.5%)   -0.0% (  -9% -   10%) 0.979
         LowSloppyPhrase       39.07      (2.3%)       39.08      (2.3%)    0.0% (  -4% -    4%) 0.970
BrowseDayOfYearSSDVFacets        4.00      (4.6%)        4.00      (5.6%)    0.1% (  -9% -   10%) 0.972
             LowSpanNear       51.01      (2.0%)       51.09      (1.6%)    0.2% (  -3% -    3%) 0.784
               LowPhrase      265.84      (2.0%)      266.25      (2.9%)    0.2% (  -4% -    5%) 0.845
                 MedTerm     1532.72      (4.0%)     1535.72      (4.5%)    0.2% (  -7% -    9%) 0.884
                HighTerm     1382.84      (5.2%)     1388.91      (4.9%)    0.4% (  -9% -   11%) 0.784
            OrHighNotMed      851.21      (3.9%)      856.51      (3.6%)    0.6% (  -6% -    8%) 0.599
            OrNotHighLow      772.85      (3.2%)      778.22      (3.6%)    0.7% (  -5% -    7%) 0.522
   BrowseMonthSSDVFacets        4.06      (6.3%)        4.09      (6.5%)    0.8% ( -11% -   14%) 0.692
            OrHighNotLow      726.91      (4.7%)      732.78      (5.0%)    0.8% (  -8% -   10%) 0.596
             MedSpanNear      101.32      (3.0%)      102.16      (2.7%)    0.8% (  -4% -    6%) 0.354
            HighSpanNear        0.90      (3.0%)        0.91      (2.9%)    0.9% (  -4% -    6%) 0.364
        HighSloppyPhrase       13.89      (4.6%)       14.02      (4.4%)    0.9% (  -7% -   10%) 0.509
            OrNotHighMed      597.86      (3.0%)      604.63      (3.8%)    1.1% (  -5% -    8%) 0.298
                 LowTerm     1874.30      (3.2%)     1898.96      (4.5%)    1.3% (  -6% -    9%) 0.287
                 Prefix3       38.02      (8.6%)       38.53     (10.1%)    1.3% ( -15% -   21%) 0.655
           OrNotHighHigh      719.81      (3.8%)      731.82      (5.3%)    1.7% (  -7% -   11%) 0.252
           OrHighNotHigh      723.65      (4.0%)      736.28      (3.7%)    1.7% (  -5% -    9%) 0.153
```

test/jdk/java/foreign/TestMemoryCopy.java
```diff
@@ -16,106 +16,390 @@
...
-        for (int i = 0 ; i < size ; i++) {
-            BYTE_HANDLE.set(s2.segment.asSlice(i), (byte)0);
+    public static MemorySegment srcSegment(int bytesLength) {
+        byte[] arr = new byte[bytesLength];
+        for (int i = 0; i < arr.length; i++) {
+            arr[i] = (byte)i;
         }
-        for (int i = 0 ; i < size ; i++) {
-            BYTE_HANDLE.set(s1.segment.asSlice(i), (byte) i);
+        return MemorySegment.ofArray(arr);
```
uschindler, line 119: This segment should be made readonly, to assert all checkAccess() calls.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryCopy.java
```diff
@@ -0,0 +1,696 @@
...
+ * <p>Unless otherwise specified, passing a {@code null} argument, or an array argument containing one or more
+ * {@code null} elements to a method in this class causes a {@link NullPointerException NullPointerException}
+ * to be thrown.</p>
+ */
+public final class MemoryCopy {
+
+    private static final ScopedMemoryAccess scopedMemoryAccess = ScopedMemoryAccess.getScopedMemoryAccess();
+    private static final Unsafe unsafe = Unsafe.getUnsafe();
+
+    private static final int BYTE_BASE = unsafe.arrayBaseOffset(byte[].class);
```
uschindler, line 58: Are there not constants for this already in `jdk.internal.misc.Unsafe`?

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryCopy.java
```diff
@@ -0,0 +1,696 @@
...
+    public static void copyFromArray(
+            char[] srcArray, int srcIndexChars, int srcCopyLengthChars,
+            MemorySegment dstSegment, long dstOffsetBytes,
+            ByteOrder order) {
+        Objects.requireNonNull(srcArray);
+        Objects.requireNonNull(dstSegment);
+        Objects.checkFromIndexSize(srcIndexChars, srcCopyLengthChars, srcArray.length);
+        AbstractMemorySegmentImpl destImpl = (AbstractMemorySegmentImpl)dstSegment;
+        destImpl.checkAccess(dstOffsetBytes, srcCopyLengthChars << 1, false);
+        if (order == ByteOrder.nativeOrder()) {
```
uschindler, line 153: I like this code much more than the previous one, as it makes clear: swap tru or false, just a boolean! The Usage of MemorySegment.copyFrom() was very complicated as it used ValueLayouts, one for each size.

uschindler (approved):
Hi,
to me this looks good now.

I like the new code much more, because the previous version of this pull request was a bit strange:
- In MemoryCopy the code checked if native order equals parameter order and then used ValueLayout (one instance for each valuecsize) to call MemorySegment.copyFrom()
- MemorySegment.copyFrom() took the valueSource and then compared it again to native byte order, just to convert it back to a boolean: swap yes/no.

All this was so much code and I was always afraid that there is some case where the ValueLayouts were used wrongly.

So much better! I love it!

mcimadamore:
> The performance is still a little bit slower for some query types (mainly facets), but I have the feeling this comes from the overhead when coyping small arrays, as well as scoping checks.
> 
> @mcimadamore: You mentioned that for small arrays, copy-loops may fit better. Do you have any suggestions what loop sizes we are talking about. Lucene's `long[]` are safely sizes <= 64 and `float[]` of sizes <= 1024 elements, so maybe it's a good idea to just use the default read loop and not specialize it to do bulk copy.

This does indeed look a lot better, thanks for testing. The issue with memory copy is that it adds unconditional memory barriers - so, in some circumstances it can cause performance degradation. If you copy in a loop, you use Unsafe::get/put and those will only add barriers if access is on-heap.

We have a bug to track this issue:
https://bugs.openjdk.java.net/browse/JDK-8269119

In terms of sizes, I think the ByteBuffer API used to avoid copying for extremely small sizes, (6 logical elements, if I'm correct).

I also think that there are other performance potholes around bound checking (as we speak, we're investigating one mentioned here [1]) and some of those will not be even addressable until we get proper support for long loop optimizations [2].

So, we know that, performance-wise, we're still not 100% there, but it seems like we're getting closer.

[1] - https://mail.openjdk.java.net/pipermail/panama-dev/2021-June/014260.html
[2] - https://github.com/openjdk/jdk/pull/2045

uschindler:
Small update: With tiered compilation enabled, perf numbers are getting closer to ByteBuffer API. The difference is now smaller than std deviation. Some queries are 8% faster, other 8% slower.

leerho (approved):

test/jdk/java/foreign/TestMemoryCopy.java
JornVernee: You're right, I seem to have misread the code earlier. For some reason I thought a copy to/from the `srcArr` to/fromthe `base` segment was happening, but looking at it now, I can see it's doing as you describe. Sorry.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryCopy.java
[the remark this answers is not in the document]
mcimadamore: Honestly, I tend to agree with Jorn here. In a copy, there is:
* a source (either an array or a segment)
* a source offset (expressed either as element index, or as byte offset)
* a destination (either an array or a segment)
* a destination offset (expressed either as element index, or as byte offset)
* a length (expressed either as number of elements, or as number of bytes)

There's only one length - it's important to distinguish between Chars vs. Bytes - e.g. copyLengthChars vs. copyLengthBytes. But adding the `src` seems confusing because the `length` parameter is, IMHO neutral, and just indicates how many "things" have to be bulk-copied.

In fact, I'll go further and say that, when editing this patch I kept finding the placement of the length argument confusing. Having it as the trailing argument would be clearer IMHO.

leerho:
The conceptual model I had in mind when I proposed this API was that it seemed sensible to order the parameters first by the data flow: {<source parameters>, <destination parameters>}, then by {Object, offset, length} with ByteOrder stuck at the end. Thus:

- copyToArray(srcSegment, srcOffsetBytes, dstArray, dstIndexChars, dstCopyLengthChars, order);
- copyFromArray(srcArray, srcIndexChars, srcCopyLengthChars, dstSegment, dstOffsetBytes, order);

But this is by no means the only mental model we could use as this {<source>, <destination>} flow model is not consistently followed even within the current FMA API.

As an alternative, we could use the conceptual model of {<segment parameters>, <array parameters>}. Thus:

- copyToArray(srcSegment, srcOffsetBytes, dstArray, dstIndexChars, dstCopyLengthChars, order);
- copyFromArray(dstSegment, dstOffsetBytes, srcArray, srcIndexChars, srcCopyLengthChars, order);

For the user, this model has the parameters by type always in the same order, the length (and order) are at the end, and because of that the prefixes for the copy length could probably be dropped.  But I would prefer to keep the other prefixes as they are as additional cues to the user of the flow direction as well as association.

JornVernee (approved):

uschindler:
Hi,
Wouldn't it make sense to change MemoryCopy test to use a read-only source segment, so we ensure that the access checks are correct? The first bug I found was that it tried to check for write access on the copy source.

I left a comment where the check could be integrated into the test.

Otherwise, looks fine to me. I have no strong preference on the parameter order, but I would prefer System.arrayCopy style.

mcimadamore:
> Hi,
> Wouldn't it make sense to change MemoryCopy test to use a read-only source segment, so we ensure that the access checks are correct? The first bug I found was that it tried to check for write access on the copy source.

I think this has been addressed?
> 
> I left a comment where the check could be integrated into the test.
> 
> Otherwise, looks fine to me. I have no strong preference on the parameter order, but I would prefer System.arrayCopy style.

mcimadamore:
> > Hi,
> > Wouldn't it make sense to change MemoryCopy test to use a read-only source segment, so we ensure that the access checks are correct? The first bug I found was that it tried to check for write access on the copy source.
> 
> I think this has been addressed?
> 
> > I left a comment where the check could be integrated into the test.
> > Otherwise, looks fine to me. I have no strong preference on the parameter order, but I would prefer System.arrayCopy style.

Note that I couldn't just add asReadOnly() to the source segment, as we're copying bits from segment to itself. What I did instead was to turn as read only the "source slice" which is used by the test.

uschindler:
> Note that I couldn't just add asReadOnly() to the source segment, as we're copying bits from segment to itself. What I did instead was to turn as read only the "source slice" which is used by the test.

Ok, missed this. When I looked at you patch after you fixed it, I may have overseen it. Sorry for this.

test/jdk/java/foreign/TestMemoryCopy.java
uschindler: Ok - you added this to slice, e.g. here. I have overseen this! https://github.com/openjdk/panama-foreign/pull/555/files#diff-cf69e75a9e792b2f93dec150609654fe2dc3b4f3b45a8fbdca1b5b450d52173bR82

uschindler (approved):
Looks fine! Performance tested.

mcimadamore:
I'll integrate for now - we'll come back to decide how to fix the signatures at a later point. Thanks!

PR #565, opened 2021-07-07
8269807: Implement the fix for JDK-8269240 in the panama-foreign repo

JornVernee:
Hi,

This patch re-writes the prologue and epilogue code of upcalls to model what is done for JNI by `JavaCallWrapper`. The old code did more or less the inverse of what was happening for downcalls, but this is not correct. The new code uses a pair of calls in the prologue and epilogue to `ProgrammableUpcallHandler::on_entry` and `ProgrammableUpcallHandler::on_exit`.

We now also need to save a little bit more information about the entry frame, namely the `JNIHandle` block that the prologue allocates, and the old handle block to restore later and for the GC to inspect. I've expanded the existing `AuxiliarySaves` struct for this and renamed and moved it to be a nested type in `OptimizedEntryBlob`. The move is mostly to avoid circular dependencies between headers.

During this rewrite, I also noticed that the return value of an upcall was not properly spilled around a call to detach a thread in the epilogue (only needed when we attached it in the prologue). Since we now always have a call in the epilogue, I've added the missing spill and fill code.

Since this was not caught by our existing testing, I've added an additional mode to TestUpcall to run test cases on a separate thread as well. I've split the three cases in TestUpcall into separate jtreg tests to keep execution times down, and to make it easier to see what failed if there is a test failure. (I'm not sure if there's a better way to ignore some testng test methods in jtreg other than throwing a `SkipException`, which also fills the console log unfortunately).

I've also added cases for testing with concurrent GCs to the test (`AsyncStackWalk`) against which the original failure was reported.

Besides that, there is a minor cleanup of some existing assert code that was copy/pasted, as well as a change to `TestByteBuffer` that was failing on WSL because of an function not being implemented, which I've now handled by skipping the test in that case.

Thanks,
Jorn

---------
### Progress

### Issue
 * [JDK-8269807](https://bugs.openjdk.java.net/browse/JDK-8269807): Implement the fix for JDK-8269240 in the panama-foreign repo

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/565/head:pull/565` \
`$ git checkout pull/565`

`$ git checkout pull/565` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/565/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 565`

`$ git pr show -t 565`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/565.diff">https://git.openjdk.java.net/panama-foreign/pull/565.diff</a>

</details>

### DIFF
--- a/make/test/JtregNativeJdk.gmk
+++ b/make/test/JtregNativeJdk.gmk
@@ -54,6 +54,7 @@ BUILD_JDK_JTREG_EXECUTABLES_CFLAGS_exeJliLaunchTest := \
     -I$(TOPDIR)/src/java.base/$(OPENJDK_TARGET_OS)/native/libjli
 
 BUILD_JDK_JTREG_LIBRARIES_LDFLAGS_libAsyncStackWalk := $(LIBCXX)
+BUILD_JDK_JTREG_LIBRARIES_LDFLAGS_libAsyncInvokers := $(LIBCXX)
 
 # Platform specific setup
 ifeq ($(call isTargetOs, windows), true)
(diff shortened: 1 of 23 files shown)


mcimadamore (approved):
I'm not a C2 expert, but from some offline discussions it seems like this fix might help with some of the issues we have seen in [1]. So I'm happy to have this integrated, so that we can test it more (which would also help this same fix being integrated into mainline).

[1] - https://mail.openjdk.java.net/pipermail/panama-dev/2021-June/014150.html

PR #566, opened 2021-07-07
Remove mem barriers to allow loop optimizations for vectors accessing off-heap

rsmogura:
This patch is to remove memory barriers, to allow better optimization of loops.

With mem barriers loops can't be optimized too much as code in compiler can not traverse loop back control through mem barriers to determine exit condition.

**Looking for feedback about possible impact.**

Partially discussed in, and extracted from https://mail.openjdk.java.net/pipermail/panama-dev/2021-June/014204.html

After
```
Benchmark                                       (size)  Mode  Cnt      Score      Error  Units
VectorCopySegments.copyWithNative                 1024  avgt   10     20.293 ?    0.436  ns/op
VectorCopySegments.copyWithNative              1048576  avgt   10  22270.840 ?  579.533  ns/op
VectorCopySegments.copyWithNativeShared           1024  avgt   10     15.854 ?    0.061  ns/op
VectorCopySegments.copyWithNativeShared        1048576  avgt   10  21948.236 ?   43.981  ns/op
VectorCopySegments.copyWithNativeToArray          1024  avgt   10     20.318 ?    0.347  ns/op
VectorCopySegments.copyWithNativeToArray       1048576  avgt   10  22142.499 ?  305.501  ns/op
VectorCopySegments.copyWithVector                 1024  avgt   10     31.240 ?    0.333  ns/op
VectorCopySegments.copyWithVector              1048576  avgt   10  25320.898 ?  118.397  ns/op
VectorCopySegments.copyWithVectorDirectBuffer     1024  avgt   10     21.605 ?    0.210  ns/op
VectorCopySegments.copyWithVectorDirectBuffer  1048576  avgt   10  23613.272 ? 1030.153  ns/op
VectorCopySegments.copyWithVectorShared           1024  avgt   10     19.897 ?    0.485  ns/op
VectorCopySegments.copyWithVectorShared        1048576  avgt   10  24719.767 ?  453.725  ns/op
VectorCopySegments.copyWithVectorShuffle          1024  avgt   10     36.364 ?    0.669  ns/op
VectorCopySegments.copyWithVectorShuffle       1048576  avgt   10  29730.528 ?  339.100  ns/op
VectorCopySegments.copyWithVectorToArray          1024  avgt   10     29.282 ?    0.338  ns/op
VectorCopySegments.copyWithVectorToArray       1048576  avgt   10  28502.004 ?  593.347  ns/op
VectorCopySegments.copyWithVectorUnroller         1024  avgt   10     36.368 ?    0.092  ns/op
VectorCopySegments.copyWithVectorUnroller      1048576  avgt   10  21528.433 ?  303.141  ns/op
```

Before
```
Benchmark                                       (size)  Mode  Cnt      Score      Error  Units
VectorCopySegments.copyWithVector                 1024  avgt   10     30.841 ?    0.119  ns/op
VectorCopySegments.copyWithVector              1048576  avgt   10  44834.639 ?  746.965  ns/op
VectorCopySegments.copyWithVectorShuffle          1024  avgt   10     47.797 ?    0.672  ns/op
VectorCopySegments.copyWithVectorShuffle       1048576  avgt   10  61171.416 ? 5656.479  ns/op
VectorCopySegments.copyWithVectorUnroller         1024  avgt   10     49.522 ?    1.027  ns/op
VectorCopySegments.copyWithVectorUnroller      1048576  avgt   10  72145.653 ?  987.111  ns/op
```

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/566/head:pull/566` \
`$ git checkout pull/566`

`$ git checkout pull/566` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/566/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 566`

`$ git pr show -t 566`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/566.diff">https://git.openjdk.java.net/panama-foreign/pull/566.diff</a>

</details>

### DIFF
--- a/make/data/symbols/java.base-H.sym.txt
+++ b/make/data/symbols/java.base-H.sym.txt
@@ -209,7 +209,8 @@ header extends java/lang/RuntimeException flags 21
 class name java/lang/runtime/SwitchBootstraps
 header extends java/lang/Object flags 21 classAnnotations @Ljdk/internal/javac/PreviewFeature;(feature=eLjdk/internal/javac/PreviewFeature$Feature;SWITCH_PATTERN_MATCHING;)
 innerclass innerClass java/lang/invoke/MethodHandles$Lookup outerClass java/lang/invoke/MethodHandles innerClassName Lookup flags 19
-method name typeSwitch descriptor (Ljava/lang/invoke/MethodHandles$Lookup;Ljava/lang/String;Ljava/lang/invoke/MethodType;[Ljava/lang/Object;)Ljava/lang/invoke/CallSite; thrownTypes java/lang/Throwable flags 89
+method name typeSwitch descriptor (Ljava/lang/invoke/MethodHandles$Lookup;Ljava/lang/String;Ljava/lang/invoke/MethodType;[Ljava/lang/Object;)Ljava/lang/invoke/CallSite; flags 89
+method name enumSwitch descriptor (Ljava/lang/invoke/MethodHandles$Lookup;Ljava/lang/String;Ljava/lang/invoke/MethodType;[Ljava/lang/Object;)Ljava/lang/invoke/CallSite; flags 89
 
 class name java/net/DatagramSocket
 -method name setDatagramSocketImplFactory descriptor (Ljava/net/DatagramSocketImplFactory;)V
@@ -454,8 +455,7 @@ innerclass innerClass java/io/ObjectInputStream$GetField outerClass java/io/Obje
 innerclass innerClass java/io/ObjectOutputStream$PutField outerClass java/io/ObjectOutputStream innerClassName PutField flags 409
 
 class name java/util/SplittableRandom
-header extends jdk/internal/util/random/RandomSupport$AbstractSplittableGenerator flags 31 runtimeAnnotations @Ljdk/internal/util/random/RandomSupport$RandomGeneratorProperties;(name="SplittableRandom",i=I64,j=I0,k=I0,equidistribution=I1)
-innerclass innerClass jdk/internal/util/random/RandomSupport$AbstractSplittableGenerator outerClass jdk/internal/util/random/RandomSupport innerClassName AbstractSplittableGenerator flags 409
+header extends java/lang/Object implements java/util/random/RandomGenerator,java/util/random/RandomGenerator$SplittableGenerator flags 31 runtimeAnnotations @Ljdk/internal/util/random/RandomSupport$RandomGeneratorProperties;(name="SplittableRandom",i=I64,j=I0,k=I0,equidistribution=I1)
 innerclass innerClass java/util/random/RandomGenerator$SplittableGenerator outerClass java/util/random/RandomGenerator innerClassName SplittableGenerator flags 609
 -method name nextInt descriptor (I)I
 -method name nextInt descriptor (II)I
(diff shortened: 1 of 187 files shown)


PaulSandoz:
@iwanowww will know more. Ideally this should be targeted at `panama-vector/vectorIntrinsics` and/or `jdk/jdk`, then it should make its way automatically to the `panama-foreign` repos.

iwanowww:
It's incorrect to completely omit the barriers. They are still needed in some cases. Here's relevant logic [1] for unsafe scalar accesses:
```
bool C2Access::needs_cpu_membar() const {
  ...
  if (anonymous) {
    // We will need memory barriers unless we can determine a unique
    // alias category for this reference.  (Note:  If for some reason
    // the barriers get omitted and the unsafe reference begins to "pollute"
    // the alias analysis of the rest of the graph, either Compile::can_alias
    // or Compile::must_alias will throw a diagnostic assert.)
    if (is_mixed || !is_unordered || (mismatched && !_addr.type()->isa_aryptr())) {
      return true;
    }
```

Similar checks are needed for vector accesses. In particular, mixed accesses and field mismatched accesses have to be surrounded by barriers.   

[1]   https://github.com/openjdk/panama-foreign/blob/foreign-memaccess%2Babi/src/hotspot/share/gc/shared/c2/barrierSetC2.cpp#L65

rsmogura:
I knew it would be too simple. Just for curiosity is this same case like with VarHandle.get / getOpaque?

PaulSandoz:
VarHandle access will use unsafe, and go through the more sophisticated analysis (Vladimir linked to) to determine whether barriers are required.

You can follow the unsafe logic from here https://github.com/openjdk/jdk/blob/master/src/hotspot/share/opto/library_call.cpp#L2212

We need to do something similar for vector load/store (just for plain accesses for now).

iwanowww:
> Just for curiosity is this same case like with VarHandle.get / getOpaque?

No, it is different.

`VH.get()` vs `VH.getOpaque()` have slightly different semantics which is currently implemented using memory barriers (but it doesn't have to). (It is`!is_unordered` case in `C2Access::needs_cpu_membar()` which covers non-plain accesses.)  

But even plain accesses (but only unsafe ones) sometimes need to be surrounded by memory barriers to preserve correctness when it is not possible to determine a unique alias category which represents the memory the access operates on. That's the case described in the comment:
```
    // We will need memory barriers unless we can determine a unique
    // alias category for this reference.  (Note:  If for some reason
    // the barriers get omitted and the unsafe reference begins to "pollute"
    // the alias analysis of the rest of the graph, either Compile::can_alias
    // or Compile::must_alias will throw a diagnostic assert.)
```

rsmogura:
Thanks I've got the point. I've been focused on off-heap. It was not hard to create a test case
```
    byte arr[] = new byte[32];
    final var bb = ByteBuffer.wrap(arr);
    final var vs = VectorSpecies.ofLargestShape(byte.class);
    final var ones = ByteVector.broadcast(vs, 1);
    var res = ByteVector.zero(vs);

    long result = 0;
    result += arr[2];
    res.add(ones).intoByteBuffer(bb, 0, ByteOrder.nativeOrder());
    result += arr[2];

    return result;
```

PR #568, opened 2021-07-13
8270376: Finalize API for memory copy

mcimadamore:
This patch ties up some loose ends with the MemoryCopy API, and it also prepares the ground for some other related refactorings in this area.

The meat of this patch is represented by the various changes in MemoryCopy, where all methods were renamed to simply `copy`, and the length parameter (now called `elementCount`) is always moved to the end.

I've also simplified naming of parameters, as I think distinguishing between `index` and `offset` is enough (e.g. an array has an index, a segment has an offset).

You will see that, at the very end of the class, three more copy methods have been added, which deal fully in terms of segments. I have also moved the complex layout-based memory segment copy operation (which does the swap) as a static method in this class, as I believe the static form makes the method more regular and usable.

I've made some changes to our uses of `copyFrom` in the linker, to use the new methods in `MemoryCopy` when the copy operation was performing slicing in the source/target segment, which I think makes the code more readable. Of course these changes are completely optional and could be omitted as well.

---------
### Progress

### Issue
 * [JDK-8270376](https://bugs.openjdk.java.net/browse/JDK-8270376): Finalize API for memory copy

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/568/head:pull/568` \
`$ git checkout pull/568`

`$ git checkout pull/568` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/568/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 568`

`$ git pr show -t 568`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/568.diff">https://git.openjdk.java.net/panama-foreign/pull/568.diff</a>

</details>


mcimadamore:
You might ask if it still makes sense to keep `MemorySegment::copyFrom` (at least I have asked myself that :-) ). I think currently our API is not very consistent in terms of static vs. non-static boundaries. For instance, all the API to do memory access is defined in terms of static methods in the `MemoryAccess` class. For bulk copy, things are more convoluted, as there is now a static class `MemoryCopy`, but also instance methods in `MemorySegment`, such as `copyFrom`, but also `toXYZArray` (which is the dual case).

To add to the confusion, the SegmentAllocator interface also defines its own suite of allocator methods that can be used to set either a single value, or an array of values - this idiom is used to facilitate native interop code, so that one can do:

```
MemorySegment segment = allocator.allocateArray(C_FLOAT, new float[] { ... });
```
or:

```
MemorySegment segment = allocator.allocateArray(C_FLOAT, 4.0f);
```
And do memory allocation *and* initialization in a single shot.

Now, I believe that it is not realistic to think that we can put *all* of the `MemoryAccess` and `MemoryCopy` API as instance methods inside `MemorySegment`. While this works in `ByteBuffer` (although there are still quite a bit of methods), it fails to scale here, as:

* every operation needs an endianness-ful overload (meaning method size goes 2x)
* here we need (at least) a pair of bulk ops per carrier, while `ByteBuffer` only provides those for `byte[]`.

For these reasons, I still think that offloading dereference and copy methods on side classes is a good move, so that the `MemorySegment` API can remain relatively simple.

So, what should we do with instance methods like `copyFrom` and `toArray`? I think these methods do have a role in the API - which is that of facilitating transfer of data from and to the Java heap. For instance:

```
MemorySegment segment = MemorySegment.allocateNative(16)
                                                                          .copyFrom(new float [] { ... }); // from Java to native
...
float[] = segment.toFloatArray(); // from native to Java
```

The above feels like a common enough pattern when marshalling data structures in and out of the Java heap that, perhaps it still justifies the cost of inserting additional instance methods in the API. Another option would be to just provide more `allocateNative` factories:

```
MemorySegment segment = MemorySegment.allocateNative(16, new float [] { ... }); // from Java to native
...
float[] = segment.toFloatArray(); // from native to Java
```
While this is nice, if we go down this path we'd have to replicate everything inside `SegmentAllocator` (otherwise clients using `SegmentAllocator` would be worse off).

Overall, I believe the best course of action would be to:

1. introduce many `MemorySegment::copyFrom` instance, in a 1-1 with `MemorySegment::toXYZArray`
2. simplify the `SegmentAllocator` interface to just be about segment allocation (and leave initialization to other APIs)
3. for complex copy cases which do not fit in the simple `copyFrom` format, users will fallback to `MemoryCopy`

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/SharedUtils.java
```diff
@@ -289,8 +290,7 @@ public static CLinker getSystemLinker() {
     public static String toJavaStringInternal(MemorySegment segment, long start) {
         int len = strlen(segment, start);
         byte[] bytes = new byte[len];
-        MemorySegment.ofArray(bytes)
-                .copyFrom(segment.asSlice(start, len));
+        MemoryCopy.copy(segment, start, MemorySegment.ofArray(bytes), 0, len);
```
uschindler, line 293: No need to wrap the byte array. Just use the other copy method taking byte[]. This spares another object allocation.

uschindler (changes requested):
Looks good. I have a minor change, because the string converter wraps the byte array instead of using the better fitting method signature taking byte array directly.

rose00:
I agree with these simplifications.  The previous iteration on naming was useful for hammering out the API, but it won't wear well:  It's too much like keeping the training wheels on the bike forever.

Also, yes, `MemoryCopy` should be right-sized to be about copies of various sorts, not just those that involve arrays only.  Perhaps there will be byte buffer copy routines added there in the future?   Perhaps, even, stream-like operations or scatter and gathers?  We don't have to decide this today, but having a suite of static methods for easy copying makes a lot of sense.

There 's a garble in the javadoc: s/see read-only/read-only segment/

It might be worth pointing out something like this:

> Note that the first five arguments of the copy methods roughly follow the convention for System.arraycopy.  The optional trailing sixth argument controls byte order, which by default is the current native byte order.

Reasons:  Everybody that approaches an API like this needs to classify it mentally:  dest or source first?  indexes or lengths?  (etc.)  Saying that it’s like arraycopy removes angst immediately.  Also, it's good to remind people of byte order; bad stuff happens eventually when we close our eyes to byte order effects.

I agree that there should be a fluent (non-static) API for the very most common cases of copying.  But I disagree slightly with where the line is drawn.  The `copyFrom` operation should take a long offset before the source operand, to place the data at a programmable position in the destination.  Also (as a second suggestion) the `copyFrom` routine should return the updated (incremented) position.  I say this because I believe that a very common pattern of copies is to gather many smaller segments into a larger buffer.  Such a loop looks like this, with and without my suggestions:

```
MemorySegment whole = …;
long pos = 0;
for (MemorySegment part : parts…) {
   pos = whole.copyFrom(pos, part);
}
for (MemorySegment part : parts…) {
   whole.copyFrom(pos, part);
   pos += part.size();  //redundant size query, bug bait
}
for (MemorySegment part : parts…) {
   var partDest = whole.asSlice(pos); //temp obj!!
   partDest.copyFrom(part);
   pos += part.size();  //redundant size query, bug bait
}
```

Regarding arrays:  I suggest replacing the byte order argument with a `ValueLayout` argument.  This makes it clearer (IMO) what is the meaning of the argument:  It's clear that it applies to the value as stored inside the MS, and not the array.  Of course, with a moment's thought, it's also obvious that's true about the ByteOrder, because the array should not ever hold anything other than native order items.  (You could also address this by changing order to dstOrder and srcOrder.)  Here's another reason to use a VL instead of a BO:  Because it's more information, you can add more extensions later on.  One that springs to mind:  a boolean value layout could imply that values copied into a Java boolean array could normalize to 0/1 values.  Another:  Vectors might use a mixed byte order convention, where the lanes are swapped one way and the lane order is swapped another way; this can be encoded in a (future) value layout but not in a BO.

Surrounding this smaller point about VL vs. BO is a larger one:  Any API that has bespoke points for each Java primitive array type is going to look out of date in Valhalla.  The way to prepare for that, IMO, is to create generic API points which use dynamic typing (such as reflection), and then define the bespoke API points as handy sugar.  That way, when we complex arrays or vector arrays, they will just play normally with the copy API.

Specific suggestions about unified array copies:

```
/// MemoryCopy.java

// these are the primitives:
void arraycopy(Object srcArray, int srcIndex, MemorySegment dstSegment, long dstOffset, int elementCount, ValueuLayout element);
void arraycopy(MemorySegment srcSegment, long srcOffset, Object dstArray, int dstIndex, int elementCount, ValueuLayout element);

// these are sugar:
void copy(char[] srcArray, int srcIndex, MemorySegment dstSegment, long dstOffset, int elementCount) {
    arraycopy(srcArray, srcIndex, dstSegment, dstOffset, elementCount, ValueLayouts.CHAR);
}
void copy(MemorySegment srcSegment, long srcOffset, char[] dstArray, int dstIndex, int elementCount) {
    arraycopy(srcSegment, srcOffset, dstArray, dstIndex, elementCount, ValueLayouts.CHAR);
}
//Omit these next two?
void copy(MemorySegment srcSegment, long srcOffset, char[] dstArray, int dstIndex, int elementCount, ByteOrder srcOrder) {
    arraycopy(srcSegment, srcOffset, dstArray, dstIndex, elementCount, ValueLayouts.CHAR.withOrder(dstOrder));
}
void copy(char[] srcArray, int srcIndex, MemorySegment dstSegment, long dstOffset, int elementCount, ByteOrder dstOrder) {
    arraycopy(srcArray, srcIndex, dstSegment, dstOffset, elementCount, ValueLayouts.CHAR.withOrder(dstOrder));
}
```

And the handling of arrays in MC interacts with the handling of arrays in MemorySegment:
```
/// MemorySegment.java

// primitive
<AT> AT toArray(Class<AT> atype, ValueLayout srcLayout);
MemorySegment ofArray(Object arr, ValueLayout destLayout) {
    MemorySegment res = …;
    MemoryCopy.arraycopy(arr, 0, res, 0, Array.getLength(arr), destLayout);
    return res;
}

// sugar
default char[] toCharArray() {
    return toArray(char[].class, ValueLayouts.CHAR);
}
static MemorySegment ofArray(char[] arr) {
    return ofArray(arr, ValueLayouts.CHAR);
}
```

Basically I'm suggesting that the API would be more coherent if you more aggressively factored all the P-array methods (where P is a primitive type) through single generic methods, akin to `System.arraycopy`.  That will make it less necessary to have multiple overloads per P-array type, since there's a way to use the generic API point to get more control.

Also, when Valhalla comes, the set of P in P-arrays will become unbounded.  You'll have flat arrays of complex float, vector128, and who knows what else.  It will be necessary at least then to rethink all of those bespoke API points for P-arrays (P in char, short, int, long etc.).


P.S. (Post-script or parting shot) Although I like and accept `asSlice` just as I like and accept `subLIst`, for reducing complexity of the API, I don't like the name `asSlice`.  We have `asList` in various places, but we don't have `asSubList`, because when you "as" something you keep its contents but change the view.  But a slice is *not* the same contents as the original.  Please consider (as an independent change) `slice` instead of `asSlice`.  Everybody knows, or can learn, that a slice is a view, so there's no need to have the extra prompt of the "as" word.  Note that everybody learned that a `subList` is a view, not a copy of a series of elements.  Same story here.

P.P.S. Unrelated suggestion about `MS::spliterator`:  It should not throw if `size()` is not a multiple of the layout size; rather, it should tolerate this and simply create the first segment as a multiple of the size.  That makes the situation no worse than before, since there is still only one segment with an “odd tail”.  I’m bringing this up because that’s *exactly* how vectorized splits should work!  I.e., if the spliterator is properly tolerant, then it can be asked to split on behalf of a vectorized loop kernel, which works best when fed an exact multiple of vector lengths, but can tolerate an “odd tail” if it must.  The spliterator should return null if the original pre-split size is less than 2*N.  That’s the only case where the client code has to look around for the “odd tail”.  Naturally, an unrolled vectorized loop is best, and the best `Layout` to hand to `MS::spliterator` in that case is the size of the vector *times the unroll count*.  Again, there is no cause to demand an exact multiple of that size; the spliterator shouudl be tolerant.

mcimadamore:
Thanks for the comments John - I agree pretty much with all the points you raise, so I think it's worth spending some more cycles to try and get this right. I do think that there is an useful distinction between primitive copy methods, and sugary ones, and we should try to make that distinction clearer in the API.

One thing I wasn't too sure about was if having a single entry point taking Object would create havoc - as the code will have to access array length reflectively, and all that - and I can guess it can get bad under profile pollution? This was the main reason as to why, few years ago, we went with the `MemorySegment::ofArray` API with all the overloads (instead of having only one taking an object) - because we had inlining issues with the monomorphic version. Maybe it's time to try again - or are there "tricks" that I overlooked?

I'm assuming the code would need to access at least:

* array length (can be done with Reflection.getLength which is an intrinsic)
* unsafe base/stride (can be done with Unsafe::arrayBaseOffset/arrayIndexScale)

In order to perform the bulk copy call. Do you foresee performance issues if we go down that path? If not, does that mean that we should retest our assumption when it comes to MemorySegment::ofArray as well?

mcimadamore:
@rose00 I believe I've addressed your comments. Now, when looking at the code it's clear, I think, what the primitives are. My only gripe is that this is not reflected in the API - e.g. all copy methods look the same-y when looking at the javadoc, and it's not clear to see which method is based on which. Of course we could clarify things (e.g. say that "this method is equivalent to..." in the javadoc) - but I'm wondering if the real copy primitives should be statics in `MemorySegment` - and leave `MemoryCopy` for strongly-typed sugary copies between segment and arrays.

More specifically, the methods that I don't believe belong in MemoryCopy are:

* `copy(MemorySegment, ValueLayout, long, MemorySegment, ValueLayout, long, long)`
* `copy(MemorySegment, ValueLayout, long, Object, int, int)`
* `copy(Object, int, MemorySegment, ValueLayout, long, long)`

But I could also claim that the following also kind of belong somewhere else:

* `copy(MemorySegment, MemorySegment, long)`
* `copy(MemorySegment, long, MemorySegment, long, long)`

As these are just telescopic shortcuts for the general segment copy primitive above.

uschindler:
Hi,
I made a quick test with Lucene, see branch: https://github.com/apache/lucene/compare/main...uschindler:draft/jdk-foreign-mmap-8270376, example implementation for reading bytes:  https://github.com/uschindler/lucene/blob/b4e76e6963c94f65a3c4348d485e66d2c5802b25/lucene/core/src/java/org/apache/lucene/store/MemorySegmentIndexInput.java#L149-L160

Lucene tests pass and performance seems OK. I did not test the `Object` syscopy-like aproach.

mcimadamore:
> Hi,
> I made a quick test with Lucene, see branch: [apache/lucene@main...uschindler:draft/jdk-foreign-mmap-8270376](https://github.com/apache/lucene/compare/main...uschindler:draft/jdk-foreign-mmap-8270376), example implementation for reading bytes: https://github.com/uschindler/lucene/blob/b4e76e6963c94f65a3c4348d485e66d2c5802b25/lucene/core/src/java/org/apache/lucene/store/MemorySegmentIndexInput.java#L149-L160
> 
> Lucene tests pass and performance seems OK. I did not test the `Object` syscopy-like aproach.

Thanks for verifying that performance did not regress with the new implementation scheme!

mcimadamore:
> @rose00 I believe I've addressed your comments. Now, when looking at the code it's clear, I think, what the primitives are. My only gripe is that this is not reflected in the API - e.g. all copy methods look the same-y when looking at the javadoc, and it's not clear to see which method is based on which. Of course we could clarify things (e.g. say that "this method is equivalent to..." in the javadoc) - but I'm wondering if the real copy primitives should be statics in `MemorySegment` - and leave `MemoryCopy` for strongly-typed sugary copies between segment and arrays.
> 
> More specifically, the methods that I don't believe belong in MemoryCopy are:
> 
>     * `copy(MemorySegment, ValueLayout, long, MemorySegment, ValueLayout, long, long)`
> 
>     * `copy(MemorySegment, ValueLayout, long, Object, int, int)`
> 
>     * `copy(Object, int, MemorySegment, ValueLayout, long, long)`
> 
> 
> But I could also claim that the following also kind of belong somewhere else:
> 
>     * `copy(MemorySegment, MemorySegment, long)`
> 
>     * `copy(MemorySegment, long, MemorySegment, long, long)`
> 
> 
> As these are just telescopic shortcuts for the general segment copy primitive above. Any thoughts?

Thinking more about this, while the new approach is, I think, beneficial in terms of making all the implementation flow into a couple of array copying routine, I don't see, as of yet, the necessity of exposing such primitives in the API. As noted, earlier, it is likely that, w/o extra work in the VM, calling these methods will be less efficient - so users will effectively have two ways for doing the same thing, one which works well, and one which doesn't.

I agree with the general idea that we should have a clear distinction between primitives and derived methods, but I think exposing an "half primitive" like we're doing now (after all the real primitive is the implementation method taking explicit byte size, stride and layout) can potentially backfire.

So, I'd like to make a suggestion to:

*  "hide" the primitive API methods - e.g. `copy(MemorySegment, ValueLayout, long, Object, int, int)` and `copy(Object, int, MemorySegment, ValueLayout, long, long)`
* move these methods under `MemorySegment`:
   - `copy(MemorySegment, ValueLayout, long, MemorySegment, ValueLayout, long, long)`
    - `copy(MemorySegment, MemorySegment, long)`
    - `copy(MemorySegment, long, MemorySegment, long, long)`

mcimadamore:
I have finished reorganizing the code (as per previous proposal). I think the new organization makes a lot of sense: MemorySegment has static copy primitives which are general, and are front and center. `MemoryCopy` is conceptually mostly convenience (e.g. avoid slicing and creating views) - although internally we do things differently to avoid performance issues - but we hope that someday we'll be able to reconcile the two paths. In this sense `MemoryCopy` and `MemoryAccess` now look similar - in that it's clear they mostly provide sugar.

I'll deal with the `copyFrom` instance methods in a different PR, but this one sets the foundation: `copyFrom` are just more fluent, sugary, methods which are implemented using the static primitives. This is now made clear both by the API (copyFrom is a `default` method) and by the javadoc, which clearly says that a call to copyFrom can be rewritten as a call to the static copy method.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/SharedUtils.java
uschindler: Thanks for the update! 👍

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -859,4 +826,150 @@ static MemorySegment globalNativeSegment() {
...
+        if (srcOffset % srcElementLayout.byteAlignment() != 0) {
+            throw new IllegalArgumentException("Source segment incompatible with alignment constraints");
+        }
+        if (dstOffset % dstElementLayout.byteAlignment() != 0) {
+            throw new IllegalArgumentException("Target segment incompatible with alignment constraints");
+        }
+        long size = elementCount * srcElementLayout.byteSize();
+        if (size % srcElementLayout.byteSize() != 0) {
+            throw new IllegalArgumentException("Segment size is not a multiple of layout size");
+        }
```
JornVernee, line 962: Given the calculation above, I don't think this can ever happen? i.e. `srcElementLayout.byteSize()` is a factor of `size`.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -859,4 +826,150 @@ static MemorySegment globalNativeSegment() {
...
+     * @throws IllegalArgumentException if the element layouts have different sizes, if the source offset is incompatible
+     * with the alignment constraints in the source element layout, or if the destination offset is incompatible with the
+     * alignment constraints in the destination element layout.
+     * @throws IllegalStateException if either the scope associated with the source segment or the scope associated
+     * with the destination segment have been already closed, or if access occurs from a thread other than the thread
+     * owning either scopes.
+     * @throws IndexOutOfBoundsException if {@code bytes > srcSegment.byteSize()} or if
+     * {@code bytes > dstSegment.byteSize()}.
+     * @throws UnsupportedOperationException if the destination segment is read-only (see {@link #isReadOnly()}).
+     */
```
JornVernee, line 860: This doesn't not specify what happens if `bytes` is negative.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -859,4 +826,150 @@ static MemorySegment globalNativeSegment() {
...
+     * @throws IllegalArgumentException if the element layouts have different sizes, if the source offset is incompatible
+     * with the alignment constraints in the source element layout, or if the destination offset is incompatible with the
+     * alignment constraints in the destination element layout.
+     * @throws IllegalStateException if either the scope associated with the source segment or the scope associated
+     * with the destination segment have been already closed, or if access occurs from a thread other than the thread
+     * owning either scopes.
+     * @throws IndexOutOfBoundsException if {@code srcOffset + bytes > srcSegment.byteSize()} or if
+     * {@code dstOffset + bytes > dstSegment.byteSize()}.
+     * @throws UnsupportedOperationException if the destination segment is read-only (see {@link #isReadOnly()}).
+     */
```
JornVernee, line 898: Again, this does not specify what happens when `srcOffset`, `dstOffset`, and/or `bytes` is negative.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -859,4 +826,150 @@ static MemorySegment globalNativeSegment() {
...
+     * with the alignment constraints in the source element layout, or if the destination offset is incompatible with the
+     * alignment constraints in the destination element layout.
+     * @throws IllegalStateException if either the scope associated with the source segment or the scope associated
+     * with the destination segment have been already closed, or if access occurs from a thread other than the thread
+     * owning either scopes.
+     * @throws IndexOutOfBoundsException if {@code srcOffset + (elementCount * S) > srcSegment.byteSize()} or if
+     * {@code dstOffset + (elementCount * S) > dstSegment.byteSize()}, where {@code S} is the byte size
+     * of the element layouts.
+     * @throws UnsupportedOperationException if the destination segment is read-only (see {@link #isReadOnly()}).
+     */
```
JornVernee, line 940: Finally here, this doesn't specify what happens when `srcOffset`, `dstOffset`, and/or `elementCount` is negative.

test/jdk/java/foreign/TestCopyFrom.java
```diff
@@ -55,7 +56,7 @@ public void testByteCopy(SegmentSlice s1, SegmentSlice s2) {
             Type.BYTE.set(s1, i, i);
         }
         //perform copy
-        s2.segment.copyFrom(s1.segment.asSlice(0, size));
+        MemorySegment.copy(s1.segment, 0, s2.segment, 0, size);
```
JornVernee, line 59: Looks like this could use the simpler overload:

```suggestion
        MemorySegment.copy(s1.segment, s2.segment, size);
```

test/jdk/java/foreign/TestNulls.java
```diff
@@ -106,7 +106,11 @@
...
             "jdk.incubator.foreign.GroupLayout/withAttribute(java.lang.String,java.lang.constant.Constable)/1/0",
             "jdk.incubator.foreign.MemoryHandles/insertCoordinates(java.lang.invoke.VarHandle,int,java.lang.Object[])/2/1",
-            "jdk.incubator.foreign.FunctionDescriptor/withAttribute(java.lang.String,java.lang.constant.Constable)/1/0"
+            "jdk.incubator.foreign.FunctionDescriptor/withAttribute(java.lang.String,java.lang.constant.Constable)/1/0",
+            "jdk.incubator.foreign.MemoryCopy/copy(jdk.incubator.foreign.MemorySegment,jdk.incubator.foreign.ValueLayout,long,java.lang.Object,int,int)/0/0",
+            "jdk.incubator.foreign.MemoryCopy/copy(jdk.incubator.foreign.MemorySegment,jdk.incubator.foreign.ValueLayout,long,java.lang.Object,int,int)/1/0",
+            "jdk.incubator.foreign.MemoryCopy/copy(java.lang.Object,int,jdk.incubator.foreign.MemorySegment,jdk.incubator.foreign.ValueLayout,long,int)/2/0",
+            "jdk.incubator.foreign.MemoryCopy/copy(java.lang.Object,int,jdk.incubator.foreign.MemorySegment,jdk.incubator.foreign.ValueLayout,long,int)/3/0"
```
JornVernee, line 113: These methods are now private, so I don't think these excludes are still needed here?

mcimadamore:
I ended up with a pretty drastic rewrite, as I was not too happy with how the code was laid out. After trying out a number of different solution, I realized that my main beefs were:

1. Operations pertaining to segments were spread into multiple class (MemoryCopy and MemoryAccess)
2. The number of operations offered by MemoryAccess seemed too big (this has been mentioned before - e.g. by index vs. by offset)
3. The role of copyFrom in the new world was not very clear

For these reasons I decided to:

1. only keep offset-based accessors in MemoryAccess - this reduces greatly the amount of methods
2. thanks to (1) we can now move MemoryCopy methods in MemoryAccess as well - the idea is that al the methods in MemoryAccess are about reading/writing Java values from/to memory segments (either a single value or in bulk), so I think the move makes a lot of sense
3. I renamed all accessors to use readByte/writeByte (for single access) and readBytes/writeBytes (for bulk access). This helps seeing the various methods as different ways to read/write a segment (the name read/write, especially in copy operation always refer to a segment - this is similar to what the ByteBuffer API already does).
4. I've turned the static copy method in MemorySegment back to instance methods - e.g. back to `copyFrom`. These are the "real" copy primitives, and everything else can be derived from there (of course the static accessors don't really do that, to avoid creation of garbage objects, but that's the spirit).

I think this makes the API more compact, more usable and easier to explain.

In a followup PR, I also plan to drop most of the allocation methods from SegmentAllocator. I think it would be better to keep SegmentAllocator a neutral abstraction, while at this point in time it still has stuff from NativeScope, in an attempt to ease the native interop use cases. I think jextract is in a better position to help there, and we will do something to counteract this loss.

A javadoc of the new MemoryAccess class is available here:
http://cr.openjdk.java.net/~mcimadamore/panama/memorycopy_finalize_javadoc/javadoc/jdk/incubator/foreign/MemoryAccess.html

PaulSandoz:
I was not following the whole discussion too closely, so I hope I am not going over old ground, but this is my impression of the latest patch:

- I can understand the appeal of consistently using the read/write prefixes, but I think it may be confusing. In our Java APIs these names are commonly used for serialization and input/output streams etc. For memory access it would be more consistent to use set/get for individual access and copy for bulk access, rather than biasing towards whether the memory segment is the src or dst. The location in `MemoryAccess` is fine. Overloading with `copy` is probably fine, rather than identifying in the name whether the array is src or dst.

- For the copy methods I would prefer consistency with `System.arraycopy`, where the src occurs first regardless of the type (rather than segment src or dst, and the src flipping argument position). It’s really easy to make stupid mistakes confusing src/dst and I think this could increase such mistakes.

- Expanding on the latter point arguably suggests that `MemorySegment.copyFrom` should be changed to `MemorySegment.copyInto` (or just `copy`), so the receiver is the src, and it returns the destination. Then there is the same consistency of arguments as in the other copy methods. However, that would bury the dst segment allocation as an argument rather than as the receiver, but at least i would be able to reason consistently about the src/dst offsets/lengths argument positions.

mcimadamore:
following @PaulSandoz  comments, I've reverted the names inside MemoryAccess to get/set and copy. Copy methods are now consistent with System::arrayCopy.

I have not touched copyFrom - as I think that one deserves some more discussion: I think under the objection there is a feeling that the instance method is worse, readability-wise, than a static method. If that's the case, we should just bite the bullet, drop copyFrom, and use a static MemorySegment::copy.

I'm also looking at MemorySegment::toArray as something that looks superseded now that we have better array copy facilities.

PaulSandoz:
Re: `MemorySegment.toXArray`, yes good point. I think we could remove these. They are a bit like the `Arrays.copyOf/Range` methods. Not suggesting we such methods.

Re: `MemorySegment.copyFrom`, my inclination is to keep the one copy arg method. and move all others to static methods on `MemorySegment` with signatures consistent with `System.arraycopy` (lets keep methods on `MemoryAccess` for accesses with primitives).
If we keep the one arg method i would rename it `copy` and flip the dst/src, and returning dst.

Now you got me looking at `MemorySegment.fill`! Arguably that is a bulk copy broadcasting a scalar, like `Arrays.fill`.

mcimadamore:
> Now you got me looking at `MemorySegment.fill`! Arguably that is a bulk copy broadcasting a scalar, like `Arrays.fill`.

Yes, it's a memset.

mcimadamore:
>  If we keep the one arg method i would rename it `copy` and flip the dst/src, and returning dst.

Honestly, `copy` alone sounds too ambiguous when used in instance mode.

mcimadamore:
> > Now you got me looking at `MemorySegment.fill`! Arguably that is a bulk copy broadcasting a scalar, like `Arrays.fill`.
> 
> Yes, it's a memset.

out of curiosity, I did some experiment:

BulkOps.segment_fill         avgt   30  60076.605 ? 702.267  ns/op
BulkOps.segment_fill_manual  avgt   30  44685.159 ? 537.964  ns/op

It seems like a manual loop with segments is faster than doing a memset (probably because of unrolling and vectorization). If that's the case, I think there's a strong case to move fill away from `MemorySegment` and into the `MemoryAccess` class.

mcimadamore:
> Re: `MemorySegment.toXArray`, yes good point. I think we could remove these. They are a bit like the `Arrays.copyOf/Range` methods. Not suggesting we such methods.

I'm less sure about this now - on the one hand these are clearly redundant. On the other hand, there is a symmetry between MemorySegment::ofArray(XYZ[]) and MemorySegment::toXYZArray().

Another option would be to just expose a `toByteArray`, e.g. just expose a basic byte view of the segment (after all a segment is just a bag of bytes), which would be consistent with the decision of only exposing `asByteBuffer` (although it's easier to go from, say, a ByteBuffer to an IntBuffer than it is to go from a byte array to an int array).

PaulSandoz:
> > Re: `MemorySegment.toXArray`, yes good point. I think we could remove these. They are a bit like the `Arrays.copyOf/Range` methods. Not suggesting we such methods.
> 
> I'm less sure about this now - on the one hand these are clearly redundant. On the other hand, there is a symmetry between MemorySegment::ofArray(XYZ[]) and MemorySegment::toXYZArray().
>

Although a key difference is `ofArray` is a factory method creating a view over the memory region covered by the primitive array, not a copy, so i would argue its not quite symmetric. Might as well do `intArray..clone()` rather than `MemorySegment.ofArray(intArray).toIntArray()`

> Another option would be to just expose a `toByteArray`, e.g. just expose a basic byte view of the segment (after all a segment is just a bag of bytes), which would be consistent with the decision of only exposing `asByteBuffer` (although it's easier to go from, say, a ByteBuffer to an IntBuffer than it is to go from a byte array to an int array).

Yeah, i think that would simplify, giving us choices to expand copying utilities in `MemoryAccess` if need be.

uschindler:
Hi,
I tested the latest variant with Lucene. Performance seems OK, heap profile is also fine (identical to our old bytebuffer code, so no extra allocations by panama). Code: https://github.com/uschindler/lucene/blob/7ed6da79239fa8f860ad58bf5568d0001927b6a2/lucene/core/src/java/org/apache/lucene/store/MemorySegmentIndexInput.java#L148-L211

I am fine with both APIs in memory access, although I liked the first variant using write/read a bit more. But I have seen the same problems with order of parameters.

mcimadamore:
In the end, I've decided to:

* only leave one `copyFrom` method taking a segment - this is a version that is friendly with slicing
* leave the toXYZArray - again these methods are friendly with slicing
* add back remaining segment copy method as statics, as I think the instance declaration makes order of argument harder to parse, and using System::arrayCopy-like parameter order is cleaner for more complex copies.

I believe this strikes a good balance. Copy operations are common enough that some special casing in the API is required - even though that comes at a slight cost in terms of API uniformity.

I also decided to leave `fill` where it is - while in some cases I'm able to see better performances with a straight loop and dereference, performance seem to be more fragile - especially as, in the general case, we have to take into account segment sizes that exceed `Integer.MAX_VALUE`.

mcimadamore:
Javadoc updated inline here:
http://cr.openjdk.java.net/~mcimadamore/panama/memorycopy_finalize_javadoc/javadoc/jdk/incubator/foreign/package-summary.html

mcimadamore:
Another update: I decided to remove toXYZArray methods (as discussed last week). And I have also consolidated the various primitive-accepting routines such as MemorySegment::ofArray and MemoryAccess::copy to work with `Object` instead of having multiple unrelated overloads. This was suggested by John some time ago - and I raised some performance concerns. I have now verified that the `Object`-accepting versions of the methods are just as fast, regardless of whether code is inlined or not. They are also easier to work with when the code doesn't know the type of the array it's working with, so these changes led to biggie simplifications in the memory copy tests. A javadoc for this iteration is here:

http://cr.openjdk.java.net/~mcimadamore/panama/memorycopy_finalize_javadoc_v2/javadoc/jdk/incubator/foreign/package-summary.html

(I put it in a different location so that it can be compared with the previous one).

It would be nice if @uschindler could test this iteration (sorry for the back and forth!), as this change might have a non trivial impact on performances. If things regress, we can go back to the previous iteration (but in my benchmarks I didn't see evidence of regression).

As John pointed out, the reason I like this version more is that it does without primitive overloads, which I think will look rather stale once Valhalla is around. That is, `MemorySegment` is Valhalla-ready as is - perhaps some adjustments will be required on `MemoryAccess` - but since that's a bag of static methods, it seems less problematic.

mcimadamore:
While the latest API still looks good - it occurred to me that if we want to do anything in MemoryAccess::copy which e.g. for small copy sizes bypasses the bulk copy and just does a for loop, having all the copy overloads merged into one could be problematic (because of pollution). That said, the threshold for switching copy mechanism in the ByteBuffer API is set at 6 bytes (!!) - which is very low (e.g. 1.5 ints). So I'm not sure how much that logic is really needed (especially since now `Unsafe.memoryCopy` is a VM intrinsic). That said, if, one day, we'd like to rewrite those routines in Java, using memory access API + vector API, using an opaque object carrier can definitively be an issue (the vector API also likes to be sharp in its carrier type).

rsmogura:
> That said, if, one day, we'd like to rewrite those routines in Java, using memory access API + vector API, using an opaque object carrier can definitively be an issue (the vector API also likes to be sharp in its carrier type).
It´s more about checking off-/on- heap than having just concrete object class.

PaulSandoz:
> That said, if, one day, we'd like to rewrite those routines in Java, using memory access API + vector API, using an opaque object carrier can definitively be an issue (the vector API also likes to be sharp in its carrier type).

I am not too concerned about this. For memory copying it is important to choose the largest shape (i.e. largest vector register) so we could choose to use Vector<Byte> or Vector<Integer> etc. even when copying arrays of floats or doubles. That is sort of how vectorized mistmatch is implemented. It will become easier to do this when we enhance the Vector API to load/store from segments, which could be backed by primitive arrays or native regions.

mcimadamore:
> > That said, if, one day, we'd like to rewrite those routines in Java, using memory access API + vector API, using an opaque object carrier can definitively be an issue (the vector API also likes to be sharp in its carrier type).
> 
> I am not too concerned about this. For memory copying it is important to choose the largest shape (i.e. largest vector register) so we could choose to use Vector or Vector etc. even when copying arrays of floats or doubles. That is sort of how vectorized mistmatch is implemented. It will become easier to do this when we enhance the Vector API to load/store from segments, which could be backed by primitive arrays or native regions.

True - I guess what I mean is that the vector shape won't be known statically in the code (while it could be if there were different oveloads).

mcimadamore:
I will close this PR. Some of the issues discussed here have resulted in a much bigger API restructuring which is described here:

https://mail.openjdk.java.net/pipermail/panama-dev/2021-September/014946.html

Among the changes described in the writeup linked in the above email, there is a proposal which makes the segment dereference API much tighter and integrated with the rest of the MemorySegment and MemoryLayout APIs. We believe that API is the way forward, and that `MemoryAccess` and `MemoryCopy` were, in the end, (useful) intermediate steps along the way.

PR #570, opened 2021-07-16
8270851: Logic for attaching/detaching native threads could be improved

mcimadamore:
For a complete description of the issue, please refer to:

https://bugs.openjdk.java.net/browse/JDK-8270851

This patch makes the logic for detaching native threads more lazy - by only doing the detach when a native thread has completed. This is achieved by using some thread local storage, which is used to keep track of the Java thread associated to a given native thread. If, by the time the thread local storage is destroyed, we see that a Java thread has been attached, we do a detach operation. This trick effectively minimizes the number of Thread instances created when interacting with multi-threaded native code.

This patch also tweaks the logic for attaching native threads to the VM by using the "daemon" attach variant. That is, native threads registered against the VM (because of Panama upcalls) should not prevent the JVM to shut down in an orderly fashion (in cases where the native threads might outlive the JVM).

---------
### Progress

### Issue
 * [JDK-8270851](https://bugs.openjdk.java.net/browse/JDK-8270851): Logic for attaching/detaching native threads could be improved

### Reviewers
 * [David Holmes](https://openjdk.java.net/census#dholmes) (@dholmes-ora - no project role)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/570/head:pull/570` \
`$ git checkout pull/570`

`$ git checkout pull/570` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/570/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 570`

`$ git pr show -t 570`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/570.diff">https://git.openjdk.java.net/panama-foreign/pull/570.diff</a>

</details>


JornVernee (approved):
LGTM

overheadhunter:
I just checked out this branch and ran some tests and the results are great. Not only did performance in multithreaded tests increase significantly (outperforming JNR) and memory usage dropped to a fraction, it is also way easier to run profiling tools due to having longer living threads. 🎉

dholmes-ora:
This change takes you from one extreme to another and so will introduce different problems. Changing to using daemon threads tries to avoid the obvious problem - that a non-terminating native thread causes the VM to never exit. But now the VM can exit in the middle of an upcall unless something else is keeping the VM alive. No implicit lifecycle management policy can suit all cases here.

David

overheadhunter:
Is switching to daemon threads required for this fix to work? When dealing with non-daemon native threads, will the JVM wait for it to _terminate_ or to _detach_, before it shuts down?

src/hotspot/share/prims/universalUpcallHandler.cpp
```diff
@@ -36,6 +36,26 @@
...
+
+  ~UpcallContext() {
+    if (attachedThread != NULL) {
+      JavaVM_ *vm = (JavaVM *)(&main_vm);
+      vm->functions->DetachCurrentThread(vm);
+    }
+  }
+};
+
+thread_local UpcallContext threadContext;
```
dholmes-ora, line 57: Use of C++11 `thread_local` violates the Hotspot rules for C++ feature usage. See:
https://github.com/openjdk/jdk/blob/master/doc/hotspot-style.md#thread_local
It is also unclear when such an object will be destroyed when C++ thread API is not used.

JornVernee:
Hi David, thanks for the comments. I think Maurizio will want to weigh in as well, but here are some thoughts from me:

> This change takes you from one extreme to another and so will introduce different problems. Changing to using daemon threads tries to avoid the obvious problem - that a non-terminating native thread causes the VM to never exit. But now the VM can exit in the middle of an upcall unless something else is keeping the VM alive. No implicit lifecycle management policy can suit all cases here.

I think this is what we expected to happen: something else would need to keep the VM alive during the upcall.

> > Is switching to daemon threads required for this fix to work? When dealing with non-daemon native threads, will the JVM wait for it to _terminate_ or to _detach_, before it shuts down?
> 
> For a non-daemon the JVM will remain alive as long as it is attached. So
> if it only detaches at termination then the JVM will wait till it
> terminates.

When I discussed this case with Maurizio offline, the main use case we had in mind was where a native application starts a JVM, makes a few upcalls, and then returns to the native application, at which point it would continue like a normal native application. In that case we didn't want to stop the VM from shutting down.

> There may also be a problem with using a thread-local to control
> detaching. The panama code is using the C++11 thread_local which we do
> not use in the VM. As the C++ thread_local relates to C++ threads and
> their lifecycle, it is not at all clear to me how it operates for
> threads that may not use the C++ thread API ie when will the object
> actually be destroyed?

As far as I can tell (for instance from [1]), the `thread_local` storage specifier is a language feature that is not directly related to the C++ `std::thread` API (assuming that's what you mean with 'C++ threads'), and the "thread" that `thread_local` storage applies to is an operating system thread. But, this might be an incorrect interpretation, as the documentation I linked to doesn't specify "thread".

Was there a prior discussion of `thread_local`?

AFAICT `thread_local` and the MSVC implementation of the `THREAD_LOCAL` macro work the same [2]. The problem with using `THREAD_LOCAL` is that the GCC implementation uses `__thread` which does not allow objects with non-trivial destructors. If I try this, the compiler suggests using `thread_local` instead:

```
<source>:16:19: error: non-local variable 'g' declared '__thread' has a non-trivial destructor
   16 | __thread struct A g(0);
      |                   ^
<source>:16:19: note: C++11 'thread_local' allows dynamic initialization and destruction
```

We essentially want to run some code when a thread terminates. The pthread API seems to allow for this by way of registering a destructor for a thread local [3], and it seemed that the C++ 11 `thread_local` keyword is the portable version of this.

Have we perhaps overlooked another mechanism for automatically detaching a thread on termination that already exists in HotSpot?

> Detaching a thread also uses the compiler native
> thread-local mechanism, as well as the library based mechanisms. Luckily
> if these things run in the wrong order (but don't otherwise interfere)
> they will actually appear detached before detach is called, and the call
> to detach is a no-op in that case.

Looking at `DetachCurrentThread`, this seems problematic? Won't there be a memory leak since the thread object is never destroyed in that case? `DetachCurrentThread` uses a thread local to get a pointer to the current thread and detach that, and if I understand correctly that thread local might be nulled out at the point when our destructor tries to detach the thread?

However, we have a pointer to the current thread saved as well, so it seems we could potentially rearrange the code to pass in that pointer to detach instead of using the thread-local to get a pointer to the current thread.

Thanks,
Jorn

[1] : https://en.cppreference.com/w/cpp/language/storage_duration
[2] : https://docs.microsoft.com/en-us/cpp/parallel/thread-local-storage-tls?view=msvc-160
[3] : https://linux.die.net/man/3/pthread_key_create

mcimadamore:
> No implicit lifecycle management policy can suit all cases here.

I believe at some point we will have options to configure this, as we do for native thread transitions in downcalls.

mcimadamore:
> As far as I can tell (for instance from [1]), the `thread_local` storage specifier is a language feature that is not directly related to the C++ `std::thread` API (assuming that's what you mean with 'C++ threads'), and the "thread" that `thread_local` storage applies to is an operating system thread. But, this might be an incorrect interpretation, as the documentation I linked to doesn't specify "thread".

I've seen similar notes too, even in the C++ standard. There's no direct mention that a "thread" is a thread coming from `std::thread` anywhere. That said, from there to inferring that `thread_local` must work regardless of how threads are created is, probably, a too optimistic take on the spec.

In Linux-land, things seem to work pretty robustly, since the C++ compiler will register destructors using certain API points:

https://code.woboq.org/userspace/glibc/stdlib/cxa_thread_atexit_impl.c.html#__cxa_thread_atexit_impl

IMHO, the code here suggests that destructors works at a pretty general level, and they are called whenever a process exits, for instance:

https://code.woboq.org/userspace/glibc/stdlib/exit.c.html#46

But also when pthreads complete:

https://code.woboq.org/userspace/glibc/nptl/pthread_create.c.html#487

(and, likely when std::threads do too).

So, yes, there needs to be cooperation between the thread API and use of `thread_local`, but it seems to me that, at least in linux, the mechanism seems general enough to common the main thread APIs being used there. Of course if a thread is created by using lower level building blocks, you are probably already in undefined behavior territory, so in that case guarantees for thread_local cannot be assumed.

dholmes-ora:
Hi Jorn, Maurizio,

Sorry I didn't see these updates due to the skara email outage.

It is hard to try and go through everything point by point! :) And I don't have definitive answers for everything, I'm just raising awareness of potential problems.

First note `thread_local` is not allowed as a C++ language feature in hotspot per the style guide on allowed C++ features - so that is the first battle that needs to be fought (email to hotspot-dev). It may be an easy battle, it may not, that depends on how much detail we have about how a language based thread-local-storage (TLS) mechanism interacts with non-C++ threads and the compiler and library TLS mechanisms. I can easily imagine that the language, compiler, and library versions of TLS all hook into the same "termination hook" in the platform thread management code, but it remains an unknown what order the different categories of "destructor" will run. That fact that gcc __thread doesn't allow non-trivial object initialization and destruction itself suggests there is some significant different between __thread and C++ thread_local in that environment.

TLS is a very complex area and it is very easy for different uses of thread-locals to interact in bad ways given that the VM itself uses thread-local for dealing with Thread::current() (and normally uses two mechanisms: compiler TLS and library TLS, because only the library TLS is considered signal-safe, but compiler TLS is (usually) much faster). 

> Have we perhaps overlooked another mechanism for automatically detaching a thread on termination that already exists in HotSpot?

No there is no such mechanism. It is in fact quite problematic. We had to introduce a hack in the VM code (threadLocalStorage_posix.cpp) to allow a pthread TLS key destructor to detach a terminating thread - with a side effect that if a thread fails to detach then it may hang on termination. :(

If you look at what detaching a thread does it is non-trivial, with further calls to Java being made; and detaching depends on both the state of the thread (any Java frames still on the stack?) and the state of the VM (see next point).

The lifecycle management aspect of this is complex. The VM is load-once and never truly unloads within a process, but is not reusable once it has been "terminated". Also note that if the VM has terminated before a daemon thread tries to detach then the terminating thread will just hang in VMExit::block_if_vm_exited() - which is unlikely what the application wants. You can't just be a casual user of the JVM this way I'm afraid. The application really needs to understand how it uses the JVM and how/when it will "terminate" it. Daemon threads have a false allure to them - in practice because they can stop at any point in their execution when the VM terminates, they really have to be doing simple non-essential tasks.

> Looking at DetachCurrentThread, this seems problematic? Won't there be a memory leak ...

As long as there is one successful detach there is no leak, but otherwise yes it will leak the JavaThread and associated resources. But things could get worse because if a thread terminates without detaching, and is visible to some of the other API's (like monitoring and management, or JFR) then we may try to interact with that terminated thread.

I don't recall what the panama programming model is in regard to application use of native threads and the use of upcalls, but perhaps for the case at hand when threads are making many upcalls and the attach/detach overhead is significant, then the application should be responsible for attaching and later detaching these threads, with the upcall mechanism checking for an already attached thread?

Anyway ... if you still want to pursue this approach then as I said thread_local needs to be approved for use in hotspot (probably fine to use it in panama repo in the meantime).

Cheers,
David

mcimadamore:
>  don't recall what the panama programming model is in regard to application use of native threads and the use of upcalls, but perhaps for the case at hand when threads are making many upcalls and the attach/detach overhead is significant, then the application should be responsible for attaching and later detaching these threads, with the upcall mechanism checking for an already attached thread?

The main issue with putting developers in charge of deciding when/how to attach/detach is that, unlike with JNI, where user code runs (albeit in C/C++) before/after the upcall (so we have a chance to attach/detach things), in Panama the code doing the upcall is some unknowing piece of native code calling a function pointer. There's nothing we can inject in the library code, of course, as that code is fixed - so the only thing we can change is what happens when the upcall starts/finishes.

I have few questions for you though.

1. What happens if an attached thread is never deattached?
2. We can imagine we have an API for attaching threads, exposed by CLinker:

```
boolean attachThread(Thread t, boolean daemon, ResourceScope)
```

This will attach a thread (if needed) to the JVM (either in daemon mode, or normal), and will register that onto a ResourceScope, meaning that when the ResourceScope is closed, the thread will be deattached.

This API is not impossible to do - what I'd like to understand is what happens if we jump into the upcall code _before_ we have attached the thread to the JVM. Is that a correctness issue? If not, we could go down that path - but if a native thread can _never ever_ execute a single line of Java code before being attached to the VM, well, that creates a chicken and egg problem, I think.

mcimadamore:
Thanks David - I think Jorn and I came up with a possible, different idea to handle this - which is to attach the lifecycle of the attached threads to the resource scope used to create an upcall. That way, when the upcall stub is destroyed, any threads that happened to call it would be deattached by the VM. We need to do some experiments, as this approach could be problematic in cases where the same native thread calls different upcall stubs - but it seems worthwhile exploring.

src/hotspot/share/prims/universalUpcallHandler.cpp
```diff
@@ -60,7 +79,7 @@ void ProgrammableUpcallHandler::detach_current_thread() {
 
 // modelled after JavaCallWrapper::JavaCallWrapper
 JavaThread* ProgrammableUpcallHandler::on_entry(OptimizedEntryBlob::FrameData* context) {
-  JavaThread* thread = maybe_attach_and_get_thread(&context->should_detach);
```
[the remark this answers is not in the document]
mcimadamore: isn't that already the case in codeBlob.hpp ?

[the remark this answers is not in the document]
JornVernee: Oh, huh. You're right. I think those changes might have been hidden when I looked at the changes, since I already looked at them before. (Now I see it).

JornVernee (approved):

src/hotspot/share/prims/universalUpcallHandler.cpp
```diff
@@ -37,18 +37,37 @@
...
+
+APPROVED_CPP_THREAD_LOCAL UpcallContext threadContext;
+
+JavaThread* ProgrammableUpcallHandler::maybe_attach_and_get_thread() {
   JavaThread* thread = JavaThread::current_or_null();
   if (thread == nullptr) {
     JavaVM_ *vm = (JavaVM *)(&main_vm);
     JNIEnv* p_env = nullptr; // unused
     jint result = vm->functions->AttachCurrentThread(vm, (void**) &p_env, nullptr);
     guarantee(result == JNI_OK, "Could not attach thread for upcall. JNI error code: %d", result);
```
dholmes-ora, line 67: Do you really want to abort if something goes wrong i.e. out-of-memory**? (Also the JNI error code won't be very illuminating as only JNI_ERR is ever returned from AttachCurrentThread.)

** Any unexpected exceptions from the java.lang.Thread construction will be silently cleared but also cause failure.

dholmes-ora (approved):
Update to new thread_local usage looks good. :)

One query below on error handling.

src/hotspot/share/prims/universalUpcallHandler.cpp
mcimadamore: For errors that occurs inside the upcall, we generally try to print a stack trace - but in this case we're a bit of a limbo between native and Java - and we still don't have a Java thread - so I'm assuming that if something goes wrong here there's not much we can do to dress it up nicely and/or recover? I agree that printing out JNI error code seems over the top (and not very useful).

I note that this code has not changed, even in the foreign-preview branch, which contains a bunch of JVM cleanups. Maybe we can file a followup (but I'd like to understand better what you are proposing - and what options we have on the table).

mcimadamore: I tend to agree - only observation is that reporting the JNI error code, as David noted, is probably useless.

dholmes-ora: I admit I don't know what the full context is here but at some level you have a piece of external code calling something and expecting something useful to happen as a result. I would have expected that API, whatever it is, to have some way of indicating that an error occurred and that the "something useful" did not in fact happen. If the initial code does not even know it is calling into a JVM I would expect it to be pretty miffed when the JVM decides to blow away the entire process due to an incidental issue the JVM had! 

Even if aborting is your only option I don't think a guarantee, with ensuing JVM core dump and hs_err file, is the way to go about it. I would not want to be fielding bug reports against the JVM in such a case.

Cheers.

mcimadamore: I think aborting is the only option. The flow starts with some native call made by Java code. At some point the native code upcalls back into Java, and, for some reason, attaching a thread fails. When this failure occurs we are deep inside a native library. Popping back until we reach some Java frame seems undesirable (what happens to all the native code in the middle?). As I pointed out yesterday, we have a similar situation when the *Java* code inside an upcall fails with some exception: but in that case at least we have a thread, so we can dump a stack trace, which makes it "look" like a normal exception (but it's not, in the sense that it cannot be caught by the Java code - as the VM exits after dumping the stack trace).
This is worse, as we can't even dump the stack trace (as we have no thread), so the best we can do is to dump some message, and then exit. I agree that generating a VM dump is too much, and we should scale that back. But I don't think there is a meaningful way to recover (as the recovery would have to happen inside the C code, which Panama cannot, by definition, control).

PR #573, opened 2021-07-24
Check for mixed access - remove mem barriers

rsmogura:
That's for previous PR, which I destroyed with bad merge from master.

I wonder if this can look like this?

```
Benchmark                               (size)  Mode  Cnt   Score    Error  Units
ByteBufferVectorAccess.directBuffers      1024  avgt   10  27.119 ?  0.186  ns/op
ByteBufferVectorAccess.heapBuffers        1024  avgt   10  28.501 ?  0.118  ns/op
ByteBufferVectorAccess.pollutedBuffers    1024  avgt   10  64.354 ? 27.778  ns/op
Finished running test 'micro:ByteBufferVectorAccess'
```

https://github.com/openjdk/panama-foreign/pull/566

---------
### Progress

### Integration blocker
&nbsp;⚠️ Whitespace errors (failed with the updated jcheck configuration)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign.git pull/573/head:pull/573` \
`$ git checkout pull/573`

`$ git checkout pull/573` \
`$ git pull https://git.openjdk.org/panama-foreign.git pull/573/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 573`

`$ git pr show -t 573`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/573.diff">https://git.openjdk.org/panama-foreign/pull/573.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/panama-foreign/pull/573#issuecomment-886079403)

### DIFF
--- a/src/hotspot/share/opto/vectorIntrinsics.cpp
+++ b/src/hotspot/share/opto/vectorIntrinsics.cpp
@@ -760,8 +760,19 @@ bool LibraryCallKit::inline_vector_mem_operation(bool is_store) {
   SafePointNode* old_map = clone_map();
 
   Node* addr = make_unsafe_address(base, offset, (is_mask ? T_BOOLEAN : elem_bt), true);
+  
+  // This check is repetition of some checks from inline_unsafe_access(), used to determine if barriers are needed
+  // Not full scope of checks is performed, we check only if access can be mixed
+  const Type *const base_type = gvn().type(base);
+
+  // Is off heap access (true implies can_access_non_heap = true)
+  const bool off_heap_access = TypePtr::NULL_PTR == base_type;
+
   // Can base be NULL? Otherwise, always on-heap access.
-  bool can_access_non_heap = TypePtr::NULL_PTR->higher_equal(gvn().type(base));
+  const bool can_access_non_heap = TypePtr::NULL_PTR->higher_equal(base_type);
+
+  // Not determined access base can and can not be null.
+  const bool mixed_access = !(off_heap_access == can_access_non_heap);
 
   const TypePtr *addr_type = gvn().type(addr)->isa_ptr();
   const TypeAryPtr* arr_type = addr_type->isa_aryptr();
@@ -823,7 +834,7 @@ bool LibraryCallKit::inline_vector_mem_operation(bool is_store) {
 
   const TypeInstPtr* vbox_type = TypeInstPtr::make_exact(TypePtr::NotNull, vbox_klass);
 
-  if (can_access_non_heap) {
+  if (mixed_access) {
     insert_mem_bar(Op_MemBarCPUOrder);
   }
 
@@ -870,7 +881,7 @@ bool LibraryCallKit::inline_vector_mem_operation(bool is_store) {
 
   old_map->destruct(&_gvn);
 
-  if (can_access_non_heap) {
+  if (mixed_access) {
     insert_mem_bar(Op_MemBarCPUOrder);
   }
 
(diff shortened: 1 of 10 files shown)


PaulSandoz:
That's an neat trick, but it will not work if the heap buffer is read only. But you could extract the array (or base) using unsafe access.

It would be preferable if we can tunnel this through the intrinsic, rather than branching on buffer state. VarHandle buffer/segment accessors uniformly go through unsafe for heap and direct. So in theory it should be possible for intrinsic vector access to achieve the same degree of functionality as intrinsic unsafe access.

(Note that for any integration we need to push first to the `panama-vector/vectorIntrinsics` branch)

rsmogura:
There's one "pollution" case I tested in ByteBufferVectorAccess.pollutedBuffers. I'm not sure if it's covered with MemorySegments - I created simillar test case, and it looks not. However I'm not sure maybe it's too extreme? [1]

From the other hand null check in ScopedMemoryAccess.storeIntoByteBufferScoped (version [2]) did not improve performance (I think loop is not unswitched and not unrolled, even if I don't see mem bars there).

(And now I see what's wrong with vectorIntrinsics branch it's in different repo... ).

--

[1] https://github.com/rsmogura/panama-foreign/blob/pollute-mem-seg-var-handle/test/micro/org/openjdk/bench/jdk/incubator/foreign/MemorySegmentPolluteUnsafe.java

[2] https://github.com/openjdk/panama-foreign/compare/foreign-memaccess+abi...rsmogura:vector-buffers-closer-to-intrinsic?expand=1

rsmogura:
Forget add benchmarks results for version with checks closer to intrinsic

```
Benchmark                               (size)  Mode  Cnt    Score    Error  Units
ByteBufferVectorAccess.directBuffers      1024  avgt   10   12.075 ?  0.178  ns/op
ByteBufferVectorAccess.heapBuffers        1024  avgt   10   15.840 ?  0.177  ns/op
ByteBufferVectorAccess.pollutedBuffers    1024  avgt   10  123.074 ? 34.672  ns/op
```
Generaly a + b < combined(a,b)

rsmogura:
I did some deeper checks about polluted case and for some reason Unsafe works differently than "get filed". I can't answer  why so. It looks very related to JIT.
```
void intoByteBuffer0(ByteBuffer bb, int offset) {
if (!bb.isDirect()) {
  return fromByteArray0(bb.array(), offset);
}
```
change to 
```
if (!bb.isDirect()) {
  intoByteArray0((byte[]) ScopedMemoryAccess.BufferAccess.bufferBase(bb), offset);
```
makes such a huge difference.

PaulSandoz:
I think we need to have a base comparison with `VarHandle` access to a byte buffer, not in terms of performance but when barriers occur or not. I believe we should be able to push the unpolluted base+offset vector heap and direct accesses into `LibraryCallKit::inline_vector_mem_operation`.

Re: repo, sorry i assumed you were aware of the different panama repos, to be precise this one: https://github.com/openjdk/panama-vector/tree/vectorIntrinsics

rsmogura:
I checked and BB view handle can create mem bars in polluted case, otherwise do not - https://gist.github.com/rsmogura/3629592b5ab90135f27e76f837693647 

I wonder if using VH to access HB can help with performance

rsmogura:
For polluted case, i think: Scoped memory access have two paths for storing vectors, one producing byte[] result one raw, those are merged together to all mem. The load in loop takes this.

As a consequence loop can't be optimized at all (it effectively works like mem barrier), so even loop limit gets executed with every loop cycle. I think this is more generic JIT case - probably worth investigating as I guess similar case can be produced not only for vectors.

However normal cases works better.

mcimadamore:

PR #576, opened 2021-09-16
8273905: Foreign API refresh

mcimadamore:
As outlined in [1], there are areas of the foreign API we'd like to improve upon, such as:

* dereference: there is a mismatch between API points which create segments (which take a layout) and dereference API points, which do not take layouts

* role of `MemoryAddress`: in Java 17, `MemoryAddress` has become a stateful carrier, which is attached to a scope. This is inconvenient, as in most cases, a memory address is just a raw pointer that arises when interacting with native code.

* resource scopes: the API for scopes has too many flavors to pick from, many of which overlap. The fact that scopes and segment allocators are unrelated forces clients to introduce ad hoc conversions from scopes to allocators, and library developers to add overloads. Finally, the API for acquiring scopes doesn't work well with try-with-resources, and could also be simplified. 

I will add separate comments to explain how the API has changed to resolve the above issues.

[1] - https://mail.openjdk.java.net/pipermail/panama-dev/2021-September/014946.html

---------
### Progress

### Issue
 * [JDK-8273905](https://bugs.openjdk.java.net/browse/JDK-8273905): Foreign API refresh

### Reviewers
 * [Uwe Schindler](https://openjdk.java.net/census#uschindler) (@uschindler - no project role) ⚠️ Review applies to 162cb936efeb7341f2beb4539545211d5bc4d4aa
 * [Paul Sandoz](https://openjdk.java.net/census#psandoz) (@PaulSandoz - Committer) ⚠️ Review applies to 162cb936efeb7341f2beb4539545211d5bc4d4aa

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/576/head:pull/576` \
`$ git checkout pull/576`

`$ git checkout pull/576` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/576/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 576`

`$ git pr show -t 576`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/576.diff">https://git.openjdk.java.net/panama-foreign/pull/576.diff</a>

</details>


mcimadamore:
### Memory dereference

The main change in this area is to attach carriers to `ValueLayout`. Because of this, we have defined several subclasses of `ValueLayout` (one per carrier), such as `ValueLayout.OfInt`, `ValueLayout.OfFloat` and so forth. Attaching layouts to carriers allows to express deference method in an elegant way - e.g. instead of:

`MemoryAccess.getIntAtOffset(segment, offset)`

simply:

```
segment.get(JAVA_INT, offset);
```

The new dereference methods are statically safe: it is not possible for clients to pass values to setters of the *wrong* type. We have used the same trick in different parts of the API, to make it more uniform and safer (see changes in `SegmentAllocator` and `VaList`). As a result, `MemoryAccess` and `MemoryCopy` go away, as the same operations are now available as instance (or static, in the case of copy) methods in `MemorySegment`. Since the new dereference API points are parameterized by a layout, it is always possible to e.g. create a layout with desired endianness and use that instead. The only caveat is that, for performance to be optimal, the layout using during dereference should be a VM constant (e.g. static final field).

### CLinker

First, we have moved most of the `CLinker` helper methods away from `CLinker`. For instance, to allocate an Utf8 string, there is now a method in `SegmentAllocator`, as well as a pair of getter/setter in `MemorySegment`. This makes the code easier to follow. 
Since carriers are attached to value layout, passing a `MethodType` to `CLinker::downcallHandle` is no longer necessary, so the API for obtaining downcall method handles has been simplified.

Some things have been removed as well; most notably, the platform-dependent layout constants, such as `C_INT` are gone. This is because we cannot type these constants sharply - given that their type would vary across platforms (is `C_LONG` a `ValueLayout.OfInt` or a `ValueLayout.OfLong` ?). Ultimately, it is the role of a tool (such as jextract) to generate a set of layout constants which work on a given platform, so we have made the API more neutral. It is still possible to use layouts such as `JAVA_INT` and `JAVA_LONG` to link downcall method handles, and that is not too different from using JNI types such as `jint` and `jlong` (that is, they should be used with care, making sure they match the underlying platform types).

Another thing we removed was the `freeMemory` and `allocateMemory` API points. These were just shortcut for commonly used downcall method handles, which in most of the cases will be included in a jextraction unit anyway. There seems to be little value in including such methods in the `CLinker` API.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -546,73 +506,149 @@ default MemorySegment asSlice(MemoryAddress newBase) {
...
+     *
+     * @param offset offset in bytes (relative to this address). The final address of this read operation can be expressed as {@code toRowLongValue() + offset}.
+     * @return a Java UTF-8 string containing all the bytes read from the given starting address ({@code toRowLongValue() + offset})
+     * up to (but not including) the first {@code '\0'} terminator character (assuming one is found).
+     * @throws IllegalArgumentException if the size of the native string is greater than the largest string supported by the platform.
+     * @throws IllegalCallerException if access to this method occurs from a module {@code M} and the command line option
+     * {@code --enable-native-access} is either absent, or does not mention the module name {@code M}, or
+     * {@code ALL-UNNAMED} in case {@code M} is an unnamed module.
+     */
+
```
PaulSandoz, line 612:
```suggestion
```

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -546,73 +506,149 @@ default MemorySegment asSlice(MemoryAddress newBase) {
...
+     * sequences with this charset's default replacement string.  The {@link
+     * java.nio.charset.CharsetDecoder} class should be used when more control
+     * over the decoding process is required.
+     * @param offset offset in bytes (relative to this segment). For instance, if this segment is a {@link #isNative()} segment,
+     *               the final address of this read operation can be expressed as {@code address().toRowLongValue() + offset}.
+     * @return a Java UTF-8 string containing all the bytes read from the given starting address up to (but not including)
+     * the first {@code '\0'} terminator character (assuming one is found).
+     * @throws IllegalArgumentException if the size of the native string is greater than the largest string supported by the platform.
+     * @throws IllegalStateException if the size of the native string is greater than the size of the segment
+     * associated with {@code addr}, or if {@code addr} is associated with a segment that is <em>not alive</em>.
```
PaulSandoz, line 626: Referral to unknown parameter `addr`

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/ResourceScope.java
```diff
@@ -56,25 +58,24 @@
...
- * <h2>Implicit resource scopes</h2>
+ * <h2>Implicit deallocation</h2>
  *
- * Resource scopes obtained from {@link #newImplicitScope()} cannot be closed explicitly. We call these resource scopes
- * <em>implicit scopes</em>. Calling {@link #close()} on an implicit resource scope always results in an exception.
- * Resources associated with implicit scopes are released once the scope instance becomes
- * <a href="../../../java/lang/ref/package.html#reachability">unreachable</a>.
+ * Resource scopes are also closed automatically once the scope instance becomes
+ * <a href="../../../java/lang/ref/package.html#reachability">unreachable</a>. This can be useful to allow for predictable,
+ * deterministic resource deallocation, while still prevent accidental native memory leaks. In case a managed resource scope
```
PaulSandoz, line 66:
```suggestion
 * deterministic resource deallocation, while still preventing accidental native memory leaks. In case a managed resource scope
```

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/package-info.java
```diff
@@ -81,6 +83,8 @@
  * which is used to <em>bind</em> the life-cycle of the segment created immediately afterwards. Note the use of the
  * <em>try-with-resources</em> construct: this idiom ensures that all the memory resources associated with the segment will be released
  * at the end of the block, according to the semantics described in Section {@jls 14.20.3} of <cite>The Java Language Specification</cite>.
+ * Since a resource scope act as a {@link jdk.incubator.foreign.SegmentAllocator segment allocators}, we can also
```
PaulSandoz, line 86:
```suggestion
 * Since a resource scope acts as a {@link jdk.incubator.foreign.SegmentAllocator segment allocator}, we can also
```

test/jdk/java/foreign/TestArrays.java
```diff
@@ -41,47 +40,55 @@
...
 public class TestArrays {
 
     static SequenceLayout bytes = MemoryLayout.sequenceLayout(100,
-            MemoryLayouts.JAVA_BYTE
+            JAVA_BYTE
     );
 
     static SequenceLayout chars = MemoryLayout.sequenceLayout(100,
-            MemoryLayouts.JAVA_CHAR
+            ValueLayout.JAVA_CHAR
```
PaulSandoz, line 62:
```suggestion
            JAVA_CHAR
```
And for other usages.

test/jdk/java/foreign/TestByteBuffer.java
```diff
@@ -102,40 +108,40 @@
...
     );
 
     static SequenceLayout shorts = MemoryLayout.sequenceLayout(100,
-            MemoryLayouts.BITS_16_BE
+            JAVA_SHORT.withOrder(ByteOrder.BIG_ENDIAN)
     );
 
     static SequenceLayout ints = MemoryLayout.sequenceLayout(100,
-            MemoryLayouts.BITS_32_BE
+            ValueLayout.JAVA_INT.withOrder(ByteOrder.BIG_ENDIAN)
```
PaulSandoz, line 128:
```suggestion
            JAVA_INT.withOrder(ByteOrder.BIG_ENDIAN)
```

src/java.base/share/classes/jdk/internal/access/foreign/MemorySegmentProxy.java
```diff
@@ -27,6 +27,7 @@
 package jdk.internal.access.foreign;
 
 import jdk.internal.misc.ScopedMemoryAccess;
+import jdk.internal.vm.annotation.ForceInline;
```
sundararajana, line 30: Is this import needed?

mcimadamore: Leftover form past experiments - good catch!

PaulSandoz:
I recommend running IntelliJ analysis over the code before integrating. It often finds small problems, typos, Java doc issues.

mcimadamore:
> I recommend running IntelliJ analysis over the code before integrating. It often finds small problems, typos, Java doc issues.

I've ran the typos and grammar analyses already (so what you see should be already relatively polished). I will run some more to catch code issues.

uschindler (approved):
I found some minor javadoc things (just by looking at it and searching for defintiions, so there may be other problems).

I did some testing in Apache Lucene with a build of your branch: https://github.com/uschindler/lucene/tree/draft/jdk-foreign-mmap-jdk18 (the relevant change to previous version using the old MemoryCopy stuff is this commit: https://github.com/uschindler/lucene/commit/2de65069a7419edd43f5fc3b63e6e9fa077e2235

The new API looks really good. I also removed all usage of `VarHandle` in the Lucene implementation and just used the versions of `MemorySegment#get(ValueLayout.Xxx,...)` for our positional reads. Interestingly the performance seemed to get a bit better without direct use of VarHandles (maybe because you use `@ForceInline`), but not significant (inside the benchmark's std dev).

The most interesting code for you is here: https://github.com/uschindler/lucene/blob/draft/jdk-foreign-mmap-jdk18/lucene/core/src/java/org/apache/lucene/store/MemorySegmentIndexInput.java

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryHandles.java
```diff
@@ -254,8 +179,8 @@ public static VarHandle asAddressVarHandle(VarHandle target) {
      * Java {@code int} to avoid dealing with negative values, which would be
      * the case if modeled as a Java {@code short}. This is illustrated in the following example:
      * <blockquote><pre>{@code
-    MemorySegment segment = MemorySegment.allocateNative(2, ResourceScope.newImplicitScope());
-    VarHandle SHORT_VH = MemoryLayouts.JAVA_SHORT.varHandle(short.class);
+    MemorySegment segment = MemorySegment.allocateNative(2, ResourceScope.newConfinedScope());
+    VarHandle SHORT_VH = ValueLayout.JAVA_SHORT.varHandle(short.class);
```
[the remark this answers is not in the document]
uschindler: There may be other similar problems at other places!

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -836,27 +888,730 @@ static MemorySegment mapFile(Path path, long bytesOffset, long bytesSize, FileCh
...
+     * Supported array types are {@code byte[]}, {@code char[]},{@code short[]},{@code int[]},{@code float[]},{@code long[]} and {@code double[]}.
+     */
+    @ForceInline
+    static void copy(
+            MemorySegment srcSegment, ValueLayout srcLayout, long srcOffset,
+            Object dstArray, int dstIndex, int elementCount) {
+        Objects.requireNonNull(srcSegment);
+        Objects.requireNonNull(dstArray);
+        Objects.requireNonNull(srcLayout);
+        long baseAndScale = getBaseAndScale(dstArray.getClass());
```
uschindler, line 1544: Should we not add a check that the srcLayout's carrier type fits the dstArray type?

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -836,27 +888,730 @@ static MemorySegment mapFile(Path path, long bytesOffset, long bytesSize, FileCh
...
+     * @throws  IllegalArgumentException if {@code srcArray} is not an array, or if it is an array but whose type is not supported.
+     * Supported array types are {@code byte[]}, {@code char[]},{@code short[]},{@code int[]},{@code float[]},{@code long[]} and {@code double[]}.
+     */
+    @ForceInline
+    static void copy(
+            Object srcArray, int srcIndex,
+            MemorySegment dstSegment, ValueLayout dstLayout, long dstOffset, int elementCount) {
+        Objects.requireNonNull(srcArray);
+        Objects.requireNonNull(dstSegment);
+        Objects.requireNonNull(dstLayout);
```
uschindler, line 1580: same here: dstLayout's carrier type should be srcArray's type.

uschindler (changes requested):

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
mcimadamore: Good point(s) - I will fix.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryHandles.java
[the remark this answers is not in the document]
mcimadamore: True - always hard to keep samples up to date with API changes - hopefully javadoc snippets will fix this issue :-)

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryAddress.java
```diff
@@ -76,155 +93,686 @@ default MemoryAddress address() {
...
+    static MemoryAddress allocateMemory(long size) {
+        Reflection.ensureNativeAccess(Reflection.getCallerClass());
+        MemoryAddress addr = SharedUtils.allocateMemoryInternal(size);
+        if (addr.equals(NULL)) {
+            throw new OutOfMemoryError();
+        } else {
+            return addr;
+        }
+    }
+
```
sundararajana, line 193: Do we still need this?

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -546,83 +505,135 @@ default MemorySegment asSlice(MemoryAddress newBase) {
...
     /**
-     * Creates a new confined buffer memory segment that models the memory associated with the given byte
+     * Reads a UTF-8 encoded, null-terminated string from this segment at given offset.
+     * <p>
+     * This method always replaces malformed-input and unmappable-character
+     * sequences with this charset's default replacement string.  The {@link
+     * java.nio.charset.CharsetDecoder} class should be used when more control
+     * over the decoding process is required.
+     * @param offset offset in bytes (relative to this segment). For instance, if this segment is a {@link #isNative()} segment,
+     *               the final address of this read operation can be expressed as {@code address().toRowLongValue() + offset}.
```
sundararajana, line 598: Mention of isNative() which is removed.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -546,83 +505,135 @@ default MemorySegment asSlice(MemoryAddress newBase) {
...
+    }
+
+    /**
+     * Writes a UTF-8 encoded, null-terminated string into this segment at given offset.
+     * <p>
+     * This method always replaces malformed-input and unmappable-character
+     * sequences with this charset's default replacement string.  The {@link
+     * java.nio.charset.CharsetDecoder} class should be used when more control
+     * over the decoding process is required.
+     * @param offset offset in bytes (relative to this segment). For instance, if this segment is a {@link #isNative()} segment,
```
sundararajana, line 616: Mention of ofNative()

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/UpcallStubs.java
```diff
@@ -52,4 +51,22 @@ private static void freeUpcallStub(long stubAddress) {
...
+                @Override
+                public void cleanup() {
+                    freeUpcallStub(entry);
+                }
+            });
+        }
+
+        @Override
+        public MemoryAddress address() {
+            return MemoryAddress.ofLong(entry);
```
PaulSandoz, line 69: Does not check if the scope is closed.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/VaList.java
```diff
@@ -0,0 +1,253 @@
+package jdk.incubator.foreign;
```
sundararajana, line 1: copyright header is missing

test/micro/org/openjdk/bench/jdk/incubator/foreign/CLayouts.java
```diff
@@ -0,0 +1,72 @@
+package org.openjdk.bench.jdk.incubator.foreign;
```
sundararajana, line 1: copyright header missing

test/micro/org/openjdk/bench/jdk/incubator/foreign/LoopOverNewHeap.java
```diff
@@ -0,0 +1,112 @@
+/*
+ * Copyright (c) 2020, Oracle and/or its affiliates. All rights reserved.
```
sundararajana, line 2: new file. (c) year 2021

PR #577, opened 2021-09-22
8274155  Foreign API refresh - jextract update

mcimadamore:
These are the jextract changes for the recent refresh of the Foreign Memory Access/Foreign Linker API. The biggest change in this API refresh comes from the removal of layout attributes. Jextract used to store several information in the form of layouts attributes. After some analysis, it turned out that in all cases, the use of layout attributes was not necessary, and it was really driven by the fact that jextract was, in certain places, using `MemoryLayout` as its main internal representation, instead of using the more flexible `Declaration`/`Type` APIs. Unfortunately, updating `RecordLayoutComputer` to work on `Declaration` rather than `MemoryLayout` caused changes in several part of the jextract parsing frontend.

Other minor changes were required, to adjust jextract to the new API changes - but most of these are small. I've dropped several now redundant overloads (so the size of the generated bindings should be smaller). There's also an additional piece of code which is used to emit typedef for common C primitive types such as `C_INT` and the likes (since the layouts for these are no longer in `CLinker`).

The test changes are mostly trivial updates to use the new API.

---------
### Progress

### Issue
 * [JDK-8274155](https://bugs.openjdk.java.net/browse/JDK-8274155): Foreign API refresh - jextract update

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/577/head:pull/577` \
`$ git checkout pull/577`

`$ git checkout pull/577` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/577/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 577`

`$ git pr show -t 577`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/577.diff">https://git.openjdk.java.net/panama-foreign/pull/577.diff</a>

</details>


mcimadamore:
Warning: there's a lot of changes in the generated libclang bindings, as I had to make these compatible with the new API. I basically bulk replaced code to use new API idioms using regular expressions, so the changes in these classes are not very interesting (as long as the thing still builds). We should probably regenerate the bindings after we're done, and clean them up.

src/jdk.incubator.jextract/share/classes/jdk/internal/clang/libclang/RuntimeHelper.java
```diff
@@ -46,17 +49,19 @@
 import java.util.stream.Stream;
 
 import static jdk.incubator.foreign.CLinker.*;
+import static jdk.internal.clang.libclang.CLayouts.C_DOUBLE;
+import static jdk.internal.clang.libclang.CLayouts.C_LONG_LONG;
+import static jdk.internal.clang.libclang.CLayouts.C_POINTER;
```
sundararajana, line 55: I assume this is transition code. RuntimeHelper is generated. It cannot use handwritten CLayouts interface (unless we plan to generate something similar from jextract every time)

src/jdk.incubator.jextract/share/classes/jdk/internal/clang/libclang/constants$0.java
```diff
@@ -32,6 +32,8 @@
 import java.nio.ByteOrder;
 import jdk.incubator.foreign.*;
 import static jdk.incubator.foreign.CLinker.*;
+import static jdk.internal.clang.libclang.CLayouts.*;
+
```
sundararajana, line 36: Same comment as for RuntimeHelper. All these are extracted code. Need to figure out how to do CLayouts equivalent from jextract.

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/ConstantBuilder.java
```diff
@@ -340,9 +329,9 @@ private Constant emitConstantSegment(String javaName, Object value) {
         append(memberMods());
         append("MemorySegment ");
         append(fieldName);
-        append(" = CLinker.toCString(\"");
+        append(" = ResourceScope.newSharedScope().allocateUtf8String(\"");
```
sundararajana, line 332: Do we need shared scope here? or confined scope will do?

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/OutputFactory.java
```diff
@@ -389,7 +367,7 @@ public Void visitVariable(Declaration.Variable tree, Declaration parent) {
 
         Type type = tree.type();
 
-        if (type instanceof Type.Declared && ((Type.Declared) type).tree().name().isEmpty()) {
+        if (type instanceof Type.Declared) {
             // anon type - let's generate something
```
sundararajana, line 371: name empty check removed. Does this comment make sense still?

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/RecordLayoutComputer.java
```diff
@@ -150,4 +181,12 @@ long offsetOf(Type parent, Cursor c) {
...
         }
     }
+
+//    public static String getName(Type type) {
+//        Cursor c = type.getDeclarationCursor();
+//        if (c.isInvalid()) {
+//            return type.spelling();
+//        }
+//        return c.spelling();
+//    }
```
sundararajana, line 191: debug code leftover?

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/ToplevelBuilder.java
```diff
@@ -177,8 +175,17 @@ String mods() {
...
+            emitPrimitiveTypedef(Type.primitive(Type.Primitive.Kind.Bool), "C_BOOL");
+            emitPrimitiveTypedef(Type.primitive(Type.Primitive.Kind.Char), "C_CHAR");
+            emitPrimitiveTypedef(Type.primitive(Type.Primitive.Kind.Short), "C_SHORT");
+            emitPrimitiveTypedef(Type.primitive(Type.Primitive.Kind.Int), "C_INT");
+            emitPrimitiveTypedef(Type.primitive(Type.Primitive.Kind.Long), "C_LONG");
+            emitPrimitiveTypedef(Type.primitive(Type.Primitive.Kind.LongLong), "C_LONG_LONG");
+            emitPrimitiveTypedef(Type.primitive(Type.Primitive.Kind.Float), "C_FLOAT");
+            emitPrimitiveTypedef(Type.primitive(Type.Primitive.Kind.Double), "C_DOUBLE");
+            emitPointerTypedef("C_POINTER");
         }
```
sundararajana, line 189: Will this clash with user's typedef?

typedef int C_INT; 

in user's header?

test/jdk/tools/jextract/JextractToolRunner.java
```diff
@@ -53,6 +54,18 @@
...
+    public static final ValueLayout.OfBoolean C_BOOL = ValueLayout.JAVA_BOOLEAN;
+    public static final ValueLayout.OfByte C_CHAR = ValueLayout.JAVA_BYTE;
+    public static final ValueLayout.OfShort C_SHORT = ValueLayout.JAVA_SHORT;
+    public static final ValueLayout.OfInt C_INT = ValueLayout.JAVA_INT;
+    public static final ValueLayout C_LONG = IS_WINDOWS ? ValueLayout.JAVA_INT : ValueLayout.JAVA_LONG;
+    public static final ValueLayout.OfLong C_LONG_LONG = ValueLayout.JAVA_LONG;
+    public static final ValueLayout.OfFloat C_FLOAT = ValueLayout.JAVA_FLOAT;
+    public static final ValueLayout.OfDouble C_DOUBLE = ValueLayout.JAVA_DOUBLE;
+    public static final ValueLayout.OfAddress C_POINTER = ValueLayout.ADDRESS;
+
```
sundararajana, line 68: where are these constant used?

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/TypeImpl.java
```diff
@@ -31,12 +31,18 @@
...
 import jdk.incubator.foreign.MemoryLayout;
 import jdk.incubator.jextract.Declaration;
 import jdk.incubator.jextract.Type;
 
+import static jdk.internal.clang.libclang.CLayouts.C_POINTER;
+
 public abstract class TypeImpl implements Type {
 
+    public static final boolean IS_WINDOWS = System.getProperty("os.name").startsWith("Windows");
+
```
sundararajana, line 45: is this used?

src/jdk.incubator.jextract/share/classes/jdk/internal/clang/libclang/CLayouts.java
```diff
@@ -0,0 +1,14 @@
+package jdk.internal.clang.libclang;
+
+import jdk.incubator.foreign.ValueLayout;
+
+public class CLayouts {
```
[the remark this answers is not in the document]
mcimadamore: yes - this can be an interface.

src/jdk.incubator.jextract/share/classes/jdk/internal/clang/libclang/RuntimeHelper.java
mcimadamore: Yep, this is transitional code - jextract will generate these layouts now.

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/ConstantBuilder.java
mcimadamore: This is a string constant - so we don't know which thread will use it. It cannot be confined.

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/OutputFactory.java
mcimadamore: I will need to double check this one.

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/RecordLayoutComputer.java
mcimadamore: good catch

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/ToplevelBuilder.java
mcimadamore: possibly, perhaps we can introduce some disambiguation later?

test/jdk/tools/jextract/JextractToolRunner.java
mcimadamore: in tests which extends this class, I believe

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/TypeImpl.java
mcimadamore: yes, ```Long("long", TypeImpl.IS_WINDOWS ? ValueLayout.JAVA_INT : ValueLayout.JAVA_LONG),```

src/jdk.incubator.jextract/share/classes/jdk/internal/clang/libclang/constants$0.java
mcimadamore: This is just a workaround I did to avoid rewriting all these constants classes (since they were using layouts like C_INT and such). So I defined the `CLayouts` class, and kept the rest identical to what it was before. If we regenerate all the bindings, all these dependencies will go away.

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/OutputFactory.java
mcimadamore: This is caused by a difference in the AST when nested structs are found. Consider this example:
```
struct Foo {
    struct Bar {
        int x, y;
    } bar;
};
```

The AST used to be:
```
Scoped: TOPLEVEL <toplevel>
    Scoped: STRUCT Foo layout = [[b32(x)[abi/kind=INT,layout/name=x]b32(y)[abi/kind=INT,layout/name=y]](bar)[layout/name=bar]](Foo)[layout/name=Foo]
        Scoped: STRUCT Bar layout = [b32(x)[abi/kind=INT,layout/name=x]b32(y)[abi/kind=INT,layout/name=y]](Bar)[layout/name=Bar]
            Variable: FIELD x type = Int(layout = b32[abi/kind=INT]), layout = Optional[b32[abi/kind=INT]]
            Variable: FIELD y type = Int(layout = b32[abi/kind=INT]), layout = Optional[b32[abi/kind=INT]]
        Variable: FIELD bar type = Declared([b32(x)[abi/kind=INT,layout/name=x]b32(y)[abi/kind=INT,layout/name=y]](Bar)[layout/name=Bar]), layout = Optional[[b32(x)[abi/kind=INT,layout/name=x]b32(y)[abi/kind=INT,layout/name=y]](Bar)[layout/name=Bar]]
```
Here you can see that both `Foo` and `Bar` declaration are present in the AST. After these changes the AST is like this:

```
Scoped: TOPLEVEL <toplevel>
    Scoped: STRUCT Foo layout = [[8%b32(x)8%b32(y)](bar)](Foo)
        Variable: FIELD bar type = Declared([8%b32(x)8%b32(y)](Bar)), layout = Optional[[8%b32(x)8%b32(y)](Bar)]
```

That is, the declaration for `Bar` is still there, but it's "hidden" behind the declared type of the struct field `bar`.

Is there some API method I can call to find a struct declared "on the fly" in this way?

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/ToplevelBuilder.java
sundararajana: sounds good

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/TypeImpl.java
sundararajana: okay. missed that.

src/jdk.incubator.jextract/share/classes/jdk/internal/jextract/impl/ConstantBuilder.java
sundararajana: got it

sundararajana (approved):

PR #581, opened 2021-09-24
8274285: By reference parameters should be kept alive during downcalls

mcimadamore:
This patch adds an adapter around the downcall method handle which keeps alive all by-reference arguments by acquiring/releasing their associated scopes.

I've implemented the adapted in a separate method in `SharedUtils`, because `SharedUtils::wrapWithAllocator` is relatively complex, and has mostly to do with creating a scope and closing it after the call.
Also, `wrapWithAllocator` is only called when specialized method handles are active, which means adding logic there doesn't fix the programmable fallback case.

Since we plan to do more implementation restructuring in this area, I preferred to keep things simple for now.

Also note that the acquire/release logic isn't too smart - if multiple arguments backed by the same scoped are passed, we do separate acquire/release for each of them. Again, I didn't want to spend too many cycles given that the implementation of this area might change, so my aim for now was to make things safe; we can improve efficiency later.

Here are some benchmarks, for now, which compare the difference between passing a struct by reference using the struct's address (e.g. calling `MemorySegment::address` before the downcall), vs. passing the struct directly. In the first case, the scope is the global scope, so acquiring/relesing is a no-op. In the latter case we have a real scope, which we need to acquire and release.

```
Benchmark                                                           Mode  Cnt   Score   Error  Units
CallOverheadConstant.panama_identity_memory_address_confined        avgt   30  13.094 ? 0.220  ns/op
CallOverheadConstant.panama_identity_memory_address_confined_3      avgt   30  16.051 ? 0.230  ns/op
CallOverheadConstant.panama_identity_memory_address_shared          avgt   30  13.287 ? 0.172  ns/op
CallOverheadConstant.panama_identity_memory_address_shared_3        avgt   30  15.812 ? 0.365  ns/op

Benchmark
CallOverheadConstant.panama_identity_struct_ref_confined    avgt   30  11.544 ? 0.170  ns/op
CallOverheadConstant.panama_identity_struct_ref_confined_3  avgt   30  12.213 ? 0.228  ns/op
CallOverheadConstant.panama_identity_struct_ref_shared      avgt   30  17.214 ? 0.560  ns/op
CallOverheadConstant.panama_identity_struct_ref_shared_3    avgt   30  33.132 ? 0.934  ns/op
```

As it can be seen, in the confined case the numbers are good. In the shared case there's more overhead - that's the price for safety. We will be able to make this faster (by only acquiring scopes which have not already been acquired) - but note that users can always opt out of safety by projecting the `Addressable` down to a raw memory address, which would remove all the costs.

---------
### Progress

### Issue
 * [JDK-8274285](https://bugs.openjdk.java.net/browse/JDK-8274285): By reference parameters should be kept alive during downcalls

### Reviewers
 * [Jorn Vernee](https://openjdk.java.net/census#jvernee) (@JornVernee - Committer) ⚠️ Review applies to 24ae457c3f3a8f8c3c24b633656ae8537bf216f6

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/581/head:pull/581` \
`$ git checkout pull/581`

`$ git checkout pull/581` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/581/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 581`

`$ git pr show -t 581`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/581.diff">https://git.openjdk.java.net/panama-foreign/pull/581.diff</a>

</details>

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/SharedUtils.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/SharedUtils.java
@@ -41,11 +41,13 @@ import jdk.internal.access.JavaLangInvokeAccess;
 import jdk.internal.access.SharedSecrets;
 import jdk.internal.foreign.CABI;
 import jdk.internal.foreign.MemoryAddressImpl;
+import jdk.internal.foreign.ResourceScopeImpl;
 import jdk.internal.foreign.Utils;
 import jdk.internal.foreign.abi.aarch64.linux.LinuxAArch64Linker;
 import jdk.internal.foreign.abi.aarch64.macos.MacOsAArch64Linker;
 import jdk.internal.foreign.abi.x64.sysv.SysVx64Linker;
 import jdk.internal.foreign.abi.x64.windows.Windowsx64Linker;
+import jdk.internal.vm.annotation.ForceInline;
 
 import java.lang.invoke.MethodHandle;
 import java.lang.invoke.MethodHandles;
@@ -66,6 +68,7 @@ import static java.lang.invoke.MethodHandles.constant;
 import static java.lang.invoke.MethodHandles.dropArguments;
 import static java.lang.invoke.MethodHandles.dropReturn;
 import static java.lang.invoke.MethodHandles.empty;
+import static java.lang.invoke.MethodHandles.filterArguments;
 import static java.lang.invoke.MethodHandles.identity;
 import static java.lang.invoke.MethodHandles.insertArguments;
 import static java.lang.invoke.MethodHandles.permuteArguments;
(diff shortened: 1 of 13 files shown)


src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/SharedUtils.java
```diff
@@ -382,20 +391,6 @@ static MethodHandle wrapWithAllocator(MethodHandle specializedHandle,
 
         closer = collectArguments(closer, insertPos++, MH_CLOSE_CONTEXT); // (Throwable, V?, Addressable?, BindingContext) -> V/void
 
-        if (!upcall) {
```
mcimadamore, line 385: This can go away, because we use acquire/release anyway to keep these args alive in the new adapter/wrapper.

JornVernee (approved):

JornVernee:
Already approved but now thinking of this: I think the javadoc in CLinker needs to be updated to specify the new behavior as well?

mcimadamore:
> Already approved but now thinking of this: I think the javadoc in CLinker needs to be updated to specify the new behavior as well? (it's currently missing for the reachability fences, but I think it would be good to add something...)

Added extra javadoc.

PR #585, opened 2021-09-28
8274157: java.foreign: Add method MemorySegment::asOverlappingSlice

FrauBoes:
This change adds a method to the MemorySegment API that detects if two segments are overlapping.

---------
### Progress

### Issue
 * [JDK-8274157](https://bugs.openjdk.java.net/browse/JDK-8274157): java.foreign: Add method MemorySegment::asOverlappingSlice

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/585/head:pull/585` \
`$ git checkout pull/585`

`$ git checkout pull/585` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/585/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 585`

`$ git pr show -t 585`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/585.diff">https://git.openjdk.java.net/panama-foreign/pull/585.diff</a>

</details>


src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -320,6 +320,20 @@ default MemorySegment asSlice(long offset) {
      */
     boolean isMapped();
 
+    /**
+     * Returns true if this segment overlaps the {@code other} segment.
```
mcimadamore, line 324: We should probably explain what "overlaps" mean. E.g. "Two segments S1 and S2 are said to overlap if it is possible to find at least two slices L1 (from S1) and L2 (from S2) such that L1 and L2 are backed by the same memory region".

This description is vague enough (as it is defined in terms of slicing). At the same time it makes it clear as to why an heap segment can never overlap with an heap segment (and vice-versa).

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -320,6 +320,20 @@ default MemorySegment asSlice(long offset) {
...
+     *
+     * If the two segments are not either both backed by native or heap memory, {@code false} is returned.
+     * Otherwise, their offset and {@linkplain #byteSize() size} (in bytes) are used to determine if an overlap occurs.
+     *
+     * @param other the segment to be tested for an overlap with this segment.
+     * @return {@code true} if this segment overlaps the other segment.
```
mcimadamore, line 330:
```suggestion
     * @return {@code true} if this segment overlaps with the other segment.
```

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -320,6 +320,20 @@ default MemorySegment asSlice(long offset) {
      */
     boolean isMapped();
 
+    /**
+     * Returns true if this segment overlaps the {@code other} segment.
+     *
+     * If the two segments are not either both backed by native or heap memory, {@code false} is returned.
```
mcimadamore, line 326: If you like the above suggestion, I'd replace this para with:

"As such, it is never possible for a native (javadoc link) segment to overlap with an heap segment".

We can drop the sentence about offset and sizes - as that seems more of an impl detail? But perhaps, if you want to explain more, a couple of examples would be better? E.g.

```
MemorySegment s1 = MemorySegment.allocateNative(100, scope);
MemorySegment s2 = MemorySegment.allocateNative(100, scope);
s1.overlap(s2); // false
MemorySegment s3 = s1.slice(0, 50);
s1.overlap(slice); // true
s2.overlap(slice); // false
```

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
```diff
@@ -253,6 +253,30 @@ public boolean isNative() {
         return false;
     }
 
+    @Override
+    public boolean isOverlapping(MemorySegment other) {
```
mcimadamore, line 257: maybe add final, so that other segment subclasses cannot accidentally override?

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
```diff
@@ -253,6 +253,30 @@ public boolean isNative() {
...
+        AbstractMemorySegmentImpl that = (AbstractMemorySegmentImpl)Objects.requireNonNull(other);
+        if (base() == that.base()) {  // both are either native or heap
+            this.checkAccess(0, this.length, true);
```
mcimadamore, line 260: I don't think `checkAccess` is needed? That operation is typically called when accessing a segment to dereference, to verify that:
* the accessed location is within the bounds of the segment
* the access mode (read/write) is compatible with the segment
Here, you don't do dereference, so these checks are not necessary, `checkValidState()` is what you want.

test/jdk/java/foreign/TestSegmentsOverlap.java
```diff
@@ -0,0 +1,112 @@
...
+import org.testng.annotations.Test;
+import org.testng.annotations.DataProvider;
+import static org.testng.Assert.*;
+
+public class TestSegmentsOverlap {
+
+    // native
+
+    @Test
+    public void testNativeBasic() {
```
mcimadamore, line 42: there are two kinds of native segments (or even three): segments created via `MemorySegment::allocateNative` and segments created via `MemorySegment.mapFile`. Perhaps it could be worth testing both (I believe `TestByteBuffer` has an example of a mapped segment).

Additionally, we have native segments which come from a direct buffer:

```
ByteBuffer b = ByteBuffer.allocateDirect(10);
MemorySegment s = MemorySegment.ofBuffer(b); // native segment!
```

mcimadamore (commented):
Good work. As I was going through the changes, it occurred to me that another way to stack this API is to return the slice that is common to both segments (or `null` if none exist). I don't know which version would be more interesting; any comment on this @leerho?

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
leerho: Returning the slice of the overlapped region is an interesting idea.

I was trying to think if this would be useful in computing intersections between A and B, but alas, the circumstances where this would be true are highly specialized. In the general case, if the returned slice is != null, then the lower bound of the size of the intersection is the size of the slice.  But if it is null, we know absolutely nothing.  So for intersections anyway, this is not too useful.  

The use-case I originally had in mind was copying non-primitive elements from A into B.  To do this efficiently, without requiring extra copying space requires that we know if there is an overlap and knowing whether A_lo <= B_lo,  or B_lo < A_lo (by address),  or if A is a subset of B.  Knowing that there exists an overlap is a big help, but to answer the rest we would need to obtain the start and end addresses of both A and B.

In my mind, what would be more useful than returning the overlapping slice, would be to return sufficient information so that an efficient copy from A to B could proceed immediately.  For example, returning a small array of 4 addresses: {A_lo, A_hi, B_lo, B_hi}.  This provides all the info we need for all different types of copies and moves.

Cheers,
Lee.

leerho: If you are reluctant to return actual addresses, you could return the 4 values as relative offsets where either A_lo == 0, or B_lo == 0, or A_lo == B_lo == 0.

I thought I saw in one of your refactorings where you eliminated both MemoryAccess and MemoryCopy, a new general purpose copy(seg A, LayoutA, offsetA, seg B, LayoutB, offsetB, length) or something like that (I can't find it at this moment).  If the layouts can describe repeating elements of any size, doesn't this solve the copy problem and possibly eliminate the need for the isOverlapping() method?

mcimadamore: Not sure I understand correctly - but it sounds almost as if you are interested in knowing the two, maximal, non-overlapping slices of A and B?

It would help to have an example with some numeric bounds, to make sure we're talking about the same things.

mcimadamore: > If you are reluctant to return actual addresses, you could return the 4 values as relative offsets where either A_lo == 0, or B_lo == 0, or A_lo == B_lo == 0.
> 
> I thought I saw in one of your refactorings where you eliminated both MemoryAccess and MemoryCopy, a new general purpose copy(seg A, LayoutA, offsetA, seg B, LayoutB, offsetB, length) or something like that (I can't find it at this moment). If the layouts can describe repeating elements of any size, doesn't this solve the copy problem and possibly eliminate the need for the isOverlapping() method?

The MemorySegment::copy you are referring to takes a pair of ValueLayouts, which cannot be arbitrarily big. In fact, not even the underlying `Unsafe.copyMemorySwap` can work with arbitrary element sizes.

That said, I'm still trying to understand what exactly you want to do; if you want to copy elements of N bytes from a slice to another - why isn't copying bytes enough (since the bulk copy already can deal with overlapping addresses)? Do you want to "swap" entire structs as part of the copy (that would be inefficient!).

mcimadamore: Reading from older emails - I'm starting to have doubts about whether the overlap predicate in this PR does anything for you. You want to protect against the case whether two segments are mapping of _the same file_, but in different address spaces. So there might not even be an address overlap. This is what the spec is calling out, I think: since there's no address overlap, Unsafe doesn't know how to protect against that. But that is a much lower level kind of overlap, which has nothing to do with the one provided here - do you agree? If the address space is the same - e.g. plain segments allocated with `malloc`, what bulk copy does is already enough.

mcimadamore: Ok, I guess I now figured out what you mean; you basically want to implement something that has same correctness guarantees as `System.arraycopy` which works on _arbitrary_ element sizes, not just 8/16/32/64 byte sizes. The copy method we have is good (in the sense that you can also copy by bytes, or by int/long words), but if you also need multi-bit swap (where multi > 64) then you are out of luck (but I'm curious about this use case of extreme swappiness :-) ).

It is possible, in the future, that with the help of the vector API, we might be able to indeed lift the restriction, and provide a bulk copy (written in Java :-) ) which works on elements bigger than 64 bits. In the meantime, we have two options:

* you "take the hit" when dealing with element sizes > 64bit (meaning in these case you will need extra buffer)
* we provide enough primitives which allow you (or others) to implement other copy routines which support similar guarantees as `System.arraycopy`

Another issue: note that the `MemoryAddress::segmentOffset` has also gone in the API refresh, and I believe your plan was probably to rely on it to do the computations; we could add it back if that was really an issue. Or we could come up with similar methods which work on segments instead (e.g. return the start/end offset of a segment into another segment - which combined with an overlapping slice region would probably do what you need to do).

I don't know how common this issue is in your API, so I can't judge. The general feeling I get is that an overlapping predicate without some other ways to compute offsets between slices is not very useful - and so, in that case it would be better to do nothing.

leerho:
I think you understand. But I haven't lost hope, that there aren't some
simple primitives that would allow copying of multiple structs from one
segment to another where a struct could be an arbitrary size in bytes (not
bits).

Imagine these segments are quite large ~ gigabytes.  Forcing me to allocate
a buffer segment of that size and perform redundant copy operations just
because I can't figure out whether these segments overlap in memory and
which segment precedes the other is a huge penalty.

I didn't realize you had already tossed the MemoryAddress:: segmentOffset. This
operation is crucial for this task.

The minimal primitive operations required would be

   - boolean Segment.isOverlapping(Segment); //if segments are in different
   spaces (heap, off-heap) the result is obviously false.
   - long startingAddress(Segment);  //this has to work whether it is
   on-heap or off-heap.

You may consider the underlying copy mechanics to do this to be
inefficient, perhaps, but it would still be orders-of-magnitude more
efficient than having to create a huge segment buffer and do all the
redundant copies.

I will construct a simulated java example using these primitives, just to
be clear. :)

Cheers,
Lee.

mcimadamore:
Hi,
some context: the main reason we dropped `segmentOffset` form `MemoryAddress` is that `MemoryAddress` now has a clearer role, which is more geared towards native interop. In other words, if you don't care about foreign functions, I believe `MemoryAddress` is not for you (unless you want to store the address of an off-heap segment in some other segment).

Now, I believe that the idea of returning the overlapping slice, combined with a way to tell the start offset of a segment into another might work?

```
MemorySegment overlap_a = a.overlap(b);
```

Now, if `overlap` is not null, can't you derive useful parameters from it, and from `a` and `b` ?

e.g.

```
long overlap_start_a = a.offsetOf(overlap_a);
long overlap_end_a = overlap_start_a + overlap_a.byteSize()
```
Then, you can obtain same info for `b`:

```
MemorySegment overlap_b = b.overlap(a);
long overlap_start_b = b.offsetOf(overlap_b);
long overlap_end_b = overlap_start_b + overlap_b.byteSize()
```
This is (more or less) the info you are looking for, right?

leerho:
Yes, I think this should work.

mcimadamore:
> Yes, I think this should work.

Thanks for confirming. Btw is it the case that you want to perform bulk copies with byte-swapping for arbitrary sizes? E.g. you want to swap bytes for entire structs? I guess I'm not super sure of what is your real use case (e.g. the reason as to why just copying bytes from one segment to another is not enough). For instance, if you want to bulk copy arbitrary structs, adjusting for endianness is not simple at all - as you would have to swap each field individually.

leerho:
Doing byte swapping (i.e, adjusting for endianness) is not required.   This
is for copying structs from one segment to another where all the structs
are the same, but arbitrary size.  If we have to restrict these sizes to
multiples of 2,4, or 8 bytes that would be OK.  But allowing these structs
to be larger than 8 bytes is critical, otherwise we would use existing copy
methods.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -321,18 +321,33 @@ default MemorySegment asSlice(long offset) {
...
+     * {@link #isNative() native} segment to overlap with a heap segment; in
+     * this case, or when no overlap occurs, {@code null} is returned.
      *
-     * @param other the segment to be tested for an overlap with this segment.
-     * @return {@code true} if this segment overlaps the other segment.
-     * @throws IllegalStateException if either the scope associated with this segment or the scope associated
-     * with the {@code other} segment have been already closed, or if access occurs from a thread other than the thread
-     * owning either scopes.
+     * @param other the segment to test for an overlap with this segment.
+     * @return a new memory segment, or {@code null} if no overlap occurs.
```
mcimadamore, line 333: Should we say a "new slice of this segment" instead of "a new segment" (both here an in the first sentence of the javadoc) ?

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -321,18 +321,33 @@ default MemorySegment asSlice(long offset) {
...
+     * {@linkplain #byteSize() size} (in bytes) of this segment (exclusive). If
+     * the two segments do not {@link #overlap(MemorySegment) overlap}, an
+     * offset can not be retrieved and {@code -1L} is returned.
+     *
+     * @param other the segment to retrieve an offset to.
+     * @return the relative offset, in bytes, from this segment to the other
+     *         segment, or -1L if no overlap occurs.
      */
-    boolean isOverlapping(MemorySegment other);
+    long offsetOf(MemorySegment other);
```
mcimadamore, line 350: I think what I had in mind was slightly different:
* return postive, or negative offsets in case of overlaps (of course in the case of sub-slices, offsets will always be positive)
* throw if asked to compute an offset from heap to non-heap (or vice-versa).

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
```diff
@@ -252,27 +252,34 @@ public boolean isNative() {
...
-        if (base() == that.base()) {  // both are either native or heap
-            this.checkAccess(0, this.length, true);
-            that.checkAccess(0, that.length, true);
-            this.checkValidState();
-            that.checkValidState();
+        long offsetFromThis = this.offsetOf(that);
+        if (offsetFromThis >= 0L) {  // this and that overlap
+            long offsetFromThat = that.offsetOf(this);
+            long newSize = Math.min(this.byteSize() - offsetFromThis, that.byteSize() - offsetFromThat);
+            return dup(offsetFromThis, newSize, mask, scope);
```
mcimadamore, line 261: nitpick: you can probably use asSlice instead of dup?

test/jdk/java/foreign/TestSegmentOffsetOf.java
```diff
@@ -0,0 +1,127 @@
+/*
```
mcimadamore, line 1: We used to have a test for MemoryAddress.segmentOffset...
https://github.com/openjdk/jdk/blob/master/test/jdk/java/foreign/TestRebase.java
I wonder if that old test can be put to new use?

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -321,7 +321,8 @@ default MemorySegment asSlice(long offset) {
     boolean isMapped();
 
     /**
-     * Returns a new memory segment that is the overlap of the two segments.
+     * Returns a new slice of this segment that is the overlap of the two
```
mcimadamore, line 324:
```suggestion
     * Returns a slice of this segment that is the overlap between this and the provided segment.
```

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -330,24 +331,22 @@ default MemorySegment asSlice(long offset) {
...
-     * @return a new memory segment, or {@code null} if no overlap occurs.
+     * @return a new slice of this segment, or {@code null} if no overlap occurs.
      */
-    MemorySegment overlap(MemorySegment other);
+    MemorySegment asOverlappingSlice(MemorySegment other);
 
     /**
      * Returns the offset, in bytes, from this segment to the other segment.
      *
      * <p>The offset is relative to the {@linkplain #address() base address} of
```
mcimadamore, line 341: Note that MemorySegment::address is only defined for native segments. Perhaps remove the link?

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -330,24 +331,22 @@ default MemorySegment asSlice(long offset) {
...
+     * this segment and can be a negative or positive value. If the segments
+     * share the same base address, {@code 0} is returned.
      *
      * @param other the segment to retrieve an offset to.
      * @return the relative offset, in bytes, from this segment to the other
-     *         segment, or -1L if no overlap occurs.
+     *         segment.
      */
-    long offsetOf(MemorySegment other);
+    long offsetTo(MemorySegment other);
```
mcimadamore, line 349: Not too sure about the name :-) I think overall I prefer `offsetOf` - although the main issue is to find a name which is clear about the direction. After looking at the test, I'm not sure `segmentOffset` would look bad?

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -330,24 +331,22 @@ default MemorySegment asSlice(long offset) {
...
      *
      * @param other the segment to test for an overlap with this segment.
-     * @return a new memory segment, or {@code null} if no overlap occurs.
+     * @return a new slice of this segment, or {@code null} if no overlap occurs.
      */
-    MemorySegment overlap(MemorySegment other);
+    MemorySegment asOverlappingSlice(MemorySegment other);
 
     /**
      * Returns the offset, in bytes, from this segment to the other segment.
```
mcimadamore, line 339:
```suggestion
     * Returns the offset, in bytes, of the provided segment, relative to this segment.
```

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
```diff
@@ -252,34 +252,30 @@ public boolean isNative() {
...
-    }
-
-    @Override
-    public final long offsetOf(MemorySegment other) {
-        AbstractMemorySegmentImpl that = (AbstractMemorySegmentImpl) Objects.requireNonNull(other);
-        if (base() == that.base()) {  // both are either native or heap
+        if (base() == that.base()) {  // both either native or heap
             final long thisStart = this.min();
             final long thatStart = that.min();
             final long thisEnd = thisStart + this.byteSize() - 1L;
```
mcimadamore, line 260: do you need the `-1L` (since then you need `<=`) ? E.g. can we drop the `- 1L` and use strict comparisons instead?

mcimadamore (commented):
Getting there - some nitpicks around docs and naming.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
FrauBoes: Yup, that's better!

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -331,22 +331,22 @@ default MemorySegment asSlice(long offset) {
...
 
     /**
-     * Returns the offset, in bytes, from this segment to the other segment.
+     * Returns the offset, in bytes, of the provided segment, relative to this
+     * segment.
      *
-     * <p>The offset is relative to the {@linkplain #address() base address} of
-     * this segment and can be a negative or positive value. If the segments
-     * share the same base address, {@code 0} is returned.
+     * <p>The offset is relative to the base address of this segment and can be
```
mcimadamore, line 342: Perhaps we could add something like:
```For instance, if both segments are native segments, the resulting offset can be computed as follows:
other.baseAddress().toRawLongValue() - segment.baseAddress().toRawLongValue()
```
So that it hints at how the method fully operates. Then you can conclude with the sentence which says we return `0` if base is the same. I'd also add another sentence to say that if `other` is a slice of this segment, then the offset is always `0 <= x <= byteSize()`.

mcimadamore (approved):
Looks good - added an optional javadoc clarification.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
```diff
@@ -252,34 +252,30 @@ public boolean isNative() {
...
             final long thatEnd = thatStart + that.byteSize() - 1L;
 
-            if (thisStart <= thatStart && thisEnd >= thatStart) {
-                return thatStart - thisStart;
-            }
-            if (thatStart <= thisStart && thatEnd >= thisStart) {
-                return 0L;
+            if (thisStart <= thatEnd && thisEnd >= thatStart) {  //overlap occurs
```
rsmogura: Hi,

Minor, I wonder if it would make sense to replace "weak" inequality with strong inequality and remove subtraction of 1 (two instructs less)?

mcimadamore: I believe that has been addresses in the latest revision?

FrauBoes: Good point, it was addressed in the latest commit b62df7e.

mcimadamore (approved):
Looks great!

PR #586, opened 2021-09-28
Make Addressable sealed again

mcimadamore:
Following the API refresh, I realized that making `Addressable` non-sealed was perhaps a step too far, at least for the time being. It forces a `scope()` accessor on the interface, which in turns pollutes the `MemoryAddress` interface (since a memory address has no scope).
While the API, as currently defined is sound, I'd prefer to make things tighter for now, as I realized that the need to define custom `Addressable` is perhaps limited. For instance, one place I thought they would have been useful was to model structs - but thinking more, a struct is probably backed by a segment anyway - and the segment is addressable. So, assuming the struct exposes a segment accessor, we still have ways to pass it around w/o necessarily having it implement the `Addressable` interface directly.

In the implementation we do have a new internal `Scoped` interface, which is used to keep all by-reference parameters alive by the linker. Basically, this interface re-adds the scope accessor which has been removed from the public API.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/586/head:pull/586` \
`$ git checkout pull/586`

`$ git checkout pull/586` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/586/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 586`

`$ git pr show -t 586`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/586.diff">https://git.openjdk.java.net/panama-foreign/pull/586.diff</a>

</details>

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/Addressable.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/Addressable.java
@@ -30,13 +30,6 @@ package jdk.incubator.foreign;
  * a {@linkplain #address() memory address}. Examples of addressable types are {@link MemorySegment},
  * {@link MemoryAddress}, {@link VaList} and {@link CLinker.UpcallStub}.
  * <p>
- * An addressable instance is associated with a {@linkplain ResourceScope resource scope}; the resource scope determines the
- * lifecycle of the addressable instance, as well as whether the instance can be used from multiple threads. Some
- * addressable instances might not be associated with a lifecycle, in which case {@link #scope()} returns the
- * {@linkplain ResourceScope#globalScope() global scope}. Attempting to obtain a memory address
- * from an addressable instance whose backing scope has already been {@linkplain ResourceScope#isAlive() closed} always
- * results in an exception.
- * <p>
  * The {@link Addressable} type is used by the {@link CLinker C linker} to model the types of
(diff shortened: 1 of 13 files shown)


src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/Addressable.java
```diff
@@ -30,33 +30,18 @@
...
  * Implementations of this interface are <a href="{@docRoot}/java.base/java/lang/doc-files/ValueBased.html">value-based</a>.
  */
-public interface Addressable {
+public sealed interface Addressable permits MemorySegment, MemoryAddress, CLinker.UpcallStub, VaList {
 
     /**
      * Returns the memory address associated with this addressable.
      * @return The memory address associated with this addressable.
-     * @throws IllegalStateException if the scope associated with this addressable has been closed, or if access occurs from
-     * a thread other than the thread owning that scope.
```
JornVernee, line 53: Assuming that this method can still throw an exception for the reason described here, is it okay for API methods to throw unspecified exceptions?

JornVernee (approved):

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/Addressable.java
mcimadamore: for runtime exception, I think that's fine. That's kind of the case with `UnsupportedOperationException`. We might add some text to the documentation, but in general I think that's within bounds.

PR #587, opened 2021-09-29
TestScopedOperations does not work as expected

mcimadamore:
When Julia started to work on the `MemorySegment::overlap` API, she noticed (thanks!) an issue with the TestScopedOperation. Essentially this test checks that calling methods on various objects should be scope-safe - that is, it shouldn't be possible to call certain methods (e.g. MemorySegment::address) if the scope associated with the object is closed, or from a thread other than the owning thread.

The test had a fatal flaw, which was probably introduced when we moved to the `ResourceScope` API. Since the scoped operation creates the desired object (e.g. a segment) and then performs an operation on it, and since that operation is performed _after_ the scope has been closed, as a result, all tests trivially fail as expected, since the object cannot be created in the first place (the scope is closed!!).

The solution is to separate the logic for creating the object from the logic for testing the scoped operation. In fixing the test I realized that the test was running two combinations which didn't make sense (prefixAllocator::allocate, and segment::spliterator, neither of which is scoped), and found a bug in the VaList implementation - for VaList::skip, which was missing a scope check (which this patch also addresses).

---------
### Progress

### Reviewers
 * [Jorn Vernee](https://openjdk.java.net/census#jvernee) (@JornVernee - Committer) ⚠️ Review applies to 2e0e4ea0f3d41763734dde440660376cf2e3dc60

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/587/head:pull/587` \
`$ git checkout pull/587`

`$ git checkout pull/587` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/587/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 587`

`$ git pr show -t 587`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/587.diff">https://git.openjdk.java.net/panama-foreign/pull/587.diff</a>

</details>

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/SegmentAllocator.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/SegmentAllocator.java
@@ -26,6 +26,7 @@
 package jdk.incubator.foreign;
 
 import jdk.internal.foreign.ArenaAllocator;
+import jdk.internal.foreign.ResourceScopeImpl;
 import jdk.internal.foreign.Utils;
 
 import java.lang.invoke.VarHandle;
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/aarch64/linux/LinuxAArch64VaList.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/aarch64/linux/LinuxAArch64VaList.java
@@ -26,6 +26,7 @@
 package jdk.internal.foreign.abi.aarch64.linux;
 
 import jdk.incubator.foreign.*;
+import jdk.internal.foreign.ResourceScopeImpl;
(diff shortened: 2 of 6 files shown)


FrauBoes:
Looks good, and you found a bug along the way, nice.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/SegmentAllocator.java
```diff
@@ -26,6 +26,7 @@
 package jdk.incubator.foreign;
 
 import jdk.internal.foreign.ArenaAllocator;
+import jdk.internal.foreign.ResourceScopeImpl;
```
JornVernee, line 29: Spurious import?

JornVernee (approved):
Looks like you have some merge conflicts as well

PR #589, opened 2021-09-30
8274602: Generalize UpcallStub into NativeSymbol

mcimadamore:
It might sometimes be useful to define a custom symbol lookup which returns symbols backed by a library that can be unloaded dynamically (e.g. using dlopen).
We indeed have all the pieces to get there, except that symbnol lookup returns a memory address, which is not a scoped entity.

To address this, I propose that we add a new sealed interface/final class (`NativeSymbol`) which implements `Addressable` and is used in two places:
* as a return value for a symbol lookup
* as a proxy for an upcall stub

In other words, `NativeSymbol` is used to reference to some symbol in some library, where the library can have a given scope. Upcall stubs are just a special case where the upcall symbol is synthetized on the fly by the VM. With this in place, I think we can drop `CLinker.UpcallStub` and just use an "anonymous` NativeSymbol to represent upcall stubs.

With this, it should be possible to define a dlopen-based lookup such as the one below:
https://github.com/sundararajana/panama-jextract-samples/blob/master/dlopen/Dlopen.java

In a way that is completely *safe* - that is, so that it is not possible for a library to be unloaded while a function is being executed (even when the scope  associated to the native symbol is shared).

---------
### Progress

### Issue
 * [JDK-8274602](https://bugs.openjdk.java.net/browse/JDK-8274602): Generalize UpcallStub into NativeSymbol

### Reviewers
 * [Athijegannathan Sundararajan](https://openjdk.java.net/census#sundar) (@sundararajana - Committer) ⚠️ Review applies to 123fb4abae71f448a2323142ce9e7cef64471198

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/589/head:pull/589` \
`$ git checkout pull/589`

`$ git checkout pull/589` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/589/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 589`

`$ git pr show -t 589`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/589.diff">https://git.openjdk.java.net/panama-foreign/pull/589.diff</a>

</details>


src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/NativeSymbolImpl.java
```diff
@@ -0,0 +1,13 @@
+package jdk.internal.foreign;
```
JornVernee, line 1: Missing copyright header

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/NativeSymbol.java
```diff
@@ -0,0 +1,60 @@
+package jdk.incubator.foreign;
```
JornVernee, line 1: missing copyright header

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/NativeSymbol.java
```diff
@@ -0,0 +1,60 @@
...
+    @Override
+    MemoryAddress address();
+
+    /**
+     * Creates a new symbol from given name, address and scope.
+     * <p>
+     * This method is <a href="package-summary.html#restricted"><em>restricted</em></a>.
+     * Restricted methods are unsafe, and, if used incorrectly, their use might crash
+     * the JVM or, worse, silently result in memory corruption. Thus, clients should refrain from depending on
+     * restricted methods, and use safe and supported functionalities, where possible.
```
JornVernee, line 46: Why is this restricted?

mcimadamore: because you can create a native symbol out of an arbitrary address (e.g. MemoryAddress.ofLong), which then, if passed to a downcall method handle, will cause havocs. That said, true that, at the moment you can pass Addressable to a downcall method handle - but maybe that's the issue? E.g. downcall method handle needs to take in a NativeSymbol? That way, all native symbols created by our APIs are known to be safe (upcall and symbol lookup) - but the user is still free to create custom symbols - although these latter would be restricted. Makes sense?

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/UpcallStubs.java
```diff
@@ -51,22 +48,13 @@ private static void freeUpcallStub(long stubAddress) {
...
-            return MemoryAddress.ofLong(entry);
-        }
+    static NativeSymbol makeUpcall(long entry, ResourceScope scope) {
+        ((ResourceScopeImpl)scope).addOrCleanupIfFail(new ResourceScopeImpl.ResourceList.ResourceCleanup() {
+            @Override
+            public void cleanup() {
+                freeUpcallStub(entry);
+            }
+        });
+        return new NativeSymbolImpl("", MemoryAddress.ofLong(entry), scope);
```
sundararajana, line 58: Could we use some symbol name for debugging purpose?

sundararajana (approved):

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/NativeSymbol.java
JornVernee: That makes sense, but we are already covered there I think because creating downcall handles is a restricted operation.

Compare it to `MemoryAddress.ofLong` which is also non-restricted: it's fine to make non-dereferencable addresses, because in order to dereference them, we need to go through a restricted API. This case seems kind of similar.

mcimadamore:
I've tweaked the API so that downcallHandle now accepts a `NativeSymbol` - which is what I should have done all along. Now all symbols generated by the API are safe - but the ones constructed by users are not (as they are backed by "random" addresses) - so I think it's fair to mark the symbol factory as restricted.

JornVernee (approved):

JornVernee:
Latest changes look very good! Thanks.

PR #590, opened 2021-10-01
8274648: Improve logic for acquiring by reference parameters in downcall handles

mcimadamore:
The current logic for acquiring/releasing scopes associated with by-reference parameters in a downcall handle call is a bit naive, in the sense that it acquires/releases each parameter in isolation (possibly redundantly).

I've been working on a way to ameliorate the situation by adapting the downcall method handle so that we can pass all addressable parameters (in bulk) to a single acquire/release function. Since this function sees all arguments at once, it can decide to only acquire the unique scopes.

How to do that proved to be challenging - as simply using a for loop didn't give the performance I expected. Turns out that as soon as we take a backward branch, there is a regression compared with the base case, even when all scopes are the same. The only case where we gain is when all scopes are shared.

But I (re)discovered a trick; we can special case the acquire/release functions so that they only use a loop for a high number of addressable parameters, and use a switch (with fallthrough) instead for low counts. This makes the logic more scrutable and the performance numbers are stricty better than what we had:

```
BEFORE

Benchmark
CallOverheadConstant.panama_identity_struct_ref_confined    avgt   30  11.544 ? 0.170  ns/op
CallOverheadConstant.panama_identity_struct_ref_confined_3  avgt   30  12.213 ? 0.228  ns/op
CallOverheadConstant.panama_identity_struct_ref_shared      avgt   30  17.214 ? 0.560  ns/op
CallOverheadConstant.panama_identity_struct_ref_shared_3    avgt   30  33.132 ? 0.934  ns/op

AFTER

Benchmark                                                   Mode  Cnt   Score   Error  Units
CallOverheadConstant.panama_identity_struct_ref_confined    avgt   30  11.613 ? 0.326  ns/op
CallOverheadConstant.panama_identity_struct_ref_confined_3  avgt   30  11.646 ? 0.333  ns/op
CallOverheadConstant.panama_identity_struct_ref_shared      avgt   30  17.763 ? 0.639  ns/op
CallOverheadConstant.panama_identity_struct_ref_shared_3    avgt   30  17.087 ? 0.475  ns/op
```

As you can see, in both cases there is no cost for passing multiple addressable arguments backed by the same scope.

---------
### Progress

### Issue
 * [JDK-8274648](https://bugs.openjdk.java.net/browse/JDK-8274648): Improve logic for acquiring by reference parameters in downcall handles

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/590/head:pull/590` \
`$ git checkout pull/590`

`$ git checkout pull/590` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/590/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 590`

`$ git pr show -t 590`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/590.diff">https://git.openjdk.java.net/panama-foreign/pull/590.diff</a>

</details>


src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/SharedUtils.java
```diff
@@ -409,14 +411,89 @@ static MethodHandle wrapWithAllocator(MethodHandle specializedHandle,
...
+            case 1:
+                scope0 = args[0].scope();
+                if (scope0 != scope1 && scope0 != scope2 && scope0 != scope3 && scope0 != scope4)
+                    acquire(scope0);
+            case 0: break;
+            // slow path, acquire all remaining addressable parameters in isolation
+            default:
+                for (int i = 5 ; i < args.length ; i++) {
+                    acquire(args[i].scope());
+                }
```
JornVernee, line 447: I think this will skip acquiring the first `5` scopes right? e.g. if `length` is 6, the switch jumps immediately into the `default` case. In other words, I think the for loop should start at `0`.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/SharedUtils.java
```diff
@@ -409,14 +411,89 @@ static MethodHandle wrapWithAllocator(MethodHandle specializedHandle,
...
+            case 1:
+                scope0 = args[0].scope();
+                if (scope0 != scope1 && scope0 != scope2 && scope0 != scope3 && scope0 != scope4)
+                    release(scope0);
+            case 0: break;
+            // slow path, release all remaining addressable parameters in isolation
+            default:
+                for (int i = 5 ; i < args.length ; i++) {
+                    release(args[i].scope());
+                }
```
JornVernee, line 485: Same here.

JornVernee (approved):

PR #591, opened 2021-10-05
API refresh - part two

mcimadamore:
After playing a bit more with the new API changes, I realized there were issues with some of the solutions proposed in [1].
I will add more details in a separate comment, to keep the header of this PR compact.

[1] - https://git.openjdk.java.net/panama-foreign/pull/576


src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemoryHandles.java
```diff
@@ -178,7 +178,7 @@ public static VarHandle varHandle(ValueLayout layout) {
      * Java {@code int} to avoid dealing with negative values, which would be
      * the case if modeled as a Java {@code short}. This is illustrated in the following example:
      * <blockquote><pre>{@code
-    MemorySegment segment = MemorySegment.allocateNative(2, ResourceScope.newConfinedScope());
+    MemorySegment segment = MemorySegment.allocateNative(2, ResourceScope.newImplicitScope());
```
mcimadamore, line 181: Most changes in this patch are like this one: since `newConfinedScope()` is no longer automatically assigned a cleaner, we have to revert those changes back to what they were previously, by using an implicit scope instead. The difference compared to 17 is that the implicit scope is just a shared scope associated with a cleaner - it can be closed. There's no such concept of "special scope that can only be closed implicitly".

mcimadamore:
This is is an example where having `ResourceScope <: SegmentAllocator` is a bit problematic:

```
try (ResourceScope scope = ResourceScope.newConfinedScope()) {
   SegmentAllocator allocator = SegmentAllocator.newArena(scope);
   MemorySegment segment = allocator.allocate(10); // (2)
}
```
Note that, in (2) the client has two allocators available: the scope and the arena - even though the code is probably meant to use the arena from that point on.

PaulSandoz (approved):
We went round and round trying to lump SA and RS together but it was a bit like magnets repelling each other. But, i think we have come away with a better appreciation of the patterns where SA should be used. (And keeping with the separation will have a good effect on how we can better make scopes be dependent on other scopes.)

PR #593, opened 2021-10-07
8274912: Eagerly generate native invokers

JornVernee:
Hi,

This patch is a refactoring of VM code pertaining to downcalls. As outlined in the JBS issue, this patch makes 'native invoker' generation currently done by C2 eager.

Most of the changes in this patch are moving around existing code:
- The native invoker generation code is moved from sharedRuntime*.cpp to universalNativeInvoker*.cpp, to match what is done for upcalls.
- C2 & nmethod scaffolding for registering native invokers has been removed (lifetime is now managed in Java code).
- Now, C2 will only intrinsify trivial downcalls, which means some of the code was simplified there.
- To reduce code duplication, several utilities have been added:
  - ArgumentShuffle (and impls of CallConvClosure): a utility class for computing argument shuffles and generating shuffling code. The shuffling code is heavily based on what is done today for optimized upcalls on x64.
  - To support the above, I've made a copy of the ComputeMoveOrder utility from SharedRuntime code (which was only implemented on x64), and adapted it to our use case.
  - RegSpillFill: similarly, a utility for computing and generating register spill and fill code (given a list of registers).
  - oop_cast: this was a pre-existing utility that was renamed (from 'cast') and moved into a separate file.
- The linkToNative method handle stub now just loads the invoker address from the appendix argument and jumps to it, instead of jumping to the fallback method handle.

The implementation passes all the jdk_foreign tests on Windows/x64, Linux/x64 (WSL), and Linux/AArch64. Though, there were some pre-exsiting test failures on AArch64 wrt variadic functions.

Thanks,
Jorn

---------
### Progress

### Issue
 * [JDK-8274912](https://bugs.openjdk.java.net/browse/JDK-8274912): Eagerly generate native invokers

### Reviewers
 * [Nick Gasson](https://openjdk.java.net/census#ngasson) (@nick-arm - no project role) ⚠️ Review applies to 493165e413d0ebec161233b9d0a74604e5b0db00

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/593/head:pull/593` \
`$ git checkout pull/593`

`$ git checkout pull/593` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/593/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 593`

`$ git pr show -t 593`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/593.diff">https://git.openjdk.java.net/panama-foreign/pull/593.diff</a>

</details>


src/hotspot/cpu/x86/macroAssembler_x86.cpp
```diff
@@ -938,12 +938,13 @@ void MacroAssembler::long_move(VMRegPair src, VMRegPair dst) {
...
     } else {
-      assert(dst.is_single_reg(), "not a stack pair");
+      assert(dst.is_single_reg(), "not a stack pair: (%s, %s), (%s, %s)",
+       src.first()->name(), src.second()->name(), dst.first()->name(), dst.second()->name());
       movq(Address(rsp, reg2offset_out(dst.first())), src.first()->as_Register());
     }
   } else if (dst.is_single_phys_reg()) {
     assert(src.is_single_reg(),  "not a stack pair");
-    movq(dst.first()->as_Register(), Address(rbp, reg2offset_out(src.first())));
+    movq(dst.first()->as_Register(), Address(rbp, reg2offset_in(src.first())));
```
JornVernee, line 947: A pre-existing bug found when testing (same below). Loads from `rbp` should always be paired with `reg2offset_in`

src/hotspot/cpu/arm/universalNativeInvoker_arm.cpp
```diff
@@ -29,3 +29,17 @@ address ProgrammableInvoker::generate_adapter(jobject jabi, jobject jlayout) {
   Unimplemented();
   return nullptr;
 }
+
+RuntimeStub* ProgrammableInvoker::make_native_invoker(BasicType* signature,
```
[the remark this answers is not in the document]
JornVernee: Well, I guess in that case each platform would still need to define a factory method for the platform specific impl, so that platform agnostic code can reach it, and that would then need to be stubbed on unsupported platforms.

But, either way, I think it's better to stick with the conventions used in other VM code, which is to stub things out.

nick-arm:
> @nick-arm if you want to, could you take a look at the AArch64 bits?

Sure, will do.

src/hotspot/cpu/aarch64/universalNativeInvoker_aarch64.cpp
```diff
@@ -125,3 +129,254 @@ address ProgrammableInvoker::generate_adapter(jobject jabi, jobject jlayout) {
...
+  Register scratch2 = r10;
+  assert(!target_uses_register(scratch1->as_VMReg()), "conflict");
+  assert(!target_uses_register(scratch2->as_VMReg()), "conflict");
+  assert(!target_uses_register(rthread->as_VMReg()), "conflict");
+
+  enum layout {
+    rbp_off,
+    rbp_off2,
+    return_off,
+    return_off2,
```
nick-arm, line 231: It's a bit weird that we use the x86 register names here (although sharedRuntime_aarch64.cpp has this already). `rfp_off` and `lr_off` would be more accurate.

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
```diff
@@ -292,20 +293,30 @@ address MacroAssembler::target_addr_for_insn(address insn_addr, unsigned insn) {
   return address(((uint64_t)insn_addr + (offset << 2)));
 }
 
-void MacroAssembler::safepoint_poll(Label& slow_path, bool at_return, bool acquire, bool in_nmethod) {
+void MacroAssembler::safepoint_poll(Label& slow_path, bool at_return, bool acquire, bool in_nmethod, Register scratch) {
```
nick-arm, line 296: Perhaps rename the `scratch` argument to `tmp` here and below as that's what most of the other MacroAssembler functions are using (and it avoids the visual similarity between `scratch` and `rscratch1`).

mcimadamore (approved):
Looks very good.

PR #594, opened 2021-10-07
API refresh - part two (jextract edition)

mcimadamore:
This patch fixes jextract to work with the latest API changes.

The patch reintroduces overloads for foreign function wrappers which take a scope (since ResourceScope is no longer a subtype of SegmentAllocator).

We will consider later whether to drop these overloads; it is likely that the majority of jextract clients will already have an allocator somewhere, so the extra code is not necessary. On the other hand, if we want overloads, then I think it would also be useful to add an overload that takes a `MemorySegment` and converts it into a `prefixAllocator` (so that clients can store result of foreign function which returns a struct-by-value in a pre-existing segments).

I've made some improvements to the string conversion logic; in a lot of places we were creating a scope, then doing a malloc to allocate a temporary CXString, then extract a Java String from the CXString, and dispose/free the CXString. Since jextract is single-threaded, I replaced all that with an high-order function which takes a segment supplier, and uses a prefix allocator under the hood. This should be way more efficient than what we had - and string conversion is quite frequent in libclang (e.g. `CXCursor::spelling`).

The patch passes all tests.

---------
### Progress

### Reviewers
 * [Jorn Vernee](https://openjdk.java.net/census#jvernee) (@JornVernee - Committer) ⚠️ Review applies to fff0fc6cc27b24f9ec666daadc540df43c8f8d46

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/594/head:pull/594` \
`$ git checkout pull/594`

`$ git checkout pull/594` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/594/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 594`

`$ git pr show -t 594`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/594.diff">https://git.openjdk.java.net/panama-foreign/pull/594.diff</a>

</details>


src/jdk.incubator.jextract/share/classes/jdk/internal/clang/LibClang.java
```diff
@@ -71,16 +77,24 @@ public static Index createIndex(boolean local) {
...
         return str;
     }
 
+    /**
+     * This is an allocator for temporary CXString structs. CXStrToString needs to save the CXString somewhere,
+     * so that we can extract a Java string out of it. Once that's done, we can dispose the CXString, and the
+     * associated segment. Since jextract is single-threaded, we can use a prefix allocator, to speed up string
+     * conversion. The size of the prefix segment is set to 256, which should be enough to hold a CXString.
+     */
+    private final static SegmentAllocator stringAllocator = SegmentAllocator.prefixAllocator(
```
PaulSandoz, line 94:
```suggestion
    private final static SegmentAllocator STRING_ALLOCATOR = SegmentAllocator.prefixAllocator(
```

src/jdk.incubator.jextract/share/classes/jdk/internal/clang/LibClang.java
```diff
@@ -44,8 +47,11 @@
     private static final boolean CRASH_RECOVERY = Boolean.getBoolean("libclang.crash_recovery");
     private static final boolean IS_WINDOWS = System.getProperty("os.name").startsWith("Windows");
 
+    final static SegmentAllocator IMPLICIT_ALLOCATOR =
```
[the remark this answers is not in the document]
mcimadamore: nice suggestion

src/jdk.incubator.jextract/share/classes/jdk/internal/clang/LibClang.java
```diff
@@ -71,16 +77,24 @@ public static Index createIndex(boolean local) {
         return index;
     }
 
-    public static String CXStrToString(MemorySegment cxstr) {
+    public static String CXStrToString(Function<SegmentAllocator, MemorySegment> segmentSupplier) {
+        MemorySegment cxstr = segmentSupplier.apply(stringAllocator);
```
JornVernee, line 81: I don't see `stringAllocator` anywhere, I guess this should be `STRING_ALLOCATOR`.
```suggestion
        MemorySegment cxstr = segmentSupplier.apply(STRING_ALLOCATOR);
```

PR #596, opened 2021-10-11
8275054: Racy updates when releasing confined scopes

mcimadamore:
As the subject says, there are cases where a close dependency on a confined parent scope can occur from an unrelated thread:

1. the confined scope is acquired by another confined scope featuring implicit cleanup - in which case the cleaner thread will do the release

2. the confined scope is acquired, in the same thread, by a shared scope. Since a shared scope can be closed by any other thread, it follows that release can also be done by any thread.

In such cases we can have races (either between release/release or between acquire/release) and the `lockCount` variable of the updated scope becomes garbage.

The solution is to add an extra variable `asyncReleaseCount`. If a release request occur from the same thread that owns the confined scope, we proceed as usual, otherwise we atomically increment the `asyncReleaseCount`. When the confined scope is later closed, if `lockCount == 0`, nothing needs to be done, so we can just close. Otherwise, we need to check that `lockCount - asyncReleaseCount == 0`. This works because a confined segment can only be acquired by the same thread owning it. So, if `lockCount == 0` there can be no other pending release.

---------
### Progress

### Issue
 * [JDK-8275054](https://bugs.openjdk.java.net/browse/JDK-8275054): Racy updates when releasing confined scopes

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/596/head:pull/596` \
`$ git checkout pull/596`

`$ git checkout pull/596` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/596/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 596`

`$ git pr show -t 596`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/596.diff">https://git.openjdk.java.net/panama-foreign/pull/596.diff</a>

</details>

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/ConfinedScope.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/ConfinedScope.java
@@ -90,9 +90,9 @@ final class ConfinedScope extends ResourceScopeImpl {
         if (Thread.currentThread() == owner) {
             lockCount--;
         } else {
-            // It is possible to end up here in two cases: this scope was the ancestor of some other confined scope
+            // It is possible to end up here in two cases: this scope was kept alive by some other confined scope
             // which is implicitly released (in which case the release call comes from the cleaner thread). Or,
-            // this scope might be the ancestor of a shared scope, which means the release call can come from any
+            // this scope might be kept alive by a shared scope, which means the release call can come from any
             // thread.
             int value;
             do {


mcimadamore:
Another solution would have been to trigger an exception when setting up a dependency from a shared scope to a confined scope - but we still need the same logic for implicitly closeable scopes anyway.

PR #601, opened 2021-10-19
8275332: Variadic functions don't work on Linux/AArch64

nick-arm:
Variable length argument lists are handled differently in the
Mac/Windows and Linux ABIs on AArch64.  Following the recent API refresh
on the foreign-memaccess+abi branch the Mac behaviour was inadvertently
applied on Linux too.  This patch restores the correct behaviour on
Linux and adds a CallArranger unit test so regressions can be more
easily caught.

I just added an extra varArgsOnStack argument to
CallArranger.getBindings() as that seemed the simplest fix, although it
could get unwieldy if there were too many configurable options.

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/aarch64/CallArranger.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/aarch64/CallArranger.java
@@ -96,7 +96,7 @@ public class CallArranger {
         }
     }
 
-    public static Bindings getBindings(MethodType mt, FunctionDescriptor cDesc, boolean forUpcall) {
+    public static Bindings getBindings(MethodType mt, FunctionDescriptor cDesc, boolean forUpcall, boolean varArgsOnStack) {
         CallingSequenceBuilder csb = new CallingSequenceBuilder(forUpcall);
 
         BindingCalculator argCalc = forUpcall ? new BoxBindingCalculator(true) : new UnboxBindingCalculator(true);
@@ -116,7 +116,7 @@ public class CallArranger {
             Class<?> carrier = mt.parameterType(i);
             MemoryLayout layout = cDesc.argumentLayouts().get(i);
             if (SharedUtils.isVarargsIndex(cDesc, i)) {
-                argCalc.storageCalculator.adjustForVarArgs();
+                argCalc.storageCalculator.adjustForVarArgs(varArgsOnStack);
             }
             csb.addArgumentBindings(carrier, layout, argCalc.getBindings(carrier, layout));
         }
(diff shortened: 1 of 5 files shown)


src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/aarch64/CallArranger.java
```diff
@@ -210,11 +210,13 @@ VMStorage nextStorage(int type, MemoryLayout layout) {
             return storage[0];
         }
 
-        void adjustForVarArgs() {
-            // This system passes all variadic parameters on the stack. Ensure
-            // no further arguments are allocated to registers.
-            nRegs[StorageClasses.INTEGER] = MAX_REGISTER_ARGUMENTS;
-            nRegs[StorageClasses.VECTOR] = MAX_REGISTER_ARGUMENTS;
+        void adjustForVarArgs(boolean varArgsOnStack) {
```
mcimadamore, line 213: If I'm right this is the only place where something different needs to happen for the two ABIs? If so - perhaps we could consider a different restacking where we use inheritance from a common abstract class - but that would change the shape of the code a bit.

nick-arm: Well I did think about that, but at the moment CallArranger is just a bag of static methods: we'd need to refactor it quite a bit to fit it into that pattern. 

Perhaps (an AArch64 extension of) AbiDescriptor is the right place to store this? That already stores information like stack alignment, shadow space, etc. We'd then have a `MACOS_C` and `LINUX_C` instead of the single `C` we have now, and then pass that into `getBindings()`, etc.

> If I'm right this is the only place where something different needs to happen for the two ABIs?

Actually I think there's one more case we don't handle correctly at the moment but isn't covered by the existing tests:

```c
void f(int, int, int, int, int, int, int, int, char x, short y);
```

In the standard ABI `x` and `y` get spilled to the stack as eight byte quantities (as if they were assigned to a register and then spilled) but in the Mac ABI they are spilled to the stack with their original size and natural alignment. So we need another flag somewhere to handle that difference. I'll make a JBS ticket for it.

mcimadamore: Maybe I'm wrong, but it doesn't seem like the refactoring would be that big. After all, the public API of CallArranger is the two static methods arrangeUpcall/Downcall. So, from a client perspective that is the only change. CallArranger could provide two already-made constants, which will give you a call arranger for linux or mac, so that clients won't need to instantiate a new call arranger all the time. This should keep the number of changes in clients under control?

nick-arm: @mcimadamore Yes you're right it's not a big change at all. What do you think of the latest version?

@JornVernee I can take look at implementing the optimised upcall stubs.

JornVernee (approved):
LGTM

PR #602, opened 2021-10-20
Address failures in jdk.incubator.foreign benchmarks

FrauBoes:
This change addresses some failures in the jdk.incubator.foreign benchmarks, mainly caused by missing casts and missing native access.


test/micro/org/openjdk/bench/jdk/incubator/foreign/BulkOps.java
```diff
@@ -138,20 +137,20 @@ public void segment_copy() {
     @Benchmark
     @OutputTimeUnit(TimeUnit.NANOSECONDS)
     public void segment_copy_static() {
-        MemorySegment.copy(bytes, 0, segment, JAVA_BYTE, 0, bytes.length);
+        MemorySegment.copy(bytes, 0, segment, JAVA_INT, 0, bytes.length);
```
mcimadamore, line 140: Good catch!

PR #603, opened 2021-10-21
8255903: Enable multi-register return values for native invokers

JornVernee:
Hi,

This patch implements support for multi-register returns in native invokers, and removes the buffered invocation strategy for downcalls. This is achieved, essentially, by using an in memory return: the caller allocates a bit of memory, and the native invoker stub writes the values of the return register to that memory. Then, the post processing code reads the register values back from there.

Currently, the target address of a downcall is handled separately from the other arguments. I initially implemented passing the IMR address the same way, but I realized the removal of the buffered invocation strategy affords us a better way of doing things: we can just make the target address and the IMR address part of the normal calling sequence, and remove a bunch of special-casing code to handle these. We now just use a binding recipe to unbox these arguments and shuffle them into registers, and the ABIDescriptor tells the native invoker stub which registers to pick up these arguments from.

Thanks,
Jorn

---------
### Progress

### Issue
 * [JDK-8255903](https://bugs.openjdk.java.net/browse/JDK-8255903): Enable multi-register return values for native invokers

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/603/head:pull/603` \
`$ git checkout pull/603`

`$ git checkout pull/603` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/603/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 603`

`$ git pr show -t 603`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/603.diff">https://git.openjdk.java.net/panama-foreign/pull/603.diff</a>

</details>


src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/CLinker.java
```diff
@@ -35,6 +35,7 @@
 import jdk.internal.reflect.Reflection;
 
 import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
```
JornVernee, line 38:
```suggestion
```

src/hotspot/cpu/aarch64/universalNativeInvoker_aarch64.cpp
```diff
@@ -217,13 +116,6 @@ RuntimeStub* ProgrammableInvoker::make_native_invoker(BasicType* signature,
 }
 
 void NativeInvokerGenerator::generate() {
-  // we can't use rscratch1 because it is r8, and used by the ABI
-  Register tmp1 = r9;
-  Register tmp2 = r10;
-  assert(!target_uses_register(tmp1->as_VMReg()), "conflict");
-  assert(!target_uses_register(tmp2->as_VMReg()), "conflict");
-  assert(!target_uses_register(rthread->as_VMReg()), "conflict");
```
JornVernee, line 225: I realized these checks were bogus now that we shuffle arguments as part of the native invoker stub. It is fine for the call to use these registers. Though, if `rthread` is clobbered by the call, we would need to save and restore it (currently that's not the case on any platform though, so we assume it doesn't happen).

PR #608, opened 2021-11-02
8275647: Enable multi-register return values for optimized upcall stubs

JornVernee:
Hi,

After we've added multi-register return support to native invokers, this patch does the same for optimized upcall stubs. The strategy is very similar: the caller (in this case the upcall stub) allocates a return buffer in which the return values are stored, and then after the upcall shuffles the return values from the buffer into the right registers.

Note, in it's current state this patch will disable regress AArch64 since optimized upcalls are not implemented there yet. i.e. this PR is blocked by: https://bugs.openjdk.java.net/browse/JDK-8275646

Jorn


src/hotspot/cpu/x86/universalUpcallHandler_x86_64.cpp
```diff
@@ -281,38 +168,16 @@ static void restore_callee_saved_registers(MacroAssembler* _masm, const ABIDescr
...
-      out_sig_bt[i++] = ss.type();  // Collect remaining bits of signature
-      if (ss.type() == T_LONG || ss.type() == T_DOUBLE)
-        out_sig_bt[i++] = T_VOID;   // Longs & doubles take 2 Java slots
-    }
-    assert(i == total_out_args, "");
-    ret_type = ss.type();
-  }
-  // skip receiver
-  BasicType* in_sig_bt = out_sig_bt + 1;
-  int total_in_args = total_out_args - 1;
```
JornVernee, line 314: I've moved this to the shared code in universalUpcallHanlder.cpp, since it can be shared between platforms.

src/hotspot/share/code/codeBlob.cpp
```diff
@@ -751,3 +751,9 @@ void OptimizedEntryBlob::oops_do(OopClosure* f, const frame& frame) {
 JavaFrameAnchor* OptimizedEntryBlob::jfa_for_frame(const frame& frame) const {
   return &frame_data_for_frame(frame)->jfa;
 }
+
+void OptimizedEntryBlob::free(OptimizedEntryBlob* blob) {
+  assert(blob != nullptr, "caller must check for NULL");
+  JNIHandles::destroy_global(blob->receiver());
+  BufferBlob::free(blob);
+}
```
JornVernee, line 759: We were freeing upcall stubs using CodeCache::free, but this doesn't track memory usage. I've added this method to properly handle freeing upcall stubs.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/ProgrammableUpcallHandler.java
```diff
@@ -155,11 +134,43 @@ private static void checkPrimitive(MethodType type) {
...
 
     private static MethodHandle specializedBindingHandle(MethodHandle target, CallingSequence callingSequence,
-                                                         Class<?> llReturn, long allocationSize) {
+                                                         Class<?> llReturn, ABIDescriptor abi) {
         MethodType highLevelType = callingSequence.methodType();
 
         MethodHandle specializedHandle = target; // initial
 
+        // we handle returns first since IMR adds an extra parameter that needs to be specialized as well
+        if (llReturn != void.class || callingSequence.needsReturnBuffer()) {
```
JornVernee, line 143: Note that the old code for handling return values is just moved from below the code that handles arguments, since the return buffer argument also needs to be handle along with the rest of arguments, and it is added here.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/abi/ProgrammableUpcallHandler.java
```diff
@@ -179,114 +190,79 @@ private static MethodHandle specializedBindingHandle(MethodHandle target, Callin
...
-            VMStorage storage = binding.storage();
-            MemorySegment ptr = bufferBase.asSlice(layout.retOffset(storage));
-            SharedUtils.writeOverSized(ptr, binding.type(), o);
-        }
+    private record InvocationData(MethodHandle leaf,
+                                  Map<VMStorage, Integer> argIndexMap,
+                                  Map<VMStorage, Integer> retIndexMap,
+                                  CallingSequence callingSequence,
+                                  Binding.VMStore[] retMoves,
+                                  ABIDescriptor abi) {}
```
JornVernee, line 203: Same as with downcalls, I've bundled all the extra arguments into a record.

nick-arm:
> 
> Note, in it's current state this patch will disable regress AArch64 since optimized upcalls are not implemented there yet. i.e. this PR is blocked by: https://bugs.openjdk.java.net/browse/JDK-8275646
> 

I've nearly finished this, I'll send the patch tomorrow.

src/hotspot/cpu/x86/foreign_globals_x86.cpp
```diff
@@ -71,51 +69,6 @@ const ABIDescriptor ForeignGlobals::parse_abi_descriptor_impl(jobject jabi) cons
...
-  layout.returns_integer = (size_t) output_offsets->long_at(INTEGER_TYPE);
-  layout.returns_vector = (size_t) output_offsets->long_at(VECTOR_TYPE);
-  layout.returns_x87 = (size_t) output_offsets->long_at(X87_TYPE);
-
-  layout.buffer_size = layout_oop->long_field(BL.size_offset);
-
-  return layout;
-}
-
-const CallRegs ForeignGlobals::parse_call_regs_impl(jobject jconv) const {
```
JornVernee, line 96: This is now also moved to shared code.

mcimadamore (approved):
Looks very good. I've tested the changes on my machine, and all works fine.

src/hotspot/cpu/arm/universalUpcallHandle_arm.cpp
```diff
@@ -25,16 +25,12 @@
 #include "prims/universalUpcallHandler.hpp"
 #include "utilities/debug.hpp"
 
-address ProgrammableUpcallHandler::generate_upcall_stub(jobject rec, jobject jabi, jobject jlayout) {
-  Unimplemented();
-  return nullptr;
-}
-
-address ProgrammableUpcallHandler::generate_optimized_upcall_stub(jobject mh, Method* entry, jobject jabi, jobject jconv) {
+address ProgrammableUpcallHandler::generate_optimized_upcall_stub(jobject receiver, Method* entry,
```
mcimadamore, line 28: is the name "optimimzed" still carrying meaning in the new VM code?

JornVernee: Not really, but I thought I'd save the renaming for: https://bugs.openjdk.java.net/browse/JDK-8275648

JornVernee: Doing it here wouldn't be such a hassle though. Though, there's also OptimizedEntryBlob which requires renaming.

mcimadamore: That's ok - ship it as is - we can rename later.

nick-arm (approved):
AArch64 changes look fine.

PR #616, opened 2021-11-23
8277645: TreeMaker should not call type() on pointer types too early

mcimadamore:
This patch fixes a couple of callsites in TreeMaker which were calling `DelegatedType::type`.

`DelegatedType` is a common supertype for both typedef types and pointer types. The reason why `TreeMaker` calls that method is to recover the *canonical* type of a typedef.

Unfortunately, as `TreeMaker` doesn't check the delegated type kind, it can sometime call `type()` on a delegated type whose kind is `POINTER`. The way in which the jextract parser is setup is such that pointee types can only be accessed *after* parsing has completed.

This issue only occurs under very rare circumstances, that is when a type qualifier (other than the usual ones, such as `const`) is used on a pointer typedef - such as `__ptr64`. In such cases, the clang AST sees an `AttributedType` and not a `QualifiedType`, which throws our parsing logic off-guard.

The solution is to check for the delegated type kind before attempting to call `type()`, and only call it when its kind is `TYPEDEF`. To test this I used the special clang flag `-fms-extensions` which allows to use the MS extensions even on Linux/Mac.

---------
### Progress

### Issue
 * [JDK-8277645](https://bugs.openjdk.java.net/browse/JDK-8277645): TreeMaker should not call type() on pointer types too early

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/616/head:pull/616` \
`$ git checkout pull/616`

`$ git checkout pull/616` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/616/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 616`

`$ git pr show -t 616`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/616.diff">https://git.openjdk.java.net/panama-foreign/pull/616.diff</a>

</details>

### DIFF
--- a/test/jdk/tools/jextract/TestAttributedPointerTypedef.java
+++ b/test/jdk/tools/jextract/TestAttributedPointerTypedef.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright (c) 2020, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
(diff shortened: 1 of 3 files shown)


test/jdk/tools/jextract/TestAttributedPointerTypedef.java
```diff
@@ -0,0 +1,40 @@
...
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+/*
+ * @test
+ * @modules jdk.incubator.jextract
+ * @library /test/lib
+ * @build BadBitfieldTest
+ * @run testng/othervm --enable-native-access=jdk.incubator.jextract BadBitfieldTest
```
JornVernee, line 29: This is not running the right test class it seems.

JornVernee (approved):

PR #617, opened 2021-11-25
8277657: OptimizedEntryBlob should extend RuntimeBlob directly

JornVernee:
This small refactoring makes OptimizedEntryBlob extend RuntimeBlob instead of BufferBlob.

The motivation for doing this is that we don't want to inherit behavior from BufferBlob by accident which is not appropriate for OptimizedEntryBlob, which has happened in the past.

This patch also reverts an earlier fix by Nick which made BufferBlob sub classes pass their header size through the BufferBlob constructor to RuntimeBlob's constructor, since this is no longer needed with this patch.

---------
### Progress

### Issue
 * [JDK-8277657](https://bugs.openjdk.java.net/browse/JDK-8277657): OptimizedEntryBlob should extend RuntimeBlob directly

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/617/head:pull/617` \
`$ git checkout pull/617`

`$ git checkout pull/617` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/617/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 617`

`$ git pr show -t 617`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/617.diff">https://git.openjdk.java.net/panama-foreign/pull/617.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/code/codeBlob.cpp
+++ b/src/hotspot/share/code/codeBlob.cpp
@@ -159,6 +159,18 @@ RuntimeBlob::RuntimeBlob(
   cb->copy_code_and_locs_to(this);
 }
 
+void RuntimeBlob::free(RuntimeBlob* blob) {
+  assert(blob != NULL, "caller must check for NULL");
+  ThreadInVMfromUnknown __tiv;  // get to VM state in case we block on CodeCache_lock
+  blob->flush();
+  {
+    MutexLocker mu(CodeCache_lock, Mutex::_no_safepoint_check_flag);
+    CodeCache::free(blob);
+  }
+  // Track memory usage statistic after releasing CodeCache_lock
+  MemoryService::track_code_cache_memory_usage();
+}
+
 void CodeBlob::flush() {
   FREE_C_HEAP_ARRAY(unsigned char, _oop_maps);
   _oop_maps = NULL;
@@ -225,8 +237,8 @@ void CodeBlob::print_code() {
 // Implementation of BufferBlob
 
 
-BufferBlob::BufferBlob(const char* name, int header_size, int size)
-: RuntimeBlob(name, header_size, size, CodeOffsets::frame_never_safe, /*locs_size:*/ 0)
+BufferBlob::BufferBlob(const char* name, int size)
+: RuntimeBlob(name, sizeof(BufferBlob), size, CodeOffsets::frame_never_safe, /*locs_size:*/ 0)
 {}
 
 BufferBlob* BufferBlob::create(const char* name, int buffer_size) {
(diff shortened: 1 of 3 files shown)


src/hotspot/share/code/codeBlob.cpp
```diff
@@ -159,6 +159,18 @@ RuntimeBlob::RuntimeBlob(
...
+  assert(blob != NULL, "caller must check for NULL");
+  ThreadInVMfromUnknown __tiv;  // get to VM state in case we block on CodeCache_lock
+  blob->flush();
+  {
+    MutexLocker mu(CodeCache_lock, Mutex::_no_safepoint_check_flag);
+    CodeCache::free(blob);
+  }
+  // Track memory usage statistic after releasing CodeCache_lock
+  MemoryService::track_code_cache_memory_usage();
+}
```
JornVernee, line 172: Note that this is just the implementation of BufferBlob::free. Now both BufferBlob::free and OptimizedEntryBlob::free delegate to this function.

src/hotspot/share/code/codeBlob.cpp
```diff
@@ -757,5 +769,23 @@ JavaFrameAnchor* OptimizedEntryBlob::jfa_for_frame(const frame& frame) const {
 void OptimizedEntryBlob::free(OptimizedEntryBlob* blob) {
   assert(blob != nullptr, "caller must check for NULL");
   JNIHandles::destroy_global(blob->receiver());
-  BufferBlob::free(blob);
+  RuntimeBlob::free(blob);
+}
+
+void OptimizedEntryBlob::preserve_callee_argument_oops(frame fr, const RegisterMap* reg_map, OopClosure* f) {
+  // do nothing for now
```
JornVernee, line 776: Will implement this in a followup.

src/hotspot/share/code/codeBlob.cpp
```diff
@@ -274,23 +286,15 @@ void* BufferBlob::operator new(size_t s, unsigned size) throw() {
...
+  RuntimeBlob::free(blob);
 }
 
 
 //----------------------------------------------------------------------------------------------------
 // Implementation of AdapterBlob
 
 AdapterBlob::AdapterBlob(int size, CodeBuffer* cb) :
-  BufferBlob("I2C/C2I adapters", sizeof(AdapterBlob), size, cb) {
+  BufferBlob("I2C/C2I adapters", size, cb) {
```
mcimadamore, line 297: the header size here used to be sizeof(AdapterBlob), but now is sizeof(BufferBlob) - is that an issue?

mcimadamore (commented):
Generally looks good - one bit of refactoring looks suspicious.

src/hotspot/share/code/codeBlob.cpp
JornVernee: This is not an issue since they are the same. Arguably the header size should be passed as `sizeof(AdapterBlob)` for clarity, but I've reverted this to how it is done in the mainline jdk to avoid unnecessary churn for the next JEP: https://github.com/openjdk/jdk/blob/master/src/hotspot/share/code/codeBlob.cpp#L292-L295

mcimadamore (approved):

PR #622, opened 2021-12-02
8278151: Heap segments should handle alignment constraints in a deterministic fashion

mcimadamore:
This patch fixes the issue with alignment checks on heap segment leading to failures which depend on VM implementation details and layout choices.

This is discussed in more details here:
https://mail.openjdk.java.net/pipermail/panama-dev/2021-November/015852.html 

The fix consists in adding a *max alignment* mask to all segments. Native segments have mask set at 0, whereas heap segment have the mask set at the element size of their backing array (e.g. 1 for `byte[]`, 2 for `short[]`/`char[]` and so forth).

The alignment check is the computed as follows:

```
if (((address | maxAlignMask) & alignmentMask) != 0) {
                throw ...
}
```

Where `address` is the long address being dereferenced, `maxAlignMask` is the max alignment associated with the segment being dereferenced, and `alignmentMask` is the alignment associated with the layout used during a dereference operation.
This patch also applies alignment checks to bulk copy operations as well, and it consolidates the way in which alignment checks are carried out (by adding a couple of helper methods, in `Utils` and `AbstractMemorySegment`, respectively).

This patch also changes alignment errors to uniformly throw IllegalArgumentException, in all cases.

Finally, this patch clarifies that during bulk copy and spliterator calls, the element layout alignment cannot be bigger than its size (as the size is used as a stride).

A new test `TestHeapAlignment` has been added to check that alignment of heap segment is correctly enforced.

Performance-wise, this patch doesn't change anything; that's because the value layout constants in `ValueLayout` are all unaligned. For code using aligned layout, some slight regression when using heap segment is possible (whereas native segment should have same performance as before). Eventually, these performance issues should disappear once [1] is fixed.

[1] - https://bugs.openjdk.java.net/browse/JDK-8277850

---------
### Progress

### Issue
 * [JDK-8278151](https://bugs.openjdk.java.net/browse/JDK-8278151): Heap segments should handle alignment constraints in a deterministic fashion

### Reviewers
 * [Paul Sandoz](https://openjdk.java.net/census#psandoz) (@PaulSandoz - Committer) ⚠️ Review applies to 765e2277587c780eef7f1ba9ca408776f9ee8ed5

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/622/head:pull/622` \
`$ git checkout pull/622`

`$ git checkout pull/622` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/622/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 622`

`$ git pr show -t 622`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/622.diff">https://git.openjdk.java.net/panama-foreign/pull/622.diff</a>

</details>


src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
```diff
@@ -124,8 +124,14 @@ private AbstractMemorySegmentImpl asSliceNoCheck(long offset, long newSize) {
         if (elementLayout.byteSize() == 0) {
             throw new IllegalArgumentException("Element layout size cannot be zero");
         }
-        if (byteSize() % elementLayout.byteSize() != 0) {
-            throw new IllegalArgumentException("Segment size is no a multiple of layout size");
+        if (elementLayout.byteAlignment() > elementLayout.byteSize()) {
```
mcimadamore, line 127: I wonder if we should add a similar check on the various `MemorySegment/MemoryAddress::[get/set]atIndex` ? I think accessing hyper-aligned elements is always going to fail, for certain offsets - in fact when deriving a var handle from a layout path, it is never possible to have an array var handle for an hyper-aligned element layout.

src/java.base/share/classes/java/lang/invoke/X-VarHandleMemoryAccess.java.template
```diff
@@ -115,14 +115,15 @@ final class MemoryAccessVarHandle$Type$Helper extends MemoryAccessVarHandleBase
     static long offsetNoVMAlignCheck(boolean skipAlignmentMaskCheck, MemorySegmentProxy bb, long offset, long alignmentMask) {
         long base = bb.unsafeGetOffset();
         long address = base + offset;
-        if (skipAlignmentMaskCheck) {
+        long maxAlignMask = bb.maxAlignMask();
+        if (skipAlignmentMaskCheck && maxAlignMask == 0) {
```
rsmogura, line 119: I've got minor comment. I wonder if we can `skipAlignmentMaskCheck` parameter and just base on `maxAlignMask`

rsmogura: Sorry, it was stupid question. Please ignore

mcimadamore: No problem - this is quite tricky. In general, `skipAlignmentMaskCheck` still does _something_ when accessing native segments (as the max align mask is zero there), but in general you are morally correct in observing that it's less useful than it used to be. My hope is that with the right VM fixes, we can get rid of `skipAlignmentMaskCheck` completely, and simplify all this.

For the records, the difference between unaligned and aligned access is as follows:

```
Benchmark                                                Mode  Cnt  Score   Error  Units
LoopOverNonConstant.segment_loop_instance                avgt   30  0.229 ? 0.002  ms/op
LoopOverNonConstant.segment_loop_instance_aligned        avgt   30  0.335 ? 0.001  ms/op
```

It's not terrible, but could of course be better.

rsmogura: Hmm...

I wonder a bit if this code can be optimized `if (((address | maxAlignMask) & alignmentMask) != 0) {`

Max `maxAlignMask`, can be 1,2,4 (if I see it correctly) so if `maxAlignMask` `alignmentMask` and will not be correct, than exception will be thrown regardless of address? I wonder if making static check could be more performant as VM move such check before loop, than we could only focus on `address & alignmentMask != 0`. Or something like this.

mcimadamore: The max align check is already moved out of the loop (as it only depends on the segment, which is a loop invariant). The problematic check, which is not moved outside the loop is exactly `address & alignmentMask != 0`. That check cannot be hoisted, because `address` is made up of `offset` and the segment base address; the latter is a loop invariant, but the former isn't.

Of course C2 should be able to detect that offset is a function of the loop variable, and prove that the check can never fail - but this pattern is currently not detected (hence JDK-8277850).

rsmogura: Yes, I was thinking about line 125 `if (((address | maxAlignMask) & alignmentMask) != 0) {` to split into to two
`if ((maxAlignMask & alignmentMask) != 0) throw...` and `if (address & alignmentMask) != 0) throw...` if this could help a bit.

test/jdk/java/foreign/TestHeapAlignment.java
```diff
@@ -0,0 +1,124 @@
...
+    public void testHeapAlignment(MemorySegment segment, int align, Object val, Object arr, ValueLayout layout, Function<Object, MemorySegment> segmentFactory) {
+        assertAligned(align, layout, () -> layout.varHandle().get(segment));
+        assertAligned(align, layout, () -> layout.varHandle().set(segment, val));
+        MemoryLayout seq = MemoryLayout.sequenceLayout(10, layout);
+        assertAligned(align, layout, () -> seq.varHandle(MemoryLayout.PathElement.sequenceElement()).get(segment, 0L));
+        assertAligned(align, layout, () -> seq.varHandle(MemoryLayout.PathElement.sequenceElement()).set(segment, 0L, val));
+        assertAligned(align, layout, () -> segment.spliterator(layout));
+        if (arr != null) {
+            assertAligned(align, layout, () -> MemorySegment.copy(arr, 0, segment, layout, 0, 1));
+            assertAligned(align, layout, () -> MemorySegment.copy(arr, 0, segment, layout, 0, 1));
```
PaulSandoz, line 57: Swap arguments for second invocation?
```suggestion
            assertAligned(align, layout, () -> MemorySegment.copy(segment, layout, 0, arr, 0, 1));
```

PaulSandoz (approved):
Excellent analysis of the problem, and a very pleasing result.

Having consistent alignment check behavior with sequenced layout construction and get/setAtIndex seems like a good thing, re: your comment.

src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
```diff
@@ -226,7 +260,9 @@
      * @param elementLayout the layout to be used for splitting.
      * @return the element spliterator for this segment
      * @throws IllegalArgumentException if the {@code elementLayout} size is zero, or the segment size modulo the
-     * {@code elementLayout} size is greater than zero.
+     * {@code elementLayout} size is greater than zero, if this segment is
+     * <a href="MemorySegment.html#segment-alignment">incompatible with the alignment constraints</a> in the provided layout,
+     * or if the {@code elementLayout} alignment is greater than its size.
```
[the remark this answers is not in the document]
mcimadamore: This is a good suggestion, but I'd prefer to handle it separately, as there are also other places where this can come in handy.

PaulSandoz (approved):

rose00:
I would prefer to reserve the term "hyper aligned" for the RFE in https://bugs.openjdk.java.net/browse/JDK-8167334 which talks about support for 128-bit alignment (and more) in the JVM.  The odd condition being called "hyper aligned" in this PR is, IIUC, an overflow condition where the alignment of a layout exceeds its size.

mcimadamore:
> I would prefer to reserve the term "hyper aligned" for the RFE in https://bugs.openjdk.java.net/browse/JDK-8167334 which talks about support for 128-bit alignment (and more) in the JVM. The odd condition being called "hyper aligned" in this PR is, IIUC, an overflow condition where the alignment of a layout exceeds its size.

I don't think the term leaks anywhere in the API - but only in an helper function and some tests. That said, I thought the term was fitting to describe a datum whose alignment constraints were bigger than its natural one (read, the size). But I'm happy to change the names there.

JornVernee (approved):

PR #633, opened 2022-01-24
8280527: Move jdk.incubator.foreign to java.lang.foreign

FrauBoes:
To prepare for preview, this change removes the jdk.incubator.foreign module and moves the public classes to java.base/java.lang.foreign, implementation classes to java.base/jdk.internal.foreign respectively. 

While here, `SystemLookup` is replaced with the method `ClassLoader::findNative` and the methods in `MemoryHandles` are moved to `MethodHandles`. The `@PreviewFeature(feature=PreviewFeature.Feature.FOREIGN)` annotation and the `--enable-preview` flag are added were needed.

---------
### Progress

### Issue
 * [JDK-8280527](https://bugs.openjdk.java.net/browse/JDK-8280527): Move jdk.incubator.foreign to java.lang.foreign

### Reviewers
 * [Maurizio Cimadamore](https://openjdk.java.net/census#mcimadamore) (@mcimadamore - Committer) ⚠️ Review applies to 01453629e5f845156d962dbf46415639a12d3262

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/633/head:pull/633` \
`$ git checkout pull/633`

`$ git checkout pull/633` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/633/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 633`

`$ git pr show -t 633`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/633.diff">https://git.openjdk.java.net/panama-foreign/pull/633.diff</a>

</details>


src/java.base/share/classes/jdk/internal/foreign/SharedScope.java
```diff
@@ -45,6 +44,16 @@
...
+    private static final VarHandle STATE;
+
+    static {
+        try {
+            STATE = MethodHandles.lookup().findVarHandle(SharedScope.class, "state", int.class);
+        } catch (Throwable ex) {
+            throw new ExceptionInInitializerError(ex);
+        }
+    }
+
```
JornVernee, line 56: This shouldn't be needed. It's already in ResourceScopeImpl

test/hotspot/jtreg/gc/shenandoah/compiler/TestLinkToNativeRBP.java
```diff
@@ -23,36 +23,36 @@
...
 public class TestLinkToNativeRBP {
     static {
         System.loadLibrary("LinkToNativeRBP");
     }
 
-    final static CLinker abi = CLinker.systemCLinker();
-    static final SymbolLookup lookup = SymbolLookup.loaderLookup();
-    final static MethodHandle foo = abi.downcallHandle(lookup.lookup("foo").get(),
-            FunctionDescriptor.of(ValueLayout.JAVA_INT));
+    final static CLinker abi = CLinker.getInstance();
```
JornVernee, line 52: This looks incorrect. This method was renamed to `systemCLinker`. Note that this test doesn't run as part of run-test-jdk_foreign.

test/jdk/java/foreign/TestMemoryHandleAsUnsigned.java
```diff
@@ -1,257 +0,0 @@
-/*
```
JornVernee, line 1: Why was this file deleted?

test/jdk/java/foreign/virtual/TestVirtualCalls.java
```diff
@@ -57,10 +55,9 @@
                 FunctionDescriptor.of(C_INT));
 
         System.loadLibrary("Virtual");
-        SymbolLookup lookup = SymbolLookup.loaderLookup();
-        funcA = lookup.lookup("funcA").get();
-        funcB = lookup.lookup("funcB").get();
-        funcC = lookup.lookup("funcC").get();
+        funcA = TestVirtualCalls.class.getClassLoader().findNative("funcA").get();
+        funcB = TestVirtualCalls.class.getClassLoader().findNative("funcB").get();
+        funcC = TestVirtualCalls.class.getClassLoader().findNative("funcC").get();
```
JornVernee, line 60: Maybe a helper method could be added to `NativeTestHelper`, look `findNativeOrThrow(...)` to make lookups a little bit simpler in the tests

test/micro/org/openjdk/bench/jdk/incubator/foreign/LoopOverSlice.java
```diff
@@ -1,259 +0,0 @@
-/*
```
JornVernee, line 1: File deleted by accident?

make/test/BuildMicrobenchmark.gmk
```diff
@@ -91,10 +91,10 @@ $(eval $(call SetupJavaCompilation, BUILD_JDK_MICROBENCHMARK, \
     TARGET_RELEASE := $(TARGET_RELEASE_NEWJDK_UPGRADED), \
     SMALL_JAVA := false, \
     CLASSPATH := $(MICROBENCHMARK_CLASSPATH), \
-    DISABLED_WARNINGS := processing rawtypes cast serial, \
+    DISABLED_WARNINGS := processing rawtypes cast serial preview, \
     SRC := $(MICROBENCHMARK_SRC), \
     BIN := $(MICROBENCHMARK_CLASSES), \
-    JAVAC_FLAGS := --add-exports java.base/sun.security.util=ALL-UNNAMED, \
+    JAVAC_FLAGS := --add-exports java.base/sun.security.util=ALL-UNNAMED --enable-preview, \
```
JornVernee, line 97: AFAIK adding --enable-preview to JAVAC_FLAGS like this would result in the class file version being poisoned for all the benchmark classes. I.e. it would require running all benchmarks with `--enable-preview` (not just the foreign ones).

This was an issue in the past as well (see https://bugs.openjdk.java.net/browse/JDK-8250669 and related).

I think at the minimum, `--enable-preview` should also be added to the default VM args when running benchmarks if we go with a javac flag. But, even that might not be good enough for mainline. I don't remember the exact details, but it was hard to have benchmarks that used preview features without effecting all benchmarks.

@cl4es Do you remember the outcome of this? I remember it was not as simple a fix as just adding `--enable-preview` to the VM args...

src/java.base/share/classes/jdk/internal/access/foreign/MemoryAddressProxy.java
```diff
@@ -1,99 +0,0 @@
...
-                return Math.multiplyExact(op1, op2);
-            } catch (ArithmeticException ex) {
-                throw overflowException(Long.MIN_VALUE, Long.MAX_VALUE);
-            }
-        }
-    }
-
-    private static IndexOutOfBoundsException overflowException(long min, long max) {
-        return new IndexOutOfBoundsException(String.format("Overflow occurred during offset computation ; offset exceeded range { %d .. %d }", min, max));
-    }
```
JornVernee, line 98: I don't see these methods back after the move, only the ones from MemorySegmentProxy. Were they not needed anymore?

JornVernee (commented):
- jdk/internal/foreign/abi/BufferLayout file is empty. It should be removed.
- same for test/jdk/java/foreign/TestUpcall.java

JornVernee:
Also, is `java.lang.foreign` the package we want? In the past we were in `java.foreign`.

test/jdk/java/foreign/TestMemoryHandleAsUnsigned.java
JornVernee: Looking closer, it seems that the `asUnsigned` combinator was removed altogether?

src/java.base/share/classes/jdk/internal/access/foreign/MemoryAddressProxy.java
JornVernee: Ah, looks like they were unused before as well :)

test/jdk/java/foreign/TestMemoryHandleAsUnsigned.java
mcimadamore: Yes, we have decided to remove the combinator, at least for now. While the other combinators are very general, this one feels very specific, and I'm not 100% sure if it's worth the complexity. We can always re-add later.

make/test/BuildMicrobenchmark.gmk
mcimadamore: I think this is something we'd need to address before we integrate into mainline - but for now I think it's tolerable to run benchmarks by passing the extra `--enable-preview`. I'd suggest to file a follow-up issue and fix that there, to keep this review simple.

test/micro/org/openjdk/bench/java/lang/foreign/BulkOps.java
```diff
@@ -175,7 +176,7 @@ public void buffer_copy() {
     @CompilerControl(CompilerControl.Mode.DONT_INLINE)
     @OutputTimeUnit(TimeUnit.NANOSECONDS)
     public void segment_copy_static_dontinline() {
-        MemorySegment.copy(bytes, 0, segment, JAVA_INT, 0, bytes.length);
+        MemorySegment.copy(bytes, 0, segment, JAVA_BYTE, 0, bytes.length);
```
mcimadamore, line 179: Note that `bytes` has type `int[]`, so using `JAVA_INT` was the right choice. Maybe you just want to rename the `bytes` field to `int` ?

mcimadamore (commented):
Looks good. I've added some comments. As for the package name, perhaps using a less nested name `java.foreign` would make it easier to update it later on (if needs be). The `lang` in `java.lang.foreign` is a bit misleading, as @JornVernee pointed out (as there's nothing directly related to the language here). But if it's too hard to fix we can look at tweaking that later.

src/java.base/share/classes/jdk/internal/foreign/SharedScope.java
FrauBoes: That's right - removed.

test/hotspot/jtreg/gc/shenandoah/compiler/TestLinkToNativeRBP.java
FrauBoes: Yepp, fixing to 
```    
final static CLinker abi = CLinker.systemCLinker();
final static MethodHandle foo = abi.downcallHandle(TestLinkToNativeRBP.class.getClassLoader().findNative("foo").get(),
            FunctionDescriptor.of(ValueLayout.JAVA_INT));
```

test/jdk/java/foreign/virtual/TestVirtualCalls.java
FrauBoes: good idea, I'll add that.

src/java.base/share/classes/java/lang/foreign/CLinker.java
```diff
@@ -108,12 +108,6 @@
  * downcall method handles (as {@link NativeSymbol} implements the {@link Addressable} interface) and,
  * when no longer required, they can be {@link ResourceScope#close() released}, via their {@linkplain NativeSymbol#scope() scope}.
  *
- * <h2>System lookup</h2>
```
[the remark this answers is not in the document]
FrauBoes: Sure, I'll add it back in.

src/java.base/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
```diff
@@ -364,15 +363,17 @@ public void checkAccess(long offset, long length, boolean readOnly) {
     }
 
     void checkValidState() {
-        scope.checkValidStateSlow();
+        try {
```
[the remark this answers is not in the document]
FrauBoes: That's wrong, `checkValidStateSlow()` is already throwing the right exception type. Fixed.

src/hotspot/share/prims/scopedMemoryAccess.cpp
```diff
@@ -171,10 +172,11 @@ static JNINativeMethod jdk_internal_misc_ScopedMemoryAccess_methods[] = {
 #undef CC
 #undef FN_PTR
```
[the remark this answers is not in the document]
FrauBoes: Will address this in https://bugs.openjdk.java.net/browse/JDK-8280596

src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -7850,4 +7856,319 @@ private static MethodType tableSwitchChecks(MethodHandle defaultCase, MethodHand
...
+     * {@code IllegalStateException} is thrown, irrespective of the access mode being used.
+     *
+     * @param layout the value layout for which a memory access handle is to be obtained.
+     * @return the new memory access var handle.
+     * @throws IllegalArgumentException if an illegal carrier type is used, or if {@code alignmentBytes} is not a power of two.
+     * @throws NullPointerException if {@code layout} is {@code null}.
+     * @since 19
+     */
+    @PreviewFeature(feature=PreviewFeature.Feature.FOREIGN)
+    public static VarHandle memoryAccessVarHandle(ValueLayout layout) {
```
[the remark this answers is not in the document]
FrauBoes: Addressing this in https://bugs.openjdk.java.net/browse/JDK-8280597

test/micro/org/openjdk/bench/jdk/incubator/foreign/LoopOverSlice.java
FrauBoes: Oops! That's added back in now.

src/java.base/share/classes/java/lang/invoke/VarHandles.java
```diff
@@ -45,6 +46,26 @@
...
 final class VarHandles {
 
+    static final MethodHandle BYTE_TO_BOOL;
+    static final MethodHandle BOOL_TO_BYTE;
+    static final MethodHandle ADDRESS_TO_LONG;
+    static final MethodHandle LONG_TO_ADDRESS;
+
+    static {
+        try {
+            BYTE_TO_BOOL = MethodHandles.lookup().findStatic(VarHandles.class, "byteToBoolean",
```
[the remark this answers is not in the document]
FrauBoes: Right, some duplication here. I dropped these MH constants and changed `MHs::memoryAccessVarHandle` to call `Utils::makeMemoryAccessVarHandle`.

test/jdk/java/foreign/NativeTestHelper.java
```diff
@@ -105,4 +106,8 @@ public static MemoryAddress allocateMemory(long size) {
             throw new IllegalStateException(ex);
         }
     }
+
+    public static NativeSymbol findNativeOrThrow(Class<?> clazz, String name) {
```
mcimadamore, line 110: note: if this was an _instance_ method, then I think you could get rid of the `clazz` parameter, and just use `this` ?

mcimadamore (approved):
Looks good. Minor (optional) nit in test helper routine.

src/java.base/share/classes/java/lang/invoke/VarHandles.java
```diff
@@ -657,6 +657,14 @@ private static boolean isCheckedException(Class<?> clazz) {
...
     }
 
+    private static boolean byteToBoolean(byte b) {
+        return b != 0;
+    }
+
+    private static byte booleanToByte(boolean b) {
+        return b ? (byte)1 : (byte)0;
+    }
+
```
JornVernee, line 667: I think these can be removed as well now?

test/jdk/java/foreign/NativeTestHelper.java
FrauBoes: good point, changed it

src/java.base/share/classes/java/lang/invoke/VarHandles.java
FrauBoes: right, removed

test/jdk/java/foreign/TestVarArgs.java
```diff
@@ -65,7 +65,7 @@
         System.loadLibrary("VarArgs");
     }
 
-    static final NativeSymbol VARARGS_ADDR = findNativeOrThrow(TestVarArgs.class, "varargs");
+    static final NativeSymbol VARARGS_ADDR = new NativeTestHelper().findNativeOrThrow("varargs");
```
mcimadamore, line 68: Maybe instead of creating `newNativeTestHelper` you can extend that class (as in other cases)

FrauBoes: Reverting back to a static method

src/java.base/share/classes/java/lang/invoke/VarHandles.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2014, 2018, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2014, 2022, Oracle and/or its affiliates. All rights reserved.
```
JornVernee, line 2: This file has a copyright year change, but no actual changes.

JornVernee (approved):
This looks good. Just one minor copyright nit inline.

Let's not get too hung up on the test changes.

mcimadamore (approved):
Looks good

src/java.base/share/classes/java/lang/invoke/VarHandles.java
FrauBoes: Good catch, thanks.

mcimadamore (approved):

JornVernee (approved):

mcimadamore:
As for the package names, let's integrate these changes, and file a follow up issue to change the names later, as we're still pondering on what the best names could be.

mcimadamore:
Are you saying that we should block progress in making this API a preview API because _your_ IDE environment happens to be broken by said changes? (Note that these preview changes are actually in a different branch - if you want stable changes, look elsewhere). Maybe file a bug against Netbeans (but, in my experience, when using a new release version like 19 this early in the cycle, all IDEs are equally "borked" and it takes one or two updates to get that up and running). Anyway, hard to see how any of this is relevant with this PR.

> _Mailing list message from [Ty Young](mailto:youngty1997@gmail.com) on [panama-dev](mailto:panama-dev@mail.openjdk.java.net):_
> 
> Thanks for this. It looks like Netbeans really doesn't like preview APIs and is going absolutely haywire with errors saying that I need to add --enable-preview when it's already added for both the compiler and runtime. It's so broken that it will no longer provide any suggestions to add a import for a class. It looks like Maven will compile it once I fix the hundred or so errors, but Netbeans is borked.
> 
> On 1/31/22 3:12 AM, Julia Boes wrote:

mcimadamore:
> But getting more back on topic, I understand Intellij is probably what
> JDK developers use for the most part, but one would think that if you
> wanted feedback on the preview that any issues with IDE support would
> have been ironed out by now. it's early and the final package name isn't
> even necessarily final, I know, which is why I'm not making too big of a
> deal about it. Just saying this is an issue in case it was missed since
> most people don't seem to use anything but Intellij.

I believe using IntelliJ with 19 and preview is similarly broken at this time (and it always is before the first IDE update _after_ the source level 19 has been in the repo). I believe these are transient issues which occur every 6 months (as IDEs have to be updated to detect the new source level).

I'd suggest perhaps to use a nightly Netbeans build, to see if the issues are being worked on (which I'm sure they are).

In other words, I don't expect this branch to be really usable in any IDE for at least another month or so (unless users are willing to work with early access IDE versions, that is - such as IntelliJ EAP).

As for the other problem you mention with IntelliJ (1) it did not stall development, (2) it was a completely different issues (use of dollar signs in names unfortunately confusing the IDE class reader). The issue you are seeing is, it seems, normal "growing pains" which trigger whenever the mainline JDK source level/classfile is bumped up one version.

PR #635, opened 2022-01-31
8278414: Replace binding recipe customization using MH combinators with bytecode spinning

JornVernee:
Hi,

This patch removes the binding recipe specialization using MethodHandle combinators in the linker implementation, and replaces it with specialization by spinning a class using the ASM bytecode API.

The main reason for doing this is simplicity of the implementation, as well as finer-grained control over the classes that are generated.

The meat of this PR is in `jdk.internal.foreign.abi.Specializer`, but there are some spin-off changes in this PR as well:
- MethodHandles are really good at sharing classes, so to get comparable performance, this PR also adds a cache behind CLinker::downcallHandle. Due to this, I've also created an AbstractLinker base class for all linker implementations, which contains the cache as well as other shared code.
- CallingSequence is used to steer the class generation. I've added some predicates to this class to check whether the CS is for downcall or upcall. As well as splitting the methodType() accessor into callerMethodType() and calleeMethodType(), reflecting the 2 method types in play during generation (one of the 2 was being computed ad-hoc during specialization previously). This is also responsible for most of the test changes. I've added some javadoc to CS which hopefully adequately explains these accessors.
- I've slightly changed the toString output of ValueLayout to include carrier types as well, in order to be able to use FunctionDescriptor.toString() as a file name in case we want to dump the generated classes to disc. This avoids conflicts between e.g. long and double, which previously were both stringified as b64.

Thanks,
Jorn

---------
### Progress

### Issue
 * [JDK-8278414](https://bugs.openjdk.java.net/browse/JDK-8278414): Replace binding recipe customization using MH combinators with bytecode spinning

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/635/head:pull/635` \
`$ git checkout pull/635`

`$ git checkout pull/635` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/635/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 635`

`$ git pr show -t 635`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/635.diff">https://git.openjdk.java.net/panama-foreign/pull/635.diff</a>

</details>


src/java.base/share/classes/jdk/internal/foreign/abi/AbstractLinker.java
```diff
@@ -0,0 +1,111 @@
...
+
+        return DOWNCALL_CACHE.get(function, fd -> {
+            MethodType type = SharedUtils.inferMethodType(fd, false);
+            MethodHandle handle = arrangeDowncall(type, fd);
+            handle = SharedUtils.maybeInsertAllocator(handle);
+            MethodHandle mh = SharedUtils.wrapDowncall(handle, fd);
+            return mh;
+        });
+    }
+
```
JornVernee, line 61:
```suggestion

```

src/java.base/share/classes/java/lang/foreign/ValueLayout.java
```diff
@@ -104,9 +104,10 @@ public ValueLayout withOrder(ByteOrder order) {
 
     @Override
     public String toString() {
-        return decorateLayoutString(String.format("%s%d",
+        return decorateLayoutString(String.format("%s%d(%s)",
                 order == ByteOrder.BIG_ENDIAN ? "B" : "b",
-                bitSize()));
+                bitSize(),
+                carrier == MemoryAddress.class ? "MA" : carrier.descriptorString()));
```
mcimadamore, line 110: I have thought separately about that. I think I'd like to change this so that we drop the `b` and we replace it with a one letter descriptor (as per descriptor string), and use `a` for `MemoryAddress. Then, as for `b` we use upper case for bigg-endian, and lower-case for little endian. That way the toString representation is compact in all cases (while with this patch it blows up on MemoryAddress).

JornVernee: Good suggestion. I went with this scheme since it seemed more extensible if we wanted to add more carriers in the future. But, it's not like the toString format is specified, so I guess we can change it up later if needed.

src/java.base/share/classes/jdk/internal/foreign/abi/Specializer.java
```diff
@@ -0,0 +1,854 @@
...
+    }
+
+    private void emitCopyBuffer(Binding.Copy copy) {
+        long size = copy.size();
+        long alignment = copy.alignment();
+
+        assert typeStack.pop() == MemorySegment.class;
+
+        // operand/srcSegment is on the stack
+        // copy(MemorySegment srcSegment, long srcOffset, MemorySegment dstSegment, long dstOffset, long bytes)
```
mcimadamore, line 600: I found those comments confusing, as they almost look like commented out code :-)

src/java.base/share/classes/jdk/internal/foreign/abi/Specializer.java
```diff
@@ -0,0 +1,854 @@
...
+import java.nio.file.Path;
+import java.util.ArrayDeque;
+import java.util.ArrayList;
+import java.util.Deque;
+import java.util.List;
+
+import static java.lang.invoke.MethodType.methodType;
+import static jdk.internal.org.objectweb.asm.Opcodes.*;
+
+public class Specializer {
```
mcimadamore, line 64: Maybe a less generic name would be better? Like BindingSpecializer?

src/java.base/share/classes/jdk/internal/foreign/abi/Specializer.java
```diff
@@ -0,0 +1,854 @@
...
+           }
+        }
+
+        mv.visitTryCatchBlock(tryStart, tryEnd, catchStart, null);
+    }
+
+    private void doBindings(List<Binding> bindings) {
+        for (Binding binding : bindings) {
+            switch (binding.tag()) {
+                case VM_STORE -> emitVMStore((Binding.VMStore) binding);
```
mcimadamore, line 399: I note that we depart from the previous model of having virtual calls on the Binding instance which does the right thing. That's fine, I guess there's a lot of coupling (esp. in VMLoad/VMStore) with the specializer state.

src/java.base/share/classes/jdk/internal/foreign/abi/Specializer.java
```diff
@@ -0,0 +1,854 @@
...
+
+        // allocator passed to us for allocating the return MS (downcalls only)
+        if (callingSequence.forDowncall()) {
+            RETURN_ALLOCATOR_IDX = 0; // first param
+        }
+
+        // create a Binding.Context for this call
+        if (callingSequence.allocationSize() != 0) {
+            emitConst(callingSequence.allocationSize());
+            mv.visitMethodInsn(INVOKESTATIC, BINDING_CONTEXT_INTRN, OF_BOUNDED_ALLOCATOR_NAME, OF_BOUNDED_ALLOCATOR_DESC, false);
```
mcimadamore, line 260: In this, and other instructions, I wonder if it would be more readable to have helper routine, and rely on class literal/strings more e.g. instead of:
```
            mv.visitMethodInsn(INVOKESTATIC, BINDING_CONTEXT_INTRN, OF_BOUNDED_ALLOCATOR_NAME, OF_BOUNDED_ALLOCATOR_DESC, false);
```
something like:

```
emitInvokeStatic(Binding.Context.class, "ofBoundedAllocator", OF_BOUNDED_ALLOCATOR_DESC);
```

mcimadamore (approved):
Overall looks a very solid change. It is now easier to look at the code and understand what kind of bytecode will be generared for a given downcall/upcall. I've left some optional comments around the code, but I didn't really found anything wrong with the new code.

Other misc comments:
* SharedUtils::wrapWithAllocator seems unused
* some of the combinators (for IMR and acquire/release) are still present in SharedUtils
* there is a lot of coupling between the specializer internal state and the recipe execution (in VMLoad/VMStore especially, but also in others) - this is made clearer by the rewrite, but I appreciate that even the old code had an element of this

src/java.base/share/classes/jdk/internal/foreign/abi/Specializer.java
JornVernee: That's a good idea. I think we can use DirectMethodHandleDesc to describe the method we want to call

src/java.base/share/classes/jdk/internal/foreign/abi/Specializer.java
```diff
@@ -0,0 +1,854 @@
...
+        return idx;
+    }
+
+    private void emitLoadInternalScope() {
+        assert CONTEXT_IDX != -1;
+        emitLoad(BasicType.L, CONTEXT_IDX);
+        mv.visitMethodInsn(INVOKEVIRTUAL, BINDING_CONTEXT_INTRN, SCOPE_NAME, SCOPE_DESC, false);
+    }
+
+    private void emitLoadInteralAllocator() {
```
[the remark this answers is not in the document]
JornVernee: Good catch

src/java.base/share/classes/jdk/internal/foreign/abi/Specializer.java
JornVernee: I tried to make the comments a bit more clear. It's intended as a mental note of the signature of the method that is being invoked in the bytecode below.

src/java.base/share/classes/jdk/internal/foreign/abi/Specializer.java
JornVernee: Done.

src/java.base/share/classes/jdk/internal/foreign/abi/BindingSpecializer.java
```diff
@@ -675,6 +659,26 @@ private static String valueLayoutConstantFor(Class<?> type) {
         }
     }
 
+    private void emitInvokeStatic(String ownerInternalName, String methodName, String descriptor) {
```
mcimadamore, line 662: Any reason as to why using a String for the owner instead of Class<?> - you want to avoid multiple computations, I guess? Having a class literal in the clients is much more readable IMHO than having a string constant.

That said, maybe we can find a compromise :-)

Why not naming the INTRN constants in a regular way, so that it's clear what they refer to?
e.g.
```
JAVA_LANG_INVOKE_MEMORYLAYOUT
```
or, if we are ok dropping the package, just:

```
MEMORY_LAYOUT
```

E.g. drop the INTRN suffix - that should make things more readable.

mcimadamore (approved):

src/java.base/share/classes/jdk/internal/foreign/abi/BindingSpecializer.java
JornVernee: Wasn't done yet :) These methods now take a class literal, and I've removed the unneeded INTRN constants

src/java.base/share/classes/java/lang/foreign/ValueLayout.java
```diff
@@ -29,6 +29,7 @@
 import java.lang.constant.DynamicConstantDesc;
 import java.lang.invoke.VarHandle;
 import java.nio.ByteOrder;
+import java.util.Locale;
```
JornVernee, line 32: Spurious import
```suggestion
```

mcimadamore (approved):
Nice!

PR #638, opened 2022-02-03
8281228: Preview branch's CLinker.downcallHandle crashes inside asm

mcimadamore:
This patch addresses a low level crash occuring in the ASM library, triggered by the new binding specialization logic. After some debugging I realized that there was no real "bug", but something subtle that we missed during review. The binding specializer uses an internal operand stack to keep track of the types maniupulated when processing ABI bindings. Most of the "pop" operation on this stack were done inside an `assert` statement, which meant that the pop operation would not be executed when running without `-esa`. Unfortunately this issue was not caught because the makefile always runs tests with assertions enabled.

After fixing this, I have also verified that the existing test (if ran without assertion enabled) would indeed have been enough to trigger the failure; in other words, the failure has not been detected because of *the way* in which tests were ran, and not because we were lacking in test coverage (for instance, TestDowncall showed 812 failures without the fix). For these reasons I did not include any test in the fix.

---------
### Progress

### Issue
 * [JDK-8281228](https://bugs.openjdk.java.net/browse/JDK-8281228): Preview branch's CLinker.downcallHandle crashes inside asm

### Reviewers
 * [Athijegannathan Sundararajan](https://openjdk.java.net/census#sundar) (@sundararajana - Committer) ⚠️ Review applies to a015dc4173c11ddf0937076e34fb5ed7b649aba5

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/638/head:pull/638` \
`$ git checkout pull/638`

`$ git checkout pull/638` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/638/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 638`

`$ git pr show -t 638`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/638.diff">https://git.openjdk.java.net/panama-foreign/pull/638.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/jdk/internal/foreign/abi/BindingSpecializer.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/abi/BindingSpecializer.java
@@ -150,7 +150,7 @@ public class BindingSpecializer {
 
         new BindingSpecializer(mv, callerMethodType, callingSequence, abi, leafHandle.type()).specialize();
 
-        mv.visitMaxs(0, 0);
+        mv.visitMaxs(-1, -1);
         mv.visitEnd();
 
         cw.visitEnd();
@@ -198,6 +198,28 @@ public class BindingSpecializer {
         }
         return sb.toString();
     }
+
+    // binding operand stack manipulation
+
+    private void pushType(Class<?> type) {
+        typeStack.push(type);
+    }
+
+    private Class<?> popType() {
+        return typeStack.pop();
+    }
+
+    private Class<?> popType(Class<?> expected) {
+        Class<?> found;
+        if ((found = popType()) != expected) {
+            throw new IllegalStateException(
+                    String.format("Invalid type on binding operand stack; found %s - expected %s",
+                            found.descriptorString(), expected.descriptorString()));
+        }
+        return found;
+    }
+    
+    // specialization
 
     private void specialize() {
         // map of parameter indexes to local var table slots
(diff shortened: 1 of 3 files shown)


src/java.base/share/classes/jdk/internal/foreign/abi/BindingSpecializer.java
```diff
@@ -150,7 +150,7 @@ static MethodHandle specialize(MethodHandle leafHandle, CallingSequence callingS
 
         new BindingSpecializer(mv, callerMethodType, callingSequence, abi, leafHandle.type()).specialize();
 
-        mv.visitMaxs(0, 0);
+        mv.visitMaxs(-1, -1);
```
sundararajana, line 153: visitMaxs arguments are ignored with COMPUTE_FRAMES flag, right? any specific reason for -1, -1 instead of 0, 0?

sundararajana (approved):
Should have turn off all assertions (of java.lang.foreign and associated internal package for foreign test runs to catch similar such issues, if any?

src/java.base/share/classes/jdk/internal/foreign/abi/BindingSpecializer.java
mcimadamore: I think you are right. I got confused initially as I thought I've seen another place in this class where -1 was used, and I was trying to experiment making ASM spit the bytecode (and not crash) so that I could look at the actual bytecode :-)

This can be reverted

mcimadamore:
I've added another small fix (spotted by @JornVernee) where the adaptation of a method type was created, and then discarded (in non specializing mode). This is only caught by TestMatrix (which is a manual-only test). When I tried to ran it, several combinations were failing because of the invalid combo of `/manual` and `/timeout`, so I've fixed that too.

JornVernee (approved):

PR #640, opened 2022-02-11
8281595: ASM-ify scope acquire/release for down call parameters

JornVernee:
Hi,

This patch rewrites the scope acquire/release logic that we do for downcalls inside ASM, in the generated specialized binding class. This also drops the fallback path we had for more than 5 scopes. The new code now always checks to see if every scope is unique. (I think that should be okay, but please let me know if otherwise).

I've also added a benchmark by Maurizio to this patch, which was showing a failure to scalar replace some scoped arguments with the old MH logic (reason discussed in [JDK-8281387](https://bugs.openjdk.java.net/browse/JDK-8281387)). With the new ASM-based logic, there is no sign of allocations (other than some noise):

```
Benchmark                                                     Mode  Cnt   Score    Error   Units
PointerInvoke.panama_call_as_address                          avgt   30  13.181 ±  0.320   ns/op
PointerInvoke.panama_call_as_address:·gc.alloc.rate           avgt   30   0.043 ±  0.049  MB/sec
PointerInvoke.panama_call_as_address:·gc.alloc.rate.norm      avgt   30   0.001 ±  0.001    B/op
PointerInvoke.panama_call_as_address:·gc.count                avgt   30     ≈ 0           counts
PointerInvoke.panama_call_as_long                             avgt   30  12.943 ±  0.287   ns/op
PointerInvoke.panama_call_as_long:·gc.alloc.rate              avgt   30   0.065 ±  0.054  MB/sec
PointerInvoke.panama_call_as_long:·gc.alloc.rate.norm         avgt   30   0.002 ±  0.001    B/op
PointerInvoke.panama_call_as_long:·gc.count                   avgt   30     ≈ 0           counts
PointerInvoke.panama_call_as_new_segment                      avgt   30  14.309 ±  0.177   ns/op
PointerInvoke.panama_call_as_new_segment:·gc.alloc.rate       avgt   30   0.173 ±  0.144  MB/sec
PointerInvoke.panama_call_as_new_segment:·gc.alloc.rate.norm  avgt   30   0.005 ±  0.004    B/op
PointerInvoke.panama_call_as_new_segment:·gc.count            avgt   30     ≈ 0           counts
PointerInvoke.panama_call_as_segment                          avgt   30  13.273 ±  0.191   ns/op
PointerInvoke.panama_call_as_segment:·gc.alloc.rate           avgt   30  ≈ 10??           MB/sec
PointerInvoke.panama_call_as_segment:·gc.alloc.rate.norm      avgt   30  ≈ 10??             B/op
PointerInvoke.panama_call_as_segment:·gc.count                avgt   30     ≈ 0           counts
```

Thanks,
Jorn

---------
### Progress

### Issues
 * [JDK-8281595](https://bugs.openjdk.java.net/browse/JDK-8281595): ASM-ify scope acquire/release for down call parameters
 * [JDK-8281387](https://bugs.openjdk.java.net/browse/JDK-8281387): Some downcall shapes show unexpected allocations

### Reviewers

### Contributors
 * Maurizio Cimadamore `<mcimadamore@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/640/head:pull/640` \
`$ git checkout pull/640`

`$ git checkout pull/640` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/640/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 640`

`$ git pr show -t 640`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/640.diff">https://git.openjdk.java.net/panama-foreign/pull/640.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/jdk/internal/foreign/abi/BindingSpecializer.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/abi/BindingSpecializer.java
@@ -251,17 +251,17 @@ public class BindingSpecializer {
 
             // for downcalls we also acquire/release scoped parameters before/after the call
             // create a bunch of locals here to keep track of their scopes (to release later)
-            scopeSlots = new int[callerMethodType.parameterCount()];
+            int[] initialScopeSlots = new int[callerMethodType.parameterCount()];
             int numScopes = 0;
             for (int i = 0; i < callerMethodType.parameterCount(); i++) {
                 if (shouldAcquire(callerMethodType.parameterType(i))) {
                     int scopeLocal = newLocal(BasicType.L);
-                    scopeSlots[numScopes++] = scopeLocal;
+                    initialScopeSlots[numScopes++] = scopeLocal;
                     emitConst(null);
                     emitStore(BasicType.L, scopeLocal); // need to initialize all scope locals here in case an exception occurs
                 }
             }
-            scopeSlots = Arrays.copyOf(scopeSlots, numScopes); // fit to size
+            scopeSlots = Arrays.copyOf(initialScopeSlots, numScopes); // fit to size
             curScopeLocalIdx = 0; // used from emitGetInput
         }
 
--- a/src/java.base/share/classes/jdk/internal/foreign/abi/BindingSpecializer.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/abi/BindingSpecializer.java
@@ -26,6 +26,7 @@ package jdk.internal.foreign.abi;
 
 import jdk.internal.foreign.MemoryAddressImpl;
 import jdk.internal.foreign.ResourceScopeImpl;
+import jdk.internal.foreign.Scoped;
 import jdk.internal.misc.VM;
 import jdk.internal.org.objectweb.asm.ClassReader;
 import jdk.internal.org.objectweb.asm.ClassWriter;
@@ -45,6 +46,7 @@ import java.lang.constant.ConstantDescs;
 import java.lang.foreign.Addressable;
 import java.lang.foreign.MemoryAddress;
 import java.lang.foreign.MemorySegment;
+import java.lang.foreign.NativeSymbol;
 import java.lang.foreign.ResourceScope;
 import java.lang.foreign.SegmentAllocator;
 import java.lang.foreign.ValueLayout;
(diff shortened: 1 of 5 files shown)


src/java.base/share/classes/jdk/internal/foreign/abi/BindingSpecializer.java
```diff
@@ -407,10 +440,61 @@ private void emitSetOutput(Class<?> storeType) {
...
+
+            mv.visitLabel(skipAcquire);
+            mv.visitInsn(POP); // drop scope
+        }
+
+        mv.visitLabel(end);
+    }
+
+    private void emitReleaseScopes() {
+        for (int scopeLocal : scopeSlots) {
```
mcimadamore, line 486: Isn't this emitting more code than required? E.g. if we pass 10 primitives, we will emit 10 scope checks at the end of the adapter? Granted, since all these locals contain "null", there's no correctness issue here - but it is odd to (i) have a local for a scope corresponding to a primitive type (even though we know we never need it) and have extra logic in the preamble/postamble because of it.

mcimadamore (commented):
Looks good - one minor comment on the handling of the local variables associated with scopes.

src/java.base/share/classes/jdk/internal/foreign/abi/BindingSpecializer.java
JornVernee: There is only a scope slot local for Addressable and NativeSymbol parameters. See the logic that initializes `scopeSlots`:

```java
scopeSlots = Arrays.copyOf(scopeSlots, numScopes); // fit to size
```

(the initial array is the size of the parameter list initially, but is adjusted downwards. Maybe that should be made clearer...)

JornVernee: I've tried to clarify the code that creates `scopeSlots` a bit by using a local variable for the intermediate array, instead of storing it in the field.

mcimadamore: Thanks - I've missed that

mcimadamore (approved):

PR #641, opened 2022-02-15
8281855: Rename ResourceScope to MemorySession

mcimadamore:
This patch implements some of the changes described in [1] and [2]. More specifically, the `ResourceScope` abstraction is renamed into `MemorySession`, and the session accessors in `MemorySegment`, `NativeSymbol` and `VaList` are tweaked to return *non-closeable* session views. To counteract this change, `MemorySession` now supports `equals` and `hashCode`, so that sessions views can be compared to each other in ways that are not identity-sensitive.

[1] - https://mail.openjdk.java.net/pipermail/panama-dev/2022-February/016152.html
[2] - https://mail.openjdk.java.net/pipermail/panama-dev/2022-February/016254.html

---------
### Progress

### Issue
 * [JDK-8281855](https://bugs.openjdk.java.net/browse/JDK-8281855): Rename ResourceScope to MemorySession

### Reviewers
 * [Paul Sandoz](https://openjdk.java.net/census#psandoz) (@PaulSandoz - Committer) ⚠️ Review applies to 2ee6f16b95e8763ef3a23df45670d06ce325bcba
 * [Jorn Vernee](https://openjdk.java.net/census#jvernee) (@JornVernee - Committer) ⚠️ Review applies to d1fc22b4061b7a007c4f56f5f0116e47cc4c8aed

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/641/head:pull/641` \
`$ git checkout pull/641`

`$ git checkout pull/641` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/641/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 641`

`$ git pr show -t 641`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/641.diff">https://git.openjdk.java.net/panama-foreign/pull/641.diff</a>

</details>


src/java.base/share/classes/jdk/internal/foreign/Scoped.java
```diff
@@ -25,8 +25,12 @@
 
 package jdk.internal.foreign;
 
-import java.lang.foreign.ResourceScope;
+import java.lang.foreign.MemorySession;
 
 public interface Scoped {
-    ResourceScope scope();
+    MemorySessionImpl sessionImpl();
```
mcimadamore, line 31: This method is used to retrieve the "root" session associated with either a resource, or a session view. This means that, to simplify the implementation, `MemorySessionImpl` has to implement `Scoped` as well (and return itself).

mcimadamore:
I've fixed javadoc references to "scopes" and improved the javadoc text in several places. A link to a javadoc corresponding to these changes can be found here:
http://cr.openjdk.java.net/~mcimadamore/panama/session_preview/javadoc/java.base/java/lang/foreign/package-summary.html

src/java.base/share/classes/java/nio/Direct-X-Buffer.java.template
```diff
@@ -300,13 +300,13 @@ class Direct$Type$Buffer$RW$$BO$
 #if[rw]
 
     public long address() {
-        ResourceScope scope = scope();
+        MemorySession scope = session();
         if (scope != null) {
-            if (scope.ownerThread() == null) {
-                throw new UnsupportedOperationException("ByteBuffer derived from shared segments not supported");
+            if (scope.ownerThread() == null && scope.isCloseable()) {
+                throw new UnsupportedOperationException("ByteBuffer derived from closeable shared segments not supported");
```
JornVernee, line 306: Reviewer note: does user have the power to create non-closeable shared segments now?

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
```diff
@@ -301,9 +300,9 @@ boolean isAligned(MemorySegment segment, long offset, MemoryLayout layout) {
     Stream<MemorySegment> elements(MemoryLayout elementLayout);
 
     /**
-     * {@return the resource scope associated with this memory segment}
+     * {@return a non-closeable view of the memory session associated with this memory segment}
      */
-    ResourceScope scope();
+    MemorySession session();
```
merykitty, line 305: Hi, could we take advantage of the type system and return a MemorySessionView which does not have a close method on it. Thanks.

mcimadamore: This is a possibility which we have explored. In the end we decided against it because we felt splitting session/scope in two would have had a negative effect on API discoverability. Note that non-closeable sessions are not just used for e.g. `MemorySegment::session`, but they are also returned by `MemorySession::global` and `MemorySession::openImplicit`. If we ever do *structured* sessions (e.g. session that give access to any threads that inherit a given Loom scope local), then it is likely that these sessions will be non-closeable too.

Also, you would need to replace most API methods accepting MemorySession with MemorySessionView - and add a subtyping relationhip between MemorySessionView and MemorySession (so that you can pass a MemorySession where a view is expected). Ultimately that all felt a bit too much.

mcimadamore: > Ultimately that all felt a bit too much.

To be clear, it felt too much not because the refactoring was complex (it was not) but because this creates a problem of how do you name the non-closeable entity (MemorySessionView is fine when returned by `MemorySegment::session`, less fine when accepted by `MemorySegment::allocateNative`). And if you give it a more general name, then you end up with two abstractions fighting for API visibility.

merykitty: I see, rethinking about it, the reason to add a `session` accessor to `MemorySegment` was to properly bind the lifetime of an **allocated** segment to an existing one and to correctly observe that lifetime. It sounds like the functionality of an allocator to me. Maybe we could instead return a `SegmentAllocator` instead (and change `MemorySession` to `SegmentDeallocator` probably) and extend the functionality of a `SegmentAllocator` to check the validity of its associating `MemorySession`. Even if it is not the case we would need to check the state of a `SegmentAllocator` anyway to ensure that we can allocate memory with it.
Thanks.

mcimadamore: > hat option was also considered, but discarded because it would make `SegmentAllocator` stateful, e.g. more complex than just a functional interface.

Note that today there is no contract which says that all segments returned by a `SegmentAllocator` must have the same session. While some allocators might behave like that, I think it would not be ok to force that behavior on _all_ allocators. And, we considered to add, perhaps in the future, some small subinterface, like `ScopedSegmentAllocator` which also has a `session` accessor - this would be a good return type for the arena allocator methods, for instance. But this is a move that is orthogonal to the changes described in this PR (and also not a binding one - we can add the sub-interface even at a later point w/o breaking compatibility).

merykitty: Thanks a lot for your clarification, I still think that having a scope that throw on `close()` sounds a lot like the problem with `UnmodifiableCollection` but it seems the alternatives have been considered already.

mcimadamore: > 

Indeed the issues are similar. However, there are some characteristics that are unique to this API. MemorySegment is a lower-level API than, say Collection. Also, while we currently do not generate method handles which require a `MemorySession`, adding intricate subtyping hierarchies always has a cost when interacting with `invokeExact` - so this API often errs on the side of "avoid unnecessary intermediate types, if we can avoid it". Not to dismiss your concern, which is a valid one - just wanted to explain some of the rationale behind some of the API decisions we've been doing.

PaulSandoz (approved):

test/jdk/java/foreign/valist/VaListTest.java
```diff
@@ -121,26 +121,26 @@ private static MethodHandle linkVaListCB(String symbol) {
...
-            = (builder, scope) -> Windowsx64Linker.newVaList(builder, scope);
-    private static final BiFunction<Consumer<VaList.Builder>, ResourceScope, VaList> sysvVaListScopedFactory
-            = (builder, scope) -> SysVx64Linker.newVaList(builder, scope);
-    private static final BiFunction<Consumer<VaList.Builder>, ResourceScope, VaList> linuxAArch64VaListScopedFactory
-            = (builder, scope) -> LinuxAArch64Linker.newVaList(builder, scope);
-    private static final BiFunction<Consumer<VaList.Builder>, ResourceScope, VaList> macAArch64VaListScopedFactory
-            = (builder, scope) -> MacOsAArch64Linker.newVaList(builder, scope);
-    private static final BiFunction<Consumer<VaList.Builder>, ResourceScope, VaList> platformVaListScopedFactory
-            = (builder, scope) -> VaList.make(builder, scope);
+            = (builder) -> VaList.make(builder, MemorySession.openConfined());
```
JornVernee, line 132: Now that the tests no longer close the scope, maybe these scopes should be implicit scopes?

test/micro/org/openjdk/bench/java/lang/foreign/BulkMismatchAcquire.java
```diff
@@ -117,7 +117,9 @@ public void setup() {
 
     @TearDown
     public void tearDown() {
-        scope.close();
+        if (session instanceof MemorySession) {
+            ((MemorySession) session).close();
+        }
```
JornVernee, line 122: Seems like this check is redundant? Or maybe this should check for isCloseable

src/java.base/share/classes/java/lang/foreign/MemorySession.java
```diff
@@ -0,0 +1,256 @@
...
+    void whileAlive(Runnable action);
+
+    /**
+     * Add a custom cleanup action which will be executed when the memory session is closed.
+     * The order in which custom cleanup actions are invoked once the memory session is closed is unspecified.
+     * @param runnable the custom cleanup action to be associated with this memory session.
+     * @throws IllegalStateException if this memory session is not {@linkplain #isAlive() alive}, or if access occurs from
+     * a thread other than the thread {@linkplain #ownerThread() owning} this memory session.
+     */
+    void addCloseAction(Runnable runnable);
```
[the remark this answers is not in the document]
JornVernee: Though, this was pre-existing... so could be done in a followup as well.

src/java.base/share/classes/jdk/internal/foreign/abi/UpcallStubs.java
```diff
@@ -48,13 +49,13 @@ private static void freeUpcallStub(long stubAddress) {
...
-        ((ResourceScopeImpl)scope).addOrCleanupIfFail(new ResourceScopeImpl.ResourceList.ResourceCleanup() {
+    static NativeSymbol makeUpcall(long entry, MemorySession session) {
+        ((MemorySessionImpl)session).addOrCleanupIfFail(new MemorySessionImpl.ResourceList.ResourceCleanup() {
             @Override
             public void cleanup() {
                 freeUpcallStub(entry);
             }
         });
-        return new NativeSymbolImpl("upcall:" + Long.toHexString(entry), MemoryAddress.ofLong(entry), scope);
+        return new NativeSymbolImpl("upcall:" + Long.toHexString(entry), MemoryAddress.ofLong(entry), ((MemorySessionImpl)session));
```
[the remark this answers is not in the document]
mcimadamore: No: `NativeSymbolImpl` wants a true, closeable MemorySessionImpl. Then, if you look at that class, its `session` method will return a non closeable view of that session. The `toSessionImpl` happens in `AbstractCLinker::upcallStub`.

test/jdk/java/foreign/TestRestricted.java
```diff
@@ -58,19 +58,19 @@ public void testDirectAccess() throws Throwable {
...
+        method.invoke(null, MemoryAddress.NULL, 4000L, MemorySession.global());
     }
 
     @Test(expectedExceptions = IllegalCallerException.class)
     public void testInvoke2() throws Throwable {
         var mh = MethodHandles.lookup().findStatic(MemorySegment.class, "ofAddress",
-            MethodType.methodType(MemorySegment.class, MemoryAddress.class, long.class, ResourceScope.class));
-        var seg = (MemorySegment)mh.invokeExact(MemoryAddress.NULL, 4000L, ResourceScope.globalScope());
+            MethodType.methodType(MemorySegment.class, MemoryAddress.class, long.class, MemorySession.class));
+        var seg = (MemorySegment)mh.invoke(MemoryAddress.NULL, 4000L, MemorySession.global());
```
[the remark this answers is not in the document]
mcimadamore: Not sure - probably just a refactoring accident

src/java.base/share/classes/jdk/internal/foreign/abi/UpcallStubs.java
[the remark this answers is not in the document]
JornVernee: Thanks, the comment helps :)

PR #642, opened 2022-02-17
8282026: Remove support for unbound sequence layouts

mcimadamore:
When doing a pass over the memory layout API, I couldn't help noting that the API implementation and specification are made more complex by the fact that we have to specify what happens when a sequence layout *without a size* is found. Support for unbounded sequence layout was added many moons ago, with the understanding that it would have made life for jextract easier.

This is not the case; all cases where jextract is using an unbounded sequence layout today can be replaced with using a simple zero-length sequence (which is allowed by the API, as are empty structs/unions).

I also found that the only real use case for using unbounded sequence layout came from a quirk in the API to obtain var handles from layout. That is, if we want to obtain an indexed var handle into a JAVA_INT, we have to do the following:

```
VarHandle intHandle = MemoryLayout.newSequenceLayout(JAVA_INT)
            .varHandle(PathElement.sequenceLayout());
```

In hindsight, this is just boilerplate: in this case the user only cares about the element layout (`JAVA_INT`), but the API wants the user to wrap that element layout in a new sequence layout, only to then provide a path inside the sequence layout (which will add a free dimension to the resulting var handle).

It is not hard to see that all the above can be simplified, by adding an extra API method on memory layouts:

```
VarHandle intHandle = JAVA_INT.arrayElementVarHandle(JAVA_INT);
```

This more explicit, succint, and provdies less opportunities for bugs to hide.

Therefore, this patch removes support for unbounded sequence element; the net effect is that now _all_ layouts have a size - which significantly simplifies the use sites where layout sizes are computed (no need to use optional there). All without losing much in terms of expressiveness: zero-length sequence layout can be used in basically the same way in which unbounded sequence layouts were used before.

---------
### Progress

### Issue
 * [JDK-8282026](https://bugs.openjdk.java.net/browse/JDK-8282026): Remove support for unbound sequence layouts

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/642/head:pull/642` \
`$ git checkout pull/642`

`$ git checkout pull/642` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/642/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 642`

`$ git pr show -t 642`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/642.diff">https://git.openjdk.java.net/panama-foreign/pull/642.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
+++ b/src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
@@ -432,7 +432,7 @@ public sealed interface MemoryLayout extends Constable permits AbstractLayout, S
      * @throws IllegalArgumentException if the layout path in {@code elements} does not select a value layout (see {@link ValueLayout}).
      */
     default VarHandle arrayElementVarHandle(PathElement... elements) {
-        return computePathOp(LayoutPath.rootPath(this, MemoryLayout::bitSize), LayoutPath::dereferenceHandle,
+        return computePathOp(LayoutPath.elementRootPath(this, MemoryLayout::bitSize), LayoutPath::dereferenceHandle,
                 Set.of(), elements);
     }
 
--- a/src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
+++ b/src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
@@ -420,7 +420,9 @@ public sealed interface MemoryLayout extends Constable permits AbstractLayout, S
 
     /**
      * Creates a <em>strided</em> memory access var handle that can be used to dereference memory at the layout selected by a given layout path,
-     * where the path is considered rooted in this layout. Equivalent to the following code:
+     * where the path is considered rooted in this layout. The returned var handle can effectively dereference multiple memory
+     * locations, using a <em>dynamic</em> index (of type {@code long}), which is multiplied by this layout size and then added
+     * to the offset of the selected layout. Equivalent to the following code:
      * {@snippet lang=java :
      * MemoryLayout.sequenceLayout(Long.MAX_VALUE, this)
      *             .varHandle(PathElement.sequenceElement());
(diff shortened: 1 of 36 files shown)


src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
```diff
@@ -481,6 +485,30 @@ default MethodHandle sliceHandle(PathElement... elements) {
...
 
+    /**
+     * Creates a <em>strided</em> method handle which, given a memory segment, returns a {@linkplain MemorySegment#asSlice(long,long) slice}
+     * corresponding to the layout selected by a given layout path, where the path is considered rooted in this layout.
+     * The returned method handle can effectively slice a given memory segment at multiple starting offsets, using a <em>dynamic</em> index
+     * (of type {@code long}), which is multiplied by this layout size and then added to the offset of the selected layout.
+     * Equivalent to the following code:
+     * {@snippet lang=java :
+     * MemoryLayout.sequenceLayout(Long.MAX_VALUE, this)
+     *             .varHandle(PathElement.sequenceElement());
```
PaulSandoz, line 496:
```suggestion
     * MemoryLayout.sequenceLayout(Long.MAX_VALUE, this)
     *             .sliceHandle(PathElement.sequenceElement());
```
?

src/java.base/share/classes/java/lang/foreign/SequenceLayout.java
```diff
@@ -34,9 +34,9 @@
```
PaulSandoz, line 34: Unused import `java.util.OptionalLong;`, suggest checking other files too.

PaulSandoz (commented):
I don't see any tests for `arrayElementSliceHandle`.

mcimadamore:
After some more thinking I decided to drop `arrayElementSliceHandle` for the time being. The "sliceHandle" method is already quite nichey, and you can always get there by wrapping into a sequence layout and then unwrap with a sequence element path, to get a strided MH.

In other words, I think the API should make it easy to express var handles, which are the most common form - other things should be possible as well (and they are) but not sure they are worth the API estate.

src/java.base/share/classes/jdk/internal/foreign/LayoutPath.java
```diff
@@ -242,7 +238,7 @@ public MemoryLayout map(UnaryOperator<MemoryLayout> op) {
     private MemoryLayout dup(MemoryLayout oldLayout, MemoryLayout newLayout) {
         newLayout = newLayout.withBitAlignment(oldLayout.bitAlignment());
         if (oldLayout.name().isPresent()) {
-            newLayout.withName(oldLayout.name().get());
+            newLayout = newLayout.withName(oldLayout.name().get());
```
JornVernee, line 241: oof, nice catch!

JornVernee (approved):

PR #643, opened 2022-02-17
8282061: Improve support for deterministic closure of shared scopes

mcimadamore:
This patch improves the logic for closing shared resource scopes, by using the following algorithm:

1. move the scope from ALIVE to CLOSED - no new thread can access segments associated with the scope
2. do an initial handshake, to collect all threads that are accessing the scope concurrently
3. if no thread is found, finish
4. if some threads T1, T2 ... Tn are found, keep doing handshakes (only against these threads) then go back to (3). 

Note that the logic no longer require *three* states for shared segment, and also this logic always succeeds - that is the close operation can never fail because of a spurious access found during an handshake.

Note that the logic converges quickly, because handshaked threads are deoptimized - meaning that they will have to re-load the liveness state of the resource they are accessing (at which point they will just throw an exception).

Implementation-wise, when looking over the code with @fisk , we realized that it is possible for multiple threads to run the handhshake closure concurrently. To collect all the problematic thread, we used a lock free stack (which was already implemented in the hotspot code). Also, to keep problematic threads alive during multiple rounds of handshaking, we use a ThreadHandleList (this is also required to be able to handshake on a specific thread).

---------
### Progress

### Issue
 * [JDK-8282061](https://bugs.openjdk.java.net/browse/JDK-8282061): Improve support for deterministic closure of shared scopes

### Reviewers
 * [Paul Sandoz](https://openjdk.java.net/census#psandoz) (@PaulSandoz - Committer) ⚠️ Review applies to 5e26ae25cec3a799fabb8cee01bdd21c20d0bcb5
 * [Jorn Vernee](https://openjdk.java.net/census#jvernee) (@JornVernee - Committer) ⚠️ Review applies to 13306945de01246b556fcdcbe7ddd9516e1863b2

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/643/head:pull/643` \
`$ git checkout pull/643`

`$ git checkout pull/643` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/643/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 643`

`$ git pr show -t 643`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/643.diff">https://git.openjdk.java.net/panama-foreign/pull/643.diff</a>

</details>


src/hotspot/share/prims/scopedMemoryAccess.cpp
```diff
@@ -138,17 +159,41 @@ class CloseScopedMemoryClosure : public HandshakeClosure {
...
+    return;
+  }
+  // now iterate on all problematic threads, until we converge
+  ThreadsListHandle tlh;
+  ThreadStackElement *element = threads.pop();
+  while (element != NULL) {
+    JavaThread* thread = element->_thread;
+    // if the thread is not in the list handle, we can safely skip further handshakes,
+    // as that means that the thread has been created when the scope state has already
+    // been set to CLOSED - meaning that the thread access will fail anyway.
```
JornVernee, line 190: I don't get this comment...

It seems true that 
1. there might be threads in the thread list which are not in the problematic stack.
2. or vice versa, there can be threads in the stack that are not in the thread list.

---

1. seems to be covered by the fact that we are iterating the threads in the thread stack, instead of the ones in the list.
2. It seems that the `includes` check is needed since a thread might have died since the initial handshake, in which case we don't want to handshake it again?

So, I think the comment should read something regarding 2. instead. e.g. 
```suggestion
    // if the thread is not in the list handle, it means that it has died since the initial handshake.
    // in that case we don't need to handshake it again.
```

mcimadamore: The logic here has changed to reflect some internal comments - I will update the comment accordingly

JornVernee (approved):
Nice! Tests also pass on Windows

src/hotspot/share/prims/scopedMemoryAccess.cpp
```diff
@@ -138,17 +159,43 @@ class CloseScopedMemoryClosure : public HandshakeClosure {
 };
 
 /*
- * This function issues a global handshake operation with all
- * Java threads. This is useful for implementing asymmetric
- * dekker synchronization schemes, where expensive synchronization
- * in performance sensitive common paths, may be shifted to
- * a less common slow path instead.
- * Top frames containg obj will be deoptimized.
+ * This functin performs a thread-local handshake against all threads running at the time
```
dholmes-ora, line 162: s/functin/function/

src/hotspot/share/prims/scopedMemoryAccess.cpp
```diff
@@ -138,17 +159,43 @@ class CloseScopedMemoryClosure : public HandshakeClosure {
...
+ * the given scope (deopt) was closed. If the handshake closure finds that a thread has
+ * safepointed inside a scoped method (that is, a method inside the ScopedMemoryAccess class
```
dholmes-ora, line 164: Do you really mean "safepointed" here, or do you just mean the point where the handshake is processed?

src/hotspot/share/prims/scopedMemoryAccess.cpp
```diff
@@ -138,17 +159,43 @@ class CloseScopedMemoryClosure : public HandshakeClosure {
...
+ * annotated with the '@Scoped' annotation), whose local variables mention the scope being closed
+ * (deopt), the thread is added to a problematic stack. After the handshake, each thread in
+ * the problematic stack is handshaked again, individually, to check that it has exited
```
dholmes-ora, line 167: I suggest saying "list" rather  than "stack" as VM developers tend to assume "stack" means thread stack.

src/hotspot/share/prims/scopedMemoryAccess.cpp
```diff
@@ -138,17 +159,43 @@ class CloseScopedMemoryClosure : public HandshakeClosure {
...
+  ThreadStack threads;
+  CloseScopedMemoryClosure cl(deopt, &threads);
+  // do a first handshake and collect all problematic threads
   Handshake::execute(&cl);
-  return !cl._found;
+  if (threads.empty()) {
+    // fast-path: return if no problematic thread is found
+    return;
+  }
+  // now iterate on all problematic threads, until we converge. Note: from this point on,
```
dholmes-ora, line 183: s/now/Now/
s/on/over/

src/hotspot/share/prims/scopedMemoryAccess.cpp
```diff
@@ -138,17 +159,43 @@ class CloseScopedMemoryClosure : public HandshakeClosure {
...
+  // we only need to focus on the problematic threads found in the previous step, as
+  // any new thread created after the initial handshake will see the scope as CLOSED,
+  // and will fail to access memory anyway.
+  ThreadsListHandle tlh;
+  ThreadStackElement *element = threads.pop();
+  while (element != NULL) {
+    JavaThread* thread = element->_thread;
+    // if the thread is not in the list handle, it means that the thread has died,
```
dholmes-ora, line 191: s/if/If/

dholmes-ora (commented):
Hi Maurizio,

A few minor nits on the comments but the handshaking protocol looks good.

Thanks,
David

PR #644, opened 2022-02-17
8282070: Drop workaround from memory segment implementation

mcimadamore:
Following some preliminary experiments (see [1]) it is now finally time to drop the various workarounds we have in the memory segment implementation. This patch removes two workarounds:

* special treatment for "small" segments (e.g. segments whose size fits in an int) - see `MemorySegment::isSmall`; these segments were special cased so that bound checks would only involve `int` comparisons. This has now been fixed in JDK-8276116, and special treatment is no longer required.

* special shortcuts to avoid extra alignment checks. Since when generating var handles from a layout we check most of the alignment constraints (except those involving the base address), in the past we have added a fastpath for these var handles, as C2 had trouble hoisting alignment checks outside of the loops. This has now been fixed in JDK-8277850 and is no longer required.

I have run benchmarks before and after and got identical results. Some benchmarkks in `ParallelSum` got a boost, but those always have noise, due to multi-threading.

---------
### Progress

### Issue
 * [JDK-8282070](https://bugs.openjdk.java.net/browse/JDK-8282070): Drop workaround from memory segment implementation

### Reviewers
 * [Jorn Vernee](https://openjdk.java.net/census#jvernee) (@JornVernee - Committer) ⚠️ Review applies to ea2d79ec918025799a80361f96e96d3de9c0fe69

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/644/head:pull/644` \
`$ git checkout pull/644`

`$ git checkout pull/644` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/644/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 644`

`$ git pr show -t 644`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/644.diff">https://git.openjdk.java.net/panama-foreign/pull/644.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
@@ -60,12 +60,10 @@ public abstract non-sealed class AbstractMemorySegmentImpl implements MemorySegm
 
     private static final ScopedMemoryAccess SCOPED_MEMORY_ACCESS = ScopedMemoryAccess.getScopedMemoryAccess();
 
-    private static final boolean enableSmallSegments =
-            Boolean.parseBoolean(GetPropertyAction.privilegedGetProperty("jdk.internal.foreign.SmallSegments", "true"));
-
     static final int READ_ONLY = 1;
-    static final int SMALL = READ_ONLY << 1;
     static final long NONCE = new Random().nextLong();
+
+    static final int DEFAULT_MODES = 0;
 
     static final JavaNioAccess nioAccess = SharedSecrets.getJavaNioAccess();
 
@@ -87,11 +85,6 @@ public abstract non-sealed class AbstractMemorySegmentImpl implements MemorySegm
     abstract AbstractMemorySegmentImpl dup(long offset, long size, int mask, ResourceScopeImpl scope);
 
     abstract ByteBuffer makeByteBuffer();
-
-    static int defaultAccessModes(long size) {
-        return (enableSmallSegments && size < Integer.MAX_VALUE) ?
-                SMALL : 0;
-    }
 
     @Override
     public AbstractMemorySegmentImpl asReadOnly() {
@@ -351,10 +344,6 @@ public abstract non-sealed class AbstractMemorySegmentImpl implements MemorySegm
         return arr;
     }
 
-    public boolean isSmall() {
-        return isSet(SMALL);
-    }
-
     public void checkAccess(long offset, long length, boolean readOnly) {
         if (!readOnly && isSet(READ_ONLY)) {
             throw new UnsupportedOperationException("Attempt to write a read-only segment");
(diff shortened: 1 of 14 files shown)


src/java.base/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
```diff
@@ -398,18 +387,14 @@ private int checkArraySize(String typeName, int elemSize) {
...
-        } else if (this != NativeMemorySegmentImpl.EVERYTHING) { // oob not possible for everything segment
-            if (
-                    length < 0 ||
-                    offset < 0 ||
-                    offset > this.length - length) { // careful of overflow
-                throw outOfBoundException(offset, length);
-            }
+    @ForceInline
+    void checkBounds(long offset, long length) {
+        if (length > 0) {
```
mcimadamore, line 392: This is the best I could do, given that `Objects.checkIndex` implements a <= semantics on the upper bound. That is, we need to protect the code from overflow, which can happen if `length == 0` (which can happen in slicing). When that happens, the upper bound in the second check `this.length - length + 1` ends up being `this.length + 1`, which is problematic. I've also tested to see if we could just use the old check bounds routine, but it seems like using `Objects.checkIndex` is required to properly apply BCE.

rsmogura: I think `Objects.checkIndex` checks if `length` && `offset` are negative, so there's no risk of overflow (actually it checks if `length` is negative, and then make unsigned compare), so maybe I miss here something.

Just minor, I wonder if it will be better to use `Preconditions.checkIndex` - one method less to inline.

mcimadamore: Consider a case of a segment whose length is Long.MAX_VALUE. Then you want to take a slice of size 0. This will throw an exception, as Long.MAX_VALUE + 1 will flip into negative territory.

Another problem is when the segment has size zero, in which case the check:

```
long checkedOffset = Objects.checkIndex(offset, this.length);
```

Will also fail. To make that pass we'd need to tweak that check to use `this.length + 1` again, and risk overflow. So I figured it was better just just special case when the accessed length was zero (which should happen rarely). Another option would be to have different checks for slicing and for access (since for access length can never be zero) - but then it duplicates the implementation, and there are also methods like `copy` which are left a bit in the middle.

As for inlining, I doubt that it would make a difference to call Preconditions directly, since it's all `@ForceInlined`.

rsmogura: I get it, it's to allow creating 0-length chunks (I've thought in context of reading value).

If we assume `length > 0`, then `this.length - length + 1 =< this.length`, (in case `this.len == 0 -> 0 <= 0`)

src/java.base/share/classes/jdk/internal/foreign/Utils.java
```diff
@@ -152,68 +151,10 @@ public static MemorySegment toCString(byte[] bytes, SegmentAllocator allocator)
...
     }
 
-    /* Helper functions for offset computations. These are required so that we can avoid issuing long opcodes
-     * (e.g. LMUL, LADD) when we're operating on 'small' segments (segments whose length can be expressed with an int).
-     * C2 BCE code is very sensitive to the kind of opcode being emitted, and this workaround allows us to rescue
-     * BCE when working with small segments. This workaround should be dropped when JDK-8259609 is resolved.
-     */
-
     @ForceInline
     public static long scaleOffset(MemorySegment segment, long index, long size) {
```
JornVernee, line 155: Seems like this method is pretty redundant now, and users can just do `index * size`? (at least I think the segment parameter could be dropped).

src/java.base/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
rsmogura: I'm not sure if I missed some case, but if it's in if (`length > 0`), then one `checkIndex` is redundant? Unless we want to disallow slicing 0 length indices?

mcimadamore: @rsmogura you are correct, the extra check is redundant now - we only need the second. I probably added before I had the check for `length > 0` and then left that there. Effectively, the second check is enough, because that checks that `offset >= 0`,  and `length > 0` by construction (and we now that segment size must be `>= 0`). So I believe we should be ok.

@merykitty yes, this is implementing the same check in spirit as `checkFromIndexSize` - but unfortunately that is not a VM intrinsics, so we have to code around that.

PaulSandoz (approved):

PR #651, opened 2022-02-22
8276648: Downcall stubs are never freed

JornVernee:
Hi,

The current code adds native invokers (downcall stubs) to a CocurrentHashMap cache, and they stay there until the application exits.

This patch replaces that cache with a SoftReferenceCache (brought to the top level from AbstractLinker), which allows at least the values to be evicted when they become unreachable.

Those value, which are NativeEntryPoint instances, are also registered with a cleaner, whose cleanup action will free the stub from the code cache.

Thanks,
Jorn

### DIFF
--- a/src/java.base/share/classes/jdk/internal/foreign/abi/NativeEntryPoint.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/abi/NativeEntryPoint.java
@@ -25,11 +25,12 @@
 
 package jdk.internal.foreign.abi;
 
+import jdk.internal.ref.CleanerFactory;
+
 import java.lang.invoke.MethodType;
+import java.lang.ref.Cleaner;
 import java.util.Arrays;
 import java.util.List;
-import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * This class describes a 'native invoker', which is used as an appendix argument to linkToNative calls.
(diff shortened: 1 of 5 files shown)


src/java.base/share/classes/jdk/internal/foreign/abi/NativeEntryPoint.java
```diff
@@ -63,16 +65,25 @@ public static NativeEntryPoint make(ABIDescriptor abi,
         assert (!needsReturnBuffer || methodType.parameterType(1) == long.class) : "return buffer address expected";
 
         CacheKey key = new CacheKey(methodType, abi, Arrays.asList(argMoves), Arrays.asList(returnMoves), needsReturnBuffer);
-        long invoker = INVOKER_CACHE.computeIfAbsent(key, k ->
-            makeInvoker(methodType, abi, argMoves, returnMoves, needsReturnBuffer));
-
-        return new NativeEntryPoint(methodType, invoker);
+        return INVOKER_CACHE.get(key, k -> {
```
mcimadamore, line 68: This uses a combination of soft references (cache) and phantom references (cleaner). So, cache entries will be evicted when the NEP is softly reachable - meaning that the GC will clear the entries when under pressure. But the cleaner won't be able to run until the entry is evicted (because the entry has a soft reference to the NEP, and to be phantom-reachable you need NOT to be soft-reachable). Assuming that's what you want, then it looks ok. E.g. the cleaner here will "just" make sure that some action is executed when the entry is evicted.

mcimadamore (approved):
Looks good

src/java.base/share/classes/jdk/internal/foreign/abi/NativeEntryPoint.java
JornVernee: Yes, that is the intent. We want to keep the downcall stub around until after the NEP is evicted from the cache. The cleanup is just there to make sure the VM stub is freed as well, and it is tied to the reachability of the associated NEP, which is strongly reachable from a downcall method handle.

PR #658, opened 2022-03-01
8282476: Drop NativeSymbol

mcimadamore:
This change drops the `NativeSymbol` interface from the API, for the following reasons:

* A `NativeSymbol` is essentially a `MemorySegment` w/o a length - following JDK-8280460, zero-length memory segment are better supported and can be used instead.

* Whenever the API takes a `NativeSymbol` (e.g. in linker downcall handles) we can take `Addressable` and have a more flexible API

* The extra level of safety added by `NativeSymbol` was questionable for such a low level API. The only difference between a `NativeSymbol` and a `MemorySegment` is the fact that the former does not support slicing and dereference. But the same is true for a zero-length segment. Also, while `NativeSymbol` made it harder for developers to pass wrong types to linker method handles (e.g. a `VaList`), it is still possible to unsafely wrap bad data into a `NativeSymbol` (e.g. the API implementation won't really check that the address used to construct a new `NativeSymbol` is really that of a "symbol"). And, `NativeSymbol` also made it harder for developers to pass raw native addresses to the linker (such addresses needed to be wrapped into a symbol instead).

Ultimately, the concept of "symbol" was a bit too fuzzy: we came to it mostly because it made sense as a return type for lookups - but it doesn't ultimately add anything to the API, other than providing two different ways to do the same thing.

---------
### Progress

### Issue
 * [JDK-8282476](https://bugs.openjdk.java.net/browse/JDK-8282476): Drop NativeSymbol

### Reviewers
 * [Jorn Vernee](https://openjdk.java.net/census#jvernee) (@JornVernee - Committer) ⚠️ Review applies to d849d6ada07af70799bb1494c9132c55bb58daee

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/658/head:pull/658` \
`$ git checkout pull/658`

`$ git checkout pull/658` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/658/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 658`

`$ git pr show -t 658`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/658.diff">https://git.openjdk.java.net/panama-foreign/pull/658.diff</a>

</details>


mcimadamore:
Javadoc for the changes here:
http://cr.openjdk.java.net/~mcimadamore/panama/no_native_symbol/javadoc/java.base/java/lang/foreign/package-summary.html

JornVernee (approved):

PaulSandoz (commented):
There are still a few references to "native symbol" in the `CLinker` docs, which i think need to refer to the function's address instead.

The JavaDoc of `ClassLoader.findNative` could start with:
```
Finds the address of a native symbol...
```

PaulSandoz (approved):

PR #661, opened 2022-03-03
8282626: Revert semantic of session accessors

mcimadamore:
This patch reverts the hook to make session accessors return a non-closeable session view.

This hook makes it harder to use segments that are backed by a one-off session; consider this case:

```java
SegmentAllocator malloc = MemorySegment::allocateNative;
MemorySegment segment = malloc.allocate(100); // no way to close this
```

In other words, the asymmetry we added to the API requires clients to perform a lot of tracking to be able to associate segments back to their sessions, creating an obvious scalability problem.

Moreover, the existing hook only protects against close. Clients of a segment can still allocate on a session they might not supposed to, or make the session uncloseable, by registering an infinite close action on it.

At the same time, when looking at how sessions might be used by APIs, usage patterns seem to fall in two categories:

* **External session**. APIs that need to create a resource (e.g. a matrix) will take as input a session, which indicates the lifecycle associated with the resource they want to create. In this case, securing the session is not necessary, as the API is getting the session from somewhere else. It is the *client* of the API that has to trust the API not to mess up with its session (but if the API does something shady with the client session, that's arguably a bug in the API). Code generated by jextract falls in this category, as well as [pooled allocators](https://github.com/openjdk/panama-foreign/pull/509).

* **Internal session**. Some APIs might want to hide sessions from clients; in such cases, the session will be created in the constructor, along with all the resources associated with it. In these cases, as the API has already provided some encapsulation, it seems sensible for the API to keep encapsulating resources associated with the session created internally. The code I've seen for both Lucene and Netty falls in this category.

In other words, making memory segments more principled only provides real benefit for APIs that combine both approaches; so far we don't have much evidence that such APIs exist (but if you have some use cases, we'd be interested to know!).

Considering all this, it makes sense to revert the API so that session accessor retain full powers.

This patch leaves the `isCloseable` predicate (still useful to query if a session can be used) as well as `equals/hashcode` (as moving away from `==` will keep our options more open).

---------
### Progress

### Issue
 * [JDK-8282626](https://bugs.openjdk.java.net/browse/JDK-8282626): Revert semantic of session accessors

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/661/head:pull/661` \
`$ git checkout pull/661`

`$ git checkout pull/661` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/661/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 661`

`$ git pr show -t 661`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/661.diff">https://git.openjdk.java.net/panama-foreign/pull/661.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/java/lang/foreign/MemorySession.java
+++ b/src/java.base/share/classes/java/lang/foreign/MemorySession.java
@@ -276,7 +276,7 @@ public sealed interface MemorySession extends AutoCloseable, SegmentAllocator pe
      * @return a non-closeable shared memory session, managed by a private {@link Cleaner} instance.
      */
     static MemorySession openImplicit() {
-        return MemorySessionImpl.createShared(CleanerFactory.cleaner()).asNonCloseable();
+        return MemorySessionImpl.createImplicit();
     }
 
     /**
--- a/src/java.base/share/classes/java/lang/foreign/MemorySession.java
+++ b/src/java.base/share/classes/java/lang/foreign/MemorySession.java
@@ -120,7 +120,7 @@ import jdk.internal.ref.CleanerFactory;
  *
  * {@snippet lang=java :
  * MemorySession session = MemorySession.openConfined();
- * MemoryLifetime nonCloseableSession = session.asNonCloseable();
+ * MemorySession nonCloseableSession = session.asNonCloseable();
  * MemorySegment segment = MemorySegment.allocateNative(100, nonCloseableSession);
  * segment.session().close(); // throws
  * session.close(); //ok
@@ -189,9 +189,8 @@ public sealed interface MemorySession extends AutoCloseable, SegmentAllocator pe
     void close();
 
     /**
-     * Returns a non-closeable view of this memory session. The returned session is the same session as this
-     * session, if this session is {@linkplain #isCloseable() non-closeable}, or a new non-closeable view of
-     * this memory session.
+     * Returns a non-closeable view of this memory session. If this session is {@linkplain #isCloseable() non-closeable},
+     * this session is returned. Otherwise, this method returns a new non-closeable view of this memory session.
      * @apiNote a non-closeable view of a memory session {@code S} keeps {@code S} reachable. As such, {@code S}
      * cannot be closed implicitly (e.g. by a {@link Cleaner}) as long as one or more non-closeable views of {@code S}
      * are reachable.
--- a/src/java.base/share/classes/java/lang/foreign/MemorySession.java
+++ b/src/java.base/share/classes/java/lang/foreign/MemorySession.java
@@ -196,6 +196,23 @@ public sealed interface MemorySession extends AutoCloseable, SegmentAllocator pe
     MemorySession asNonCloseable();
 
     /**
+     * Compares the specified object with this memory session for equality. Returns {@code true} if and only if the specified
+     * object is also a memory session, and it refers to the same memory session as this memory session. This method
+     * is especially useful when operating on {@linkplain #asNonCloseable() non-closeable} session views.
+     *
+     * @param that the object to be compared for equality with this memory session.
+     * @return {@code true} if the specified object is equal to this memory session.
+     */
+    @Override
+    boolean equals(Object that);
+
+    /**
+     * {@return the hash code value for this memory session}
+     */
+    @Override
+    int hashCode();
+
+    /**
      * Allocates a new native segment, using this session. Equivalent to the following code:
      * {@snippet lang=java :
      * MemorySegment.allocateNative(size, align, this);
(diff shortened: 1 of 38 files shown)


test/jdk/java/foreign/TestSegments.java
```diff
@@ -129,26 +129,35 @@ public void testSmallSegmentMin() {
...
-                () -> SegmentFactory.ofArray(() -> MemorySegment.ofArray(new char[] {'a', 'b', 'c', 'd' })),
-                () -> SegmentFactory.ofArray(() -> MemorySegment.ofArray(new double[] { 1d, 2d, 3d, 4d} )),
-                () -> SegmentFactory.ofArray(() -> MemorySegment.ofArray(new float[] { 1.0f, 2.0f, 3.0f, 4.0f })),
-                () -> SegmentFactory.ofArray(() -> MemorySegment.ofArray(new int[] { 1, 2, 3, 4 })),
-                () -> SegmentFactory.ofArray(() -> MemorySegment.ofArray(new long[] { 1l, 2l, 3l, 4l } )),
-                () -> SegmentFactory.ofArray(() -> MemorySegment.ofArray(new short[] { 1, 2, 3, 4 } )),
-                () -> SegmentFactory.ofImplicitSession(session -> MemorySegment.allocateNative(4, 8, session)),
-                () -> SegmentFactory.ofImplicitSession(session -> MemorySegment.allocateNative(JAVA_INT, session)),
-                () -> SegmentFactory.ofImplicitSession(session -> MemorySegment.allocateNative(4, session))
+        List<Supplier<MemorySegment>> l = List.of(
```
mcimadamore, line 146: You will see changes like these in tests - basically I've reverted the test to a simpler form (as per before the API change being reverted here).

mcimadamore:
I've updated the patch so that we retain some of the spirit of the original goal, but without making the API less usable as a result. `MemorySession` now has  a new method, namely `asNonCloseable` which returns a non-closeable view of the session. An API is then free to create a non-closeable view, and returns segments based on that, if it so wishes (e.g. to prevent calls to `close()` from external clients). This seems like an acceptable compromise between API surface and flexibility.

If we need to restrict more (e.g. have session views which don't support anything, like allocation, or close action), we might add another "unmodifiable" view - but doing so would at least require another method in MemorySegment and VaList to "rebase" the resources onto a new unmodifiable session (since it would not be possible to create resources in the unmodifiable session to begin with!).

mcimadamore:
> If we need to restrict more (e.g. have session views which don't support anything, like allocation, or close action), we might add another "unmodifiable" view - but doing so would at least require another method in `MemorySegment` and `VaList` to "rebase" the resources onto a new unmodifiable session (since it would not be possible to create resources in the unmodifiable session to begin with!).

A slightly simpler way to do this, would be to provide some method in `MemorySegment` which would return another `MemorySegment` whose `session` accessor throws (or make the method return `Optional`). This will effectively give a memory segment that cannot be used to go back at the session capabilities, and therefore cannot be used to close, allocate, or add cleanup actions.

src/java.base/share/classes/java/lang/foreign/MemorySession.java
```diff
@@ -101,13 +102,33 @@
...
+ * There are situations in which it might not be desirable for a memory session to be reachable from one or
+ * more resources associated with it. For instance, an API might create a private memory session, and allocate
+ * a memory segment, and then expose one or more slices of this segment to its clients. Since the API's memory session
+ * would be reachable from the slices (using the {@link MemorySegment#session()} accessor), it might be possible for
+ * clients to compromise the API (e.g. by closing the session prematurely). To avoid leaking private memory sessions
+ * to untrusted clients, an API can instead return segments based on a non-closeable view of the session it created, as follows:
+ *
+ * {@snippet lang=java :
+ * MemorySession session = MemorySession.openConfined();
+ * MemoryLifetime nonCloseableSession = session.asNonCloseable();
```
PaulSandoz, line 123:
```suggestion
 * MemorySession nonCloseableSession = session.asNonCloseable();
```

src/java.base/share/classes/java/lang/foreign/MemorySession.java
```diff
@@ -167,11 +188,21 @@
      */
     void close();
 
+    /**
+     * Returns a non-closeable view of this memory session. The returned session is the same session as this
```
PaulSandoz, line 192: By "same" do you mean sames as in `==` rather than `equals`?

Might be clearer to say:
"
If this session is non-closeable then this session is returned.
Otherwise, ...
In the latter case the returned session will be equal to this session as specified by MemorySession.equals.
"
?

PaulSandoz (approved):

PR #667, opened 2022-03-10
8282953: Drop MemoryLayout::map

mcimadamore:
This patch removes the `MemoryLayout::map` method. This method was introduced to add some support for VLA (back when we had unbounded sequence layouts).

This method is quite limited when it comes to define transformations, as it can only alter one layout per call. As such, it is not useful to define global transformations to e.g. flip the endianness of all value layouts nested in a given layout.

We believe we can do much better than this (and have few ideas floating around) - but for the time being we'd like not to take the "good" name for a method that just doesn't seem all that useful.

When removing the method, I realized that the `LayoutPath` class had a lot of unneeded complexity:

* the size function is no longer needed, given that all layouts have a size
* keeping track of selected indices while navigating layout was only useful for the map operation, so I removed it
* we had two static methods `addStride` and `addScaledOffset` which did same thing; I dropped the first - the duplication was introduced when we dropped workarounds for "small" segments, in fact the leading `MemorySegment` parameter in the first method is now unused;
* I have greatly simplified the logic to create var handles from layouts; there is no need to do permutations (a similar simpler scheme was also used in the logic for computing offset method handles)

---------
### Progress

### Issue
 * [JDK-8282953](https://bugs.openjdk.java.net/browse/JDK-8282953): Drop MemoryLayout::map

### Reviewers
 * [Paul Sandoz](https://openjdk.java.net/census#psandoz) (@PaulSandoz - Committer) ⚠️ Review applies to b0d0f4393ecba0fa218a8b898bdccddd76c21304

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/667/head:pull/667` \
`$ git checkout pull/667`

`$ git checkout pull/667` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/667/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 667`

`$ git pr show -t 667`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/667.diff">https://git.openjdk.java.net/panama-foreign/pull/667.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/jdk/internal/foreign/LayoutPath.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/LayoutPath.java
@@ -189,7 +189,7 @@ public class LayoutPath {
     // Layout path construction
 
     public static LayoutPath rootPath(MemoryLayout layout) {
-        return new LayoutPath(layout, 0L, EMPTY_STRIDES,null);
+        return new LayoutPath(layout, 0L, EMPTY_STRIDES, null);
     }
 
     private static LayoutPath nestedPath(MemoryLayout layout, long offset, long[] strides, LayoutPath encl) {
--- a/src/java.base/share/classes/jdk/internal/foreign/LayoutPath.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/LayoutPath.java
@@ -138,7 +138,7 @@ public class LayoutPath {
         checkAlignment(this);
 
         VarHandle handle = Utils.makeSegmentViewVarHandle(valueLayout);
-        for (int i = strides.length - 1; i >=0; i--) {
+        for (int i = strides.length - 1; i >= 0; i--) {
             MethodHandle collector = MethodHandles.insertArguments(MH_ADD_SCALED_OFFSET, 2,
                     Utils.bitsToBytesOrThrow(strides[i], IllegalArgumentException::new));
             // (J, ...) -> J to (J, J, ...) -> J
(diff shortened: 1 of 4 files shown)


src/java.base/share/classes/jdk/internal/foreign/LayoutPath.java
```diff
@@ -215,42 +186,14 @@ public MemoryLayout layout() {
...
-        }
-        return newLayout;
-    }
-
     // Layout path construction
 
-    public static LayoutPath rootPath(MemoryLayout layout, ToLongFunction<MemoryLayout> sizeFunc) {
-        return new LayoutPath(layout, 0L, EMPTY_STRIDES, -1, null, sizeFunc);
+    public static LayoutPath rootPath(MemoryLayout layout) {
+        return new LayoutPath(layout, 0L, EMPTY_STRIDES,null);
```
PaulSandoz, line 192:
```suggestion
        return new LayoutPath(layout, 0L, EMPTY_STRIDES, null);
```

src/java.base/share/classes/jdk/internal/foreign/LayoutPath.java
```diff
@@ -154,28 +137,16 @@ public VarHandle dereferenceHandle() {
...
-            handle = MethodHandles.collectCoordinates(handle, 1 + i,
-                    MethodHandles.insertArguments(ADD_STRIDE, 1, Utils.bitsToBytesOrThrow(strides[strides.length - 1 - i], IllegalStateException::new))); // MS, long, MS_n, long_n, long
-        }
-        //add offset
-        handle = MethodHandles.insertCoordinates(handle, 1 + strides.length, Utils.bitsToBytesOrThrow(offset, IllegalStateException::new));
-
-        if (strides.length > 0) {
-            // remove duplicate MS args
-            handle = MethodHandles.permuteCoordinates(handle, expectedCoordinates, perms.stream().mapToInt(i -> i).toArray());
+        for (int i = strides.length - 1; i >=0; i--) {
```
PaulSandoz, line 141:
```suggestion
        for (int i = strides.length - 1; i >= 0; i--) {
```

PaulSandoz (approved):

PR #668, opened 2022-03-10
8281412: MemorySegment::map should live in FileChannel

FrauBoes:
This change moves the factory for mapped memory segments to `FileChannel`, as a new instance method `map`, which moves it closer to the existing `FileChannel::map` that returns a `MappedByteBuffer`.  

Tests are adjusted accordingly and a new test for custom file channels (that do/don't override the method) is added.

---------
### Progress

### Issue
 * [JDK-8281412](https://bugs.openjdk.java.net/browse/JDK-8281412): MemorySegment::map should live in FileChannel

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/668/head:pull/668` \
`$ git checkout pull/668`

`$ git checkout pull/668` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/668/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 668`

`$ git pr show -t 668`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/668.diff">https://git.openjdk.java.net/panama-foreign/pull/668.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/java/nio/channels/FileChannel.java
+++ b/src/java.base/share/classes/java/nio/channels/FileChannel.java
@@ -1017,30 +1017,32 @@ public abstract class FileChannel
      * @return A new mapped memory segment.
      *
      * @throws IllegalArgumentException
-     *         If {@code bytesOffset < 0}, {@code bytesSize < 0},
-     *         or if {@code path} is not associated with the default file system.
+     *         If {@code offset < 0}, {@code size < 0} or
+     *         {@code offset + size < 0}.
      *
      * @throws IllegalStateException
-     *         If {@code session} is not
+     *         If the {@code session} is not
      *         {@linkplain MemorySession#isAlive() alive}, or if access occurs
      *         from a thread other than the thread
-     *         {@linkplain MemorySession#ownerThread() owning} {@code session}.
+     *         {@linkplain MemorySession#ownerThread() owning} the
+     *         {@code session}.
+     *
+     * @throws NonReadableChannelException
+     *         If the {@code mode} is {@link MapMode#READ_ONLY READ_ONLY} or
+     *         an implementation specific map mode requiring read access
+     *         but this channel was not opened for reading.
+     *
+     * @throws NonWritableChannelException
+     *         If the {@code mode} is {@link MapMode#READ_WRITE READ_WRITE}.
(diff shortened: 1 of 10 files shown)


test/micro/org/openjdk/bench/java/lang/foreign/LoopOverNonConstantMapped.java
```diff
@@ -95,7 +95,8 @@ public void setup() throws IOException {
             }
             ((MappedByteBuffer)byteBuffer).force();
         }
-        segment = MemorySegment.mapFile(tempPath, 0L, ALLOC_SIZE, FileChannel.MapMode.READ_WRITE, MemorySession.openConfined());
+        FileChannel fileChannel = FileChannel.open(tempPath, StandardOpenOption.READ, StandardOpenOption.WRITE);
```
mcimadamore, line 98: This should be closed in a tearDown method?

src/java.base/share/classes/java/nio/channels/FileChannel.java
```diff
@@ -967,6 +970,87 @@ public String toString() {
...
+     *         If {@code bytesOffset < 0}, {@code bytesSize < 0},
+     *         or if {@code path} is not associated with the default file system.
+     *
+     * @throws IllegalStateException
+     *         If {@code session} is not
+     *         {@linkplain MemorySession#isAlive() alive}, or if access occurs
+     *         from a thread other than the thread
+     *         {@linkplain MemorySession#ownerThread() owning} {@code session}.
+     *
+     * @throws UnsupportedOperationException
```
mcimadamore, line 1029: I think the javadoc might have gotten out of sync with that of FileChannel::map -> ByteBuffer. In the ByteBuffer map method there is no SecurityException anymore, and there is a NonReadableChannelException and NonWriteableChannelException which can also be thrown. I would suspect that this method can throw exactly the same exceptions in exactly the same ways. These exception ate triiggered by `checkMode` which is called by `mapInternal`, so I'd suspect that they can occur with the memory segment map too.

mcimadamore (commented):
Some javadoc tweaks are probably required.

mcimadamore (approved):
Looks good!

PR #671, opened 2022-03-15
8283175: Add equals/hashcode to MemorySegment

mcimadamore:
When looking at the API, we realized that `MemorySegment` is the only class not to expose equals/hashcode (`MemoryLayout`, `MemoryAddress` and `MemorySession` do). This makes it difficult to e.g. use a `MemorySegment` as a map key.

This patch adds equals/hashcode; the behavior is that of a shallow comparison - e.g. equals makes sure that the two segments point to the same memory region and have same spatial and temporal bounds. Read-only status is ignored by the comparison (on the basis that the segment is still pointing to the same memory region). And, since comparing memory session uses `MemorySession::equals`, segments backed by different "views" of the same session will also turn out to be equal (which also makes sense).

To perform a deeper comparison, clients can go through the `MemorySegment::mismatch` path, but I don't think that's a useful implementation for equals/hashCode (even though that's how byte buffers behave - but then, byte buffers do not provide a mismatch method).

When fixing this I realized that the equals method for `FunctionDescriptor` did not take into account the variable arity index, so I've fixed that too (and also tweaked the toString method to take that into account as well). Doing this revealed an issue in some of the calling convention tests, where we passed a variable arity descriptor as the expected descriptor, where in reality the calling convention "flattens" the variable arity descriptor into a specialized, fixed-arity one.

---------
### Progress

### Issue
 * [JDK-8283175](https://bugs.openjdk.java.net/browse/JDK-8283175): Add equals/hashcode to MemorySegment

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/671/head:pull/671` \
`$ git checkout pull/671`

`$ git checkout pull/671` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/671/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 671`

`$ git pr show -t 671`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/671.diff">https://git.openjdk.java.net/panama-foreign/pull/671.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/java/lang/foreign/FunctionDescriptor.java
+++ b/src/java.base/share/classes/java/lang/foreign/FunctionDescriptor.java
@@ -186,8 +186,9 @@ public sealed class FunctionDescriptor implements Constable permits FunctionDesc
      * Compares the specified object with this function descriptor for equality. Returns {@code true} if and only if the specified
      * object is also a function descriptor, and all the following conditions are met:
      * <ul>
-     *     <li>the two function descriptors have equals return layouts (see {@link MemoryLayout#equals(Object)}), or both have no return layout</li>
-     *     <li>the two function descriptors have argument layouts that are pair-wise equal (see {@link MemoryLayout#equals(Object)})
+     *     <li>the two function descriptors have equals return layouts (see {@link MemoryLayout#equals(Object)}), or both have no return layout;</li>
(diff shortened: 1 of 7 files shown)


sundararajana (approved):
LGTM

PR #24, opened 2020-02-19
8239345: need an enum for standard C types and a way to get a memory layout for those C types

sundararajana:
SystemABI.Type enum added and ABI_CLASS attribute of layout removed (inferred from Type instead)

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Change must be properly reviewed

### Issue
 * [JDK-8239345](https://bugs.openjdk.java.net/browse/JDK-8239345): need an enum for standard C types and a way to get a memory layout for those C types

### Reviewers
 * Jorn Vernee ([jvernee](@JornVernee) - Committer)
 * Maurizio Cimadamore ([mcimadamore](@mcimadamore) - Committer)

### Download
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/24/head:pull/24`
`$ git checkout pull/24`

PR #676, opened 2022-04-01
8284041: Explicitly reject unsupported call shapes on macos-aarch64

JornVernee:
Hi,

This patch explicitly rejects calls shapes that we can not link correctly on macos-aarch64, rather than silently linking them, and causing potential issues when the actual call later occurs.

This also means that some tests now fail explicitly instead of silently passing (such as StdLibTest's printf test, for which we don't check the output). These tests have been added to the problem list.

I've also split TestDowncall into 2 separate test classes, one for normal downcall tests (TestDowncallScope), and one for downcall tests that use stack arguments (TestDowncallStack). This means that only the latter will need to be problem listed.

Thanks,
Jorn

---------
### Progress

### Issue
 * [JDK-8284041](https://bugs.openjdk.java.net/browse/JDK-8284041): Explicitly reject unsupported call shapes on macos-aarch64

### Reviewers
 * [Nick Gasson](https://openjdk.java.net/census#ngasson) (@nick-arm - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/676/head:pull/676` \
`$ git checkout pull/676`

`$ git checkout pull/676` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/676/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 676`

`$ git pr show -t 676`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/676.diff">https://git.openjdk.java.net/panama-foreign/pull/676.diff</a>

</details>


src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/CallArranger.java
```diff
@@ -183,8 +188,19 @@ public StorageCalculator(boolean forArguments) {
 
         VMStorage stackAlloc(long size, long alignment) {
             assert forArguments : "no stack returns";
-            alignment = Math.max(alignment, STACK_SLOT_SIZE);
-            stackOffset = Utils.alignUp(stackOffset, alignment);
+            // Implementation limit: each arg must take up at least an 8 byte stack slot (on the Java side)
+            // There is currently no way to address stack offsets that are not multiples of 8 bytes
+            // The VM can only address multiple-of-4-bytes offsets, which is also not good enough for some ABIs
+            // see JDK-8283462 and related issues
+            long stackSlotAlignment = Math.max(alignment, STACK_SLOT_SIZE);
```
mcimadamore, line 195: Reading this fix makes me wonder a bit. In reality size and alignment are determined by the layout. So a user could, in principle, say "JAVA_INT.withBitAlignment(64)" and this check would be satisfied. I guess what I'm trying to get at is that the set of allowed size/alignment combos is set by the platform ABI. For instance, SysV specifies how to pass a 4-byte long, 4-byte aligned, little-endian `int` - but does not specify what happens if e.g. the alignment constraints of the value are different from that described in the ABI document. Now, in the case of SysV, the only case where alignment matters is for struct classification (structs that have unaligned fields are always passed in memory). But in the case of the MacOS ABI, altering the layout alignment of one or more value can result in subtle differences in how the function is invoked - and if the alignment constraint specified by the user does not agree with that of the ABI, while the linker will do _something_, it will almost certainly lead to an error.

So, I believe that, in addition to this, for all the ABIs that have special alignment-dependent treatment, we should apply some checks, to make sure that the layouts in the function descriptor conforms to the alignments specified in the ABI. What do you think?

JornVernee: Yes that makes sense, good catch!

We need to do the same on x64 platforms, since we use the same logic to align stack values. So, for instance, someone could have a hyper-aligned `long` aligned to 128 bits, which would mean it is placed at the wrong point on the stack.

JornVernee:
Looking some more at what the compiler does with this, it seems that for varargs, the stack args _are_ alligned to 8 bytes: https://godbolt.org/z/4x3PTW9x5

Which would explain why TestVarArgs.java (which also verifies the result in a problematic case) was passing. I guess I need to tinker with the fix a little more to allow varargs to be passed. (and probably add some more test cases to that test).

mcimadamore:
> Looking some more at what the compiler does with this, it seems that for varargs, the stack args _are_ alligned to 8 bytes: https://godbolt.org/z/4x3PTW9x5
> 
> For instance, the code for reading an int from a va_list:
> 
> ```
> ldr     x8, [sp, #24]
> ldr     w10, [x8]
> add     x8, x8, #8                      ; =8
> str     x8, [sp, #24]
> str     w10, [sp, #16]
> ldr     w10, [sp, #16]
> str     w10, [sp, #20]
> ```
> 
> `[sp, #24]` looks like the va_list pointer, which is then bumped by `8` every time an int is read (even though an int is only 4 bytes).
> 
> Which would explain why TestVarArgs.java (which also verifies the result in a problematic case) was passing. I guess I need to tinker with the fix a little more to allow varargs to be passed. That should hopefully mean we can un-problem list StdLibTest and TestVarArgs.

This seems consistent with what said here:

https://developer.apple.com/documentation/xcode/writing-arm64-code-for-apple-platforms

```
For functions that contain a variable number of parameters, Apple initializes the relevant registers (Stage A) and determines how to pad or extend arguments (Stage B) as usual. When it’s time to assign arguments to registers and stack slots, Apple platforms use the following rules for each variadic argument:

    Round up the Next SIMD and Floating-point Register Number (NSRN) to the next multiple of 8 bytes.

    Assign the variadic argument to the appropriate number of 8-byte stack slots.

Because of these changes, the type va_list is an alias for char*, and not for the struct type in the generic procedure call standard.
```

mcimadamore:
Re. problem lists - honestly, if these tests are passing now, that would seem to suggest that the patch is disallowing too much stuff. Reading the macos ABI section again, it almost seems to suggest that VaList is 8-byte aligned, while arguments spilled on the stack are packed.

JornVernee:
varargs need to be allowed through, yes, but the tests passing now is more a case of not checking the faulty arguments. e.g. for TestDowncallStack, we always return the first argument, which can never be mis-aligned. It might be interesting to change that test to return the last argument instead, since it depends more on correct alignment computations.

JornVernee:
I've converted the PR to a draft for now, while I investigate the varargs thing further. Thanks for the comments so far.

mcimadamore:
> I've converted the PR to a draft for now, while I investigate the varargs thing further. Thanks for the comments so far.

Good point on TestDowncall returning first argument - that is indeed a big life saver in a lot of cases for these thorny issues. I agree returning the last arg would beef up tests a lot (even on non arm64/mac).

src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/CallArranger.java
JornVernee: Looking more at this. I think we already solve this by manually recomputing the alignment in `SharedUtils.alignment` called from the other overload of this method.

mcimadamore: I think `SharedUtils::alignment` overrides whatever alignment constraint is specified in the layout. This function has existed for a very long time (probably from before layouts even supported alignment constraints). Some pieces of this code seems to be influenced by how SysV handles alignment constraints - especially the array alignment implementation.
IMHO, an ABI should check that the incoming layouts have the correct alignment constraints, rather than drop the constraints, and then recomputing them from scratch.

test/jdk/java/foreign/libTestDowncall.h
```diff
@@ -1,12210 +0,0 @@
...
-struct S_PDI { void* p0; double p1; int p2; };
-struct S_PDF { void* p0; double p1; float p2; };
-struct S_PDD { void* p0; double p1; double p2; };
-struct S_PDP { void* p0; double p1; void* p2; };
-struct S_PPI { void* p0; void* p1; int p2; };
-struct S_PPF { void* p0; void* p1; float p2; };
-struct S_PPD { void* p0; void* p1; double p2; };
-struct S_PPP { void* p0; void* p1; void* p2; };
-EXPORT void f0_V__(void) ;
-EXPORT void f0_V_I_(int p0) ;
```
mcimadamore, line 115: I assume you removed the functions from the header because they are not needed - e.g. the definition in the C file is sufficient?

src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/CallArranger.java
JornVernee: Yes, I agree. I think in the past that would have meant that users would always have to adjust the alignment of JAVA_* layouts before being able to link with them, but not that those layouts have the right alignment constraints we can more explicitly reject layouts that have the wrong alignment during linking.

JornVernee: I've filed a followup for this: https://bugs.openjdk.java.net/browse/JDK-8284300

test/jdk/java/foreign/libTestDowncall.h
JornVernee: Right, only the struct definitions are needed

test/jdk/ProblemList.txt
```diff
@@ -533,6 +533,8 @@ java/beans/XMLEncoder/Test6570354.java 8015593 macosx-all
 
 java/foreign/TestMismatch.java 8249684 macosx-all
 java/foreign/TestUpcallStack.java 8275584 macosx-aarch64
+java/foreign/TestIntrinsics.java 8275584 macosx-aarch64
```
mcimadamore, line 536: What happened to this test? This seems to be working today

JornVernee:
I've updated the patch to let varargs through. I've also significantly beefed up `TestVarArgs` to now test all the signatures we are also testing for TestDown/Upcalls.

test/jdk/ProblemList.txt
JornVernee: Yeah, that one should work I think. Let me try to re-add it.

mcimadamore (approved):
Looks good - thanks for looking into this!

PR #679, opened 2022-05-06
8286306: Upcall wrapper class sharing

JornVernee:
This patch adds upcall wrapper class sharing.

I recently realized that these classes can be shared if they have the same input FunctionDescriptor, but a different target method handle.

This is implemented by generating a wrapper class that takes the target method handles as a leading argument. The target method handle is then inserted into the parameter list using `MethodHandles::insertArguments` to create the final method handle that is returned by the specializer. After MH customization kicks in, the target will be a constant again (so shouldn't affect peak performance).

Doing this doesn't effect peak performance, as verified through the Upcalls benchmark, but does significantly improve linking performance.

Before:

    Benchmark              Mode  Cnt   Score   Error  Units
    LinkUpcall.link_blank  avgt   30  69.323 � 2.522  us/op

After:

    Benchmark              Mode  Cnt   Score   Error  Units
    LinkUpcall.link_blank  avgt   30  11.787 � 0.744  us/op

---------
### Progress
- [x] Change must be properly reviewed (1 review required, with at least 1 committer)

### Issue
 * [JDK-8286306](https://bugs.openjdk.java.net/browse/JDK-8286306): Upcall wrapper class sharing

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/panama-foreign pull/679/head:pull/679` \
`$ git checkout pull/679`

`$ git checkout pull/679` \
`$ git pull https://git.openjdk.java.net/panama-foreign pull/679/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 679`

`$ git pr show -t 679`

</details>

<a href="https://git.openjdk.java.net/panama-foreign/pull/679.diff">https://git.openjdk.java.net/panama-foreign/pull/679.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/jdk/internal/foreign/abi/SharedUtils.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/abi/SharedUtils.java
@@ -209,6 +209,10 @@ public class SharedUtils {
 
         if (dropReturn) { // no handling for return value, need to drop it
             target = dropReturn(target);
+        } else {
+            // adjust return type so it matches the inferred type of the effective
+            // function descriptor
+            target = target.asType(target.type().changeReturnType(Addressable.class));
         }
 
         return target;
--- a/src/java.base/share/classes/jdk/internal/foreign/abi/BindingSpecializer.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/abi/BindingSpecializer.java
@@ -204,16 +204,13 @@ public class BindingSpecializer {
         }
 
         try {
-            if (callingSequence.forUpcall()) {
-                // We must initialize the class since the upcall stubs don't have a clinit barrier, and the slow
-                // path in the c2i adapter we end up calling can not handle the particular code shape where the
-                // caller is an upcall stub.
-                MethodHandles.Lookup lookup = MethodHandles.lookup().defineHiddenClass(bytes, true);
-                return lookup.findStatic(lookup.lookupClass(), METHOD_NAME, callerMethodType);
-            } else { // downcall
-                MethodHandles.Lookup lookup = MethodHandles.lookup().defineHiddenClassWithClassData(bytes, leafHandle, false);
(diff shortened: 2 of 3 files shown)


src/java.base/share/classes/jdk/internal/foreign/abi/SharedUtils.java
```diff
@@ -209,6 +209,10 @@ public static MethodHandle adaptUpcallForIMR(MethodHandle target, boolean dropRe
 
         if (dropReturn) { // no handling for return value, need to drop it
             target = dropReturn(target);
+        } else {
+            // adjust return type so it matches the inferred type of the effective
+            // function descriptor
+            target = target.asType(target.type().changeReturnType(Addressable.class));
```
JornVernee, line 215: Note that this is needed to prevent a situation on Windows where the effective function descriptor of `()[<some IMR struct>]` is the same as a plain `(a64)a64` function descriptor. The `target` will currently have the type `(MemoryAddress)MemoryAddress` in the former case, but `(MemoryAddress)Addressable` in the latter case. However, they share an upcall wrapper class because they have the same function descriptor, and this leads to a `WrongMethodTypeException` when invoking one of them.

mcimadamore: I think you mean in the latter case is `(Addressable)MemoryAddress` ?

JornVernee: No, for upcalls it is `(MemoryAddress)Addressable` i.e. the return type is changed.

mcimadamore: aaah contravariance :-)

mcimadamore (approved):
Looks solid!

PR #690, opened 2022-06-28
8289285: Use records for binding classes

JornVernee:
This PR switches the Binding classes to use records. This results in a clean simplification of the code due to the no longer having the need to implement equals/hashcode/toString as well. When we have type patterns for switch, we will also be able to remove the tag, and just switch on the type.

I also noticed that the UnboxAddress binding had 2 fields that were not being used. In practice this binding always operates on Addressable instances (both in the interpreter and specializer). So, I've cleaned up these 2 fields as well, and related factory method.

---------
### Progress

### Issue
 * [JDK-8289285](https://bugs.openjdk.org/browse/JDK-8289285): Use records for binding classes

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/690/head:pull/690` \
`$ git checkout pull/690`

`$ git checkout pull/690` \
`$ git pull https://git.openjdk.org/panama-foreign pull/690/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 690`

`$ git pr show -t 690`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/690.diff">https://git.openjdk.org/panama-foreign/pull/690.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/jdk/internal/foreign/abi/SharedUtils.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/abi/SharedUtils.java
@@ -371,7 +371,7 @@ public class SharedUtils {
     }
 
     static void checkType(Class<?> actualType, Class<?> expectedType) {
-        if (expectedType != actualType) {
+        if (!expectedType.isAssignableFrom(actualType)) {
             throw new IllegalArgumentException(
                     String.format("Invalid operand type: %s. %s expected", actualType, expectedType));
         }
--- a/src/java.base/share/classes/jdk/internal/foreign/abi/Binding.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/abi/Binding.java
@@ -339,11 +339,7 @@ public interface Binding {
     }
 
     static UnboxAddress unboxAddress() {
-        return unboxAddress(MemoryAddress.class);
-    }
-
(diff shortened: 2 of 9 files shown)


src/java.base/share/classes/jdk/internal/foreign/abi/SharedUtils.java
```diff
@@ -371,7 +371,7 @@ public static VaList emptyVaList() {
     }
 
     static void checkType(Class<?> actualType, Class<?> expectedType) {
-        if (expectedType != actualType) {
+        if (!expectedType.isAssignableFrom(actualType)) {
```
JornVernee, line 374: This change is needed to accommodate for the fact that UnboxAddress now accepts an Addressable, but a previous binding still outputs either a MemoryAddress or MemorySegment.

mcimadamore (approved):
Nice use of records :+1:

src/java.base/share/classes/jdk/internal/foreign/abi/SharedUtils.java
mcimadamore: Thanks for cleaning up these (and corresponding binding classes)

PR #691, opened 2022-07-22
8290880: Add static MemorySegment::mismatch method

mcimadamore:
As discussed in [1], it would be useful if the `MemorySegment` class provided a static mismatch method, similarly to what we have provided for `copy`.

This patch adds that method, and rewrites the existing instance `mismatch` method to be based on that new static method.
It then adds new tests to cover the new static method.

[1] - https://mail.openjdk.org/pipermail/panama-dev/2022-June/017099.html

---------
### Progress

### Issue
 * [JDK-8290880](https://bugs.openjdk.org/browse/JDK-8290880): Add static MemorySegment::mismatch method

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/691/head:pull/691` \
`$ git checkout pull/691`

`$ git checkout pull/691` \
`$ git pull https://git.openjdk.org/panama-foreign pull/691/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 691`

`$ git pr show -t 691`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/691.diff">https://git.openjdk.org/panama-foreign/pull/691.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/java/lang/foreign/MemorySegment.java
+++ b/src/java.base/share/classes/java/lang/foreign/MemorySegment.java
@@ -40,6 +40,7 @@ import java.util.Spliterator;
 import java.util.stream.Stream;
 import jdk.internal.foreign.AbstractMemorySegmentImpl;
 import jdk.internal.foreign.HeapMemorySegmentImpl;
+import jdk.internal.foreign.MemorySessionImpl;
 import jdk.internal.foreign.NativeMemorySegmentImpl;
 import jdk.internal.foreign.Utils;
 import jdk.internal.foreign.abi.SharedUtils;
@@ -48,7 +49,10 @@ import jdk.internal.misc.ScopedMemoryAccess;
 import jdk.internal.misc.Unsafe;
 import jdk.internal.reflect.CallerSensitive;
 import jdk.internal.reflect.Reflection;
+import jdk.internal.util.ArraysSupport;
 import jdk.internal.vm.annotation.ForceInline;
+
+import static java.lang.foreign.ValueLayout.JAVA_BYTE;
 
 /**
  * A memory segment models a contiguous region of memory. A memory segment is associated with both spatial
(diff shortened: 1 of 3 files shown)


JornVernee (approved):

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
```diff
@@ -1966,4 +1980,69 @@ private static long getBaseAndScale(Class<?> arrayType) {
...
+     * {@linkplain MemorySession#isAlive() alive}.
+     * @throws WrongThreadException if this method is called from a thread other than the thread owning
+     * the {@linkplain #session() session} associated with {@code dstSegment}.
+     * @throws IndexOutOfBoundsException if {@code srcOffset + bytes > srcSegment.byteSize()} or if
+     * {@code dstOffset + bytes > dstSegment.byteSize()}, or if either {@code srcOffset}, {@code dstOffset}
+     * or {@code bytes} are {@code < 0}.
+     *
+     * @see MemorySegment#mismatch(MemorySegment)
+     */
+    static long mismatch(MemorySegment srcSegment, long srcOffset, MemorySegment dstSegment, long dstOffset, long bytes) {
```
PaulSandoz, line 2015: Should we make the signature the same as that for Arrays.mismatch e.g.:
```
    public static int mismatch(byte[] a, int aFromIndex, int aToIndex,
                               byte[] b, int bFromIndex, int bToIndex) {
```
?
It's tempting to make both mismatch methods static.

PaulSandoz: Re: static vs instance, apologies I don't mean to reopen that discussion.

IIRC the motivation for the signatures in `Arrays` was in part because that is the pattern used by all the other methods. You are right that the underlying primitive is passing in a common length, which is what the supporting internal mismatch methods use, but it may not be as helpful to require the user to compute the minimum length. To be honest I am torn between keeping close to the `copy` signature or using absolute ranges :-)

mcimadamore: One advantage to push the length computation inside is that explaining what the instance method does become trivial. An option could be to expose both, but seems like overkill. I think I'm fine with exposing the Arrays signature - after all the user can always get what he wants either way. That would at least make the API point consistent with Arrays.mismatch (in the same way MemorySegment::copy is consistent with System::arraycopy).

PaulSandoz: I agree with that positioning, `System::arraycopy` is to `MemorySegment::copy` as `Arrays::mistmatch` is to `MemorySegment::mistmatch`.

JornVernee (approved):
Still good

PR #694, opened 2022-07-28
8291473: Unify MemorySegment and MemoryAddress

mcimadamore:
This patch implements the changes described in [1].

The main things to notice are the removal of `MemoryAddress` and `Addressable`. Everything else falls from that.

There are some changes to the `MemorySegment` API as well, as some methods such as `segmentOffset` and `asOverlappingSlice` have been removed, and instead `MemorySegment::address` now works on heap segments too (and, clients can obtain the array of an heap segment using its `array` method). So, these methods can, if needed, be implemented outside the core API.

`MemorySegment::equals` now compares the segments address, which is the most useful comparison when working with native interop (again, client can implement deeper comparison using `mismatch`).

Finally, restricted factories accepting raw addresses (in `MemorySegment` and `VaList`) have been tweaked to accept a long value instead.

`VaList` also needed some adjustments, since it builds on top of segments. The API is similar to before - but instead of having an address accessor, it has a zero-length segment accessor.

`ValueLayout.OfAddress` has a new method to create *unsafe* address layouts which return segment with maximal size (`Long.MAX_VALUE`). I've used these layouts in a lot of places in the implementation internal, which simplifies things quite a bit (e.g. removing the need to create a new segment from an address).

There are several test and microbenchmark updates, but relatively minor, and all caused by removal of `MemoryAddress`/`Addressable`. Changes here should be relatively simple to follow.

While the javadoc is ok, I didn't put too much effort in trying to provide a complete and cohesive story on how zero-length segments are used to model raw addresses. I'm working on some more extensive doc changes on the side, but I wanted to split the javadoc changes from the impl changes, so as to simplify the review work. I hope that makes sense.

[1] - https://mail.openjdk.org/pipermail/panama-dev/2022-July/017181.html

---------
### Progress

### Issue
 * [JDK-8291473](https://bugs.openjdk.org/browse/JDK-8291473): Unify MemorySegment and MemoryAddress

### Reviewers
 * [Jorn Vernee](https://openjdk.org/census#jvernee) (@JornVernee - Committer) ⚠️ Review applies to [142dd031](https://git.openjdk.org/panama-foreign/pull/694/files/142dd03164a6c21ecdb35d19f47ba67bb17b2935)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/694/head:pull/694` \
`$ git checkout pull/694`

`$ git checkout pull/694` \
`$ git pull https://git.openjdk.org/panama-foreign pull/694/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 694`

`$ git pr show -t 694`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/694.diff">https://git.openjdk.org/panama-foreign/pull/694.diff</a>

</details>


JornVernee:
>  but I wanted to split the javadoc changes from the impl changes, so as to simplify the review work. I hope that makes sense.

This is very appreciated.

test/micro/org/openjdk/bench/java/lang/foreign/pointers/Struct.java
```diff
@@ -0,0 +1,9 @@
+package org.openjdk.bench.java.lang.foreign.pointers;
```
JornVernee, line 1: missing copyright header

test/micro/org/openjdk/bench/java/lang/foreign/pointers/PointerBench.java
```diff
@@ -0,0 +1,122 @@
+package org.openjdk.bench.java.lang.foreign.pointers;
```
JornVernee, line 1: missing copyright header

test/micro/org/openjdk/bench/java/lang/foreign/pointers/Pointer.java
```diff
@@ -0,0 +1,65 @@
+package org.openjdk.bench.java.lang.foreign.pointers;
```
JornVernee, line 1: missing copyright header

test/micro/org/openjdk/bench/java/lang/foreign/pointers/Point.java
```diff
@@ -0,0 +1,40 @@
+package org.openjdk.bench.java.lang.foreign.pointers;
```
JornVernee, line 1: missing copyright header

test/micro/org/openjdk/bench/java/lang/foreign/pointers/NativeType.java
```diff
@@ -0,0 +1,59 @@
+package org.openjdk.bench.java.lang.foreign.pointers;
```
JornVernee, line 1: missing copyright header

test/jdk/java/foreign/valist/VaListTest.java
```diff
@@ -183,7 +186,8 @@ public static Object[][] sumDoubles() {
         Function<ValueLayout.OfDouble, BiFunction<Integer, VaList, Double>> sumDoublesJavaFact  = layout ->
                 (num, list) -> DoubleStream.generate(() -> list.nextVarg(layout)).limit(num).sum();
         BiFunction<Integer, VaList, Double> sumDoublesNative
-                = MethodHandleProxies.asInterfaceInstance(BiFunction.class, MH_sumDoubles);
+                = MethodHandleProxies.asInterfaceInstance(BiFunction.class,
+                MethodHandles.filterArguments(MH_sumDoubles, 1, VALIST_TO_ADDRESS));
```
JornVernee, line 190: I suggest maybe moving this adaptation to `linkVaList` (the VaList should always be the last parameter)

test/micro/org/openjdk/bench/java/lang/foreign/pointers/Point.java
```diff
@@ -0,0 +1,40 @@
...
+
+    Point(Pointer<Point> ptr) {
+        super(ptr);
+    }
+
+    int x() {
+        return ptr.segment.getAtIndex(NativeType.C_INT.layout(),  0);
+    }
+    int y() {
+        return ptr.segment.getAtIndex(NativeType.C_INT.layout(),  4);
```
JornVernee, line 18: Shouldn't this just be `get`? (otherwise this seems like it would access at offset = 4 * 4)

JornVernee (commented):
There seems to be some comments (and 1 string)  that mention `MemoryAddress`. Please update those as well.

src/java.base/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
```diff
@@ -219,33 +211,6 @@ public boolean isNative() {
...
-
-    @Override
-    public final long segmentOffset(MemorySegment other) {
-        AbstractMemorySegmentImpl that = (AbstractMemorySegmentImpl) Objects.requireNonNull(other);
-        if (unsafeGetBase() == that.unsafeGetBase()) {
-            return that.unsafeGetOffset() - this.unsafeGetOffset();
-        }
-        throw new UnsupportedOperationException("Cannot compute offset from native to heap (or vice versa).");
-    }
-
```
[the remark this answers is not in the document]
mcimadamore: Fair point, I can separate out. As you noted, though, presence of array() and totalization of address() does change things considerably in this space. We added these methods because they could not have been written by hand. But dealing with that in a separate patch sounds fine.

src/java.base/share/classes/java/lang/foreign/VaList.java
```diff
@@ -175,50 +204,47 @@ sealed public interface VaList extends Addressable permits WinVaList, SysVVaList
...
      * @throws IllegalCallerException if access to this method occurs from a module {@code M} and the command line option
      * {@code --enable-native-access} is specified, but does not mention the module name {@code M}, or
      * {@code ALL-UNNAMED} in case {@code M} is an unnamed module.
      */
     @CallerSensitive
-    static VaList ofAddress(MemoryAddress address, MemorySession session) {
-        Reflection.ensureNativeAccess(Reflection.getCallerClass(), VaList.class, "ofAddress");
-        Objects.requireNonNull(address);
-        Objects.requireNonNull(session);
+    static VaList ofAddress(long address, MemorySession session) {
```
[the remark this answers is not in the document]
mcimadamore: Whoops - yes. This is the result on some back and forth on my part that I forgot to restore.

src/java.base/share/classes/jdk/internal/foreign/HeapMemorySegmentImpl.java
```diff
@@ -59,9 +61,19 @@ public abstract class HeapMemorySegmentImpl extends AbstractMemorySegmentImpl {
...
 
+    @Override
+    public Optional<Object> array() {
+        return Optional.of(base);
+    }
+
+    @Override
+    public long maxAlignMask() {
+        return 0;
+    }
```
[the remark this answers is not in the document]
mcimadamore: I see - odd.

src/java.base/share/classes/jdk/internal/foreign/abi/BindingSpecializer.java
```diff
@@ -447,7 +444,8 @@ private boolean needsSession() {
     }
 
     private static boolean shouldAcquire(Class<?> type) {
-        return type == Addressable.class;
+        // by this time, the only MemorySegment left are those for by-ref args
+        return type == MemorySegment.class;
```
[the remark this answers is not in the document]
mcimadamore: I see - I might have been confused into thinking that by-value args were already lowered, which is clearly not the case. We need to piggy back on the function descriptor.

src/java.base/share/classes/jdk/internal/foreign/abi/SharedUtils.java
```diff
@@ -200,14 +193,14 @@ public static MethodHandle adaptUpcallForIMR(MethodHandle target, boolean dropRe
...
+            target = target.asType(target.type().changeReturnType(MemorySegment.class));
         }
 
         return target;
     }
 
-    private static MemoryAddress bufferCopy(MemoryAddress dest, MemorySegment buffer) {
-        MemoryAddressImpl.ofLongUnchecked(dest.toRawLongValue(), buffer.byteSize()).copyFrom(buffer);
+    private static MemorySegment bufferCopy(MemorySegment dest, MemorySegment buffer) {
+        NativeMemorySegmentImpl.makeNativeSegmentUnchecked(dest.address(), buffer.byteSize()).copyFrom(buffer);
```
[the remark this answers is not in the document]
mcimadamore: I'll try

src/java.base/share/classes/jdk/internal/foreign/abi/x64/windows/CallArranger.java
```diff
@@ -275,14 +274,14 @@ public List<Binding> getBindings(Class<?> carrier, MemoryLayout layout, boolean
                     assert carrier == MemorySegment.class;
                     VMStorage storage = storageCalculator.nextStorage(StorageClasses.INTEGER, layout);
                     bindings.vmLoad(storage, long.class)
-                            .boxAddress()
+                            .boxAddress(Long.MAX_VALUE)
                             .toSegment(layout);
```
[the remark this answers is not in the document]
mcimadamore: Ok, so the problem here is that we have a `boxAddress` but we don't really need it, and instead we just want `toSegment` ?

test/jdk/java/foreign/TestUpcallBase.java
```diff
@@ -145,7 +145,8 @@ static Addressable makeCallback(MemorySession session, Ret ret, List<ParamType>
 
     static Object passAndSave(Object[] o, AtomicReference<Object[]> ref, int retArg) {
         for (int i = 0; i < o.length; i++) {
-            if (o[i] instanceof MemorySegment) {
+            if (o[i] instanceof MemorySegment &&
+                    !((MemorySegment) o[i]).session().equals(MemorySession.global())) {
```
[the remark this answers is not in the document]
mcimadamore: I had what you suggested, and then got back to this to minimize diffs (adding an extra param ended up being quire pervasive)  - I'll add a comment

test/jdk/java/foreign/handles/invoker_module/handle/invoker/MethodHandleInvoker.java
```diff
@@ -71,11 +69,10 @@ static <Z> void addDefaultMapping(Class<Z> carrier, Z value) {
         addDefaultMapping(MethodHandle.class, MethodHandles.identity(int.class));
         addDefaultMapping(Charset.class, Charset.defaultCharset());
         addDefaultMapping(MethodType.class, MethodType.methodType(void.class));
-        addDefaultMapping(MemoryAddress.class, MemoryAddress.NULL);
-        addDefaultMapping(Addressable.class, MemoryAddress.NULL);
+        addDefaultMapping(MemorySegment.class, MemorySegment.NULL);
         addDefaultMapping(MemoryLayout.class, ValueLayout.JAVA_INT);
         addDefaultMapping(FunctionDescriptor.class, FunctionDescriptor.ofVoid());
-        addDefaultMapping(MemorySession.class, MemorySession.openImplicit());
+        addDefaultMapping(MemorySession.class, MemorySession.openShared());
```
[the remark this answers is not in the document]
mcimadamore: Not needed - more a result of other shuffling - but then when coming back to it I realized that it was probably more correct to use implicit

test/micro/org/openjdk/bench/java/lang/foreign/BulkMismatchAcquire.java
```diff
@@ -1,168 +0,0 @@
...
-import java.util.concurrent.TimeUnit;
-import java.util.function.Supplier;
-
-@BenchmarkMode(Mode.AverageTime)
-@Warmup(iterations = 5, time = 500, timeUnit = TimeUnit.MILLISECONDS)
-@Measurement(iterations = 10, time = 500, timeUnit = TimeUnit.MILLISECONDS)
-@State(org.openjdk.jmh.annotations.Scope.Thread)
-@OutputTimeUnit(TimeUnit.MILLISECONDS)
-@Fork(value = 3, jvmArgsAppend = "--enable-preview")
-public class BulkMismatchAcquire {
```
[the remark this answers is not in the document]
mcimadamore: no - thanks for the catch

src/java.base/share/classes/jdk/internal/foreign/HeapMemorySegmentImpl.java
```diff
@@ -59,9 +61,19 @@ public abstract class HeapMemorySegmentImpl extends AbstractMemorySegmentImpl {
     final long offset;
     final Object base;
 
+    @Override
+    public Optional<Object> array() {
+        return Optional.of(base);
```
rsmogura, line 66: I wonder if we should use here `ofNullable`

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
```diff
@@ -226,36 +224,41 @@
...
+     * @return the base address of the memory region associated with this segment.
      */
-    @Override
-    MemoryAddress address();
+    long address();
+
+    /**
+     * {@return the Java array associated with this memory segment, if any}
+     */
+    Optional<Object> array();
```
rsmogura, line 261: I wonder if array word is not too strict, I think what if we would use value object as a base for struct (sometime in future).

mcimadamore: > I wonder if array word is not too strict, I think what if we would use value object as a base for struct (sometime in future).

I think you are correct, the word might indeed be too strict. "Heap container" is what we're after, but I'm not sure if having a method named "heapContainer" would fly?

rsmogura: I'm not good at finding names. I got used to `base`. Maybe heapBase or heapObject. I'll think about it and come back if I invite something worth of mentioning.

rsmogura: I think heapContainer is good, I just thought as well about something like holder, or value holder if we would like to have something shorter - I'm really not an expert at naming.

mcimadamore: what about `heapBase` ?

mcimadamore:
I've finished addressing the review comments. Now `Binding.ToSegment` has been removed, and `Binding.BoxAddress` does all the work. I think it looks clean.

src/java.base/share/classes/jdk/internal/foreign/HeapMemorySegmentImpl.java
rsmogura: I would like to recall this comment. Even if it would look better from language perspective, having not specialized `Optional` can make `Unsafe` operations slower, as the VM can make issues with detecting if it's on heap or off-heap access.

JornVernee (approved):
I think this is mostly good now.

The only thing is that there's still some occurrence of `MemoryAddress` (when I search for it in `src/java.base/classes`), mostly in comments. Some of these are already stale, so not that important to fix, but there's also a reference to `MemoryAddress` in the class javadoc of `SymbolLookup` which I think should be updated.

mcimadamore:
> I think this is mostly good now.
> 
> The only thing is that there's still some occurrence of `MemoryAddress` (when I search for it in `src/java.base/classes`), mostly in comments. Some of these are already stale, so not that important to fix, but there's also a reference to `MemoryAddress` in the class javadoc of `SymbolLookup` which I think should be updated.

I'll take a pass

mcimadamore:
> > I think this is mostly good now.
> > The only thing is that there's still some occurrence of `MemoryAddress` (when I search for it in `src/java.base/classes`), mostly in comments. Some of these are already stale, so not that important to fix, but there's also a reference to `MemoryAddress` in the class javadoc of `SymbolLookup` which I think should be updated.
> 
> I'll take a pass

I've fixed most of the spurious references to the now removed classes. I've left the main Binding javadoc alone, because it's full of references to stuff that's no longer valid, and should probably be rewritten in a separate pass.

JornVernee (approved):
Thanks. Looks good

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
mcimadamore: I'll keep the name (`array`) for now, we can always revise that later. Thanks.

PR #695, opened 2022-08-02
8291639: Improve the ability to visualize a MemorySegment in human readable forms

minborg:
This PR improves the ability to visualize a memory abstraction (e.g. `MemorySegment`, `ByteBuffer`, `byte[]`) in human-readable forms.

The existing `HexFormat` class will get a new configurable `MemoryDumper` that can produce dumps of MemorySegments, byte arrays and ByteBuffers:

```
0000000000000000 54 68 65 20 71 75 69 63 6B 20 62 72 6F 77 6E 20 |The quick brown |
0000000000000010 66 6F 78 20 6A 75 6D 70 65 64 20 6F 76 65 72 20 |fox jumped over |
0000000000000020 74 68 65 20 6C 61 7A 79 20 64 6F 67 0A 53 65 63 |the lazy dog.Sec|
0000000000000030 6F 6E 64 20 6C 69 6E 65 09 3A 68 65 72 65 00 00 |ond line.:here..|
0000000000000040 00 00 00 00                                     |....            |
```

A MemoryDumper can contain any number of columns (either index or data) which can be formatted in a large number of ways using composable `HexFormat` instances. Even more configuration possibilities can be achieved by providing custom formatting lambdas.

---------
### Progress

### Issue
 * [JDK-8291639](https://bugs.openjdk.org/browse/JDK-8291639): Improve the ability to visualize a MemorySegment in human readable forms (**Enhancement** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign.git pull/695/head:pull/695` \
`$ git checkout pull/695`

`$ git checkout pull/695` \
`$ git pull https://git.openjdk.org/panama-foreign.git pull/695/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 695`

`$ git pr show -t 695`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/695.diff">https://git.openjdk.org/panama-foreign/pull/695.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/panama-foreign/pull/695#issuecomment-1204101897)


src/java.base/share/classes/jdk/internal/foreign/MemorySegmentRenderUtil.java
```diff
@@ -0,0 +1,367 @@
...
+            if (!sb.isEmpty()) {
+                sb.append(System.lineSeparator());
+            }
+            sb.append(line);
+        };
+        renderView(memorySegment, lens, action, new ViewState(), "");
+        return sb.toString();
+    }
+
+    public static void renderView(MemorySegment memorySegment,
```
mcimadamore, line 156: Suggestion, instead of keeping track of an explicit index, it would be perhaps better to construct a layout path using `MemoryLayout.PathElement`. That way, access to the correct element can be made w/o index. It will be slower, as it will compute a new VH each time, but at least it's guaranteed to be correct. Another approach could be to construct the layout path, and then use it to compute the offset, and then use the offset as you are doing now. I'm worried about duplicating offsetting logic.

src/java.base/share/classes/jdk/internal/foreign/MemorySegmentRenderUtil.java
```diff
@@ -0,0 +1,367 @@
...
+            return;
+        }
+        if (memoryLayout instanceof PaddingLayout paddingLayout) {
+            action.accept(state.indentSpaces() + paddingLayout.bitSize() + " padding bits");
+            state.indexAndAdd(paddingLayout);
+            return;
+        }
+        if (memoryLayout instanceof GroupLayout groupLayout) {
+
+            // Strictly, we should provide all permutations of nested unions.
```
mcimadamore, line 207: Not sure I get this comment

src/java.base/share/classes/jdk/internal/foreign/MemorySegmentRenderUtil.java
```diff
@@ -0,0 +1,367 @@
...
+        return indentedLabel(state, layout) + "=" + value + suffix;
+    }
+
+    static String indentedLabel(ViewState state,
+                                MemoryLayout layout) {
+        return state.indentSpaces() + layout.name()
+                .orElseGet(layout::toString);
+    }
+
+    static final class HexStreamState {
```
mcimadamore, line 262: While this class is simple - I note that there are many others hex utils in the JDK like src/java.base/share/classes/sun/security/util/HexDumpEncoder.java

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
```diff
@@ -1874,6 +1872,73 @@ default void setAtIndex(ValueLayout.OfAddress layout, long index, MemorySegment
...
+     * of a memory segment.
+     *
+     * @return a Stream of human-readable, lines with hexadecimal values
+     */
+    default Stream<String> hexStream() {
+        return MemorySegmentRenderUtil.hexStream(this);
+    }
+
+    /**
+     * Returns a human-readable view of this memory segment viewed through
```
mcimadamore, line 1908: I strongly suggest dropping all reference to "lenses". I see what you mean, but I think the term seems to exotic. E.g. "viewing a memory segment through a memory layout" is fine.

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
```diff
@@ -1874,6 +1872,73 @@ default void setAtIndex(ValueLayout.OfAddress layout, long index, MemorySegment
...
+     * the provided memory layout {@code lens}.
+     * <p>
+     * Lines are separated with the system-dependent line separator {@link System#lineSeparator() }.
+     * Otherwise, the exact format of the returned view is unspecified and should not
+     * be acted upon programmatically.
+     * <p>
+     * As an example, a memory segment viewed though the following memory layout lens
```
mcimadamore, line 1915: As an example, consider the following segment (show a segment initialized to (1, 2)).
We can view the above segment through the layout defined as follows (show layout you have now).
This will render the segment as follows (the exact contents are not specified):...

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
```diff
@@ -1874,6 +1872,73 @@ default void setAtIndex(ValueLayout.OfAddress layout, long index, MemorySegment
...
+     *   y=2
+     * }
+     *}
+     * <p>
+     * This method is intended to view memory segments through small and medium-sized memory layout
+     * lenses and is, in all cases, restricted by the inherent String capacity limit.
+     *
+     * @param lens  to use as a lens when viewing the memory segment
+     * @return a view of the memory segment viewed through a memory layout lens
+     * @throws OutOfMemoryError if the view exceeds the array size VM limit
```
mcimadamore, line 1935: we have some verbiage for this in `MemorySegment::getUtf8String`:

```
     * @throws IllegalArgumentException if the size of the UTF-8 string is greater than the largest string supported by the platform.
     ```

File: src/java.base/share/classes/java/lang/foreign/MemorySegment.java [hunk 7]
```diff
@@ -1874,6 +1872,73 @@ default void setAtIndex(ValueLayout.OfAddress layout, long index, MemorySegment
...
+     *}
+     * might be rendered to something like this:
+     * {@snippet lang = text:
+     * Point {
+     *   x=1,
+     *   y=2
+     * }
+     *}
+     * <p>
+     * This method is intended to view memory segments through small and medium-sized memory layout
```
mcimadamore, line 1930: I would leave this out, since it's covered in `@throws`

mcimadamore (commented):
The new functionalities seem helpful. From an API perspective I'm a bit concerned in that we are probably not able to say a lot in terms of specification. For instance, how many String are returned in the hex dump stream? Or what is the format of the `viewThrough` string? While you call some of these out, part of me is a bit wary of adding methods whose semantics cannot be fully described in the javadoc, as that creates an issue when it comes to testing and conformance testing of the API. Are there similar analogue functionalities in public JDK APIs (e.g. ByteBuffer) ?

Separately, I think viewThrough is more segment/layout specific, whereas hexDump really is a general functionality that just so happens to be defined by MemorySegment. I'm worried about this because it leaves users of ByteBuffer, String, arrays and what not in the cold, while we provide a specific functionality for memory segments. And, if we had a more general functionality available somewhere else (e.g. that worked on a byte[]) I'm not sure we would be adding this method?

For these reason I'd veer towards keeping the layout method in the API, but make the hex dump method private for now.

src/java.base/share/classes/jdk/internal/foreign/MemorySegmentRenderUtil.java
```diff
@@ -0,0 +1,367 @@
...
+    }
+
+    static char viewByteAsAscii(byte b) {
+        final int value = Byte.toUnsignedInt(b);
+        return (value >= 32 && value < 127)
+                ? (char) value
+                : '.';
+    }
+
+}
```
mcimadamore, line 367: Note missing newline here

src/java.base/share/classes/java/lang/foreign/PaddingLayout.java
```diff
@@ -35,7 +35,7 @@
  * @implSpec
  * This class is immutable, thread-safe and <a href="{@docRoot}/java.base/java/lang/doc-files/ValueBased.html">value-based</a>.
  */
-/* package-private */ final class PaddingLayout extends AbstractLayout implements MemoryLayout {
+public final class PaddingLayout extends AbstractLayout implements MemoryLayout {
```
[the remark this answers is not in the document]
minborg: We can re-write the check using a replacement for `instanceof` and revert the PaddingLayout change.

src/java.base/share/classes/jdk/internal/foreign/MemorySegmentRenderUtil.java
minborg: So, suppose there is a nested union such that: U = (A|B)|(C|D). Then the actual layout U could be either of AC, AD, BC or BD. Hence, we could present all these combinations.

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
```diff
@@ -1874,6 +1872,73 @@ default void setAtIndex(ValueLayout.OfAddress layout, long index, MemorySegment
...
+     * 0000000000000030  6F 6E 64 20 6C 69 6E 65  09 3A 68 65 72 65 00 00  |ond line.:here..|
+     * 0000000000000040  00 00 00 00                                       |....|
+     *}
+     * <p>
+     * Use a {@linkplain MemorySegment#asSlice(long, long) slice} to inspect a specific region
+     * of a memory segment.
+     *
+     * @return a Stream of human-readable, lines with hexadecimal values
+     */
+    default Stream<String> hexStream() {
```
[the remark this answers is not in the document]
minborg: I think it is possible to specify the `hexDump()` method in detail and also make it more general so it can be applied to virtually any memory representation such as `byte[]` or `ByteBuffer`.

src/java.base/share/classes/jdk/internal/foreign/MemorySegmentRenderUtil.java
minborg: True. This class operates on input/output streams and so there is an impedance mismatch. But it would be possible to use it via an adaptor.

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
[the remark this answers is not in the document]
mcimadamore: Yep, I do think that hex dump doesn't really belong here - but should be a common functionality shared between all byte array views.

src/java.base/share/classes/java/lang/foreign/PaddingLayout.java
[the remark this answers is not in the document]
mcimadamore: Note that there's a predicate `isPadding`

src/java.base/share/classes/jdk/internal/foreign/MemorySegmentRenderUtil.java
mcimadamore: Ah I see. I think that, perhaps, it would be better to collect all fields in a union, and present them all (like if it was a struct). I don't see any particular problem with that, and I don't think the API should "pick" a field.

src/java.base/share/classes/jdk/internal/foreign/MemorySegmentRenderUtil.java
```diff
@@ -133,14 +133,14 @@ public static Stream<String> hexStream(MemorySegment memorySegment) {
      * and is, in all cases, restricted by the inherent String capacity limit.
      *
      * @param memorySegment to be viewed
-     * @param lens  to use as a lens when viewing the memory segment
+     * @param memoryLayout  to use as a memoryLayout when viewing the memory segment
```
mcimadamore, line 136: bad find/replace?

mcimadamore: I suggest calling `memoryLayout` just `layout`

src/java.base/share/classes/jdk/internal/foreign/MemorySegmentRenderUtil.java
mcimadamore: Ah ok, I misread your comment. I think what the impl does now is actually sensible, as it presents the full `(A|B)|(C|D)`.

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
```diff
@@ -1874,6 +1872,61 @@ default void setAtIndex(ValueLayout.OfAddress layout, long index, MemorySegment
...
+     *     its {@link MemoryLayout#toString()} representation.</li>
+     *     <li> Values are rendered differently depending on their type:
+     *         <ul>
+     *             <li>Numeric values are rendered in decimal form (e.g 1 or 1.2).</li>
+     *             <li>Boolean values are rendered as {@code true} or {@code false}.</li>
+     *             <li>Character values are rendered as {@code char}.</li>
+     *             <li>Address values are rendered according to their {@code toString()} values.</li>
+     *         </ul>
+     *     </li>
+     *     <li>Grouped layouts are enclosed in curly brackets.</li>
```
mcimadamore, line 1892: Shouldn't it be "values that are part of a group layout" ? Also maybe add a linkplain to `GroupLayout`. Same for sequence layout. I also note that you use the "are rendered" in the above text. Maybe you can lean on it - e.g.

"values in a group layout are rendered with enclosing curly braces".

mcimadamore: Possible questions: why are address values represented as toString? Or numbers in decimal form? What if I want hex form? This is the part that makes me uncomfortable about this method: in a way, you almost want it to generate some kind of XML or json schema - and then leave the toString mapping to the user - but this method is doing everything, so it has to make opinionated choices at every step of the way, and I can see how some of those choices would be good for some and bad for others.

minborg: > Possible questions: why are address values represented as toString? Or numbers in decimal form? What if I want hex form? This is the part that makes me uncomfortable about this method: in a way, you almost want it to generate some kind of XML or json schema - and then leave the toString mapping to the user - but this method is doing everything, so it has to make opinionated choices at every step of the way, and I can see how some of those choices would be good for some and bad for others.

I think, that once we have the Data Oriented Programming pattern in place, users can easily roll their own renderers and this renderer would only provide reasonable defaults. I think addresses should be rendered in hex form by default as suggested.

GavinRay97:
This is incredibly useful, thank you!

One question though -- do you know why it might fail for certain values after printing the stream?
For example, using this with JDK 19 early access release (I copy-pasted the util file):

```java
try (MemorySession session = MemorySession.openConfined()) {
    MemorySegment segment = session.allocate(4096);
    MemorySegmentRenderUtil.hexDump(segment).forEach(System.out::println);
}
```

Gives:
```java
0000000000000000  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
0000000000000010  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
0000000000000020  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
0000000000000030  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
0000000000000040  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
0000000000000050  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
0000000000000060  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
0000000000000070  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
Exception in thread "main" java.lang.ArrayIndexOutOfBoundsException: Index 268435448 out of bounds for length 16
	at com.example.MemorySegmentRenderUtil.appendHexTo(MemorySegmentRenderUtil.java:392)
	at com.example.MemorySegmentRenderUtil.appendHexTo(MemorySegmentRenderUtil.java:385)
	at com.example.MemorySegmentRenderUtil$HexStreamState.appendIndex(MemorySegmentRenderUtil.java:307)
	at com.example.MemorySegmentRenderUtil.lambda$hexDump$0(MemorySegmentRenderUtil.java:120)
	at java.base/java.util.stream.LongPipeline$1$1.accept(LongPipeline.java:177)
	at java.base/java.util.stream.Streams$RangeLongSpliterator.forEachRemaining(Streams.java:228)
	at java.base/java.util.Spliterator$OfLong.forEachRemaining(Spliterator.java:775)
```

test/jdk/java/foreign/RenderTest.java
```diff
@@ -0,0 +1,375 @@
+/*
+ * Copyright (c) 2020, 2022, Oracle and/or its affiliates. All rights reserved.
```
sundararajana, line 2: new test. why 2020?

minborg:
> This is incredibly useful, thank you!
> 
> One question though -- do you know why it might fail for certain values after printing the stream? For example, using this with JDK 19 early access release (I copy-pasted the util file):
> 
> ```java
> try (MemorySession session = MemorySession.openConfined()) {
>     MemorySegment segment = session.allocate(4096);
>     MemorySegmentRenderUtil.hexDump(segment).forEach(System.out::println);
> }
> ```
> 
> Gives:
> 
> ```java
> 0000000000000000  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
> 0000000000000010  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
> 0000000000000020  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
> 0000000000000030  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
> 0000000000000040  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
> 0000000000000050  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
> 0000000000000060  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
> 0000000000000070  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
> Exception in thread "main" java.lang.ArrayIndexOutOfBoundsException: Index 268435448 out of bounds for length 16
> 	at com.example.MemorySegmentRenderUtil.appendHexTo(MemorySegmentRenderUtil.java:392)
> 	at com.example.MemorySegmentRenderUtil.appendHexTo(MemorySegmentRenderUtil.java:385)
> 	at com.example.MemorySegmentRenderUtil$HexStreamState.appendIndex(MemorySegmentRenderUtil.java:307)
> 	at com.example.MemorySegmentRenderUtil.lambda$hexDump$0(MemorySegmentRenderUtil.java:120)
> 	at java.base/java.util.stream.LongPipeline$1$1.accept(LongPipeline.java:177)
> 	at java.base/java.util.stream.Streams$RangeLongSpliterator.forEachRemaining(Streams.java:228)
> 	at java.base/java.util.Spliterator$OfLong.forEachRemaining(Spliterator.java:775)
> ```
> 
> Specifically it seems to be this line:
> 
> ```java
>     static void appendHexTo(StringBuilder sb,
>                             byte value) {
>         sb.append((char) UPPERCASE_HEX_DIGITS[(value >>> 4)]);  // <---
>         sb.append((char) UPPERCASE_HEX_DIGITS[(value & (byte) 0x0f)]);
>     }
> ```
> 
> EDIT: Oh okay, I logged out `value` and it goes from `0x00, 0x10 ... 0x70 -> 0xffffff80` and that's where it explodes. This seems to work:
> 
> ```java
> static void appendHexTo(StringBuilder sb,
>                         byte value) {
>     String hex = Integer.toHexString(value & 0xff).toUpperCase();
>     if (hex.length() == 1) {
>         sb.append('0');
>     }
>     sb.append(hex);
> }
> ```
The issue identified has been fixed now. Thanks.

test/jdk/java/foreign/RenderTest.java
minborg: The test class was copied from an existing test class. I will fix it.

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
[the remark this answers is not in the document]
RogerRiggs: I worked on a more comprehensive HexPrinter API that was downsized to what became java.util.HexFormat.
The HexPrinter API was sidelined into the test test library: jdk/test/lib/hexdump/HexPrinter.

I agree that support for other memory buffers/segments is desirable as a generalization, even if only to HexFormat but expands the scope beyond Panama (and could be done separately).

A more general human readable formatting/layout capability is typically beyond what's included in JDK APIs. I might look to IntelliJ or the debugger to support the inspection functionality.

minborg:
Should we have a util class `MemoryInspection` with a single static method or should we add this method to `MemorySegment` itself (e.g. segment.toString())?

GavinRay97:
Sorry for the noob question, but with the most recent changes is there any way others can test them in their projects outside of the JDK source code?

Before, I was putting the files in `com.example.foreign`, which worked because it used no private/internal classes

But now it references IE `PaddingLayout` and `StructLayout` which are package-private (are they supposed to be private?)

I tried googling and found this, saying you can use `extension classes` as a JVM arg, but I'm not sure if that'd work because Maven/Gradle would need to be able to compile them to `.jar`s in the first place and that compilation would fail (presumably)

https://stackoverflow.com/questions/4838301/how-to-overwrite-classes-from-jdk

src/java.base/share/classes/java/lang/foreign/MemoryInspection.java
```diff
@@ -0,0 +1,210 @@
...
+        }
+
+        /**
+         * Renders the provided {@code addressLayout} and {@code value} to a String.
+         *
+         * @param addressLayout the layout to render
+         * @param value         the value to render
+         * @return rendered String
+         */
+        default String render(ValueLayout.OfAddress addressLayout, MemorySegment value) {
```
GavinRay97, line 187: I think you can make it take `Addressable` so it works for both `MemorySegment` and `MemoryAddress`:

```java
default String render(ValueLayout.OfAddress addressLayout, Addressable value) {
    requireNonNull(addressLayout);
    return String.format("0x%0" + (ValueLayout.ADDRESS.byteSize() * 2) + "X", value.address());
}
```

In JDK 20 release I get an error:
image


Trying to do `(MemorySegment)` cast fails:

image

But if I make it take `Addressable` instead, then it works:
image

mcimadamore: Note that the API in the Panama repo is different - MemoryAddress (and Addressable) have been removed as part of:
https://github.com/openjdk/panama-foreign/pull/694

GavinRay97: Oh, well that'd make sense why it didn't work OOTB then, thanks for the info!

GavinRay97:
I just want to say thanks again for how useful this functionality is

With latest changes, I've got a demo of visualizing the members of a database's Heap File pages buffer 

This is saving me so much headache in debugging and developing.

image

image

minborg:
@GavinRay97 We are going to split this PR into two. This one will only contain the HexFormat-specific parts. The other is here: https://github.com/openjdk/panama-foreign/pull/715

mcimadamore:
This PR seems superseded by https://github.com/openjdk/panama-foreign/pull/833.
If so, it should be closed.

minborg:
We anticipate much more general functionality with https://github.com/openjdk/panama-foreign/pull/833 and so, I am going to close this PR.

PR #699, opened 2022-08-05
8275644: Replace VMReg in shuffling code with something more fine grained.

JornVernee:
This patch replaces `VMReg` with a new `VMStorage` class in the C++ code, which can be used to represent stack arguments of different sizes, as well as allowing finer grained stack addressing.

The old masm code that used `VMReg` for moving values is also replaced with new code that uses `VMStorage`. And `VMReg` is changed to `VMStorage` everywhere.

The `vmstorageBase` class was added in C++ to mirror the `jdk.internal.foreign.VMStorage`. That class now also has an additional field to encode the size of a stack argument, or a register mask, which can also be used to encode the size of a register argument. This also leads to some stubs needing to be updated on unsupported platforms.

Lastly, instead of encoding stack indexes as 64-bit slots as we did before, they are now encoded as byte offsets, which allows finer grained stack addressing. (this leads to a bunch of values being multiplied by 8 in the Java code as well).

---

This change is needed to address https://bugs.openjdk.org/browse/JDK-8275584 as well as adding support for `long double` as well as more exotic ABIs that take arguments as vectors.

---------
### Progress

### Issue
 * [JDK-8275644](https://bugs.openjdk.org/browse/JDK-8275644): Replace VMReg in shuffling code with something more fine grained.

### Reviewers
 * [Maurizio Cimadamore](https://openjdk.org/census#mcimadamore) (@mcimadamore - Committer) ⚠️ Review applies to [513ccc02](https://git.openjdk.org/panama-foreign/pull/699/files/513ccc02ed15e3b51d73458514e7f33d4d376445)
 * [Nick Gasson](https://openjdk.org/census#ngasson) (@nick-arm - Committer) ⚠️ Review applies to [7d54e3d5](https://git.openjdk.org/panama-foreign/pull/699/files/7d54e3d59b3521ff69f8a107f7aba887fe48890e)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/699/head:pull/699` \
`$ git checkout pull/699`

`$ git checkout pull/699` \
`$ git pull https://git.openjdk.org/panama-foreign pull/699/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 699`

`$ git pr show -t 699`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/699.diff">https://git.openjdk.org/panama-foreign/pull/699.diff</a>

</details>


src/java.base/share/classes/jdk/internal/foreign/abi/VMStorage.java
```diff
@@ -27,21 +27,35 @@
 import java.util.Objects;
 
 public class VMStorage {
-    private final int type;
+    private final byte type;
+    private final short segmentMaskOrSize;
```
mcimadamore, line 31: We have a couple of fields here which do double duty - e.g. depending on whether the storage models a stack argument or a record, we end up with different meanings. One way is to use (as you did) names like `segmentMaskOrSize`.

But perhaps a better (?) way would be to make VMStorage an abstract class/interface and then have two implementations, for stack slots and registers. Although I realize this might impact the VM code in a bad way (as I think this class is mirrored there in some way).

mcimadamore (approved):
Java code looks good - left some minor comment.
I like the generality that this approach buys, I think it will be helpful when modelling more exotic ABIs/calling conventions.

src/java.base/share/classes/jdk/internal/foreign/abi/VMStorage.java
mcimadamore: That's ok, I suspected something similar was at play. In that case I suggest a uniform naming scheme, plus perhaps some documents explaining the semantics of the fields in the various cases.

JornVernee: I can add some comments to the fields. What do you mean by 'uniform naming scheme'?

mcimadamore: > I can add some comments to the fields. What do you mean by 'uniform naming scheme'?

Sorry, I meant that one field is called `segmentMaskOrSize` but the other is just called `index`. Using `or` on both might be a good idea.

JornVernee: Ah, ok. That sounds good.

JornVernee:
I'm still testing but I added some comments, and changed the name to indexOrOffset/index_or_offset on both sides. (I also got rid of the union in the C++ code, since it didn't really buy us anything)

JornVernee:
@nick-arm Could you take a look as well?

I've also taken a stab at fixing [Incorrect stack spilling in CallArranger on MacOS/AArch64](https://bugs.openjdk.org/browse/JDK-8275584)
Here: https://github.com/JornVernee/panama-foreign/compare/VMStorage2...JornVernee:panama-foreign:Mac_Stack
But there are still some failing tests.

src/hotspot/cpu/aarch64/foreignGlobals_aarch64.cpp
```diff
@@ -42,112 +43,165 @@ bool ABIDescriptor::is_volatile_reg(FloatRegister reg) const {
...
-
-    masm->block_comment(err_msg("bt=%s", null_safe_string(type2name(arg_bt))));
-    switch (arg_bt) {
-      case T_BOOLEAN:
-      case T_BYTE:
-      case T_SHORT:
-      case T_CHAR:
-      case T_INT:
-        masm->move32_64(from_vmreg, to_vmreg, tmp_reg);
+constexpr int RFP_BIAS = 16; // skip old rfp and lr
```
nick-arm, line 100: Should this be `static constexpr`?

src/hotspot/cpu/aarch64/vmstorage_aarch64.inline.hpp
```diff
@@ -0,0 +1,83 @@
...
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+#ifndef SHARE_PRIMS_VMSTORAGE_AARCH64
+#define SHARE_PRIMS_VMSTORAGE_AARCH64
```
nick-arm, line 25: Should this be `CPU_AARCH64_VMSTORAGE_AARCH64_HPP`?

nick-arm (approved):
The AArch64-specific changes look sensible to me!

src/hotspot/cpu/aarch64/foreignGlobals_aarch64.cpp
JornVernee: Can do.

src/hotspot/cpu/aarch64/vmstorage_aarch64.inline.hpp
JornVernee: Yes. I messed up this pattern for all the new header files. Thanks for catching!

PR #700, opened 2022-08-06
Prevent maxAlign virtual calls for polluted accesses

rsmogura:
In case of polluted accesses (when different kinds of segments are accessed
from same code), `maxAlign()` can get virtual call which would prevent
effective inlining and loop optimizations.

This patch moves `maxAlign` to `AbstractMemorySegmentImpl` field, and makes method
final. The value of align is passed as constructor argument.

_Note: This patch can cause slightly bigger memory usage, as memory segment will carry `maxAlign` value, this can optimizaed by using smaller container for value i. e. `byte` or `short`_

After
```
Benchmark                                  (size)  Mode  Cnt       Score     Error  Units
MixedAccessBenchmarks.directCopy          1048576  avgt   10   16410.733 ±  79.901  ns/op
MixedAccessBenchmarks.pollutedAccessCopy  1048576  avgt   10  168497.502 ± 632.578  ns/op
```

Before
```
Benchmark                                  (size)  Mode  Cnt        Score        Error  Units
MixedAccessBenchmarks.directCopy          1048576  avgt   10    18336.054 ±     63.133  ns/op
MixedAccessBenchmarks.pollutedAccessCopy  1048576  avgt   10  2069032.456 ± 167512.633  ns/op
```

---------
### Progress

### Integration blockers
&nbsp;⚠️ Too few reviewers with at least role committer found (have 0, need at least 1) (failed with the updated jcheck configuration)
&nbsp;⚠️ Whitespace errors (failed with the updated jcheck configuration)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign.git pull/700/head:pull/700` \
`$ git checkout pull/700`

`$ git checkout pull/700` \
`$ git pull https://git.openjdk.org/panama-foreign.git pull/700/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 700`

`$ git pr show -t 700`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/700.diff">https://git.openjdk.org/panama-foreign/pull/700.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/panama-foreign/pull/700#issuecomment-1207177155)

### DIFF
--- a/src/java.base/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
@@ -69,6 +69,8 @@ public abstract non-sealed class AbstractMemorySegmentImpl implements MemorySegm
 
     private static final ScopedMemoryAccess SCOPED_MEMORY_ACCESS = ScopedMemoryAccess.getScopedMemoryAccess();
 
+    private final long maxAlignMask;
+
     static final JavaNioAccess nioAccess = SharedSecrets.getJavaNioAccess();
 
     final long length;
@@ -76,9 +78,10 @@ public abstract non-sealed class AbstractMemorySegmentImpl implements MemorySegm
     final MemorySession session;
 
     @ForceInline
-    AbstractMemorySegmentImpl(long length, boolean readOnly, MemorySession session) {
+    AbstractMemorySegmentImpl(long length, boolean readOnly, long maxAlignMask, MemorySession session) {
         this.length = length;
         this.readOnly = readOnly;
+        this.maxAlignMask = maxAlignMask;
         this.session = session;
     }
 
@@ -319,7 +322,9 @@ public abstract non-sealed class AbstractMemorySegmentImpl implements MemorySegm
 
     // Helper methods
 
-    public abstract long maxAlignMask();
+    public final long maxAlignMask() {
+        return maxAlignMask;
+    }
 
     @ForceInline
     public final boolean isAlignedForElement(long offset, MemoryLayout layout) {
--- a/src/java.base/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
@@ -66,10 +66,12 @@ import static java.lang.foreign.ValueLayout.JAVA_BYTE;
  * {@link MappedMemorySegmentImpl}.
  */
 public abstract non-sealed class AbstractMemorySegmentImpl implements MemorySegment, SegmentAllocator, Scoped, BiFunction<String, List<Number>, RuntimeException> {
+    private static final long MAX_ALIGN_1 = 1;
+    private static final long MAX_ALIGN_2 = 2;
+    private static final long MAX_ALIGN_4 = 4;
+    private static final long MAX_ALIGN_8 = 8;
 
     private static final ScopedMemoryAccess SCOPED_MEMORY_ACCESS = ScopedMemoryAccess.getScopedMemoryAccess();
-
-    private final long maxAlignMask;
 
     static final JavaNioAccess nioAccess = SharedSecrets.getJavaNioAccess();
 
@@ -78,10 +80,9 @@ public abstract non-sealed class AbstractMemorySegmentImpl implements MemorySegm
     final MemorySession session;
 
     @ForceInline
-    AbstractMemorySegmentImpl(long length, boolean readOnly, long maxAlignMask, MemorySession session) {
+    AbstractMemorySegmentImpl(long length, boolean readOnly, MemorySession session) {
         this.length = length;
         this.readOnly = readOnly;
-        this.maxAlignMask = maxAlignMask;
         this.session = session;
     }
 
@@ -322,8 +323,26 @@ public abstract non-sealed class AbstractMemorySegmentImpl implements MemorySegm
 
     // Helper methods
 
+    @ForceInline
     public final long maxAlignMask() {
-        return maxAlignMask;
+        if (this instanceof NativeMemorySegmentImpl) {
+            return 0;
+        } else if (this instanceof HeapMemorySegmentImpl.OfByte) {
+            return MAX_ALIGN_1;
+        } else if (this instanceof HeapMemorySegmentImpl.OfChar) {
+            return MAX_ALIGN_2;
+        } else if (this instanceof HeapMemorySegmentImpl.OfShort) {
+            return MAX_ALIGN_2;
+        } else if (this instanceof HeapMemorySegmentImpl.OfInt) {
+            return MAX_ALIGN_4;
+        } else if (this instanceof HeapMemorySegmentImpl.OfLong) {
+            return MAX_ALIGN_8;
+        } else if (this instanceof HeapMemorySegmentImpl.OfFloat) {
+            return MAX_ALIGN_4;
+        } else if (this instanceof HeapMemorySegmentImpl.OfDouble) {
+            return MAX_ALIGN_8;
+        }
+        throw new IllegalStateException("Unknown memory segment");
     }
 
     @ForceInline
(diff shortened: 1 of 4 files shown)


src/java.base/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
```diff
@@ -319,7 +323,27 @@ public void checkValidState() {
 
     // Helper methods
 
-    public abstract long maxAlignMask();
+    @ForceInline
```
mcimadamore, line 326: When adding `maxAlign` I have played with something similar as what you have done. My recollection is that doing this doesn't 100% solve the issue, as it's possible to have polluted profile for `maxAlignMask` (e.g. the JVM will bias the method implementation towards the 1-2 layouts that seem to be more common, and put everything else in an uncommon branch). Maybe what you see is that, since the virtual call is gone, there is still a net gain.

Does the problem only manifest with bulk copy? Or also with plain memory access? If the former, perhaps some other inlining issue with bulk copy (which is a big method) could be at play.

rsmogura: I just checked it with vector operations only, there's check in `ScopedMemoryAccess` `loadFromMemorySegment` which checks max align in case of polluted access this will create polymorphic call.

- However when I look right now I think there's second option. Instead of changing `maxAlign`, the check in `loadFromMemorySegment`, could be replaced by checking if segment instance is native or `Bytes`.?

mcimadamore: is this the affected code?
```
// @@@ Smarter alignment checking if accessing heap segment backing non-byte[] array
        if (msp.maxAlignMask() > 1) {
            throw new IllegalArgumentException();
        }
 ```

Comment (rsmogura) [reply src/java.base/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java:326]: Exactly!

Comment (mcimadamore) [reply src/java.base/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java:326]: I guess which solution is taken depends on how deep of a problem this is. I suggest writing a benchmark using e.g. bulk copy between segments, and see if the same issue occurs. We do have tests for polluted memory segment access and we have not seen issues with maxAlignMask showing up there, so I'm curious as to why the vector use-site seems to be more problematic. I wonder if part of the issue is the lack of argument type profiling - e.g. profile info based on the type of arguments in a static call. I think Method/VarHandle have that enabled by default (for obvious reasons) and there are hacks in the JVM to allow that for Unsafe as well. But that support is not enabled for this particular vector call, and I wonder if that could lead to the problem you describe.

https://github.com/openjdk/jdk/blob/master/src/hotspot/share/oops/methodData.cpp#L1583

Comment (rsmogura) [reply src/java.base/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java:326]: I created test with core method like this
```java
    @CompilerControl(DONT_INLINE)
    private static long sum(MemorySegment in) {
        long sz = in.byteSize() / Integer.BYTES;
        long result = 0L;
        for (long i = 0; i < sz; i += Integer.BYTES) {
            result += (int) JAVA_INT.get(in, i);
        }

        return result;
    }
```
But results after / before patch does not differs so much. 

Interesting I see some additional optimization turns on after some time (it should be fully JITed during warm-up)
```
Iteration   3: 228720.552 ns/op
Iteration   4: 220640.677 ns/op
Iteration   5: 12248.776 ns/op
Iteration   6: 9470.916 ns/op
Iteration   7: 9422.982 ns/op
```

Later I can try a bit with `TypeProfileLevel` looking at code setting it to `2` should enable mentioned profiling always.

rsmogura:
I made tests with using this flag `-XX:TypeProfileLevel=2` my understanding is that this flag should enable all profiling. Results for polluted case [1] are similar
Here are results for polluted case:
```
Iteration   5: 212232.300 ns/op
Iteration   6: 11157.340 ns/op
```
It does not matter if flag is turned on or off, or if it's with or without PR changes.

[1] https://gist.github.com/rsmogura/42bcb22ca55a730c0c46a4035f179422

src/java.base/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
rsmogura: @mcimadamore,

I've updated JVM method to allow profiling arguments as you hinted, for vector operations
```
bool MethodData::profile_unsafe(const methodHandle& m, int bci) {
  Bytecode_invoke inv(m , bci);
  if (inv.is_invokevirtual()) {
    Symbol* klass = inv.klass();
    if (klass == vmSymbols::jdk_internal_misc_Unsafe() ||
        klass == vmSymbols::sun_misc_Unsafe() ||
        klass == vmSymbols::jdk_internal_misc_ScopedMemoryAccess()) {
      Symbol* name = inv.name();
      if (name->starts_with("get") || name->starts_with("put") || name->starts_with("load") || name->starts_with("store")) {
        return true;
      }
    }
  }
  return false;
}
```
This improved performance with benchmarks, in same way as a patch:

Before
```
MixedAccessBenchmarks.directCopy          1048576  avgt   10    22763.442 ±   2145.833  ns/op
MixedAccessBenchmarks.pollutedAccessCopy  1048576  avgt   10  3536082.952 ± 309959.183  ns/op
```
After
```
Benchmark                                  (size)  Mode  Cnt       Score      Error  Units
MixedAccessBenchmarks.directCopy          1048576  avgt   10   23006.005 ± 2497.308  ns/op
MixedAccessBenchmarks.pollutedAccessCopy  1048576  avgt   10  254049.440 ± 5218.658  ns/op
```

mcimadamore: Thanks for the experiment - that seems a better way to handle this situation, but I'll leave further comments to @PaulSandoz or @iwanowww.

PaulSandoz: Expanding the profiling to for method names starting with "loadFrom" and "storeInto" seems reasonable, given the most other access methods are also profiled (although not all the read-modify-write methods are).

I think we should separate out the comparison of method names thereby we don't inadvertently profile other methods just in case new methods get added. Plus we can add some comments to the class doc of `Unsafe` and `ScopedMemoryAccess` to indicate such methods are profiled. Ideally I would prefer an HotSpot-specific annotation indicated that all arguments of an annotated method should be profiled.
 
The profiling of all arguments is a blunter hammer that necessary in this case but so be it.

rsmogura: Adding annotation for this could be useful for profiling other methods too. I'll try to find some time and prepare some draft.

One thing I see that adding such annotation could consume one bit here (still there will be 5 spare): 
https://github.com/openjdk/jdk/blob/master/src/hotspot/share/oops/method.hpp#L98

PaulSandoz: Ok. I think it may be better to propose the more focused fix to JDK 20, and then use that PR as the opportunity to ask the question to the HotSpot devs whether they think an annotation would be useful before diving in.

rsmogura: So you are suggesting to create PR with annotations and ask on hotspot-dev what they think about it?

PaulSandoz: No, i am suggesting to create a PR with changes to https://github.com/openjdk/jdk/blob/master/src/hotspot/share/oops/methodData.cpp#L1583 as you proposed. Then before you dive into any annotation driven approach ask on the hotspot email list, or ask in that PR.

PaulSandoz:
Linking to https://bugs.openjdk.org/browse/JDK-8292710 so we don't forget it.

mcimadamore:

PR #705, opened 2022-08-10
8292174: Add unaligned layouts to ValueLayout

minborg:
The layouts are placed in a special `Unaligned` class. This prevents the classes to be seen on the top level (for less cluttering) and allows lazy initialization of items that are likely not so much used.

---------
### Progress

### Issue
 * [JDK-8292174](https://bugs.openjdk.org/browse/JDK-8292174): Add unaligned layouts to ValueLayout

### Reviewers
 * [Maurizio Cimadamore](https://openjdk.org/census#mcimadamore) (@mcimadamore - Committer) ⚠️ Review applies to [ab34e8a0](https://git.openjdk.org/panama-foreign/pull/705/files/ab34e8a0af2f5251844e1aef3c37d74cfe66a0a9)
 * [Paul Sandoz](https://openjdk.org/census#psandoz) (@PaulSandoz - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/705/head:pull/705` \
`$ git checkout pull/705`

`$ git checkout pull/705` \
`$ git pull https://git.openjdk.org/panama-foreign pull/705/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 705`

`$ git pr show -t 705`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/705.diff">https://git.openjdk.org/panama-foreign/pull/705.diff</a>

</details>


src/java.base/share/classes/java/lang/foreign/ValueLayout.java
```diff
@@ -649,6 +649,126 @@ public boolean isUnbounded() {
...
+         * An unaligned value layout constant whose size is the same as that of a machine address ({@code size_t}),
+         * and byte order set to {@link ByteOrder#nativeOrder()}.
+         * Equivalent to the following code:
+         * {@snippet lang=java :
+         * MemoryLayout.valueLayout(MemorySegment.class, ByteOrder.nativeOrder());
+         * }
+         * @apiNote Care should be taken when using unaligned value layouts as they may induce
+         *          performance and portability issues.
+         */
+        public static final OfAddress UNALIGNED_ADDRESS = new OfAddress(ByteOrder.nativeOrder());
```
JornVernee, line 672: This seems to create an address layout with the alignment set to the size of the layout? For unaligned layouts we want the alignment set to 1 byte/8 bits I think.

mcimadamore: I'd like not to have a separate nested class if possible. That will show up in javadoc and make those constants less discoverable.

Also, I think it would be better to use UNALIGNED as a suffix - e.g. JAVA_INT_UNALIGNED.

mcimadamore (commented):
I suggest you make a search for _UNALIGNED in the API implementation and test/benchmarks. There some places where we used hand-written unaligned layouts - e.g. BindingSpecializer and test/micro/org/openjdk/bench/java/lang/foreign/JavaLayouts.java to name a few.

minborg:
`JAVA_BYTE` and `JAVA_BOOLEAN` are implicitly unaligned. We could either:

* provide `JAVA_BYTE_UNALIGNED` for symmetry which is identical to `JAVA_BYTE` (and the same for boolean).
* provide neither `JAVA_BYTE_UNALIGNED` nor `JAVA_BYTE_UNALIGNED`

src/java.base/share/classes/java/lang/foreign/ValueLayout.java
minborg: > This seems to create an address layout with the alignment set to the size of the layout? For unaligned layouts we want the alignment set to 1 byte/8 bits I think.

You are right. So, parts like `new OfInt(ByteOrder.nativeOrder()).withBitAlignment(32)` could be simplified to `new OfInt(ByteOrder.nativeOrder())`.

mcimadamore:
> `JAVA_BYTE` and `JAVA_BOOLEAN` are implicitly unaligned. We could either:
> 
>     * provide `JAVA_BYTE_UNALIGNED` for symmetry which is identical to `JAVA_BYTE` (and the same for boolean).
> 
>     * provide neither `JAVA_BYTE_UNALIGNED` nor `JAVA_BYTE_UNALIGNED`

Leave alone UNALIGNED for 8-bit entities. I think that's fine.

mcimadamore (approved):
Looks good

mcimadamore (commented):
I think one of my comment got lost here. There are places where we define unaligned layouts in the implementation. Look at:

src/java.base/share/classes/jdk/internal/foreign/abi/BindingSpecializer.java (seem unused)
src/java.base/share/classes/jdk/internal/foreign/abi/SharedUtils.java (these are used)
test/jdk/java/foreign/TestMemoryDereference.java
test/micro/org/openjdk/bench/java/lang/foreign/JavaLayouts.java

For the last one, my suggestion is to drop the JavaLayouts class, and inline the use where needed.

src/java.base/share/classes/java/lang/foreign/ValueLayout.java
```diff
@@ -649,6 +649,8 @@ public boolean isUnbounded() {
         }
     }
 
+    // Some constants below are accessed from generated bytecode, see BindingSpecializer::emitLoadLayoutConstant
```
mcimadamore, line 652: Which constants "below" ? The comment would have been perfect in the old code, but now it seems the constants have been removed?

minborg: So, before they were all in a special class (Platform) but now the public layouts are used.

mcimadamore: What I meant is: the comment you copied referred to the layout definitions which have now been removed. So I think it's better to remove the comment.

minborg: OK. I thought the comment meant: "Do not change, move or remove any of these constants as they are invisibly used in bytecode generated in BindingSpecializer.". But I can remove the comment.

mcimadamore (approved):
Looks good

src/java.base/share/classes/java/lang/foreign/ValueLayout.java
```diff
@@ -740,5 +738,102 @@ public boolean isUnbounded() {
...
-    public static final OfDouble JAVA_DOUBLE = new OfDouble(ByteOrder.nativeOrder()).withBitAlignment(64);
+    public static final OfDouble JAVA_DOUBLE = new OfDouble(ByteOrder.nativeOrder());
+
+    /**
+     * An unaligned value layout constant whose size is the same as that of a machine address ({@code size_t}),
+     * and byte order set to {@link ByteOrder#nativeOrder()}.
+     * Equivalent to the following code:
+     * {@snippet lang=java :
+     * MemoryLayout.valueLayout(MemorySegment.class, ByteOrder.nativeOrder())
+     *             .withBitAlignment(8);
```
PaulSandoz, line 749: An alternative snippet is to leverage the exiting aligned constant and make it unaligned, like what we would commonly do in internal code:
```
ADDRESS.withBitAlignment(8);
```
I think expresses the intent more clearly.

PaulSandoz (approved):

PR #708, opened 2022-08-18
Micro optimizations and tidy

minborg:
This small PR adds some optimizations and tidies up some imports.

---------
### Progress

### Reviewers
 * [Paul Sandoz](https://openjdk.org/census#psandoz) (@PaulSandoz - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/708/head:pull/708` \
`$ git checkout pull/708`

`$ git checkout pull/708` \
`$ git pull https://git.openjdk.org/panama-foreign pull/708/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 708`

`$ git pr show -t 708`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/708.diff">https://git.openjdk.org/panama-foreign/pull/708.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/java/lang/foreign/FunctionDescriptor.java
+++ b/src/java.base/share/classes/java/lang/foreign/FunctionDescriptor.java
@@ -26,7 +26,6 @@ package java.lang.foreign;
 
 import java.lang.invoke.MethodHandle;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
@@ -81,8 +80,7 @@ public sealed class FunctionDescriptor {
      */
     public static FunctionDescriptor of(MemoryLayout resLayout, MemoryLayout... argLayouts) {
         Objects.requireNonNull(resLayout);
-        Objects.requireNonNull(argLayouts);
-        Arrays.stream(argLayouts).forEach(Objects::requireNonNull);
+        assertVarargsNonNull(argLayouts);
         return new FunctionDescriptor(resLayout, List.of(argLayouts));
     }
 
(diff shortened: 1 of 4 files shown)


src/java.base/share/classes/java/lang/foreign/FunctionDescriptor.java
```diff
@@ -81,8 +80,7 @@ public List<MemoryLayout> argumentLayouts() {
      */
     public static FunctionDescriptor of(MemoryLayout resLayout, MemoryLayout... argLayouts) {
         Objects.requireNonNull(resLayout);
-        Objects.requireNonNull(argLayouts);
-        Arrays.stream(argLayouts).forEach(Objects::requireNonNull);
+        assertVarargsNonNull(argLayouts);
```
PaulSandoz, line 83: This check is redundant, since it is performed by `List.of` (granted that the stack trace is different but i think that is ok for NPEs).

src/java.base/share/classes/java/lang/foreign/FunctionDescriptor.java
```diff
@@ -107,8 +104,7 @@ public static FunctionDescriptor ofVoid(MemoryLayout... argLayouts) {
      * @return a variadic function descriptor, or this descriptor if {@code variadicLayouts.length == 0}.
      */
     public FunctionDescriptor asVariadic(MemoryLayout... variadicLayouts) {
-        Objects.requireNonNull(variadicLayouts);
-        Arrays.stream(variadicLayouts).forEach(Objects::requireNonNull);
+        assertVarargsNonNull(variadicLayouts);
         return variadicLayouts.length == 0 ? this : new VariadicFunction(this, variadicLayouts);
```
PaulSandoz, line 108: We could change the VariadicFunction constructor to accept a List of non-null elements.

src/java.base/share/classes/java/lang/foreign/FunctionDescriptor.java
minborg: Well spotted!

src/java.base/share/classes/java/lang/foreign/FunctionDescriptor.java
minborg: It already throws an NPE if any of the args or variadicLayouts elements are `null`. So, the check is redundant also here.

minborg: Well, it turned out it didn't. But soon it will...

PaulSandoz (approved):

PR #711, opened 2022-08-22
Split foreign vector load and store by null or not null base

rsmogura:
Split store / load operation by if checking if base is null
or not null.

When this happens base in Unsafe is not perceived with mixed
access by VM, and VM does not insert barriers.

Test results gives the expected values where the case of polluted access is 2x multiplication of normal access.

After
```
Benchmark                                    (size)  Mode  Cnt    Score     Error  Units
MemorySegmentVectorAccess.arrayCopy            1024  avgt   10    7.437 ±   0.195  ns/op
MemorySegmentVectorAccess.directSegments       1024  avgt   10   15.593 ±   0.371  ns/op
MemorySegmentVectorAccess.heapSegments         1024  avgt   10   16.997 ±   0.118  ns/op
MemorySegmentVectorAccess.pollutedSegments2    1024  avgt   10   58.673 ± 105.783  ns/op
MemorySegmentVectorAccess.pollutedSegments3    1024  avgt   10   67.216 ±  16.157  ns/op
MemorySegmentVectorAccess.pollutedSegments4    1024  avgt   10  122.567 ± 263.950  ns/op
MemorySegmentVectorAccess.pollutedSegments5    1024  avgt   10  114.725 ± 209.183  ns/op
```

Before
```
Benchmark                                    (size)  Mode  Cnt    Score   Error  Units
MemorySegmentVectorAccess.arrayCopy            1024  avgt   10    8.547 ± 0.115  ns/op
MemorySegmentVectorAccess.directSegments       1024  avgt   10   15.536 ± 0.082  ns/op
MemorySegmentVectorAccess.heapSegments         1024  avgt   10   15.818 ± 0.101  ns/op
MemorySegmentVectorAccess.pollutedSegments2    1024  avgt   10  146.380 ± 1.127  ns/op
MemorySegmentVectorAccess.pollutedSegments3    1024  avgt   10  290.784 ± 7.274  ns/op
MemorySegmentVectorAccess.pollutedSegments4    1024  avgt   10  297.187 ± 5.096  ns/op
MemorySegmentVectorAccess.pollutedSegments5    1024  avgt   10  310.166 ± 9.310  ns/op
```

Additonally with profiling `load` and `store` method arguments as
described in [1]
```
Benchmark                                    (size)  Mode  Cnt    Score   Error  Units
MemorySegmentVectorAccess.arrayCopy            1024  avgt   10    7.480 ± 0.169  ns/op
MemorySegmentVectorAccess.directSegments       1024  avgt   10   15.497 ± 0.062  ns/op
MemorySegmentVectorAccess.heapSegments         1024  avgt   10   16.829 ± 0.132  ns/op
MemorySegmentVectorAccess.pollutedSegments2    1024  avgt   10  145.436 ± 1.081  ns/op
MemorySegmentVectorAccess.pollutedSegments3    1024  avgt   10  291.081 ± 2.297  ns/op
MemorySegmentVectorAccess.pollutedSegments4    1024  avgt   10  305.388 ± 7.518  ns/op
MemorySegmentVectorAccess.pollutedSegments5    1024  avgt   10  303.931 ± 3.412  ns/op
```

[1] https://github.com/openjdk/panama-foreign/pull/700

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign.git pull/711/head:pull/711` \
`$ git checkout pull/711`

`$ git checkout pull/711` \
`$ git pull https://git.openjdk.org/panama-foreign.git pull/711/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 711`

`$ git pr show -t 711`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/711.diff">https://git.openjdk.org/panama-foreign/pull/711.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/panama-foreign/pull/711#issuecomment-1221969570)


PaulSandoz:
Are the third set of performance results produced by measuring the changes from this PR *and* changes from #700?

rsmogura:
Yes the third result set compares against https://github.com/openjdk/panama-foreign/pull/700.

I'm not sure how to interpret this results and explain that avoiding barriers gives similar results as profiling arguments. Maybe it's because barrier prevents some other optimizations.

rsmogura:
One more thing worth noticing is that in https://github.com/openjdk/panama-foreign/pull/700 profiling really give performance boost, and single difference is that loop bound was calculated like this
```
long sz = in.byteSize();
for (long i = 0; i < SPECIES_BYTE.loopBound(sz); i += SPECIES_BYTE.vectorByteSize()) {
```
instead of
```
for (long i = 0; i < SPECIES.loopBound(in.byteSize()); i += SPECIES.vectorByteSize()) {
```
So loop bound could definitely be treated as loop invariant even with mem barriers added by polluted (mixed) access, however profiling looks like not solving all issues well in 2nd case.

PaulSandoz:
It's frustrating we have to do this. @iwanowww any opinions on this?

rsmogura:
I can try to add it to VM, I wonder if it would be enough to create such If in VM and than duplicate current intrinsic call for each branch. However I'm not sure how far I'll go with my paramount ;) VM skills

rsmogura:
I created something to move it to VM, but there's huge amount of boilerplate code (surprisingly it not crashed) - and this is only support for store vector.

https://github.com/openjdk/panama-foreign/compare/foreign-memaccess+abi...rsmogura:panama-foreign:split-load-store-by-null-base-in-vm?expand=1

rsmogura:
I think I would need help, as I've found that large number of deoptimizations happens when I execute following code (both in case of Java split and VM split):
```java
    public static int test3(MemorySegment in, MemorySegment out, MemorySegment out2, byte[] arr) {
        long sz = in.byteSize();
        var zero = ByteVector.zero(SPECIES_BYTE);
        for (long i = 0; i < SPECIES_BYTE.loopBound(in.byteSize()); i += SPECIES_BYTE.vectorByteSize()) {
            var v1 = ByteVector.fromMemorySegment(SPECIES_BYTE, in, i, ByteOrder.nativeOrder());
//            arr[i] = (byte) 0;
            v1.intoMemorySegment(out, i, ByteOrder.nativeOrder());
        }

        return 0;
    }

    public static void main(String[] args) throws Exception {
        var session = MemorySession.openConfined();
        MemorySegment heapIn = MemorySegment.ofArray(new byte[size]);
        MemorySegment heapOu = MemorySegment.ofArray(new byte[size]);

        MemorySegment directIn = MemorySegment.allocateNative(size, session);
        MemorySegment directOu = MemorySegment.allocateNative(size, session);

        for (int i=0; i < 30_000; i++) {
            test3(heapIn, heapOu, heapOu, (byte[]) heapOu.array().get());
            test3(directIn, directOu, directOu, (byte[]) heapOu.array().get());
        }
    }
```
In compilation log I have huge amount of entries like
```
<deoptimized thread='31917' reason='constraint' pc='0x00007fffe134cdc7' compile_id='895' compiler='c1' level='3'>
<jvms bci='78' method='eu.smogura.panama.tests.vectorscopy.Main test3 (Ljava/lang/foreign/MemorySegment;Ljava/lang/foreign/MemorySegment;Ljava/lang/foreign/MemorySegment;[B)I' bytes='83' count='6061' backedge_count='39955677' iicount='6061' decompiles='95' profile_predicate_traps='100' overflow_recompiles='92'/>
```
The VM options I use
```
"-XX:+UnlockDiagnosticVMOptions", "-XX:CompileCommand=dontinline,\\*::test3\\*", "-XX:+LogCompilation",
```

Other thing I noticed, there's huge number of _PhaseIdealLoop_ phases (hits allowed maximum) and it create 64 CountedLoopNodes for main part of loop (there's should be at most 4 unswitched branches).

I wonder if someone could help me with this concern?

PaulSandoz:
@rsmogura you need some guidance from HotSpot engineers such as @iwanowww

rsmogura:
I moved forward with trying to implement this inside VM, but I hit the wall: mainly I get often a method (`pollutedSegments4`) to be recompiled. I narrowed it to switch `UseProfiledLoopPredicate` and on graph, I see that there's profile_predicate connected to check `class == NativeMemorySegmentImpl`. This causes the method to be recompiled as we use both classes, however after recompilation, the same graph is generated - so I think this mainly runs in interpreter mode.

Shortly, for now, this PR is the single working idea I have.

mcimadamore:

PR #713, opened 2022-08-25
8292918: Split FunctionDescriptor into an interface and implementation(s)

minborg:
This PR splits the current FunctionDescription class into an interface and two internal implementing classes.

---------
### Progress

### Issue
 * [JDK-8292918](https://bugs.openjdk.org/browse/JDK-8292918): Split FunctionDescriptor into an interface and implementation(s)

### Reviewers
 * [Athijegannathan Sundararajan](https://openjdk.org/census#sundar) (@sundararajana - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/713/head:pull/713` \
`$ git checkout pull/713`

`$ git checkout pull/713` \
`$ git pull https://git.openjdk.org/panama-foreign pull/713/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 713`

`$ git pr show -t 713`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/713.diff">https://git.openjdk.org/panama-foreign/pull/713.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/jdk/internal/foreign/FunctionDescriptorImpl.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/FunctionDescriptorImpl.java
@@ -105,7 +105,8 @@ public sealed class FunctionDescriptorImpl implements FunctionDescriptor {
      * @throws IllegalArgumentException if {@code index < 0 || index > argumentLayouts().size()}.
      */
     public FunctionDescriptorImpl insertArgumentLayouts(int index, MemoryLayout... addedLayouts) {
-        checkIndex(index);
+        if (index < 0 || index > argLayouts.size())
+            throw new IllegalArgumentException("Index out of bounds: " + index);
         List<MemoryLayout> added = List.of(addedLayouts); // null check on array and its elements
         List<MemoryLayout> newLayouts = new ArrayList<>(argLayouts.size() + addedLayouts.length);
         newLayouts.addAll(argLayouts.subList(0, index));
@@ -176,11 +177,6 @@ public sealed class FunctionDescriptorImpl implements FunctionDescriptor {
         return Objects.hash(argLayouts, resLayout, firstVariadicArgumentIndex());
     }
 
-    final void checkIndex(int index) {
-        if (index < 0 || index > argLayouts.size())
-            throw new IllegalArgumentException("Index out of bounds: " + index);
-    }
-
     public static FunctionDescriptor of(MemoryLayout resLayout, List<MemoryLayout> argLayouts) {
         return new FunctionDescriptorImpl(resLayout, argLayouts);
     }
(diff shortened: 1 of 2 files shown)


src/java.base/share/classes/jdk/internal/foreign/FunctionDescriptorImpl.java
```diff
@@ -0,0 +1,238 @@
+/*
+ *  Copyright (c) 2020, 2022, Oracle and/or its affiliates. All rights reserved.
```
sundararajana, line 2: New source. No need for 2020

src/java.base/share/classes/jdk/internal/foreign/FunctionDescriptorImpl.java
```diff
@@ -0,0 +1,238 @@
...
+        public int firstVariadicArgumentIndex() {
+            return firstVariadicIndex;
+        }
+
+        @Override
+        public FunctionDescriptorImpl insertArgumentLayouts(int index, MemoryLayout... addedLayouts) {
+            checkIndex(index);
+            for (MemoryLayout layout : addedLayouts) {
+                requireNonNull(layout);
+            }
```
sundararajana, line 219: Do we need to enforce other problems (index, null) before throwing UnsupportedOperationException? Existing code unconditionally throws UnsupportedOperationException.

src/java.base/share/classes/jdk/internal/foreign/FunctionDescriptorImpl.java
```diff
@@ -0,0 +1,238 @@
...
+            throw newUnsupportedOperationException();
+        }
+
+        @Override
+        public FunctionDescriptorImpl changeReturnLayout(MemoryLayout newReturn) {
+            requireNonNull(newReturn);
```
sundararajana, line 225: Do we need to enforce other problems (index, null) before throwing UnsupportedOperationException? Existing code unconditionally throws UnsupportedOperationException.

src/java.base/share/classes/jdk/internal/foreign/FunctionDescriptorImpl.java
minborg: I think this is a philosophical question and how strictly we want to adhere to principles. Maybe this is overdoing it so I will revert this section.

src/java.base/share/classes/jdk/internal/foreign/FunctionDescriptorImpl.java
minborg: The code in this class is basically copied from the original FunctionalDescription concrete class from 2020 so I think 2020 can stay here.

sundararajana (approved):
LGTM

PR #717, opened 2022-09-07
8292037: Consider moving Linker::methodType as FunctionDescriptor::toMethodType

JornVernee:
This patch moves `Linker::methodType` to `FunctionDescriptor` as `carrierMethodType`.

I went with the name `carrierMethodType` instead of `toMethodType` (as suggested by the JBS issue) since I think it's a bit more descriptive. Please let me know your thoughts.

This issue also suggest adding support for `SequenceLayout` in the linker implementation. I think adding support in FunctionDescriptor could be a good idea, but since passing arrays by value is not supported in C, I feel like we should reject it in the linker implementations we have (which are all C as well). Of course, we could make it work any way by interpreting such cases as a pointer layout instead, but it's a case of being able to do the same thing in multiple ways (which I think can be confusing), and might actually be surprising if someone passes the wrong layout to the linker by accident, but it doesn't get "caught".

I've added support for `SequenceLayout` to `FunctionDescriptor`, but it is still rejected in the type classification logic.

---------
### Progress

### Issue
 * [JDK-8292037](https://bugs.openjdk.org/browse/JDK-8292037): Consider moving Linker::methodType as FunctionDescriptor::toMethodType

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/717/head:pull/717` \
`$ git checkout pull/717`

`$ git checkout pull/717` \
`$ git pull https://git.openjdk.org/panama-foreign pull/717/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 717`

`$ git pr show -t 717`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/717.diff">https://git.openjdk.org/panama-foreign/pull/717.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/java/lang/foreign/FunctionDescriptor.java
+++ b/src/java.base/share/classes/java/lang/foreign/FunctionDescriptor.java
@@ -25,6 +25,7 @@
 package java.lang.foreign;
 
 import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodType;
 import java.util.*;
 
 import jdk.internal.foreign.FunctionDescriptorImpl;
@@ -93,6 +94,22 @@ public sealed interface FunctionDescriptor permits FunctionDescriptorImpl {
      * @return the new function descriptor.
      */
     FunctionDescriptor dropReturnLayout();
+
+    /**
+     * Returns the method type consisting of the carrier types of the layouts in this function descriptor.
+     * <p>
+     * The carrier type of a layout is determined as follows:
+     * <ul>
+     * <li>If the layout is a {@link ValueLayout} the carrier type is determined through {@link ValueLayout#carrier()}.</li>
+     * <li>If the layout is a {@link GroupLayout} the carrier type is {@link MemorySegment}.</li>
+     * <li>If the layout is a {@link PaddingLayout}, or {@link SequenceLayout} an {@link IllegalArgumentException} is thrown.</li>
+     * </ul>
+     *
+     * @return the method type consisting of the carrier types of the layouts in this function descriptor
+     * @throws IllegalArgumentException if one or more layouts in the function descriptor are not supported
+     * (e.g. if they are sequence layouts or padding layouts).
+     */
+    MethodType carrierMethodType();
 
     /**
      * Creates a specialized variadic function descriptor, by appending given variadic layouts to this
--- a/src/java.base/share/classes/java/lang/foreign/FunctionDescriptor.java
+++ b/src/java.base/share/classes/java/lang/foreign/FunctionDescriptor.java
@@ -101,15 +101,15 @@ public sealed interface FunctionDescriptor permits FunctionDescriptorImpl {
      * The carrier type of a layout is determined as follows:
      * <ul>
      * <li>If the layout is a {@link ValueLayout} the carrier type is determined through {@link ValueLayout#carrier()}.</li>
-     * <li>If the layout is a {@link GroupLayout}, or {@link SequenceLayout} the carrier type is {@link MemorySegment}.</li>
-     * <li>If the layout is a {@link PaddingLayout} an {@link IllegalArgumentException} is thrown.</li>
+     * <li>If the layout is a {@link GroupLayout} the carrier type is {@link MemorySegment}.</li>
+     * <li>If the layout is a {@link PaddingLayout}, or {@link SequenceLayout} an {@link IllegalArgumentException} is thrown.</li>
      * </ul>
      *
      * @return the method type consisting of the carrier types of the layouts in this function descriptor
-     * @throws IllegalArgumentException if one or more layouts in the function descriptor are not supported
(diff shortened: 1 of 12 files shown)


src/java.base/share/classes/java/lang/foreign/FunctionDescriptor.java
```diff
@@ -94,6 +95,22 @@ public sealed interface FunctionDescriptor permits FunctionDescriptorImpl {
...
     FunctionDescriptor dropReturnLayout();
 
+    /**
+     * Returns the method type consisting of the carrier types of the layouts in this function descriptor.
+     * <p>
+     * The carrier type of a layout is determined as follows:
+     * <ul>
+     * <li>If the layout is a {@link ValueLayout} the carrier type is determined through {@link ValueLayout#carrier()}.</li>
+     * <li>If the layout is a {@link GroupLayout}, or {@link SequenceLayout} the carrier type is {@link MemorySegment}.</li>
+     * <li>If the layout is a {@link PaddingLayout} an {@link IllegalArgumentException} is thrown.</li>
```
mcimadamore, line 105: This comment seems to be inconsistent with the @throws clause

mcimadamore: The implementation seems to support sequence layouts, and return MemorySegment. Not super sure about that (given the linker doesn't support that).

mcimadamore (commented):
I think for now your choice is ok. Depending on whether we land with GroupLayout we might need to revisit though. Note that a sequence layout is just another way to describe a struct layout - so I don't think (as you seem to suggest) that we should pass by pointer, but we should pass by value instead (expanding the sequence layout to the corresponding struct layout).

So, if GroupLayout only covers structs and unions (as now), what we have now makes sense. If GroupLayout is enhanced to cover all composite layouts (including SequenceLayout), then I think the current position seems a tad more inconsistent, as one might like to be able to use GroupLayout for all things that I can pass by value to the linker.

src/java.base/share/classes/java/lang/foreign/FunctionDescriptor.java
```diff
@@ -94,6 +95,22 @@ public sealed interface FunctionDescriptor permits FunctionDescriptorImpl {
...
+     * </ul>
+     *
+     * @return the method type consisting of the carrier types of the layouts in this function descriptor
+     * @throws IllegalArgumentException if one or more layouts in the function descriptor are not supported
+     * (e.g. if they are sequence layouts or padding layouts).
+     */
+    MethodType carrierMethodType();
```
mcimadamore, line 112: this name seems long, and the extra wording doesn't seem to add a lot (at least IMHO).

src/java.base/share/classes/java/lang/foreign/FunctionDescriptor.java
JornVernee: In the interest of moving this forward, for now I'll keep the current mapping behavior from Linker.methodType, and just do the move to FunctionDescriptor in this patch.

mcimadamore: Note that even if we keep behavior, the javadoc `@throws` seems still inconsistent with what the method javadoc says.

src/java.base/share/classes/java/lang/foreign/FunctionDescriptor.java
```diff
@@ -94,6 +95,22 @@ public sealed interface FunctionDescriptor permits FunctionDescriptorImpl {
...
+     * The carrier type of a layout is determined as follows:
+     * <ul>
+     * <li>If the layout is a {@link ValueLayout} the carrier type is determined through {@link ValueLayout#carrier()}.</li>
+     * <li>If the layout is a {@link GroupLayout} the carrier type is {@link MemorySegment}.</li>
+     * <li>If the layout is a {@link PaddingLayout}, or {@link SequenceLayout} an {@link IllegalArgumentException} is thrown.</li>
+     * </ul>
+     *
+     * @return the method type consisting of the carrier types of the layouts in this function descriptor
+     * @throws IllegalArgumentException if one or more layouts in the function descriptor can not be mapped to carrier
+     *                                  types (e.g. if they are sequence layouts or padding layouts).
```
mcimadamore, line 110:
```suggestion
     *                                  types (e.g. if they are padding layouts).
```

JornVernee: I've removed the support for SequenceLayout now, so I think this is currently correct?

mcimadamore: Ah - of course. I didn't notice the bullet being rewritten. I think this is a good move, and one that still give us the option to change things later, if needed (it's always relatively ok to turn throwing cases into non-throwing ones).

mcimadamore (approved):

PR #719, opened 2022-09-13
Refactor compound layouts such as SequenceLayout and StructLayout

minborg:
This PR proposes harmonizations between `SequenceLayout` and `GroupLayout` as well as some improvements to the API exposing fewer details.

In this PR, there is an extra level that unifies and deduplicates the common methods in Sequence and Group Layouts. This level is visible in the sealed hierarchy. It would be possible to make it a separate interface that is not in the sealed hierarchy as an alternative.

graphviz (23)

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/719/head:pull/719` \
`$ git checkout pull/719`

`$ git checkout pull/719` \
`$ git pull https://git.openjdk.org/panama-foreign pull/719/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 719`

`$ git pr show -t 719`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/719.diff">https://git.openjdk.org/panama-foreign/pull/719.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/java/lang/foreign/CompoundLayout.java
+++ b/src/java.base/share/classes/java/lang/foreign/CompoundLayout.java
@@ -1,0 +1,46 @@
+package java.lang.foreign;
+
+import java.util.NoSuchElementException;
+import java.util.stream.Stream;
+
+/**
+ * A compound layout aggregates multiple <em>memory layout</em> elements. There are two ways in which
+ * element layouts can be combined: if element layouts can be heterogeneous, the resulting compound
+ * layout is said to be a <em>group layout</em> (see {@link GroupLayout});
+ * conversely, if all element layouts are homogeneous, the resulting compound layout is said to be a
+ * <em>sequence layout</em> (see {@link SequenceLayout}).
+ *
+ * @implSpec
+ * Implementing classes are immutable, thread-safe and <a href="{@docRoot}/java.base/java/lang/doc-files/ValueBased.html">value-based</a>.
+ *
+ * @since 20
+ */
+public sealed interface CompoundLayout extends MemoryLayout, Iterable<MemoryLayout> permits GroupLayout, SequenceLayout {
+
+    /**
+     * {@return the size of this group layout (i.e. the number of elements in this group).}
+     */
+    long elementCount();
+
+    /**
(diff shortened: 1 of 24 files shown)


src/java.base/share/classes/java/lang/foreign/CompoundLayout.java
```diff
@@ -0,0 +1,46 @@
+package java.lang.foreign;
+
+import java.util.NoSuchElementException;
+import java.util.stream.Stream;
+
+/**
```
mcimadamore, line 6: I believe there's no reason to keep both CompoundLayout and GroupLayout in the public API, since they are both umbrellas for the same sets of public types (for now at least). I think we can just move these methods down to `GroupLayout` and remove `CompoundLayout`.

It's true that a `SequenceLayout` can also be viewed as a compound layout with the elements replicated N times, but, I'm having an hard time as to why a client would e.g. treat a SequenceLayout as an Iterable. I think SequenceLayout should mostly stay as is for now. As long as the `elementCount` name matches, we have room at a later point (if we so decide) to bring `SequenceLayout` under `GroupLayout` (or under some other common supertype).

mcimadamore: I guess that, more fundamentally, even though the idea of treating sequence layout and group layouts more uniformly sounded cool (at least to me), it also occurred to me that clients tend to deal with sequence layouts in very differently from how they deal with group layouts (the way you recurse on the components is fundamentally different).

For these reasons, I'm not sure having e.g. `Iterable` on `SequenceLayout` will be useful at all. And, as for other changes, such as removing `memberLayouts` and replacing it with a iterator/stream accessor, an element accessor and a length accessor works well in some cases (e.g. enhanced for), but might work less well in others (e.g. if a client needs to pass a list of the member layouts to some other API, in which case an extra `toList` is needed). Overall, none of these things is a big deal, but, at the same time, I guess I'm no longer convinced of the benefits in terms of client code.

minborg:
We decided not to go down this route to I am closing this PR.

PR #720, opened 2022-09-14
8291873: Implement return value normalization in Java

JornVernee:
This patch moves value normalization code from the downcall stub (where it was hard coded) to Java, and also adds it for upcall arguments where it was missing.

~There is a slight change in behavior with this. The previous hard-coded conversion from a native value to Java `boolean` checked if the least significant byte was non-zero, i.e. `boolean result = value & 0xFF != 0`. While the new conversion only looks at the least significant bit, i.e. `boolean result = value & 0b1 != 0`. I think the new behavior is more correct. It means that the Linker will just do the "dumb" thing when mapping anything to `boolean`, and the native representation is really expected to be like a Java `boolean` i.e. only using the least significant bit. On the other hand I think it means that e.g. jextract will have to map the C `bool` as `byte` and then do a non-zero check to convert to `boolean`.~

---------
### Progress

### Issue
 * [JDK-8291873](https://bugs.openjdk.org/browse/JDK-8291873): Implement return value normalization in Java

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/720/head:pull/720` \
`$ git checkout pull/720`

`$ git checkout pull/720` \
`$ git pull https://git.openjdk.org/panama-foreign pull/720/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 720`

`$ git pr show -t 720`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/720.diff">https://git.openjdk.org/panama-foreign/pull/720.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/jdk/internal/foreign/abi/Binding.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/abi/Binding.java
@@ -30,6 +30,9 @@ import java.lang.foreign.MemoryLayout;
 import java.lang.foreign.MemorySegment;
 import java.lang.foreign.MemorySession;
 import java.lang.foreign.SegmentAllocator;
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodType;
 import java.util.ArrayList;
 import java.util.Deque;
 import java.util.List;
@@ -281,7 +284,8 @@ public interface Binding {
         ALLOC_BUFFER,
         BOX_ADDRESS,
         UNBOX_ADDRESS,
-        DUP
+        DUP,
+        CAST
     }
 
     Tag tag();
@@ -351,6 +355,10 @@ public interface Binding {
         return Dup.INSTANCE;
     }
 
+    static Binding cast(Class<?> fromType, Class<?> toType) {
+        return new Cast(fromType, toType);
+    }
+
 
     static Binding.Builder builder() {
         return new Binding.Builder();
@@ -362,13 +370,28 @@ public interface Binding {
     class Builder {
         private final List<Binding> bindings = new ArrayList<>();
 
+        private static boolean isSubIntType(Class<?> type) {
+            return type == boolean.class || type == byte.class || type == short.class || type == char.class;
+        }
+
         public Binding.Builder vmStore(VMStorage storage, Class<?> type) {
+            if (isSubIntType(type)) {
+                bindings.add(Binding.cast(type, int.class));
+                type = int.class;
+            }
             bindings.add(Binding.vmStore(storage, type));
             return this;
         }
 
         public Binding.Builder vmLoad(VMStorage storage, Class<?> type) {
-            bindings.add(Binding.vmLoad(storage, type));
+            Class<?> loadType = type;
+            if (isSubIntType(type)) {
+                loadType = int.class;
+            }
+            bindings.add(Binding.vmLoad(storage, loadType));
+            if (isSubIntType(type)) {
+                bindings.add(Binding.cast(int.class, type));
+            }
             return this;
         }
 
@@ -670,6 +693,41 @@ public interface Binding {
         public void interpret(Deque<Object> stack, BindingInterpreter.StoreFunc storeFunc,
                               BindingInterpreter.LoadFunc loadFunc, Context context) {
             stack.push(stack.peekLast());
+        }
+    }
+
+    /**
+     * CAST([fromType], [toType])
+     *   Pop a [fromType] from the stack, convert it to [toType], and push the resulting
+     *   value onto the stack.
+     *
+     */
+    record Cast(Class<?> fromType, Class<?> toType) implements Binding {
+
+        @Override
+        public Tag tag() {
+            return Tag.CAST;
+        }
+
+        @Override
+        public void verify(Deque<Class<?>> stack) {
+            Class<?> actualType = stack.pop();
+            SharedUtils.checkType(actualType, fromType);
+            stack.push(toType);
+        }
+
+        @Override
+        public void interpret(Deque<Object> stack, BindingInterpreter.StoreFunc storeFunc,
+                              BindingInterpreter.LoadFunc loadFunc, Context context) {
+            Object arg = stack.pop();
+            MethodHandle converter = MethodHandles.explicitCastArguments(MethodHandles.identity(toType),
+                    MethodType.methodType(toType, fromType));
+            try {
+                Object result = converter.invoke(arg);
+                stack.push(result);
+            } catch (Throwable e) {
+                throw new InternalError(e);
+            }
         }
     }
 }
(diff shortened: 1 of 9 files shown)


mcimadamore:
I find the change in semantics a bit unsettling. In the VarHandle code we use this normalization:

```
    private static boolean byteToBoolean(byte b) {
        return b != 0;
    }
```

In this email [1] @rose00 said that JDK uses either `!= 0` or `&1`, and I think the difference is that the former works with external boolean values (e.g. not in the JVM), whereas the latter can be used when you know that you are dealing with normalized boolean values. In this case, we're dealing with native function, so I believe `!= 0` should be used for normalization (in a way that's consistent with what var handle does).

[1] - https://mail.openjdk.org/pipermail/panama-dev/2021-March/012580.html

JornVernee:
Thanks for the link. John says this:

```
Thus, Java booleans
in non-Java data structures are by convention represented as
8-bit containers containing either zero (for false) or any non-zero
value (for true).
```

I think that convinces me.

Though, my first instinct was to say that `segment.get(JAVA_BOOLEAN, 0)` is reading a Java boolean, therefore the value should be 0 or 1, therefore we normalize it by masking off all the other bits. But, it seems that even Java booleans have different storage conventions :)

rose00:
If this change is refactoring the logic that is intended to be a maximally compatible JNI replacement, and the encoded boolean appears as a return value from a JNI function, then we have to use the non-zero byte test (not the low-order bit test).

For other occurrences of unconditioned bytes, when they originate outside the JVM, we could in principle make other choices, but it seems safer to stick with the pre-existing JNI convention (non-zero byte test) rather than make a new one.

The pre-existing convention is arguably correct  from a C-language point of view, because a JNI boolean (`jboolean`, specifically) is an 8-bit type.  C tests that type as a boolean by performing a zero comparison, not a bit selection.  So the JVM does what C code would do *for that particular type*.  It's important to follow this model, because JNI wrapper code (written in C) will sometimes perform boolean tests on `jboolean` variables, using that non-zero byte test.

(For other types, C might test *more than 8 bits*, but JNI doesn't ever treat other types as booleans.  There is no `jboolean32` type, for example.  So the question doesn't come up with JNI of a more general test of other types as booleans.)

Meanwhile, the convention of testing the low-order bit (and clearing other bits) is enforced for the JVM's internal storage of booleans.  It's an internal matter within the JVM heap, fully disjoint from whatever contract JNI defines about `jboolean`.

It would be nice to use the bit-test other places, since it is efficient and immune to effects from word size.  But I don't think we have any occasion to do so, unless we are intentionally inventing another kind of boolean (not JNI's `jboolean`) to use as an interchange type between Panama and native code.

If we could prove that Panama's use of `JAVA_BOOLEAN` types is fully disjoint from JNI's use of `jboolean` then it would be a new, independent boolean type.   We could change the convention for testing the boolean value as originally suggested.  There might be some confusion as users converted their JNI code to Panama code.

In the end, I think it is safer to use the non-zero test for any value (of any size) that might have been in contact with C code "in the wild", since that C code might be throwing around non-zero values freely, and intending us to see them all as true values.  (The JVM's heap-management code is tightly controlled, so is not affected by this consideration.)

If Panama ever defines a direct implicit conversion from a multi-byte type (pointer, int, long) to boolean (JAVA_BOOLEAN), then I recommend aligning that conversion semantics with those of C, which is to test the whole value (not just its low byte) against zero (the all-zero-bits value).  The "take the low byte" part of JNI's treatment of booleans is just an artifact of `jboolean` being a single byte value, plus an assumption that any ABI-carried value of a single byte will be placed in the low byte of its register container.  (This is not actually true in all ABIs, IIRC some put bytes in the high part of the register.)

src/java.base/share/classes/jdk/internal/foreign/abi/Binding.java
```diff
@@ -362,13 +370,28 @@ static Binding.Builder builder() {
     class Builder {
         private final List<Binding> bindings = new ArrayList<>();
 
+        private static boolean isSubIntType(Class<?> type) {
+            return type == boolean.class || type == byte.class || type == short.class || type == char.class;
+        }
+
         public Binding.Builder vmStore(VMStorage storage, Class<?> type) {
+            if (isSubIntType(type)) {
```
mcimadamore, line 378: So, this is interesting, in the sense that a cast binding is never emitted in isolation, but always in combination with a vm load /store binding. Does it mean that maybe creating a separate binding for this is conceptually wrong, and maybe it's the semantics of vmstore/load that should be updated? It's more of a theoretical question, in the sense that the code is obviously doing the correct thing as is.

mcimadamore (commented):
Code looks good, except for the boolean normalization, as discussed. I'm unsure about cast being its own binding, but willing to go along with it for the time being.

src/java.base/share/classes/jdk/internal/foreign/abi/Binding.java
JornVernee: I think it makes sense to have cast as a separate binding. Because we don't also do a cast, I think it's nice that it is somehow reflected in the recipe.

I mostly but the code here for the sake of convenience. The alternative would be to add this kind code in each CallArranger implementation, which seemed like unnecessary repetition.

Since this is a builder class, it made sense to me to put it here (as in the class encapsulates some construction logic). Though, maybe now that these methods do more than just a VMStore and VMLoad, they should be renamed as well to something that reflects that. (maybe `erasedVMStore` / `erasedVMLoad`).

src/java.base/share/classes/jdk/internal/foreign/abi/BindingSpecializer.java
```diff
@@ -682,8 +682,21 @@ private void emitCast(Binding.Cast cast) {
             popType(int.class);
 
             if (toType == boolean.class) {
-                emitConst(0b1);
+                // implement least significant byte non-zero test
+                Label falseLabel = new Label();
```
mcimadamore, line 686: Optional: couldn't we call `Utils.byteToBoolean`?

mcimadamore (approved):

src/java.base/share/classes/jdk/internal/foreign/abi/BindingSpecializer.java
JornVernee: Yes we can, good point! (stuck in my head with that we couldn't call internal code)

JornVernee:
I've implemented the non-zero byte test.

Thinking more on the discussion, the way it makes sense to me to think about is:
- carrier types of layouts determine classification/semantics
- different linker implementations can have different interpretations for the same carrier type
- our current linker implementations, which all target C, choose to interpret `boolean` as having the semantics of `jboolean`
- `jboolean` has non-zero-first-byte-true semantics, so that's what we implement

(On a side note: if we want other semantics in the future for other linkers, we can add an additional bit of steering info to the `Cast` operator)

mcimadamore (approved):

PR #721, opened 2022-09-16
8293510: Code style issues in the foreign API

minborg:
This PR fixes numerous code style issues. PRs like this are work intensive to review and contain subjective improvements which could be debated. I think it is better to make these changes now rather than later.

---------
### Progress

### Issue
 * [JDK-8293510](https://bugs.openjdk.org/browse/JDK-8293510): Code style issues in the foreign API

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/721/head:pull/721` \
`$ git checkout pull/721`

`$ git checkout pull/721` \
`$ git pull https://git.openjdk.org/panama-foreign pull/721/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 721`

`$ git pr show -t 721`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/721.diff">https://git.openjdk.org/panama-foreign/pull/721.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
+++ b/src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
@@ -278,8 +278,10 @@ public sealed interface MemoryLayout permits SequenceLayout, GroupLayout, Paddin
                 EnumSet.of(PathKind.SEQUENCE_ELEMENT, PathKind.SEQUENCE_RANGE), elements);
     }
 
-    private static <Z> Z computePathOp(LayoutPath path, Function<LayoutPath, Z> finalizer,
-                                       Set<PathKind> badKinds, PathElement... elements) {
+    private static <Z> Z computePathOp(LayoutPath path,
+                                       Function<LayoutPath, Z> finalizer,
+                                       Set<PathKind> badKinds,
+                                       PathElement... elements) {
         Objects.requireNonNull(elements);
         for (PathElement e : elements) {
             LayoutPath.PathElementImpl pathElem = (LayoutPath.PathElementImpl)Objects.requireNonNull(e);
--- a/src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
+++ b/src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
@@ -170,7 +170,8 @@ import jdk.internal.javac.PreviewFeature;
  * @since 19
  */
 @PreviewFeature(feature=PreviewFeature.Feature.FOREIGN)
-public sealed interface MemoryLayout permits SequenceLayout, GroupLayout, PaddingLayout, ValueLayout {
+public sealed interface MemoryLayout
+        permits SequenceLayout, GroupLayout, PaddingLayout, ValueLayout {
 
     /**
      * {@return the layout size, in bytes}
--- a/src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
+++ b/src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
@@ -173,11 +173,6 @@ import jdk.internal.javac.PreviewFeature;
 public sealed interface MemoryLayout permits SequenceLayout, GroupLayout, PaddingLayout, ValueLayout {
 
     /**
-     * {@return the layout size, in bits}
-     */
-    long bitSize();
-
-    /**
      * {@return the layout size, in bytes}
      * @throws UnsupportedOperationException if {@code bitSize()} is not a multiple of 8.
      */
@@ -198,24 +193,6 @@ public sealed interface MemoryLayout permits SequenceLayout, GroupLayout, Paddin
      * @see MemoryLayout#name()
      */
     MemoryLayout withName(String name);
-
-    /**
-     * Returns the alignment constraint associated with this layout, expressed in bits. Layout alignment defines a power
-     * of two {@code A} which is the bit-wise alignment of the layout. If {@code A <= 8} then {@code A/8} is the number of
-     * bytes that must be aligned for any pointer that correctly points to this layout. Thus:
-     *
-     * <ul>
-     * <li>{@code A=8} means unaligned (in the usual sense), which is common in packets.</li>
-     * <li>{@code A=64} means word aligned (on LP64), {@code A=32} int aligned, {@code A=16} short aligned, etc.</li>
-     * <li>{@code A=512} is the most strict alignment required by the x86/SV ABI (for AVX-512 data).</li>
-     * </ul>
-     *
-     * If no explicit alignment constraint was set on this layout (see {@link #withBitAlignment(long)}),
-     * then this method returns the <a href="#layout-align">natural alignment</a> constraint (in bits) associated with this layout.
-     *
-     * @return the layout alignment constraint, in bits.
-     */
-    long bitAlignment();
 
     /**
      * Returns the alignment constraint associated with this layout, expressed in bytes. Layout alignment defines a power
@@ -240,6 +217,24 @@ public sealed interface MemoryLayout permits SequenceLayout, GroupLayout, Paddin
     }
 
     /**
+     * Returns the alignment constraint associated with this layout, expressed in bits. Layout alignment defines a power
+     * of two {@code A} which is the bit-wise alignment of the layout. If {@code A <= 8} then {@code A/8} is the number of
+     * bytes that must be aligned for any pointer that correctly points to this layout. Thus:
+     *
+     * <ul>
+     * <li>{@code A=8} means unaligned (in the usual sense), which is common in packets.</li>
+     * <li>{@code A=64} means word aligned (on LP64), {@code A=32} int aligned, {@code A=16} short aligned, etc.</li>
+     * <li>{@code A=512} is the most strict alignment required by the x86/SV ABI (for AVX-512 data).</li>
+     * </ul>
+     *
+     * If no explicit alignment constraint was set on this layout (see {@link #withBitAlignment(long)}),
+     * then this method returns the <a href="#layout-align">natural alignment</a> constraint (in bits) associated with this layout.
+     *
+     * @return the layout alignment constraint, in bits.
+     */
+    long bitAlignment();
+
+    /**
      * Returns a memory layout of the same type with the same size and name as this layout,
      * but with the specified alignment constraints (in bits).
      *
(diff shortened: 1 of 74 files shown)


src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
```diff
@@ -571,160 +673,4 @@ static PathElement sequenceElement() {
...
-     *     <li>{@link ValueLayout.OfLong}, for {@code long.class}</li>
-     *     <li>{@link ValueLayout.OfDouble}, for {@code double.class}</li>
-     *     <li>{@link ValueLayout.OfAddress}, for {@code MemorySegment.class}</li>
-     * </ul>
-     * @param carrier the value layout carrier.
-     * @param order the value layout's byte order.
-     * @return a value layout with the given Java carrier and byte-order.
-     * @throws IllegalArgumentException if the carrier type is not supported.
-     */
-    static ValueLayout valueLayout(Class<?> carrier, ByteOrder order) {
```
mcimadamore, line 636: This method should not be removed. It is very useful when working with layouts programmatically (e.g. when a client wants to create a value layout from a class carrier). Even if it's unused in code (but it is used in one test), we should not treat this factory as useless. Also, I was only able to find this by looking at commits one by one. So, as a general comment, perhaps it would be a good idea to split the various changes.

src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
```diff
@@ -275,23 +274,10 @@ default long byteOffset(PathElement... elements) {
      * in {@code elements} is {@code null}.
      */
     default long bitOffset(PathElement... elements) {
-        return computePathOp(LayoutPath.rootPath(this), LayoutPath::offset,
-                EnumSet.of(PathKind.SEQUENCE_ELEMENT, PathKind.SEQUENCE_RANGE), elements);
-    }
-
-    private static <Z> Z computePathOp(LayoutPath path,
```
mcimadamore: I'm generally not too sure about this. Yes, if there's a lot of code it might be useful to move it out - but in this case it's a single method used in several other methods in this interface. Why is it bad to use a private static interface method for something like this?

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
```diff
@@ -1920,6 +1789,111 @@ static MemorySegment allocateNative(long byteSize, MemorySession session) {
...
+                     long dstOffset,
+                     long elementCount) {
+
+        // Check the trivial invariants here
+        Objects.requireNonNull(srcSegment);
+        Objects.requireNonNull(srcElementLayout);
+        Objects.requireNonNull(dstSegment);
+        Objects.requireNonNull(dstElementLayout);
+
+        Utils.copy(srcSegment, srcElementLayout, srcOffset, dstSegment, dstElementLayout, dstOffset, elementCount);
```
mcimadamore, line 1908: I think it would make more sense to move the `copy` method to AbstractMemorySegmentImpl - where other helper stuff for segment is defined.

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
```diff
@@ -2095,40 +2009,11 @@ static long mismatch(MemorySegment srcSegment,
...
-            if (srcImpl.get(JAVA_BYTE, srcFromOffset + i) != dstImpl.get(JAVA_BYTE, dstFromOffset + i)) {
-                return i;
-            }
-        }
-        return srcBytes != dstBytes ? bytes : -1;
+        // Check the trivial invariants here
+        Objects.requireNonNull(srcSegment);
+        Objects.requireNonNull(dstSegment);
+
+        return Utils.mismatch(srcSegment,srcFromOffset,srcToOffset, dstSegment, dstFromOffset, dstToOffset);
```
mcimadamore, line 2030: Same here - mismatch should live in AbstractMemorySegment

src/java.base/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
```diff
@@ -74,7 +74,9 @@
     final MemorySession session;
 
     @ForceInline
-    AbstractMemorySegmentImpl(long length, boolean readOnly, MemorySession session) {
+    AbstractMemorySegmentImpl(long length,
```
mcimadamore, line 77: Here I'm a bit puzzled. Looking at other JDK code, I don't see this rule being applied consistently. I think I'm ok with breaking up long parameter lists (an example of that below) - but doing that for all method decls seems a bit too much IMHO.

src/java.base/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
```diff
@@ -133,10 +135,13 @@ public AbstractMemorySegmentImpl asSlice(long offset) {
     /**
      * Mismatch over long lengths.
      */
-    public static long vectorizedMismatchLargeForBytes(MemorySessionImpl aSession, MemorySessionImpl bSession,
-                                                        Object a, long aOffset,
-                                                        Object b, long bOffset,
-                                                        long length) {
+    public static long vectorizedMismatchLargeForBytes(MemorySessionImpl aSession,
```
mcimadamore, line 138: So, this is a decision whether to break up one per line, or break up so that we never exceed 100 chars per line. I don't have any strong opinion here, but I note that the JDK code is not consistent about this at all.

test/jdk/java/lang/Thread/jni/AttachCurrentThread/ImplicitAttach.java
```diff
@@ -62,7 +62,7 @@ public static void main(String[] args) throws Throwable {
 
         // void start_threads(int count, void *(*f)(void *))
         SymbolLookup symbolLookup = SymbolLookup.loaderLookup();
-        MemorySegment symbol = symbolLookup.lookup("start_threads").orElseThrow();
+        MemorySegment symbol = symbolLookup.find("start_threads").orElseThrow();
```
mcimadamore, line 65: This is dealt with in another PR: https://github.com/openjdk/panama-foreign/pull/722

src/java.base/share/classes/java/lang/foreign/SegmentAllocator.java
```diff
@@ -325,24 +325,24 @@ default MemorySegment allocateArray(MemoryLayout elementLayout, long count) {
...
-     * @param bytesAlignment the alignment (in bytes) of the block of memory to be allocated.
+     * @param byteSize the size (in bytes) of the block of memory to be allocated.
+     * @param byteAlignment the alignment (in bytes) of the block of memory to be allocated.
      * @return a segment for the newly allocated memory block.
-     * @throws IllegalArgumentException if {@code bytesSize < 0}, {@code alignmentBytes <= 0},
+     * @throws IllegalArgumentException if {@code byteSize < 0}, {@code byteAlignment <= 0},
      * or if {@code alignmentBytes} is not a power of 2.
      */
-    MemorySegment allocate(long bytesSize, long bytesAlignment);
+    MemorySegment allocate(long byteSize, long byteAlignment);
```
mcimadamore: Good stuff!

src/java.base/share/classes/jdk/internal/foreign/abi/VMStorage.java
```diff
@@ -26,32 +26,22 @@
...
+/**
+ * @param type              the type of storage. e.g. stack, or which register type (GP, FP, vector)
+ * @param segmentMaskOrSize The (on stack) size in bytes when type = stack, a register mask, otherwise
+ *                          the register mask indicates which segments of a register are used.
+ * @param indexOrOffset     The index is either a register number within a type, or
+ *                          a stack offset in bytes if type = stack.
+ *                          (a particular platform might add a bias to this in generate code)
+ * @param debugName         the debug name to use
+ */
+public record VMStorage(byte type,
```
mcimadamore, line 38: This is well spotted - please separate it out

src/java.base/share/classes/jdk/internal/foreign/abi/x64/sysv/CallArranger.java
```diff
@@ -78,17 +78,10 @@ public final class CallArranger {
...
-        public final boolean isInMemoryReturn;
-        public final int nVectorArgs;
-
-        Bindings(CallingSequence callingSequence, boolean isInMemoryReturn, int nVectorArgs) {
-            this.callingSequence = callingSequence;
-            this.isInMemoryReturn = isInMemoryReturn;
-            this.nVectorArgs = nVectorArgs;
-        }
-    }
+    public record Bindings(
```
mcimadamore: Also good

src/java.base/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
```diff
@@ -240,21 +240,25 @@ public final long segmentOffset(MemorySegment other) {
 
     @Override
     public void load() {
-        throw new UnsupportedOperationException("Not a mapped segment");
+        throw newUnsupportedOperationExceptionNotAMappedSegment();
```
mcimadamore, line 401: Nice

mcimadamore (commented):
As said in one of the comments, we should try to separate out the various changes. I did a pass on the code, by filtering out one commit at a time, which was useful. What I liked:

* general better consistency on use of `final`
* parameter naming consistency
* moving some of the bigger methods `copy` `mismatch` out (although I think the destination should be different)
* use of `record` in places where we didn't use them
* breaking up long method call chains
* fixing of imports

What I find more questionable:

* alphabetically reordering all members
* breaking apart method parameters no matter what

There is also one API change (removal of MemoryLayout::valueLayout) which is a red herring in a code style PR.

src/java.base/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
JornVernee: Agree, we should follow the style used in the rest of the jdk code, at least in java.base.

minborg: The principle used in this PR is the same as in `java.util.stream.Stream` where methods with 0 <= x <= 2 parameters are on the same line and methods with > 2 parameters are on separate vertical lines.

JornVernee: But that does not make it the norm. The majority of code in java.base does not do this. Even in `java.util.stream.Stream` it is not done for every method (e.g. `iterate` puts 3 parameters on the same line).

minborg: It is true that this does not make it the norm. It is just an explanation of the principle used in this PR.

minborg:
This PR will be replaced by several others.

PR #725, opened 2022-09-19
Change classes into records and fix finality

minborg:
This PR converts some of the regular classes into records. It also fixes some finality issues.

Note: Some methods with a very long method parameter list have their parameter list broken up into separate lines.

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/725/head:pull/725` \
`$ git checkout pull/725`

`$ git checkout pull/725` \
`$ git pull https://git.openjdk.org/panama-foreign pull/725/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 725`

`$ git pr show -t 725`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/725.diff">https://git.openjdk.org/panama-foreign/pull/725.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/jdk/internal/foreign/abi/ABIDescriptor.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/abi/ABIDescriptor.java
@@ -43,9 +43,14 @@ public class ABIDescriptor {
     final VMStorage targetAddrStorage;
     final VMStorage retBufAddrStorage;
 
-    public ABIDescriptor(Architecture arch, VMStorage[][] inputStorage, VMStorage[][] outputStorage,
-                         VMStorage[][] volatileStorage, int stackAlignment, int shadowSpace,
-                         VMStorage targetAddrStorage, VMStorage retBufAddrStorage) {
+    public ABIDescriptor(Architecture arch,
+                         VMStorage[][] inputStorage,
+                         VMStorage[][] outputStorage,
+                         VMStorage[][] volatileStorage,
+                         int stackAlignment,
+                         int shadowSpace,
+                         VMStorage targetAddrStorage,
+                         VMStorage retBufAddrStorage) {
         this.arch = arch;
         this.inputStorage = inputStorage;
         this.outputStorage = outputStorage;
--- a/src/java.base/share/classes/jdk/internal/foreign/abi/VMStorage.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/abi/VMStorage.java
@@ -26,32 +26,22 @@ package jdk.internal.foreign.abi;
 
 import java.util.Objects;
 
-public class VMStorage {
-    /**
-     * Type of storage. e.g. stack, or which register type (GP, FP, vector)
-     */
-    private final byte type;
+/**
+ * @param type              the type of storage. e.g. stack, or which register type (GP, FP, vector)
+ * @param segmentMaskOrSize The (on stack) size in bytes when type = stack, a register mask, otherwise
+ *                          the register mask indicates which segments of a register are used.
+ * @param indexOrOffset     The index is either a register number within a type, or
+ *                          a stack offset in bytes if type = stack.
+ *                          (a particular platform might add a bias to this in generate code)
+ * @param debugName         the debug name to use
+ */
+public record VMStorage(byte type,
+                        short segmentMaskOrSize,
+                        int indexOrOffset,
+                        String debugName) {
 
-    /**
-     * The (on stack) size in bytes when type = stack, a register mask otherwise.
-     * The register mask indicates which segments of a register are used.
-     */
-    private final short segmentMaskOrSize;
-
-    /**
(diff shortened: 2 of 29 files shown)


src/java.base/share/classes/jdk/internal/foreign/abi/VMStorage.java
```diff
@@ -26,32 +26,22 @@
...
+/**
+ * @param type              the type of storage. e.g. stack, or which register type (GP, FP, vector)
+ * @param segmentMaskOrSize The (on stack) size in bytes when type = stack, a register mask, otherwise
+ *                          the register mask indicates which segments of a register are used.
+ * @param indexOrOffset     The index is either a register number within a type, or
+ *                          a stack offset in bytes if type = stack.
+ *                          (a particular platform might add a bias to this in generate code)
+ * @param debugName         the debug name to use
+ */
+public record VMStorage(byte type,
```
JornVernee, line 38: I think this should stay a `class`. VMStorage should not be constructed directly, but one of the factory methods should be used. But, the change to `record` forces the constructor to be public.

src/java.base/share/classes/jdk/internal/foreign/abi/ABIDescriptor.java
```diff
@@ -43,9 +43,14 @@ public class ABIDescriptor {
...
-                         VMStorage[][] volatileStorage, int stackAlignment, int shadowSpace,
-                         VMStorage targetAddrStorage, VMStorage retBufAddrStorage) {
+    public ABIDescriptor(Architecture arch,
+                         VMStorage[][] inputStorage,
+                         VMStorage[][] outputStorage,
+                         VMStorage[][] volatileStorage,
+                         int stackAlignment,
+                         int shadowSpace,
+                         VMStorage targetAddrStorage,
+                         VMStorage retBufAddrStorage) {
```
JornVernee, line 53: Please move the line splitting changes to a separate PR. I think more discussion is needed on that. But we can move forward with some of the changes in this PR I think.

JornVernee:
I'd also like to hear your motivation for doing the finality changes. This is not an obvious "fix" to me, but seems like a rather superfluous change (especially for internal classes).

src/java.base/share/classes/jdk/internal/foreign/abi/ABIDescriptor.java
minborg: I will hold back the other splittings. Can we accept the ones here to move forward?

minborg:
Some of the internal classes that are not designed to be subclassed are marked as `final` to signal this fact. Admittedly, it is less important to keep internal classes non-final compared to classes in the public API. However, I think these changes merit the effort as the internal classes are likely to live for decades.

src/java.base/share/classes/jdk/internal/foreign/abi/VMStorage.java
mcimadamore: Doesn't the same reasoning hold for Binding? E.g. there's a builder there, no?

JornVernee: For Binding there's a builder, but it just calls the constructor of the respective Binding class. i.e. there's no additional instantiation logic encapsulated that can lead to a malformed object. Although, it is possible to lead to malformed recipes.

VMStorage does encapsulate some instantiation logic, namely the debug name for stack storage. Though, to be fair, there used to be more of that in an earlier version where I also checked that the wrong `type` was not being used. So, maybe it's time to revisit 🤔 

Thinking about it now, I think the real encapsulation happens in the architecture classes (e.g. https://github.com/openjdk/panama-foreign/blob/cb272592d5043aaa2d9f013eac37620de50418ef/src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/AArch64Architecture.java#L139) where the type of stack storage is actually known.

Maybe it would be better to make `VMStorage` just a dumb record, and drop the factory methods altogether.

src/java.base/share/classes/jdk/internal/foreign/abi/ABIDescriptor.java
JornVernee: No. We don't have consensus on whether these changes are good or bad, so we should not just move forward with them. We're not in a hurry.

src/java.base/share/classes/jdk/internal/foreign/abi/VMStorage.java
mcimadamore: although this seems to deserve its own targeted PR

minborg: Yes. Let's do this in a separate PR.

minborg:
This PR is closed in favour of https://github.com/openjdk/panama-foreign/pull/731

PR #728, opened 2022-09-19
8284300: Explicitly reject layouts that don't have the correct alignment during linking

JornVernee:
I realized that rejecting layouts that don't have natural alignment also prevents us from supporting packed structs in the future. I tested out the packed struct support and it works on Windows pretty much out of the box. It doesn't work on SysV, some changes to CallArranger are needed. So, there are some options:
1. reject layouts that don't have natural alignment on all platforms (Windows won't be able to use packed structs)
2. reject layouts that don't have natural alignment on platforms besides Windows (yay Windows users)
3. align the status quos: Windows and AArch64 re-compute the alignment. It's not used on Windows, so it can be dropped there. On AArch64 we could "trust" the alignment that is attached to the layout given by the user, same as SysV, but it would fail for packed structs. With the intention of fixing the failure in the future.

I'm leaning most towards 1 at the moment, and that's what this patch implements. We can change that approach when we add reliable support for packed structs across platforms.

---------
### Progress

### Issue
 * [JDK-8284300](https://bugs.openjdk.org/browse/JDK-8284300): Explicitly reject layouts that don't have the correct alignment during linking

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/728/head:pull/728` \
`$ git checkout pull/728`

`$ git checkout pull/728` \
`$ git pull https://git.openjdk.org/panama-foreign pull/728/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 728`

`$ git pr show -t 728`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/728.diff">https://git.openjdk.org/panama-foreign/pull/728.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/jdk/internal/foreign/abi/AbstractLinker.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/abi/AbstractLinker.java
@@ -30,10 +30,12 @@ import jdk.internal.foreign.abi.aarch64.macos.MacOsAArch64Linker;
 import jdk.internal.foreign.abi.x64.sysv.SysVx64Linker;
 import jdk.internal.foreign.abi.x64.windows.Windowsx64Linker;
 
+import java.lang.foreign.MemoryLayout;
 import java.lang.foreign.MemorySession;
 import java.lang.foreign.FunctionDescriptor;
 import java.lang.foreign.Linker;
 import java.lang.foreign.MemorySegment;
+import java.lang.foreign.ValueLayout;
 import java.lang.invoke.MethodHandle;
 import java.lang.invoke.MethodType;
 import java.util.Objects;
--- a/src/java.base/share/classes/jdk/internal/foreign/abi/AbstractLinker.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/abi/AbstractLinker.java
@@ -29,11 +29,17 @@ import jdk.internal.foreign.abi.aarch64.linux.LinuxAArch64Linker;
 import jdk.internal.foreign.abi.aarch64.macos.MacOsAArch64Linker;
 import jdk.internal.foreign.abi.x64.sysv.SysVx64Linker;
 import jdk.internal.foreign.abi.x64.windows.Windowsx64Linker;
+import jdk.internal.foreign.layout.AbstractLayout;
 
+import java.lang.foreign.GroupLayout;
+import java.lang.foreign.MemoryLayout;
 import java.lang.foreign.MemorySession;
 import java.lang.foreign.FunctionDescriptor;
 import java.lang.foreign.Linker;
 import java.lang.foreign.MemorySegment;
+import java.lang.foreign.PaddingLayout;
+import java.lang.foreign.SequenceLayout;
+import java.lang.foreign.ValueLayout;
 import java.lang.invoke.MethodHandle;
 import java.lang.invoke.MethodType;
 import java.util.Objects;
@@ -46,6 +52,7 @@ public abstract sealed class AbstractLinker implements Linker permits LinuxAArch
     @Override
     public MethodHandle downcallHandle(FunctionDescriptor function) {
         Objects.requireNonNull(function);
+        checkHasNaturalAlignment(function);
 
         return DOWNCALL_CACHE.get(function, fd -> {
             MethodType type = fd.toMethodType();
@@ -61,6 +68,7 @@ public abstract sealed class AbstractLinker implements Linker permits LinuxAArch
         Objects.requireNonNull(scope);
         Objects.requireNonNull(target);
         Objects.requireNonNull(function);
+        checkHasNaturalAlignment(function);
         SharedUtils.checkExceptions(target);
 
         MethodType type = function.toMethodType();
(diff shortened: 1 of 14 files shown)


src/java.base/share/classes/jdk/internal/foreign/abi/SharedUtils.java
```diff
@@ -94,14 +94,6 @@ private SharedUtils() {
         throw new IllegalStateException("Cannot get here");
     };
 
-    /**
-     * Align the specified type from a given address
-     * @return The address the data should be at based on alignment requirement
-     */
-    public static long align(MemoryLayout t, boolean isVar, long addr) {
```
JornVernee, line 101: Notice this code was dead, so I removed it.

mcimadamore:
I agree with choice #1. Unless some well defined strategy is defined in some ABI, it doesn't seem like a slam dunk.

src/java.base/share/classes/jdk/internal/foreign/abi/x64/sysv/TypeClass.java
```diff
@@ -174,6 +176,7 @@ private static List<ArgumentClassImpl> classifyStructType(GroupLayout type) {
     }
 
     static TypeClass classifyLayout(MemoryLayout type) {
+        SharedUtils.checkHasNaturalAlignment(type);
```
mcimadamore, line 179: Isn't this check too deep in the call chain? E.g. isn't this some validation we could do immediately in the linker, by looking at the function descriptor being passed? Does it need to live in platform dependent code?

src/java.base/share/classes/jdk/internal/foreign/abi/SharedUtils.java
```diff
@@ -375,6 +367,21 @@ public static NoSuchElementException newVaListNSEE(MemoryLayout layout) {
...
+    public static void checkHasNaturalAlignment(MemoryLayout layout) {
+        long naturalByteAlignmnet = naturalByteAlignment(layout);
+        if (naturalByteAlignmnet != layout.byteAlignment()) {
+            throw new IllegalArgumentException("Layout bit alignment must be natural alignment: "
+                    + layout.byteAlignment() + " != " + naturalByteAlignmnet);
+        }
+    }
+
+    private static long naturalByteAlignment(MemoryLayout layout) {
+        return alignment(layout, true);
```
mcimadamore, line 382: Can you please take a look at that `alignment` routine? I'm not 100% sure it's correct (e.g. for arrays it returns 16 - is that true on all platforms?). Also, note that `AbstractLayout` has an `hasNaturalAlignment` predicate, which seems close to what you need.

src/java.base/share/classes/jdk/internal/foreign/abi/x64/sysv/TypeClass.java
JornVernee: Good point, it should be moved up. I went through an iteration where I only checked ValueLayouts, for which it made more sense to put here.

src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/CallArranger.java
```diff
@@ -264,24 +263,27 @@ protected BindingCalculator(boolean forArguments) {
...
             long offset = 0;
+            long align = layout.byteAlignment();
             while (offset < layout.byteSize()) {
-                long copy = Math.min(layout.byteSize() - offset, STACK_SLOT_SIZE);
-                VMStorage storage =
-                    storageCalculator.stackAlloc(copy, STACK_SLOT_SIZE);
-                if (offset + STACK_SLOT_SIZE < layout.byteSize()) {
+                long copy = Math.min(layout.byteSize() - offset, COPY_CHUNK_SIZE);
+                VMStorage storage = storageCalculator.stackAlloc(copy, align);
+                align = 1; // only first copy needs aligning
```
mcimadamore, line 279: so, this was an issue in the old code?

JornVernee: Sorry, forgot to explain this.

In the old code we always passed STACK_SLOT_SIZE. This works out in the current implementation since we only allow values to be copied to the stack at stack slot aligned offset. Though, technically, we might need a more precise alignment in the case of Mac/AArch64. For instance, if a struct has 4 byte alignment, we might need to spill it at a 4-byte aligned stack offset.

I noticed this when looking through the code and decided to fix this. But, I'm think now that it kinda doesn't belong in this patch. I can remove it. All we need is the change to drop `SharedUtils.alignment` for `layout.byteAlignment` above.

JornVernee: I think this might have slipped over from when I was working on the Mac/AArch64 stack spill issue, and I didn't think to remove it :/

src/java.base/share/classes/jdk/internal/foreign/abi/SharedUtils.java
JornVernee: Hmm yeah, IIRC the array alignment is only 16 if it is a global variable, so I think we can drop that here in this case. I'll use `hasNaturalAlignment` instead. Good suggestion!

JornVernee:
I've addressed the review comments. I've reverted the changes to the aarch64 CallArranger, since they really don't do anything at the moment.

I also switched to using `AbstractLayout::hasNaturalAlignment` called from AbstractLinker. For aggregates we also have to recursively check the member layouts.

The old alignment computing code was no longer used, so I removed.

I also had a look at the layouts clang gives us for the different platforms, and I don't see any 16 byte alignments for arrays larger than 16 bytes or incomplete arrays, so I think we don't need to take that into account.

src/java.base/share/classes/jdk/internal/foreign/abi/AbstractLinker.java
```diff
@@ -77,4 +85,29 @@ protected abstract MemorySegment arrangeUpcall(MethodHandle target, MethodType t
...
+    private static void checkHasNaturalAlignment(FunctionDescriptor descriptor) {
+        descriptor.returnLayout().ifPresent(AbstractLinker::checkHasNaturalAlignmentRecursive);
+        descriptor.argumentLayouts().forEach(AbstractLinker::checkHasNaturalAlignmentRecursive);
+    }
+
+    private static void checkHasNaturalAlignmentRecursive(MemoryLayout layout) {
+        checkHasNaturalAlignment(layout);
+        if (layout instanceof GroupLayout gl) {
+            for (MemoryLayout member : gl.memberLayouts()) {
+                checkHasNaturalAlignment(member);
```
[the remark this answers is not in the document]
JornVernee: Ah, good catch

JornVernee:
Should be good now. I've amended the test as well to check the recursive case. Sorry for the messy patch.

mcimadamore (approved):
Looks good!

PR #729, opened 2022-09-20
8293367: Enable native access for modules not in the boot layer

sundararajana:
Adding a new API ModuleLayer.Controller.enableNativeAccess(Module).
Reworked thread safety of enable native access flag reading and setting.

---------
### Progress

### Issue
 * [JDK-8293367](https://bugs.openjdk.org/browse/JDK-8293367): Enable native access for modules not in the boot layer

### Reviewers
 * [Maurizio Cimadamore](https://openjdk.org/census#mcimadamore) (@mcimadamore - Committer) ⚠️ Review applies to [c98e661c](https://git.openjdk.org/panama-foreign/pull/729/files/c98e661c4e1e8391490d20b4ce12c6faacbacbd5)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/729/head:pull/729` \
`$ git checkout pull/729`

`$ git checkout pull/729` \
`$ git pull https://git.openjdk.org/panama-foreign pull/729/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 729`

`$ git pr show -t 729`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/729.diff">https://git.openjdk.org/panama-foreign/pull/729.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/java/lang/Module.java
+++ b/src/java.base/share/classes/java/lang/Module.java
@@ -269,9 +269,8 @@ public final class Module implements AnnotatedElement {
      * @return {@code true} if this module can access <em>restricted</em> methods.
      */
     @PreviewFeature(feature=PreviewFeature.Feature.FOREIGN)
-    public boolean isNativeAccessEnabled() {
-        return enableNativeAccess ||
-                (!isNamed() && ALL_UNNAMED_MODULE.enableNativeAccess);
+    public synchronized boolean isNativeAccessEnabled() {
+        return implIsEnableNativeAccess();
     }
 
     /**
--- a/src/java.base/share/classes/java/lang/Module.java
+++ b/src/java.base/share/classes/java/lang/Module.java
@@ -52,6 +52,7 @@ import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import jdk.internal.javac.PreviewFeature;
 import jdk.internal.loader.BuiltinClassLoader;
 import jdk.internal.loader.BootLoader;
 import jdk.internal.loader.ClassLoaders;
@@ -259,6 +260,18 @@ public final class Module implements AnnotatedElement {
     Module implAddEnableNativeAccess() {
         enableNativeAccess = true;
         return this;
+    }
+
+    /**
+     * Returns {@code true} if this module can access
+     * <a href="foreign/package-summary.html#restricted"><em>restricted</em></a> methods.
+     *
+     * @return {@code true} if this module can access <em>restricted</em> methods.
+     */
+    @PreviewFeature(feature=PreviewFeature.Feature.FOREIGN)
+    public boolean isNativeAccessEnabled() {
+        return enableNativeAccess ||
+                (!isNamed() && ALL_UNNAMED_MODULE.enableNativeAccess);
     }
 
     /**
--- a/src/java.base/share/classes/java/lang/Module.java
+++ b/src/java.base/share/classes/java/lang/Module.java
@@ -57,6 +57,7 @@ import jdk.internal.loader.BuiltinClassLoader;
 import jdk.internal.loader.BootLoader;
 import jdk.internal.loader.ClassLoaders;
 import jdk.internal.misc.CDS;
+import jdk.internal.module.ModuleBootstrap;
 import jdk.internal.module.ModuleLoaderMap;
 import jdk.internal.module.ServicesCatalog;
 import jdk.internal.module.Resources;
@@ -269,24 +270,60 @@ public final class Module implements AnnotatedElement {
      * @return {@code true} if this module can access <em>restricted</em> methods.
      */
     @PreviewFeature(feature=PreviewFeature.Feature.FOREIGN)
-    public synchronized boolean isNativeAccessEnabled() {
-        return implIsEnableNativeAccess();
+    public boolean isNativeAccessEnabled() {
+        boolean enabled = isNamed()? enableNativeAccess :
+            ALL_UNNAMED_MODULE.enableNativeAccess;
+        if (enabled) {
+            return enabled;
+        }
+        // slow path
+        if (isNamed()) {
+            synchronized(this) {
+                return enableNativeAccess;
+            }
+        } else {
+            synchronized(ALL_UNNAMED_MODULE) {
+                return ALL_UNNAMED_MODULE.enableNativeAccess;
+            }
+        }
     }
+
+    // This is invoked from Reflection.ensureNativeAccess
+    void ensureNativeAccess(Class<?> owner, String methodName) {
+        boolean isNativeAccessEnabled = isNativeAccessEnabled();
+        if (!isNativeAccessEnabled) {
+            if (ModuleBootstrap.hasEnableNativeAccessFlag()) {
+                throw new IllegalCallerException("Illegal native access from: " + this);
+            } else {
+                synchronized (this) {
+                    // warn and set flag, so that only one warning is reported per module
+                    String cls = owner.getName();
+                    String mtd = cls + "::" + methodName;
+                    String mod = isNamed() ? "module " + getName() : "the unnamed module";
+                    String modflag = isNamed() ? getName() : "ALL-UNNAMED";
+                    System.err.printf("""
+                        WARNING: A restricted method in %s has been called
+                        WARNING: %s has been called by %s
+                        WARNING: Use --enable-native-access=%s to avoid a warning for this module
+                        %n""", cls, mtd, mod, modflag);
+                    if (isNamed()) {
+                        implAddEnableNativeAccess();
+                    } else {
+                        implAddEnableNativeAccessAllUnnamed();
+                    }
+                }
+            }
+        }
+    }
+
 
     /**
      * Update all unnamed modules to allow access to restricted methods.
      */
     static void implAddEnableNativeAccessAllUnnamed() {
-        ALL_UNNAMED_MODULE.enableNativeAccess = true;
-    }
-
-    /**
-     * Returns true if module m can access restricted methods.
-     */
-    boolean implIsEnableNativeAccess() {
-        return isNamed() ?
-                enableNativeAccess :
-                ALL_UNNAMED_MODULE.enableNativeAccess;
+        synchronized (ALL_UNNAMED_MODULE) {
+            ALL_UNNAMED_MODULE.enableNativeAccess = true;
+        }
     }
 
     // --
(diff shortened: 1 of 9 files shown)


src/java.base/share/classes/java/lang/Module.java
```diff
@@ -256,25 +258,73 @@ public ModuleLayer getLayer() {
...
+        Module target = enableNativeAccessHolder();
+        synchronized(target) {
+            return target.enableNativeAccess;
+        }
+    }
+
+    // Returns the Module object holds the enableNativeAccess
+    // flag for this module.
+    private Module enableNativeAccessHolder() {
+        return isNamed()? this : ALL_UNNAMED_MODULE;
```
mcimadamore, line 283:
```suggestion
        return isNamed() ? this : ALL_UNNAMED_MODULE;
```

src/java.base/share/classes/java/lang/Module.java
```diff
@@ -256,25 +258,73 @@ public ModuleLayer getLayer() {
...
+        Module target = enableNativeAccessHolder();
+        // racy read of the enable native access flag
+        boolean isNativeAccessEnabled = target.enableNativeAccess;
+        if (!isNativeAccessEnabled) {
+            synchronized(target) {
+                // safe read of the enableNativeAccess of the target module
+                isNativeAccessEnabled = target.enableNativeAccess;
+                // check again with the safely read flag
+                if (isNativeAccessEnabled) {
+                   return;
```
mcimadamore, line 298: I'd suggest adding a comment like "another thread beat us to it - nothing to do"

src/java.base/share/classes/java/lang/ModuleLayer.java
```diff
@@ -297,6 +299,32 @@ public Controller addOpens(Module source, String pn, Module target) {
...
+         * Restricted methods are unsafe, and, if used incorrectly, their use might crash
+         * the JVM or, worse, silently result in memory corruption. Thus, clients should refrain
+         * from depending on restricted methods, and use safe and supported functionalities,
+         * where possible.
+         *
+         * @param target the module that's given native access
+         *
+         * @return This controller
+         *
+         * @throws IllegalCallerException If the caller module doesn't have native access.
```
mcimadamore, line 317: This is a restricted method, and needs the verbiage javadoc around restricted method. See e.g. `Linker::nativeLinker`. It should also probably specify what happens when called with no module in the call stack (possible if call comes from JNI). See `SymbolLookup::libraryLookup` for some hints. (note - the code already does the correct thing in Reflection.java)

src/java.base/share/classes/jdk/internal/access/JavaLangAccess.java
```diff
@@ -273,9 +273,10 @@ public interface JavaLangAccess {
     void addEnableNativeAccessAllUnnamed();
 
     /**
-     * Returns true if module m can access restricted methods.
+     * Ensure that the given module has native access. If not, warn
```
mcimadamore, line 276: We have a section on restricted methods in the foreign package javadoc. Perhaps we could link to that?

src/java.base/share/classes/java/lang/Module.java
```diff
@@ -256,25 +258,74 @@ public ModuleLayer getLayer() {
...
+     * @return {@code true} if this module can access <em>restricted</em> methods.
      */
-    static void implAddEnableNativeAccessAllUnnamed() {
-        ALL_UNNAMED_MODULE.enableNativeAccess = true;
+    @PreviewFeature(feature=PreviewFeature.Feature.FOREIGN)
+    public boolean isNativeAccessEnabled() {
+        Module target = enableNativeAccessHolder();
+        synchronized(target) {
+            return target.enableNativeAccess;
+        }
```
AlanBateman, line 277: enableNativeAccess is `@Stable` and I'm trying to understand the access. ensureNativeAccess will only synchronize when false, isNativeAccessEnabled synchronizes always. Is that intentional?

src/java.base/share/classes/java/lang/ModuleLayer.java
```diff
@@ -297,6 +299,35 @@ public Controller addOpens(Module source, String pn, Module target) {
             source.implAddOpens(pn, target);
             return this;
         }
+
+        /***
+         * Enables native access for a module in the layer if the caller's module
+         * already has native access.
```
AlanBateman, line 305: The API looks okay, ModuleLayer.Controller is the right place for this. I think you can drop "already" from the javadoc.

src/java.base/share/classes/java/lang/ModuleLayer.java
```diff
@@ -297,6 +299,35 @@ public Controller addOpens(Module source, String pn, Module target) {
...
+         *
+         * <p> This method is <a href="foreign/package-summary.html#restricted"><em>restricted</em></a>.
+         * Restricted methods are unsafe, and, if used incorrectly, their use might crash
+         * the JVM or, worse, silently result in memory corruption. Thus, clients should refrain
+         * from depending on restricted methods, and use safe and supported functionalities,
+         * where possible.
+         *
+         * @param target the module that's given native access
```
AlanBateman, line 313: The other methods use the phrase "Updates module ..." so I think target should probably be described here as "The module to update".

src/java.base/share/classes/java/lang/Module.java
```diff
@@ -256,25 +258,74 @@ public ModuleLayer getLayer() {
...
+    }
+
+    // Returns the Module object holds the enableNativeAccess
```
AlanBateman, line 280: There's a typo, should be "that holds".  Also the method name is a bit misleading as it doesn't enable anything.  The isNativeAccessEnabled and enableNativeAccess are deeply coupled so the method isn't strictly needed. If you keep it then something like moduleForNativeAccess() in the usage by both methods might be clearer.

src/java.base/share/classes/java/lang/Module.java
mcimadamore: That is intentional. `ensureNativeAccess` can be called in hot paths, so it's crucial we only synchronized when needed (so we do an optimistic plain read and lock if we see `false`). The new API predicate on `Module` is a public API, so we do not have the same performance constraints there, and we can make the implementation more intuitive.

src/java.base/share/classes/java/lang/Module.java
```diff
@@ -256,25 +258,74 @@ public ModuleLayer getLayer() {
...
+     * <a href="foreign/package-summary.html#restricted"><em>restricted</em></a> methods.
+     *
+     * @return {@code true} if this module can access <em>restricted</em> methods.
      */
-    static void implAddEnableNativeAccessAllUnnamed() {
-        ALL_UNNAMED_MODULE.enableNativeAccess = true;
+    @PreviewFeature(feature=PreviewFeature.Feature.FOREIGN)
+    public boolean isNativeAccessEnabled() {
+        Module target = enableNativeAccessHolder();
+        synchronized(target) {
```
JornVernee, line 275: I'm not sure how much of a problem this is, but I guess since the module object is publicly accessible, it's theoretically possible that some user code is also locking on it.

For the record; alternative could be to add a private field to module with a dedicated lock Object for enabling native access. Though, I don't think that's needed at the moment.

mcimadamore: That's a thought that occurred to me as well. E.g. declare the variable w/o `volatile`, but then use either Unsafe, or VarHandle to access the variable in `volatile` mode when required. That said, note that this will not get rid of all the synchronization needs: we would still need to synchronize on a module in the native access check to make sure that we issue at most one warning. But perhaps we can have a private lock just for that.

JornVernee: I think it would be possible to emit only one warning by using a compare and exchange.

```
boolean shouldEnable = !ModuleBootstrap.hasEnableNativeAccessFlag();
isNativeAccessEnabled = Unsafe.compareAndExchangeBoolean(target, OFFSET, false, shouldEnable);
// check again with the safely read flag
if (isNativeAccessEnabled) {
    // another thread beat us to it - nothing to do
    return;
} else if (!shouldEnable) {
    throw new IllegalCallerException("Illegal native access from: " + this);
} else {
    // emit watning
}
```

JornVernee: To be clear. The current implementation looks correct to me as well. So I'm fine with going ahead as is.

src/java.base/share/classes/java/lang/ModuleLayer.java
```diff
@@ -297,6 +299,36 @@ public Controller addOpens(Module source, String pn, Module target) {
...
+         * @param  target
+         *         The module to update
+         *
+         * @return This controller
+         *
+         * @throws IllegalCallerException if access to this method occurs from a module {@code M}
+         *         and the command line option {@code --enable-native-access} is specified, but
+         *         does not mention the module name {@code M}, or {@code ALL-UNNAMED} in case
+         *         {@code M} is an unnamed module.
+         */
```
AlanBateman, line 322: The implementation changes are okay but I think the javadoc, particularly the description for IllegalCallerException, will need a bit of word smithing as we can't use informal language like "does not mention the module name" or talk about command line options or operands that the RI provides. The method description links to the "Restricted methods" section in the package description and maybe lend on that rather than trying to repeat in the exception description. It may be that you can just say that the caller is in a module that does not have native access enabled. That will work for module layers because the caller may had its native access enabled via the API rather than the command line.

The method will also needs a @throws IllegalArgumentException for when the module is not in the module layer.

mcimadamore: I agree with the comment - e.g. not specifying command line options in the API javadoc. However this is the standard text we have on all restricted methods, so for now I think this change seems at least consistent with all the other restricted methods, and we'll need to do another pass to fix them all.

src/java.base/share/classes/java/lang/ModuleLayer.java
```diff
@@ -297,6 +299,37 @@ public Controller addOpens(Module source, String pn, Module target) {
             source.implAddOpens(pn, target);
             return this;
         }
+
+        /***
```
mcimadamore, line 303: Watch out, three asterisks

mcimadamore (approved):
Looks good

src/java.base/share/classes/java/lang/ModuleLayer.java
sundararajana: javadoc blurb on IllegalCallerException is from other similar Panama APIs. We need to revisit all those (perhaps as separate patch). I've fixed that and added IllegalArgumentException as you suggested.

src/java.base/share/classes/java/lang/ModuleLayer.java
sundararajana: Will remove it.

PR #730, opened 2022-09-21
8292034: Improve javadoc after memory segment/memory address unification

mcimadamore:
This is a rather big patch which overhauls the javadoc of the `MemorySegment` class.
A big thanks to Alex Buckley, who helped review the API javadoc and contributed many of the changes you see here.

The main things touched in this PR are:

* the section on unsafe segments is gone from the package-level javadoc. In its place, there's a new section on zero-length memory segments in the `MemorySegment` javadoc
* the class javadoc for `MemorySegment` has been overhauled greatly. It now focusses on two kinds of segments (native segments, and heap segments) and define how address, size, alignment of both is handled by the API.
* we introduce a distinction between memory segment and "the region of memory" which backs the segment, which helps spelling out a lot of these properties in a cleaner way
* perhaps the biggest change in the javadoc is the section on memory segment alignment, which now is much more clearly defined, and provides lots of examples (as that's a complex topic).
* some of the javadoc, in both this class and other classes have been tweaked to reflect the new terminology (for instance, references to the term `base address` are gone)

Other javadoc changes might follow, at some point later. The `Linker` class might need some attention, as we don't really spell out how zero-length segments interacts with downcalls and upcalls.
But, given the size of the changes, I'd rather deal with that separately.

---------
### Progress

### Issue
 * [JDK-8292034](https://bugs.openjdk.org/browse/JDK-8292034): Improve javadoc after memory segment/memory address unification

### Reviewers
 * [Jorn Vernee](https://openjdk.org/census#jvernee) (@JornVernee - Committer) ⚠️ Review applies to [0ce10944](https://git.openjdk.org/panama-foreign/pull/730/files/0ce10944e1f2b653cfdeb73222a81b92d43676a4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/730/head:pull/730` \
`$ git checkout pull/730`

`$ git checkout pull/730` \
`$ git pull https://git.openjdk.org/panama-foreign pull/730/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 730`

`$ git pr show -t 730`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/730.diff">https://git.openjdk.org/panama-foreign/pull/730.diff</a>

</details>


mcimadamore:
Rather than looking at code changes, it might be perhaps more valuable to look at the javadoc, here:
http://cr.openjdk.java.net/~mcimadamore/panama/memory_segment_refresh_javadoc/javadoc/java.base/module-summary.html

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
```diff
@@ -55,65 +56,81 @@
...
- * {@linkplain #load() loaded} to and from the underlying memory-mapped file;</li>
- *     <li>{@linkplain MemorySegment#ofArray(int[]) array segments}, wrapping an existing, heap-allocated Java array; and</li>
- *     <li>{@linkplain MemorySegment#ofBuffer(Buffer) buffer segments}, wrapping an existing {@link Buffer} instance;
- * buffer memory segments might be backed by either off-heap memory or on-heap memory, depending on the characteristics of the
- * wrapped buffer instance. For instance, a buffer memory segment obtained from a byte buffer created with the
- * {@link ByteBuffer#allocateDirect(int)} method will be backed by off-heap memory.</li>
+ *     <li>A <em>heap segment</em> is backed by, and provides access to, a region of memory inside the Java heap (an "on-heap" region).</li>
+ *     <li>A <em>native segment</em> is backed by, and provides access to, a region of memory outside the Java heap (an "off-heap" region).</li>
  * </ul>
+ * Heap segments can be obtained by calling one of the {@link MemorySegment#ofArray(int[])} factory methods.
```
minborg, line 66: There is just one MemorySegment#ofArray(int[]) method. Is there a way of expressing a link to the method reference `MemorySegment::ofArray` (which covers all the overloads)?

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
```diff
@@ -55,65 +56,81 @@
...
+ * These methods return a memory segment backed by the on-heap region that holds the specified Java array.
+ * <p>
+ * Native segments can be obtained by calling one of the {@link MemorySegment#allocateNative(long, long)}
```
minborg, line 69: DItto here

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
```diff
@@ -176,28 +196,103 @@
...
  * }
  * }
  *
- * If, however, the segment being dereferenced is a heap segment, the above function will not work: a heap
- * segment's {@linkplain #address() base address} is <em>virtualized</em> and, as such, cannot be used to construct an
- * alignment check. Instead, heap segments are assumed to produce addresses which are never more aligned than the element
- * size of the Java array from which they have originated from, as shown in the following table:
+ * For example:
+ * <ul>
+ * <li>A native segment at address 1000 can be accessed at offsets 0, 8, 16, 24, etc under an 8-byte alignment constraint,
```
minborg, line 224: Here, we have to be more specific about the alignment of the segment address and the alignment of the layout otherwise there might be confusion.

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
```diff
@@ -769,122 +920,138 @@ default void setUtf8String(long offset, String str) {
...
-     * Creates an array memory segment that models the memory associated with the given heap-allocated int array.
-     * The returned segment is associated with the {@linkplain MemorySession#global() global} memory session.
+     * Creates a heap segment that models the region of memory associated with the given int array.
+     * The returned segment is associated with the {@linkplain MemorySession#global() global} memory session, and
+     * its {@link #address()} is set to zero.
      *
-     * @param intArray the primitive array backing the array memory segment.
-     * @return an array memory segment.
+     * @param intArray the primitive array backing the heap segment.
+     * @return a heap memory segment backed by a int array.
```
minborg, line 991: "an int array"

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
mcimadamore: not that I know of

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
mcimadamore: I think this includes all the information? E.g. address 1000 is a number, which is can be divided by 2/4/8.  In a paragraph above we explain that:

```
For example, the address 1000 is aligned according to an 8-byte alignment constraint (because 1000 is an integer multiple of 8), and to a 4-byte alignment constraint, and to a 2-byte alignment constraint; in contrast, the address 1004 is aligned according to a 4-byte alignment constraint, and to a 2-byte alignment constraint, but not to an 8-byte alignment constraint.
 ```

File: src/java.base/share/classes/java/lang/foreign/MemorySegment.java [hunk 5]
```diff
@@ -769,122 +920,138 @@ default void setUtf8String(long offset, String str) {
...
+     *     <li>if the buffer is a direct buffer, then {@code S} is an
+     *     {@linkplain MemorySession#openImplicit() implicit session} that keeps the buffer reachable.
+     *     Therefore, the off-heap memory associated with the buffer instance will remain available as long as the
+     *     returned segment is reachable.</li>
+     * </ul>
      *
-     * @param buffer the buffer instance backing the buffer memory segment.
-     * @return a buffer memory segment.
+     * @param buffer the buffer instance to be wrapped by a new memory segment.
+     * @return a memory segment, wrapping the given buffer instance.
```
JornVernee, line 943: Since there's not always a wrapping happening, maybe:
```suggestion
     * @param buffer the buffer instance to be turned into a new memory segment.
     * @return a memory segment, derived from the given buffer instance.
```

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
```diff
@@ -963,10 +1135,10 @@ static MemorySegment ofAddress(long address, long byteSize, MemorySession sessio
      *         .allocate(layout.bytesSize(), layout.bytesAlignment());
      * }
      * <p>
-     * The block of off-heap memory associated with the returned native memory segment is initialized to zero.
+     * The region of off-heap region of memory associated with the returned native segment is initialized to zero.
```
JornVernee, line 1138:
```suggestion
     * The off-heap region of memory associated with the returned native segment is initialized to zero.
```

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
```diff
@@ -975,22 +1147,26 @@ static MemorySegment allocateNative(MemoryLayout layout) {
...
+     * of the returned segment is guaranteed to be at least 1-byte aligned.
      * <p>
      * The returned memory segment is associated with a new {@linkplain MemorySession#openImplicit implicit}
-     * memory session. As such, the native memory region associated with the returned segment is
+     * memory session. As such, the off-heap region which backs the returned segment is
      * freed <em>automatically</em>, some unspecified time after it is no longer referenced.
      * <p>
      * Native segments featuring deterministic deallocation can be obtained using the
-     * {@link MemorySession#allocate(long)}} method.
+     * {@link MemorySession#allocate(MemoryLayout)} method.
```
JornVernee, line 1161: Is this the right method to reference? Since this method takes a `long`, I'd expect the referenced method to be `allocate(long)` as well.

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
```diff
@@ -1005,22 +1181,26 @@ static MemorySegment allocateNative(long byteSize) {
...
+     * of the returned segment will be aligned according to the provided alignment constraint.
      * <p>
      * The returned memory segment is associated with a new {@linkplain MemorySession#openImplicit implicit}
-     * memory session. As such, the native memory region associated with the returned segment is
+     * memory session. As such, the off-heap region which backs the returned segment is
      * freed <em>automatically</em>, some unspecified time after it is no longer referenced.
      * <p>
      * Native segments featuring deterministic deallocation can be obtained using the
-     * {@link MemorySession#allocate(long, long)} method.
+     * {@link MemorySession#allocate(MemoryLayout)} method.
```
JornVernee, line 1195: Same here.

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
```diff
@@ -1891,20 +2053,20 @@ default void setAtIndex(ValueLayout.OfAddress layout, long index, MemorySegment
...
+     * object is also a memory segment, and if the two segments refer to the same location, in some region of memory.
+     * More specifically, for two segments {@code s1} and {@code s2} to be considered equals, all the following must be true:
      * <ul>
      *     <li>{@code s1.array().equals(s2.array())}, that is, the two segments must be of the same kind;
      *     either both are {@linkplain #isNative() native segments}, backed by off-heap memory, or both are backed by
      *     the same on-heap Java array;
-     *     <li>{@code s1.address() == s2.address()}, that is, the base address of the two segments should be the same.
-     *     This means that the two segments either refer at the same off-heap memory location, or they refer
+     *     <li>{@code s1.address() == s2.address()}, that is, the address of the two segments should be the same.
+     *     This means that the two segments either refer at the same location in some off-heap region, or they refer
```
JornVernee, line 2063:
```suggestion
     *     This means that the two segments either refer to the same location in some off-heap region, or they refer
```

JornVernee (commented):
Mostly minor nits. I think my main gripe is the "physical address" terminology. I think depending on the situation, the "physical" could maybe be left out. Or maybe replaced with another term, like off-heap, native, or maybe real/underlying.

src/java.base/share/classes/java/lang/foreign/MemorySession.java
```diff
@@ -215,14 +215,17 @@ public sealed interface MemorySession extends AutoCloseable, SegmentAllocator pe
...
+     * <p>
      * A client is responsible for ensuring that this memory session is closed when the
      * segment is no longer in use. Failure to do so will result in off-heap memory leaks.
      * <p>
-     * The off-heap memory associated with the returned native memory segment is initialized to zero.
+     * The off-heap region of memory associated with the returned native memory segment is initialized to zero.
      *
      * @param byteSize the size (in bytes) of the off-heap memory block backing the native memory segment.
-     * @param byteAlignment the alignment constraint (in bytes) of the off-heap memory block backing the native memory segment.
+     * @param byteAlignment the alignment constraint (in bytes) of the off-heap region of memory backing the native memory segment.
```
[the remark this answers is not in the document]
mcimadamore: well spotted - "backing" is the way to go. Memory sessions are "associated" to segments. Regions of memory "back" segments. Note that we deliberately use "region of memory" instead of "memory region" to avoid confusion between the concept and API construct (e.g. memory segment - there are many memory XYZ in the API).

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
```diff
@@ -55,65 +56,81 @@
...
- * {@link ByteBuffer#allocateDirect(int)} method will be backed by off-heap memory.</li>
+ *     <li>A <em>heap segment</em> is backed by, and provides access to, a region of memory inside the Java heap (an "on-heap" region).</li>
+ *     <li>A <em>native segment</em> is backed by, and provides access to, a region of memory outside the Java heap (an "off-heap" region).</li>
  * </ul>
+ * Heap segments can be obtained by calling one of the {@link MemorySegment#ofArray(int[])} factory methods.
+ * These methods return a memory segment backed by the on-heap region that holds the specified Java array.
+ * <p>
+ * Native segments can be obtained by calling one of the {@link MemorySegment#allocateNative(long, long)}
+ * factory methods, which return a memory segment backed by a newly allocated off-heap region with the given size
+ * and alignment constraint. Alternatively, native segments can be obtained by
```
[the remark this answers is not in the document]
mcimadamore: Yeah, I agree  that works better.

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
```diff
@@ -176,28 +196,103 @@
  *
  * <h2 id="segment-alignment">Alignment</h2>
  *
- * When dereferencing a memory segment using a layout, the runtime must check that the segment address being dereferenced
- * matches the layout's {@linkplain MemoryLayout#byteAlignment() alignment constraints}. If the segment being
- * dereferenced is a native segment, then it has a concrete {@linkplain #address() base address}, which can
- * be used to perform the alignment check. The pseudo-function below demonstrates this:
+ * Access operations on a memory segment are constrained not only by the spatial and temporal bounds of the segment,
+ * but also by the <em>alignment constraint</em> of the value layout specified to the operation. An access operation can
+ * access only those offsets in the segment that denote addresses in physical memory which are <em>aligned</em> according
```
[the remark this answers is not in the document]
mcimadamore: I note here that we use physical address, but here we say address in physical memory - which seems like a mismatch (we never defined the term physical memory).

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
```diff
@@ -176,28 +196,103 @@
...
+ * constraint; in contrast, the address 1004 is aligned according to a 4-byte alignment constraint, and to a 2-byte alignment
+ * constraint, but not to an 8-byte alignment constraint.
+ * Access operations are required to respect alignment because it can impact the performance of access operations, and
+ * can also determine which access operations are available at a given physical address. For instance,
+ * {@linkplain java.lang.invoke.VarHandle#compareAndSet(Object...) atomic access operations} operations using
+ * {@link java.lang.invoke.VarHandle} are only permitted at aligned addresses. In addition, alignment
+ * applies to an access operation whether the segment being accessed is a native segment or a heap segment.
+ * <p>
+ * If the segment being accessed is a native segment, then its {@linkplain #address() address} in physical memory can be
+ * combined with the offset to obtain the <em>target address</em> in physical memory. The pseudo-function below demonstrates this:
```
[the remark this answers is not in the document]
mcimadamore: Ok, it seems there's a lot of those.

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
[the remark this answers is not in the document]
mcimadamore: Agree

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
```diff
@@ -176,28 +196,103 @@
...
+ * <li>A native segment at address 1006 can be accessed at offsets 0, 2, 4, 6, etc under a 2-byte alignment constraint.
+ * Under a 4-byte alignment constraint, it can be accessed at offsets 2, 6, 10, 14, etc.
+ * Under an 8-byte alignment constraint, it can be accessed at offsets 2, 10, 18, 26, etc.
+ * <li>A native segment at address 1007 can be accessed at offsets 0, 1, 2, 3, etc under a 1-byte alignment constraint.
+ * Under a 2-byte alignment constraint, it can be accessed at offsets 1, 3, 5, 7, etc.
+ * Under a 4-byte alignment constraint, it can be accessed at offsets 1, 5, 9, 13, etc.
+ * Under an 8-byte alignment constraint, it can be accessed at offsets 1, 9, 17, 25, etc.</li>
+ * </ul>
+ * <!-- Under a 1-byte alignment constraint, any offset address in a native segment can be accessed. -->
+ * <!-- In effect, every target address is at least 1-byte aligned. -->
```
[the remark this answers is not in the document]
mcimadamore: whoops - that should be removed.

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
mcimadamore: Good suggestion, we've been generally trying to avoid "wrapping"

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
```diff
@@ -55,65 +56,81 @@
...
+ * Every memory segment has an {@linkplain #address() address}, expressed as a {@code long} value.
+ * The nature of a segment's address depends on the kind of the segment:
+ * <ul>
+ * <li>The address of a heap segment is not a physical address, but rather an offset within the region of memory
+ * which backs the segment. The region is inside the Java heap, so garbage collection might cause the region to be
+ * relocated in physical memory over time, but this is not exposed to clients of the {@code MemorySegment} API who
+ * see a stable <em>virtualized</em> address for a heap segment backed by the region.
+ * A heap segment obtained from one of the {@link #ofArray(int[])} factory methods has an address of zero.</li>
+ * <li>The address of a native segment (including mapped segments) denotes the physical address of the region of
+ * memory which backs the segment.</li>
```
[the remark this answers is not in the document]
JornVernee: I get what is meant by it, but I don't think "physical" is a great term wrt intuitive meaning. It would maybe be nice to define the term somewhere up front if we want to stick with "physical".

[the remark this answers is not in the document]
JornVernee: I think for regular Java developers, there is probably a much less clear distinction between heap/off-heap and "physical" memory in the first place.

i.e. if heap memory is the part of memory that is used to store Java objects, and off-heap is the part of memory that _isn't_ used to store Java objects, then I guess "physical memory" is just all of memory combined? Maybe just saying "memory" is enough? In the above, that could make:

    The {@linkplain #address() address} of a native segment (including mapped segments) denotes the starting address of the region of memory which backs the segment.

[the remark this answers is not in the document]
JornVernee: If none of these options seem better to you though, I don't mind sticking with "physical address/memory".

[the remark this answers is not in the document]
mcimadamore: I think `physical memory` is the problem, more so than `physical address`. We can explain `physical address` as something different from `MemorySegment::address`, but I don't think we need a qualifier for memory.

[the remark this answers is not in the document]
mcimadamore: I looked at this again, and decided to leave as is. `MappedByteBuffer` is also quite full of references to terms like `physical memory` (which we inherited in methods like `MemorySegment::isLoaded`.

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
mcimadamore: I've decided to leave this as is for the time being.

JornVernee (approved):

PR #734, opened 2022-09-22
8292047: Consider ways to add linkage parameters to downcall handles

JornVernee:
This patch implements a linker option interface, that can be using to indicate additional linkage requirements to the linker.

The variadic function descriptor support is dropped in favor of using a linker option to indicate the first variadic argument index. I think the switch turned out to be straight forward for users of this feature as well, though a bit more verbose.

I also experimented with putting linker options on FunctionDescriptor directly, but it increases the complexity of the combinators of FunctionDescriptor. For instance, if an argument layout is inserted somewhere, should the variadic index option also be updated? (a similar issue arises if we support other linker options that include parameter list indexes). As a result of that, it seemed simpler to just move the linker options as an additional argument to a link request, since at that point we know  the descriptor won't change anymore.

Another design decision is to only allow a single instance of a linker option of a certain type (and they are stored internally in a `Map<Class<?>, Option>` as well). This makes it easier to determine whether a combination of linker options is valid.

Other option kinds that could be added in the future are listed in the JBS issue, but not part of this patch.

---------
### Progress

### Issue
 * [JDK-8292047](https://bugs.openjdk.org/browse/JDK-8292047): Consider ways to add linkage parameters to downcall handles

### Reviewers
 * [Maurizio Cimadamore](https://openjdk.org/census#mcimadamore) (@mcimadamore - Committer) ⚠️ Review applies to [8bb13e3d](https://git.openjdk.org/panama-foreign/pull/734/files/8bb13e3dca358ffd1a6160822653d186d62de393)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/734/head:pull/734` \
`$ git checkout pull/734`

`$ git checkout pull/734` \
`$ git pull https://git.openjdk.org/panama-foreign pull/734/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 734`

`$ git pr show -t 734`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/734.diff">https://git.openjdk.org/panama-foreign/pull/734.diff</a>

</details>


src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -282,4 +287,21 @@ default MethodHandle downcallHandle(MemorySegment symbol, FunctionDescriptor fun
...
+
+    /**
+     * A linker option that can be used to indicate additional linking requirements to the linker,
+     * besides what is described by a function descriptor.
+     */
+    sealed interface Option
+            permits LinkerOptions.FirstVariadicArg {
+
+        /**
+         * {@return A linker characteristic used to denote the index of the first variadic argument layout in a
```
JornVernee, line 299:
```suggestion
         * {@return A linker option used to denote the index of the first variadic argument layout in a
```

src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -235,11 +238,13 @@ default MethodHandle downcallHandle(MemorySegment symbol, FunctionDescriptor fun
      * associated with the {@link MemorySegment#NULL} address, or a {@link NullPointerException} if that parameter is {@code null}.
      *
      * @param function the function descriptor of the target function.
+     * @param options  any linker options.
      * @return a downcall method handle. The method handle type is <a href="Linker.html#downcall-method-handles"><em>inferred</em></a>
      * from the provided function descriptor.
      * @throws IllegalArgumentException if the provided function descriptor is not supported by this linker.
+     * @throws IllegalArgumentException if an invalid combination of linker options is given
```
mcimadamore, line 245:
```suggestion
     * @throws IllegalArgumentException if an invalid combination of linker options is given.
```

src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -210,15 +211,17 @@ static Linker nativeLinker() {
...
      *
-     * @param symbol the address of the target function.
+     * @param symbol   the address of the target function.
      * @param function the function descriptor of the target function.
+     * @param options  any linker options.
      * @return a downcall method handle. The method handle type is <a href="Linker.html#downcall-method-handles"><em>inferred</em></a>
      * @throws IllegalArgumentException if the provided function descriptor is not supported by this linker.
-     * or if the symbol is {@link MemorySegment#NULL}
+     *                                  or if the symbol is {@link MemorySegment#NULL}
+     * @throws IllegalArgumentException if an invalid combination of linker options is given
```
mcimadamore, line 220:
```suggestion
     * @throws IllegalArgumentException if an invalid combination of linker options is given.
```

src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/CallArranger.java
```diff
@@ -140,7 +145,7 @@ public Bindings getBindings(MethodType mt, FunctionDescriptor cDesc, boolean for
         for (int i = 0; i < mt.parameterCount(); i++) {
             Class<?> carrier = mt.parameterType(i);
             MemoryLayout layout = cDesc.argumentLayouts().get(i);
-            if (varArgsOnStack() && SharedUtils.isVarargsIndex(cDesc, i)) {
+            if (varArgsOnStack() && options.isVarargsIndex(i)) {
```
mcimadamore, line 148: nice!

test/jdk/java/foreign/TestLinker.java
```diff
@@ -0,0 +1,50 @@
...
+import java.lang.foreign.Linker;
+import java.lang.invoke.MethodHandle;
+
+import static org.testng.Assert.assertNotSame;
+
+public class TestLinker extends NativeTestHelper {
+
+    @Test
+    public void testLinkerOptionsCache() {
+        Linker linker = Linker.nativeLinker();
```
mcimadamore, line 42: Good test

mcimadamore (approved):
I think this looks good. As you say, it's a quite straightforward change, but one that gives us a lot of extra flexibility when linking downcalls/upcalls.

PR #737, opened 2022-10-03
8294309: Downcall and Upcall unboxing code should reject heap segments

JornVernee:
Recently I noticed that we don't check for heap segments anymore when unboxing memory segments to be passed to native code as addresses. This allows someone to pass a heap segment, which will be unboxed simply by calling `MemorySegment::address` which for heap segments returns the byte offset into the array. This is of course not a valid pointer.

This patch changes the unboxing logic to reject heap segments with an `IllegalArgumentException`.

---------
### Progress

### Issue
 * [JDK-8294309](https://bugs.openjdk.org/browse/JDK-8294309): Downcall and Upcall unboxing code should reject heap segments

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/737/head:pull/737` \
`$ git checkout pull/737`

`$ git checkout pull/737` \
`$ git pull https://git.openjdk.org/panama-foreign pull/737/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 737`

`$ git pr show -t 737`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/737.diff">https://git.openjdk.org/panama-foreign/pull/737.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/jdk/internal/foreign/abi/SharedUtils.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/abi/SharedUtils.java
@@ -28,6 +28,7 @@ import jdk.internal.access.JavaLangAccess;
 import jdk.internal.access.JavaLangInvokeAccess;
 import jdk.internal.access.SharedSecrets;
 import jdk.internal.foreign.CABI;
+import jdk.internal.foreign.HeapMemorySegmentImpl;
 import jdk.internal.foreign.abi.aarch64.linux.LinuxAArch64Linker;
 import jdk.internal.foreign.abi.aarch64.macos.MacOsAArch64Linker;
 import jdk.internal.foreign.abi.x64.sysv.SysVx64Linker;
@@ -258,6 +259,13 @@ public final class SharedUtils {
             t.printStackTrace();
             JLA.exit(1);
         }
+    }
+
+    static long unboxSegment(MemorySegment segment) {
+        if (segment instanceof HeapMemorySegmentImpl) {
+            throw new IllegalArgumentException("Heap segment not allowed: " + segment);
+        }
+        return segment.address();
     }
 
     public static void checkExceptions(MethodHandle target) {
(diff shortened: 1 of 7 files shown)


test/jdk/java/foreign/TestUpcallException.java
```diff
@@ -33,71 +33,26 @@
...
 import java.io.IOException;
-import java.io.InputStream;
-import java.io.InputStreamReader;
-import java.nio.file.Paths;
-import java.util.List;
 
-import static org.testng.Assert.assertFalse;
-import static org.testng.Assert.assertNotEquals;
-import static org.testng.Assert.assertTrue;
+public class TestUpcallException extends UpcallTestHelper {
```
JornVernee, line 41: I refactored this class to use a common utility method in the super class, which is also used by the new test

test/jdk/java/foreign/UpcallTestHelper.java
```diff
@@ -0,0 +1,87 @@
...
+import java.io.InputStreamReader;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import static org.testng.Assert.assertNotEquals;
+import static org.testng.Assert.assertTrue;
+
+public class UpcallTestHelper extends NativeTestHelper {
```
JornVernee, line 38: I thought about just adding this to NativeTestHelper, but didn't want to require every dependent test to have to add `@library /test/lib`

src/java.base/share/classes/jdk/internal/foreign/abi/SharedUtils.java
```diff
@@ -260,6 +261,13 @@ static void handleUncaughtException(Throwable t) {
         }
     }
 
+    static long unboxSegment(MemorySegment segment) {
+        if (segment instanceof HeapMemorySegmentImpl) {
```
mcimadamore, line 265: there's MemorySegment.isNative, which maybe is better?

JornVernee: Yeah, that works. It also includes mapped segments

mcimadamore (approved):

PR #739, opened 2022-10-11
8294986: Possible mismatched acquire/release in binding specializer

JornVernee:
We acquire sessions of address parameters before a downcall, and release them afterwards. However, if acquire fails with an exception, we still call release in the finally block.

Fixing this is a simple case of switching 2 lines: when acquire fails, we don't store the session in a local, and so the cleanup in the finally block will not call release on it either.

For testing, I doesn't seem possible to make this bug result in an error with the current code. So, I've added an assert to the implementation that checks whether release was called too many times.

---------
### Progress

### Issue
 * [JDK-8294986](https://bugs.openjdk.org/browse/JDK-8294986): Possible mismatched acquire/release in binding specializer

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/739/head:pull/739` \
`$ git checkout pull/739`

`$ git checkout pull/739` \
`$ git pull https://git.openjdk.org/panama-foreign pull/739/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 739`

`$ git pr show -t 739`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/739.diff">https://git.openjdk.org/panama-foreign/pull/739.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/jdk/internal/foreign/ConfinedSession.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/ConfinedSession.java
@@ -81,12 +81,12 @@ final class ConfinedSession extends MemorySessionImpl {
 
     void justClose() {
         checkValidState();
-        if (state == 0 || state - ((int)ASYNC_RELEASE_COUNT.getVolatile(this)) == 0) {
-            assert state != 0 || ((int)ASYNC_RELEASE_COUNT.getVolatile(this)) == 0
-                    : "if state was 0, there should be no async releases";
+        int asyncCount = (int)ASYNC_RELEASE_COUNT.getVolatile(this);
+        if ((state == 0 && asyncCount == 0)
+                || ((state - asyncCount) == 0)) {
             state = CLOSED;
         } else {
-            throw alreadyAcquired(state);
+            throw alreadyAcquired(state - asyncCount);
         }
     }
 
--- a/src/java.base/share/classes/jdk/internal/foreign/ConfinedSession.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/ConfinedSession.java
@@ -82,6 +82,8 @@ final class ConfinedSession extends MemorySessionImpl {
     void justClose() {
         checkValidState();
         if (state == 0 || state - ((int)ASYNC_RELEASE_COUNT.getVolatile(this)) == 0) {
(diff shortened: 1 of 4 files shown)


mcimadamore (approved):
Looks good. Not super sure about the `assert`, which will only fire with assertion enabled. Historically I've been using `assert` keywords sparingly and using `throw new AssertError` more (especially in the javac code base), as that is guaranteed to fail. But that is subjective.

JornVernee:
I used an assert instead of an explicit exception since this is a case that should never happen in a sane implementation, and didn't want to introduce a performance penalty to the code for the common case where someone just has a confined session that is closed. Since the assert condition always does a volatile read in that case.

mcimadamore:
> Looks good. Not super sure about the `assert`, which will only fire with assertion enabled. Historically I've been using `assert` keywords sparingly and using `throw new AssertError` more (especially in the javac code base), as that is guaranteed to fail. But that is subjective.

Also, in the `SharedSession` implementation we throw a runtime exception in case of too many releases.

JornVernee:
I've removed the use of assert in favor of an explicit check. In the case of too many releases the exception message will be: `Session is acquired by -1 clients` I think that is probably good?

src/java.base/share/classes/jdk/internal/foreign/ConfinedSession.java
```diff
@@ -81,10 +81,12 @@ public void release0() {
...
     void justClose() {
         checkValidState();
-        if (state == 0 || state - ((int)ASYNC_RELEASE_COUNT.getVolatile(this)) == 0) {
+        int asyncCount = (int)ASYNC_RELEASE_COUNT.getVolatile(this);
+        if ((state == 0 && asyncCount == 0)
+                || ((state - asyncCount) == 0)) {
             state = CLOSED;
         } else {
-            throw alreadyAcquired(state);
+            throw alreadyAcquired(state - asyncCount);
```
JornVernee, line 89: I figured the exception message should take asyncCount into account as well. either both are 0, or state  - asynCount should be 0 for close to succeed. So, if we get here, we should always get a non-zero value in the exception message.

mcimadamore (approved):
Looks good!

PR #740, opened 2022-10-13
8295265: Refactor handling of special values passed to stubs

JornVernee:
This patch refactors the handling of special values passed between stubs and Java code, such as the target address and the return buffer address.

Currently the Java code determines some registers that these values are stored in. This refactor changes that to use placeholders, and lets the stub generation code fill in the actual locations. This allows more special values (such as locations to store preserved thread locals), as well as storing things to a location in the frame, as opposed to a register.

To do this, this patch adds a Java enum and C++ helper class called StubLocations, which gives each special value an index (ordinal). The C++ helper class also has an array which maps this index to the actual location the stub gen code decides to use. This location is then used in shuffling code.

Instead of the ABIDescriptor being used to determine a register for the target address and return buffer address, it now has 2 scratch registers, which don't conflict with arguments being passed. They can be used by the stub gen code.

The patch also contains a minor renaming of a boolean value being used. I've also renamed StorageClasses, and RegType in Java and C++ code respectively to StorageType, to align the 2, since they also need to be kept in sync. The latter is responsible for a lot of changes in this patch.

The stub gen code now also has a special FRAME_DATA storage type, which is used to indicate locations in the stub frame. This is for instance used to store the return buffer address.

---------
### Progress

### Issue
 * [JDK-8295265](https://bugs.openjdk.org/browse/JDK-8295265): Refactor handling of special values passed to stubs

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/740/head:pull/740` \
`$ git checkout pull/740`

`$ git checkout pull/740` \
`$ git pull https://git.openjdk.org/panama-foreign pull/740/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 740`

`$ git pr show -t 740`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/740.diff">https://git.openjdk.org/panama-foreign/pull/740.diff</a>

</details>

### DIFF
--- a/src/hotspot/cpu/aarch64/downcallLinker_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/downcallLinker_aarch64.cpp
@@ -215,10 +215,10 @@ void DowncallStubGenerator::generate() {
     int offset = 0;
     for (int i = 0; i < _output_registers.length(); i++) {
       VMStorage reg = _output_registers.at(i);
-      if (reg.type() == RegType::INTEGER) {
+      if (reg.type() == StorageType::INTEGER) {
         __ str(as_Register(reg), Address(tmp1, offset));
         offset += 8;
-      } else if(reg.type() == RegType::VECTOR) {
+      } else if(reg.type() == StorageType::VECTOR) {
         __ strd(as_FloatRegister(reg), Address(tmp1, offset));
         offset += 16;
       } else {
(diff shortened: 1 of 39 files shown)


mcimadamore (approved):
The Java code and test changes look good. I eyeballed the VM changes, which seems ok - e.g. they seem to add an extra degree of freedom/indirection where certain locations are "virtualized", so each ABI can decide how to model them using scratch registers or something else. It obviously seems a bit convoluted for now, but reading the description in JDK-8295265 helps a bit understanding why this was needed, and where this is going.

PR #741, opened 2022-10-14
8294621: [Panama] Copying from a StringCharBuffer-backed MemorySegment results in crash

minborg:
This PR prevents `StringCharBuffer` arguments in `MemorySegment::ofBuffer`.

---------
### Progress

### Issue
 * [JDK-8294621](https://bugs.openjdk.org/browse/JDK-8294621): [Panama] Copying from a StringCharBuffer-backed MemorySegment results in crash

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/741/head:pull/741` \
`$ git checkout pull/741`

`$ git checkout pull/741` \
`$ git pull https://git.openjdk.org/panama-foreign pull/741/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 741`

`$ git pr show -t 741`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/741.diff">https://git.openjdk.org/panama-foreign/pull/741.diff</a>

</details>


src/java.base/share/classes/java/lang/foreign/MemorySegment.java
```diff
@@ -939,6 +939,7 @@ default void setUtf8String(long offset, String str) {
      *
      * @param buffer the buffer instance to be turned into a new memory segment.
      * @return a memory segment, derived from the given buffer instance.
+     * @throws IllegalArgumentException if the provided {@code buffer} is a StringCharBuffer.
```
mcimadamore, line 942: Note that StringCharBuffer is not a public API type. It is something we create with `CharBuffer,wrap(CharSequence)` so the doc should reflect that.

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
```diff
@@ -939,6 +940,9 @@ default void setUtf8String(long offset, String str) {
      *
      * @param buffer the buffer instance to be turned into a new memory segment.
      * @return a memory segment, derived from the given buffer instance.
+     * @throws IllegalArgumentException if the provided {@code buffer} is obtained by calling either
+     *                                  ({@link CharBuffer#wrap(CharSequence)} or {@link CharBuffer#wrap(char[], int, int)}
+     *                                  because these Buffer objects are not backed directly by a char array.
```
mcimadamore, line 945:
```suggestion
     *                                  because these buffer objects are not backed directly by a char array.
```

src/java.base/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
```diff
@@ -469,6 +469,10 @@ public int hashCode() {
 
     public static AbstractMemorySegmentImpl ofBuffer(Buffer bb) {
         Objects.requireNonNull(bb);
+        // StringCharBuffer is packet-private
+        if ("java.nio.StringCharBuffer".equals(bb.getClass().getName())) {
+            throw new IllegalArgumentException("StringCharBuffer objects are not supported because they are not directly backed by a char[] array.");
```
mcimadamore, line 474:
```suggestion
            throw new IllegalArgumentException("CharBuffer is not backed by a char[].");
```

test/jdk/java/foreign/TestOfBufferIssue.java
```diff
@@ -0,0 +1,53 @@
...
+        try {
+            final CharBuffer cb = CharBuffer.wrap("Hello");
+            MemorySegment src2 = MemorySegment.ofBuffer(cb);
+            fail("A StringCharBuffer is not allowed as an argument.");
+        } catch (IllegalArgumentException iae) {
+            // Ignored. Happy path
+        }
+    }
+
+}
```
mcimadamore, line 53: Watch out here missing newline

minborg: Why do we need a newline there?

JornVernee: Unix file specification says that every line in a file should end with a newline. I believe this is also why GitHub warns about it.

mcimadamore: It's one of the custom checks we have always applied to JDK code. E.g. no trailing spaces, newline at the end. It's odd that jcheck says it's fine. I recall these things being enforced more strictly in the past (e.g. via HG hooks).

src/java.base/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
```diff
@@ -469,6 +469,10 @@ public int hashCode() {
 
     public static AbstractMemorySegmentImpl ofBuffer(Buffer bb) {
         Objects.requireNonNull(bb);
+        // StringCharBuffer is packet-private
+        if ("java.nio.StringCharBuffer".equals(bb.getClass().getName())) {
```
mcimadamore, line 473: IMHO would be better to call Buffer::isDirect, and then check if Buffer::array is == null.

mcimadamore: Nevermind, the array is not given away for read-only buffers... grrr

JornVernee: We get the array in the code below. I guess we can check if it's `null` there?

mcimadamore: yep - that works as `NIO_ACCESS.getBufferBase` goes straight to `Buffer.hb`, which is what we need.

PR #742, opened 2022-10-14
8294970: Add linker option for saving thread-locals that the VM can overwrite

JornVernee:
DRAFT

### DIFF
--- a/src/java.base/share/classes/java/lang/foreign/Linker.java
+++ b/src/java.base/share/classes/java/lang/foreign/Linker.java
@@ -27,12 +27,19 @@ package java.lang.foreign;
 
 import jdk.internal.foreign.abi.AbstractLinker;
 import jdk.internal.foreign.abi.LinkerOptions;
+import jdk.internal.foreign.abi.PreservableValues;
 import jdk.internal.foreign.abi.SharedUtils;
 import jdk.internal.javac.PreviewFeature;
 import jdk.internal.reflect.CallerSensitive;
 import jdk.internal.reflect.Reflection;
 
 import java.lang.invoke.MethodHandle;
+import java.util.Arrays;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 /**
  * A linker provides access to foreign functions from Java code, and access to Java code from foreign functions.
@@ -293,7 +300,8 @@ public sealed interface Linker permits AbstractLinker {
      * besides what is described by a function descriptor.
      */
     sealed interface Option
-            permits LinkerOptions.FirstVariadicArg {
+            permits LinkerOptions.LinkerOptionImpl,
+                    Option.PreserveValue {
 
         /**
          * {@return A linker option used to denote the index of the first variadic argument layout in a
@@ -302,6 +310,39 @@ public sealed interface Linker permits AbstractLinker {
          */
         static Option firstVariadicArg(int index) {
             return new LinkerOptions.FirstVariadicArg(index);
+        }
+
+        /**
+         * {@return A linker option used to preserve values that might be overwirtten by the runtime
+         *          before they can be read through conventional means}
+         * @param preservedValues the names of the values to preserve.
+         * @see PreserveValue#supported()
+         */
+        static PreserveValue preserveValue(String... preservedValues) {
+            Set<String> set = Stream.of(preservedValues).collect(Collectors.toSet());
+            return new LinkerOptions.PreserveValueImpl(set);
+        }
+
+        /**
+         * A linker option for preserving a value that might be overwritten by the runtime
+         * before it can be read through conventional means.
+         */
+        sealed interface PreserveValue extends Option
+                                       permits LinkerOptions.PreserveValueImpl {
+            /**
+             * {@return A struct layout that should be used to allocate the segment for,
+             *          and retrieve preserved values}
+             */
+            StructLayout layout();
+
+            /**
+             * {@return the names of the values that can be preserved by this implementation}
+             */
+            static Set<String> supported() {
+                return Arrays.stream(PreservableValues.values())
+                             .map(PreservableValues::valueName)
+                             .collect(Collectors.toSet());
+            }
         }
     }
 }
--- a/src/java.base/share/classes/jdk/internal/foreign/abi/AbstractLinker.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/abi/AbstractLinker.java
@@ -54,6 +54,7 @@ public abstract sealed class AbstractLinker implements Linker permits LinuxAArch
         Objects.requireNonNull(options);
         checkHasNaturalAlignment(function);
         LinkerOptions optionSet = LinkerOptions.of(options);
+        optionSet.validateForDowncall(function);
 
         return DOWNCALL_CACHE.get(new LinkRequest(function, optionSet), linkRequest ->  {
             FunctionDescriptor fd = linkRequest.descriptor();
--- a/src/java.base/share/classes/jdk/internal/foreign/abi/PreservableValues.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/abi/PreservableValues.java
@@ -24,13 +24,46 @@
  */
 package jdk.internal.foreign.abi;
 
-// must keep in sync with StubLocations in VM code
-public enum StubLocations {
-    TARGET_ADDRESS,
-    RETURN_BUFFER,
-    SAVE_THREAD_LOCAL;
+import java.lang.foreign.ValueLayout;
+import java.util.stream.Stream;
 
-    public VMStorage storage(byte type) {
-        return new VMStorage(type, (short) 8, ordinal());
+import static java.lang.foreign.ValueLayout.JAVA_INT;
+
+public enum PreservableValues {
+    GET_LAST_ERROR    ("GetLastError",    JAVA_INT, 1 << 0),
+    WSA_GET_LAST_ERROR("WSAGetLastError", JAVA_INT, 1 << 1),
+    ERRNO             ("errno",           JAVA_INT, 1 << 2);
+
+    private final String valueName;
+    private final ValueLayout layout;
+    private final int mask;
+
+    PreservableValues(String valueName, ValueLayout layout, int mask) {
+        this.valueName = valueName;
+        this.layout = layout.withName(valueName);
+        this.mask = mask;
+    }
+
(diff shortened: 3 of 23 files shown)


test/jdk/java/foreign/preservevalue/libPreserveValue.c
```diff
@@ -0,0 +1,34 @@
...
+
+#include <errno.h>
+
+#ifdef _WIN64
+#define EXPORT __declspec(dllexport)
+#else
+#define EXPORT
+#endif
+
+EXPORT void set_errno(int value) {
```
JornVernee, line 32: Simple wrapper to be able to set errno easily on different platforms

src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -303,5 +311,67 @@ default MethodHandle downcallHandle(MemorySegment symbol, FunctionDescriptor fun
...
+         *          before they can be read through conventional means}
+         * <p>
+         * A downcall method handle linked with this option will feature an additional {@link MemorySegment}
+         * parameter directly following the target address parameter. This memory segment represents
+         * the off-heap memory into which the preserved values are to be written.
+         *
+         * @param preservedValues the names of the values to preserve.
+         * @see PreserveValue#supported()
+         */
+        static PreserveValue preserveValue(String... preservedValues) {
```
mcimadamore, line 326: I think I prefer `saveValue` - it's simpler, but means more or less the same thing.
Also, I think we should use the plural "saveValues" both in the factory and in the option interface.

src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -303,5 +311,67 @@ default MethodHandle downcallHandle(MemorySegment symbol, FunctionDescriptor fun
...
+         *     handle.invoke(preservedValues);
+         *     int errno = errnoHandle.get(preservedValues);
+         *     // use errno
+         * }
+         * }
+         */
+        sealed interface PreserveValue extends Option
+                                       permits LinkerOptions.PreserveValueImpl {
+            /**
+             * {@return A struct layout that represents the layout of the memory region passed
```
mcimadamore, line 362: maybe "layout of the native segment" instead of region? If you stick with region, I suggest "off-heap region of memory" which is what we use elsewhere,

src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -303,5 +311,67 @@ default MethodHandle downcallHandle(MemorySegment symbol, FunctionDescriptor fun
         static Option firstVariadicArg(int index) {
             return new LinkerOptions.FirstVariadicArg(index);
         }
+
+        /**
+         * {@return A linker option used to preserve values that might be overwritten by the runtime
+         *          before they can be read through conventional means}
+         * <p>
+         * A downcall method handle linked with this option will feature an additional {@link MemorySegment}
```
mcimadamore, line 319: We should say somewhere that the additional segment should be a native segment (right?)

src/java.base/share/classes/jdk/internal/foreign/abi/AbstractLinker.java
```diff
@@ -54,6 +54,7 @@ public MethodHandle downcallHandle(FunctionDescriptor function, Option... option
         Objects.requireNonNull(options);
         checkHasNaturalAlignment(function);
         LinkerOptions optionSet = LinkerOptions.of(options);
+        optionSet.validateForDowncall(function);
```
mcimadamore, line 57: Does it make sense to have a validation step that is disjoint from the creation of the LinkerOptions? E.g. should LinkerOptions factory just take the descriptor and perform all the required validation in one go?

src/java.base/share/classes/java/lang/foreign/Linker.java
mcimadamore: I also find it a bit odd that I can create a LinkerOption whose preservedValue is unrelated from the supported ones? E.g. I'd like to get an exception early, while it seems like the implementation will only give me an exception when I'm creating the downcall, which seems a tad late.

src/java.base/share/classes/jdk/internal/foreign/abi/PreservableValues.java
```diff
@@ -0,0 +1,69 @@
...
+    WSA_GET_LAST_ERROR("WSAGetLastError", JAVA_INT, 1 << 1),
+    ERRNO             ("errno",           JAVA_INT, 1 << 2);
+
+    private final String valueName;
+    private final ValueLayout layout;
+    private final int mask;
+
+    PreservableValues(String valueName, ValueLayout layout, int mask) {
+        this.valueName = valueName;
+        this.layout = layout.withName(valueName);
```
mcimadamore, line 43: Ok, layouts get the name attached - which can be used by developers to generate the correct getter. Another choice, I realize, would have been for the PreserveValue linker option to expose a static getter which takes a name. Then you don't need to expose a layout. Not sure which one is better. The current approach is probably cleaner in that if the layout doesn't have a certain named layout, then you cannot obtain a getter for it.

test/jdk/java/foreign/TestLinker.java
```diff
@@ -24,16 +24,20 @@
 /*
  * @test
  * @enablePreview
- * @run testng TestLinker
+ * @run junit TestLinker
```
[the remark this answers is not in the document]
mcimadamore: Not sure - but I can only find one test  that has imports for `org.junit.jupiter.xyz`. Wouldn't it be easy to have a testng data provider with the two values, even if slightly more verbose?

mcimadamore (commented):
A very good start - I left some comments/questions, especially around the API

src/java.base/share/classes/java/lang/foreign/Linker.java
JornVernee: True. I think we can validate the names here. (I think I had that in a previous iteration as well).

src/java.base/share/classes/jdk/internal/foreign/abi/AbstractLinker.java
JornVernee: I don't really like that idea, because it requires the descriptor to be specified twice, which opens up the possibility of them going out of sync.

JornVernee: I think we should have 2 validation steps, one when creating the option, where we can check if the option itself is valid, and then another when linking, where we check whether the option makes sense for a particular linkage request.

mcimadamore: I understand that the steps are logically two, and are implemented separately. But if all the code we have is doing:

```
        LinkerOptions optionSet = LinkerOptions.of(options);
        optionSet.validateForDowncall(function);
```

Then, by exposing them as separate steps, we don't gain much, but it might be the case that, at some point later, another client forgets about calling the validation step.

JornVernee: Ah, sorry, I miss read what you were saying. I thought the suggestion was to add the validation + descriptor param when creating Linker.Options.

I agree, doing the validation as part of the factory for LinkerOptions would be better.

src/java.base/share/classes/jdk/internal/foreign/abi/PreservableValues.java
JornVernee: Besides getting, you'd also need to know the layout in order to allocate the segment to pass to a downcall. So, the layout is needed either way. A _static_ getter wouldn't work, since a layout can be different depending on the combination of values saved. An instance getter could be an option, but then I don't think the var handle underneath will be seen as constant.

mcimadamore: Good point

src/java.base/share/classes/java/lang/foreign/Linker.java
JornVernee: I've changed the text a bit to specify that a native segment should be used.

JornVernee:
I've addressed all the review comments so far.

src/java.base/share/classes/java/lang/foreign/Linker.java
mcimadamore: Should we add a test for this?

JornVernee: It's already covered by `TestPassHeapSegment` since we use the same code everywhere to unbox these segments.

I don't mind adding an explicit test for this case as well though.

src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -303,5 +309,71 @@ default MethodHandle downcallHandle(MemorySegment symbol, FunctionDescriptor fun
...
+         */
+        static CaptureCallState captureCallState(String... capturedState) {
+            Set<CapturableState> set = Stream.of(capturedState)
+                    .map(CapturableState::forName)
+                    .collect(Collectors.toSet());
+            return new LinkerOptions.CaptureCallStateImpl(set);
+        }
+
+        /**
+         * A linker option for portions of the execution state immediately
```
mcimadamore, line 333:
```suggestion
         * A linker option for saving portions of the execution state immediately
```

src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -303,5 +309,71 @@ default MethodHandle downcallHandle(MemorySegment symbol, FunctionDescriptor fun
...
+         * calling a foreign function associated with a downcall method handle,
```
mcimadamore, line 334:
```suggestion
         * after calling a foreign function associated with a downcall method handle,
```

src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -303,5 +309,71 @@ default MethodHandle downcallHandle(MemorySegment symbol, FunctionDescriptor fun
         static Option firstVariadicArg(int index) {
             return new LinkerOptions.FirstVariadicArg(index);
         }
+
+        /**
+         * {@return A linker option used to portions of the execution state immediately after
```
mcimadamore, line 314:
```suggestion
         * {@return A linker option used to save portions of the execution state immediately after
```

mcimadamore (approved):
Looks good - some javadoc nits

JornVernee:
Thanks for spotting those!

PR #746, opened 2022-10-21
8275584: Incorrect stack spilling in CallArranger on MacOS/AArch64 

nick-arm:
This patch adds special handling for argument spilling on M1 Mac which does not follow the standard AArch64.  In the standard ABI arguments are first extended to the full 64-bit register width and then spilled but on macOS the arguments are spilled according to their original width and packed using theior natural.  @JornVernee  did most of the work to support this but there were a few issues remaining related to structspilling which I've fixed up in the last commit here.

---------
### Progress

### Issue
 * [JDK-8275584](https://bugs.openjdk.org/browse/JDK-8275584): Incorrect stack spilling in CallArranger on MacOS/AArch64

### Reviewers

### Contributors
 * Jorn Vernee `<jvernee@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/746/head:pull/746` \
`$ git checkout pull/746`

`$ git checkout pull/746` \
`$ git pull https://git.openjdk.org/panama-foreign pull/746/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 746`

`$ git pr show -t 746`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/746.diff">https://git.openjdk.org/panama-foreign/pull/746.diff</a>

</details>

### DIFF
--- a/src/hotspot/cpu/aarch64/foreignGlobals_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/foreignGlobals_aarch64.cpp
@@ -131,8 +131,6 @@ static void move_stack(MacroAssembler* masm, Register tmp_reg, int in_stk_bias, 
     case StorageType::INTEGER:
       assert(to_reg.segment_mask() == REG64_MASK, "only moves to 64-bit registers supported");
       switch (from_reg.stack_size()) {
-        // FIXME these loads zero upper bits of the register
-        // should we sign extend instead?
         case 8: masm->ldr (as_Register(to_reg), from_addr); break;
         case 4: masm->ldrw(as_Register(to_reg), from_addr); break;
         case 2: masm->ldrh(as_Register(to_reg), from_addr); break;
--- a/src/hotspot/cpu/aarch64/foreignGlobals_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/foreignGlobals_aarch64.cpp
@@ -109,15 +109,16 @@ static void move_reg64(MacroAssembler* masm, int out_stk_bias,
       break;
     case StorageType::STACK:
       out_bias = out_stk_bias;
-    case StorageType::FRAME_DATA:
+    case StorageType::FRAME_DATA: {
+      Address dest(sp, to_reg.index() + out_bias);
       switch (to_reg.stack_size()) {
-        // FIXME use correctly sized stores
-        case 8: case 4: case 2: case 1:
-          masm->str(from_reg, Address(sp, to_reg.offset() + out_stk_bias));
-        break;
+        case 8: masm->str (from_reg, dest); break;
+        case 4: masm->strw(from_reg, dest); break;
+        case 2: masm->strh(from_reg, dest); break;
+        case 1: masm->strb(from_reg, dest); break;
         default: ShouldNotReachHere();
       }
-      break;
+    } break;
     default: ShouldNotReachHere();
   }
 }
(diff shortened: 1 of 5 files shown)


src/hotspot/cpu/aarch64/foreignGlobals_aarch64.cpp
```diff
@@ -130,10 +131,12 @@ static void move_stack(MacroAssembler* masm, Register tmp_reg, int in_stk_bias,
     case StorageType::INTEGER:
       assert(to_reg.segment_mask() == REG64_MASK, "only moves to 64-bit registers supported");
       switch (from_reg.stack_size()) {
-        // FIXME use correctly sized loads
-        case 8: case 4: case 2: case 1:
-          masm->ldr(as_Register(to_reg), from_addr);
-        break;
+        // FIXME these loads zero upper bits of the register
```
nick-arm, line 134: I'm not sure about this: I'll try creating a test next week.

nick-arm: I think this is correct as-is (I tested with spilled negative values).

src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/CallArranger.java
```diff
@@ -271,7 +287,7 @@ protected void spillStructUnbox(Binding.Builder bindings, MemoryLayout layout) {
             while (offset < layout.byteSize()) {
                 long copy = Math.min(layout.byteSize() - offset, STACK_SLOT_SIZE);
                 VMStorage storage =
-                    storageCalculator.stackAlloc(copy, STACK_SLOT_SIZE);
+                    storageCalculator.stackAlloc(copy, layout.byteAlignment(), false);
```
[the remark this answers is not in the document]
JornVernee: I think maybe passing `layout.byteAlignment` here is a little misleading, as it is unconditionally overwritten to `STACK_SLOT_SIZE` in `stackAlloc` any ways.

I know that's a change I made, but I think it might be better to just use `STACK_SLOT_SIZE` as alignment here, and then I don't think the extra `subSlotPacking` parameter is needed either.

src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/CallArranger.java
```diff
@@ -186,21 +186,16 @@ public StorageCalculator(boolean forArguments) {
...
+        VMStorage stackAlloc(long size, long alignment, boolean subSlotPacking) {
             assert forArguments : "no stack returns";
-            // Implementation limit: each arg must take up at least an 8 byte stack slot (on the Java side)
-            // There is currently no way to address stack offsets that are not multiples of 8 bytes
-            // The VM can only address multiple-of-4-bytes offsets, which is also not good enough for some ABIs
-            // see JDK-8283462 and related issues
-            long stackSlotAlignment = Math.max(alignment, STACK_SLOT_SIZE);
+            long stackSlotAlignment = subSlotPacking && !forVarArgs
+                    ? alignment
+                    : Math.max(alignment, STACK_SLOT_SIZE);
```
JornVernee, line 197: With spillStruct(Un)box, and `nextStorageForHFA` now also accounting for sub slot packing, I think that maybe this is not the right level to adjust the alignment (again).

I suggest removing the conditional expression here, and simply using the `alignment` passed to this method in place of `stackSlotAlignment`. The alignment can be adjusted based on `requiresSubSlotStackPacking()` in the other  `stackAlloc(MemorryLayout)` overload instead, which looks to be the only place that doesn't pass a constant boolean value. Then, `nextStorageForHFA` can also just call that overload for each member layout.

JornVernee (approved):
Thanks again for fixing.

src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/CallArranger.java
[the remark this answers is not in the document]
nick-arm: > Ok. So it looks like only HFAs are 'packed', and other structs follow the classic approach of passing everything in stack slot sized chunks :/

Yes but I only figured that out by looking at what Clang does. Apple's page at [1] isn't very clear on this.

[1] https://developer.apple.com/documentation/xcode/writing-arm64-code-for-apple-platforms

JornVernee (approved):

nick-arm:
I'll integrate this one now unless there's any more reviews/comments?

JornVernee:
I think you're good to go :)

PR #750, opened 2022-11-04
8296417: Make memory session a pure lifetime abstraction

mcimadamore:
Add description

---------
### Progress

### Issue
 * [JDK-8296417](https://bugs.openjdk.org/browse/JDK-8296417): Make memory session a pure lifetime abstraction

### Reviewers
 * [Jorn Vernee](https://openjdk.org/census#jvernee) (@JornVernee - Committer) ⚠️ Review applies to [0809daa7](https://git.openjdk.org/panama-foreign/pull/750/files/0809daa79b46305f3a8022e65c58d9639924aded)
 * [Per Minborg](https://openjdk.org/census#pminborg) (@minborg - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/750/head:pull/750` \
`$ git checkout pull/750`

`$ git checkout pull/750` \
`$ git pull https://git.openjdk.org/panama-foreign pull/750/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 750`

`$ git pr show -t 750`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/750.diff">https://git.openjdk.org/panama-foreign/pull/750.diff</a>

</details>


src/java.base/share/classes/java/lang/foreign/SegmentAllocator.java
```diff
@@ -345,85 +346,19 @@ default MemorySegment allocate(long byteSize) {
...
-     */
-    static SegmentAllocator newNativeArena(long arenaSize, MemorySession session) {
-        return newNativeArena(arenaSize, arenaSize, session);
-    }
-
-    /**
-     * Creates an arena-based allocator used to allocate native segments. The returned allocator features
-     * the given block size {@code B} and the given arena size {@code A}, and the native segments
-     * it allocates are associated with the provided memory session.
+     * Returns a segment allocator which responds to allocation requests by returning consecutive slices
```
mcimadamore, line 349: Note: newNativeArena is now gone, and replaced with a more general "slicing allocator". The reason I did this was two-fold: the arena factory seemed to be fairly specific, and not very useful to build upon (slicing allocator greatly improve on that aspect). The second problem is the name: now that we have a toplevel Arena, I think having an arena allocator factory here would be too confusing. I don't think the loss in functionality is significant (newNativeArena can be trivially implemented in terms of a slicing allocator), but if we really need it, we could always add another Arena subclass (e.g. SlicingArena which does more or less what the old allocator did.

src/java.base/share/classes/java/lang/foreign/Arena.java
```diff
@@ -0,0 +1,121 @@
...
+ *     MemorySegment nativeString = arena.allocateUtf8String("Hello!");
+ *     MemorySegment upcallStub = linker.upcallStub(handle, desc, arena.session());
+ *     ...
+ * } // memory released here
+ *}
+ *
+ * @since 20
+ */
+@PreviewFeature(feature=PreviewFeature.Feature.FOREIGN)
+public sealed interface Arena extends SegmentAllocator, AutoCloseable permits ArenaImpl { //@@@: This can also be non-sealed!
```
rsmogura, line 59: I wonder if this have to be sealed? Having this unsealed would allow creating custom arenas.

(I see that this comment is already in code)

src/java.base/share/classes/java/lang/foreign/Arena.java
```diff
@@ -0,0 +1,121 @@
...
+    /**
+     * Creates a native memory segment with the given size (in bytes), alignment constraint (in bytes) associated with
+     * this memory session. The {@link MemorySegment#address()} of the returned memory segment is the starting address of
+     * the newly allocated off-heap memory region backing the segment. Moreover, the {@linkplain MemorySegment#address() address}
+     * of the returned segment will be aligned according the provided alignment constraint.
+     * <p>
+     * Clients are responsible for ensuring that this arena is closed when the segments returned by this method are no
+     * longer in use. Failure to do so will result in off-heap memory leaks.
+     * <p>
+     * This is equivalent to the following code:
```
rsmogura, line 70: If we would allow creating custom arenas than this statement may not be true (ie. of pooling arena)

test/jdk/java/foreign/TestMemorySession.java
```diff
@@ -249,83 +224,25 @@ public void testCloseConfinedLock() {
...
                 acquireRecursive(session, acquireCount - 1);
             }
-            if (session.isCloseable()) {
-                assertThrows(IllegalStateException.class, session::close);
-            }
-        }
-    }
-
-    @Test
-    public void testConfinedSessionWithImplicitDependency() {
```
JornVernee, line 275: Is it not possible to keep these 2 tests? Might be nice as a stress test for a naked acquire.

src/java.base/share/classes/java/lang/foreign/MemorySession.java
```diff
@@ -25,41 +25,42 @@
...
 
 /**
- * A memory session manages the lifecycle of one or more resources. Resources (e.g. {@link MemorySegment}) associated
- * with a memory session can only be accessed while the memory session is {@linkplain #isAlive() alive},
- * and by the {@linkplain #ownerThread() thread} associated with the memory session (if any).
- * <p>
- * Memory sessions can be closed. When a memory session is closed, it is no longer {@linkplain #isAlive() alive},
- * and subsequent operations on resources associated with that session (e.g. attempting to access a {@link MemorySegment} instance)
- * will fail with {@link IllegalStateException}.
+ * A memory session manages the lifecycle of memory segments.
```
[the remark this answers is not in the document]
mcimadamore: Not sure I agree - in the sense that a session is still the lifecycle of a segment. I'll defer for now.

test/jdk/java/foreign/TestSegmentAllocators.java
```diff
@@ -498,9 +480,8 @@ public double[] toArray(MemorySegment segment, ValueLayout layout) {
     @DataProvider(name = "allocators")
     static Object[][] allocators() {
         return new Object[][] {
-                { SegmentAllocator.implicitAllocator() },
```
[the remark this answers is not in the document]
mcimadamore: yes, but except for the different session, it's not like the test is going to stress a different allocation implementation?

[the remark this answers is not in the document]
JornVernee: Yeah, that's true

test/jdk/java/foreign/TestSegmentAllocators.java
```diff
@@ -87,16 +85,14 @@ public <Z, L extends ValueLayout> void testAllocation(Z value, AllocationFactory
                     try {
                         allocationFunction.allocate(allocator, alignedLayout, value);
                         assertFalse(isBound);
-                    } catch (OutOfMemoryError ex) {
+                    } catch (IndexOutOfBoundsException | IllegalStateException ex) {
                         //failure is expected if bound
                         assertTrue(isBound);
```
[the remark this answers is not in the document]
mcimadamore: You are correct in the sense that slicing allocator returns NULL when it cannot allocate. However, if we allocate using e.g. `SegmentAllocator.allocate(ValueLayout.OfInt, int)` there will be issues because the returned segment has zero size and we will be trying to set a value on it. I think returning NULL here is perhaps too clever, we should just throw IOOBE on the allocate method - agree?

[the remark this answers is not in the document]
JornVernee: Ah, I see. But, returning `NULL` seems nice for composing with other allocators. I guess you'd need to overwrite all the methods of SlicingAllocator to forward the `NULL` returned by the base `allocate`.

If you go for throwing an exception, I suggest sticking with OOME though.

test/micro/org/openjdk/bench/java/lang/foreign/StrLenTest.java
```diff
@@ -154,17 +155,17 @@ public RingAllocator(MemorySession session) {
...
         public MemorySegment allocate(long byteSize, long byteAlignment) {
-            MemorySegment res = current.allocate(byteSize, byteAlignment);
-            if (res.equals(MemorySegment.NULL)) {
-                // reached the end of the segment, start again
+            if (rem < byteSize) {
                 reset();
-                res = current.allocate(byteSize, byteAlignment);
             }
+            MemorySegment res = current.allocate(byteSize, byteAlignment);
+            rem -= segment.segmentOffset(res);
```
JornVernee, line 162: I think this should be `rem = segment.byteSize() - segment.segmentOffset(res);` right? Otherwise more and more is subtracted on subsequent allocations.

mcimadamore: good catch!

src/java.base/share/classes/java/lang/foreign/Arena.java
```diff
@@ -0,0 +1,133 @@
...
+     * {@linkplain MemorySession#isOwnedBy(Thread) owning} the session associated with this arena.
+     * @see MemorySegment#allocateNative(long, long, MemorySession)
+     */
+    @Override
+    default MemorySegment allocate(long byteSize, long byteAlignment) {
+        return MemorySegment.allocateNative(byteSize, byteAlignment, session());
+    }
+
+    /**
+     * {@return the session associated with this arena}
```
rsmogura, line 89: Minor, curly bracers around `@return`

src/java.base/share/classes/java/lang/foreign/SegmentAllocator.java
```diff
@@ -345,85 +346,19 @@ default MemorySegment allocate(long byteSize) {
...
-     * </ul>
-     * <p>
-     * This segment allocator can be useful when clients want to perform multiple allocation requests while avoiding the
-     * cost associated with allocating a new off-heap memory region upon each allocation request.
-     * <p>
-     * The returned allocator might throw an {@link OutOfMemoryError} if the total memory allocated with this allocator
-     * exceeds the arena size {@code A}, or the system capacity. Furthermore, the returned allocator is not thread safe.
-     * Concurrent allocation needs to be guarded with synchronization primitives.
+     * When the returned allocator cannot satisfy an allocation request, e.g. because a slice of the provided
+     * segment with the requested size cannot be found, an {@link IndexOutOfBoundsException} is thrown.
```
minborg, line 354: This type of exception appears a bit strange to me.

src/java.base/share/classes/java/lang/foreign/Arena.java
mcimadamore: That's the inline javadoc return tag?

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
```diff
@@ -1101,121 +1133,115 @@ static MemorySegment ofAddress(long address, long byteSize) {
...
-     * is not a power of 2.
-     * @see MemorySession#allocate(long, long)
+     *                                  is not a power of 2.
+     * @throws IllegalStateException if {@code session} is not {@linkplain MemorySession#isAlive() alive}.
+     * @throws WrongThreadException if this method is called from a thread other than the thread
+     * {@linkplain MemorySession#isOwnedBy(Thread) owning} {@code session}.
      */
-    static MemorySegment allocateNative(long byteSize, long byteAlignment) {
+    static MemorySegment allocateNative(long byteSize, long byteAlignment, MemorySession session) {
+        Objects.requireNonNull(session);
```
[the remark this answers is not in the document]
mcimadamore: Some of our tests (TestNulls) rely on nulls being checked first. I'm ok with swapping when there are two null checks (as in `allocateNative(MemoryLayout, MemorySession)`).

src/java.base/share/classes/java/lang/foreign/SegmentAllocator.java
mcimadamore: Given that the implementation of this method is based on `MemorySegment::asSlice`, I think the exception type is not too bad.

JornVernee (approved):

test/micro/org/openjdk/bench/java/lang/foreign/StrLenTest.java
```diff
@@ -142,4 +142,30 @@ static String makeString(int size) {
...
+            reset();
+        }
+
+        @Override
+        public MemorySegment allocate(long byteSize, long byteAlignment) {
+            if (rem < byteSize) {
+                reset();
+            }
+            MemorySegment res = current.allocate(byteSize, byteAlignment);
+            rem -= segment.byteSize() - segment.segmentOffset(res);
```
JornVernee, line 162:
```suggestion
            rem = segment.byteSize() - segment.segmentOffset(res);
```

minborg (approved):

merykitty:
I think that `Arena` is a bad name. And if I read the document correctly, `MemorySession` is now a bad name, too, since essentially a `MemorySession` only allows read operations, it acts more like a view than a true session. Thanks a lot.

mcimadamore:
> I think that `Arena` is a bad name. And if I read the document correctly, `MemorySession` is now a bad name, too, since essentially a `MemorySession` only allows read operations, it acts more like a view than a true session. Thanks a lot.

Names can be changed. I agree that `MemorySession` seems suboptimal now. We were thinking of other candidates like `SegmentScope`.

That said, I'd like to understand more where the problem is with `Arena`, can you please expand on why you think it's not a great name? Thanks.

merykitty:
@mcimadamore I think `Arena` is a word indicating mainly spacial properties, using it as a pure temporal abstraction seems unrelated and potentially misleading.

Moreover, in memory contexts, an arena is often used to indicate a contiguous piece of memory that can be sliced out to satisfy requested allocations. As a result, using it in for an orthogonal responsibility may be confusing.

Thank you very much.

merykitty:
Personally I think we could use `MemoryContext` or reuse the existing `MemorySession` name instead.

mcimadamore:
> @mcimadamore I think `Arena` is a word indicating mainly spacial properties, using it as a pure temporal abstraction seems unrelated and potentially misleading.

But the pure temporal abstraction is MemorySession, not Arena (hence why you need to go from Arena to a session to e.g. map a segment).

> 
> Moreover, in memory contexts, an arena is often used to indicate a contiguous piece of memory that can be sliced out to satisfy requested allocations. As a result, using it for an orthogonal responsibility may be confusing.

Ok, so your angle is: Arena is an allocator with very specific properties. I think that's fair - although Arena is also an allocator - and preserves a main properties or arena allocators - which is that when the arena is closed, all the memory is invalidated at once. It seems like the fact that the arena is contiguous or not under the hood (which is an impl change which we might in fact improve over time) should not be too much of a distraction?

> 
> Thank you very much.

merykitty:
My view is that an `Arena` is the temporal abstraction of memory, as it can create and free memory. On the other hand, a `MemorySession` is merely a view on its owning `Arena`, allowing queries of the arena's status. An implicit session can be understood as a session that has an implicit `Arena`.

While I agree with your definition of an arena from the theoretical point of view, practically, an arena allocator is often used to indicate a more specific kind of allocator and the arena is often used to indicate the underlying segment that gets sliced. For example, protobuf C++ API specifies an Arena as: [1]

> With arena allocation, new objects are allocated out of a large piece of preallocated memory called the arena.

In contrast, in C++ standard, to specify a kind of memory source that "releases the allocated memory only when the resource is destroyed." is called `std::pmr::monotonic_buffer_resource`.

Thanks a lot.

JornVernee:
`Arena` in HotSpot is like `monotonic_buffer_resource`, i.e. a linked list of slabs that grows over time, where the slabs get sliced into pieces to fulfill allocation requests. But in general I understand `Arena` to just be something that groups multiple allocations. The particular allocation strategy seems less important to understanding what an arena is than the grouping property.

PR #753, opened 2022-11-17
8296973: saving errno on a value-returning function crashes the JVM

JornVernee:
There are 2 bugs:

- Frame allocation code is not in the right order. There's an optimization where the area used for spilling return values around slow path calls is shared with the shadow space and stack args area. This is possible since the latter are only used before the native call, and the former only after. But, the order of the code that allocates this space is incorrect, which means that the spill area for return values is shared with _all_ the data in the frame. This also recently includes the address of the captured state segment (for `errno`). In case there are no stack arguments and no shadow space (as on linux in most cases), this means in practice that a spill of the return value will overwrite the capture state address. (previously we've also had the return buffer address, but use of a return buffer and spilling of return values are mutually exclusive, so it was never a problem).
- The `CaptureCallState` linker option adds an additional leading parameter (the captured state segment), which can mess up `SharedUtils::adaptDowncallForIMR`.

I've extended the existing `CaptureCallState` test to include the most interesting cases: simple scalar returns, and various struct returns that are: returned in a single register (no return buffer), returned in multiple registers (yes return buffer), or returned through ABI-level in-memory-return pointers (using `SharedUtils::adaptDowncallForIMR`).

I've also done a big sanity test where I adapted `TestDowncallScope` to also set and capture `errno` with each call. That seemed overkill to include in the repo though (?)

Thanks

---------
### Progress

### Issue
 * [JDK-8296973](https://bugs.openjdk.org/browse/JDK-8296973): saving errno on a value-returning function crashes the JVM

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/753/head:pull/753` \
`$ git checkout pull/753`

`$ git checkout pull/753` \
`$ git pull https://git.openjdk.org/panama-foreign pull/753/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 753`

`$ git pr show -t 753`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/753.diff">https://git.openjdk.org/panama-foreign/pull/753.diff</a>

</details>

### DIFF
--- a/src/hotspot/cpu/aarch64/downcallLinker_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/downcallLinker_aarch64.cpp
@@ -161,6 +161,20 @@ void DowncallStubGenerator::generate() {
   assert(_abi._shadow_space_bytes == 0, "not expecting shadow space on AArch64");
   allocated_frame_size += arg_shuffle.out_arg_bytes();
 
+  bool should_save_return_value = !_needs_return_buffer;
+  RegSpiller out_reg_spiller(_output_registers);
+  int spill_offset = -1;
+
+  if (should_save_return_value) {
+    spill_offset = 0;
+    // spill area can be shared with shadow space and out args,
+    // since they are only used before the call,
+    // and spill area is only used after.
+    allocated_frame_size = out_reg_spiller.spill_size_bytes() > allocated_frame_size
+      ? out_reg_spiller.spill_size_bytes()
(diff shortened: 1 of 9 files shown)


mcimadamore (approved):
Looks good - testing coverage is also reasonable. Let's see how it goes and decide later if we need a more comprehensive test.

PR #755, opened 2022-12-02
Move Linux & MacOs CallArranger tests into separate files

swesonga:
This separation was suggested in https://github.com/openjdk/panama-foreign/pull/754#discussion_r1032658310


JornVernee (approved):
LGTM, thanks for taking care of this

PR #758, opened 2022-12-09
Update foreign memory doc

mcimadamore:
This patch updates the foreign memory document to reflect the latest API changes in Java 20.

I've streamlined the document, removed some obsolete sections (as well as sections that didn't seem to add much), and re-ordered the various sections so that the narrative is now easier to follow.

---------
### Progress

### Reviewers
 * [Jorn Vernee](https://openjdk.org/census#jvernee) (@JornVernee - Committer) ⚠️ Review applies to [84093026](https://git.openjdk.org/panama-foreign/pull/758/files/840930264d4d8442db05e4396ed9383706599254)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/758/head:pull/758` \
`$ git checkout pull/758`

`$ git checkout pull/758` \
`$ git pull https://git.openjdk.org/panama-foreign pull/758/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 758`

`$ git pr show -t 758`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/758.diff">https://git.openjdk.org/panama-foreign/pull/758.diff</a>

</details>


doc/panama_memaccess.md
```diff
@@ -1,31 +1,88 @@
...
+
+Most programs, though, require off-heap memory to be deallocated  while the program is running, and thus need memory segments with *bounded* lifetimes. The simplest way to obtain a segment with bounded lifetime is to use an *automatic scope*:
+
+```java
+MemorySegment segment = MemorySegment.allocateNative(100, SegmentScope.auto());
+```
+
+Segments associated with an automatic scope are alive as long as they are determined to be reachable by the garbage collector. In other words, the above snippet creates a native segment whose behavior closely matches that of a `ByteBuffer` allocated with the `allocateDirect` factory.
+
+There are cases, however, where automatic deallocation is not enough: consider the case where a large memory segment is mapped from a file (this is possible using `MemorySegment::map`); in this case, an application would probably prefer to release (e.g. `unmap`) the memory associated with this segment in a *deterministic* fashion, to ensure that memory doesn't remain available for longer than in needs to.
```
JornVernee, line 29:
```suggestion
There are cases, however, where automatic deallocation is not enough: consider the case where a large memory segment is mapped from a file (this is possible using `MemorySegment::map`); in this case, an application would probably prefer to release (e.g. `unmap`) the memory associated with this segment in a *deterministic* fashion, to ensure that memory doesn't remain available for longer than it needs to.
```

doc/panama_memaccess.md
```diff
@@ -1,31 +1,88 @@
...
+```
+try (Arena arena = Arena.openConfined()) {
+    MemorySegment segment1 = MemorySegment.allocateNative(100, arena.scope());
+    MemorySegment segment2 = MemorySegment.allocateNative(100, arena.scope());
+    ...
+    MemorySegment segmentN = MemorySegment.allocateNative(100, arena.scope());
+} // all segments are deallocated here
+```
+
+When the arena is closed (above, this is done with the *try-with-resources* construct) the arena scope is no longer alive, all the segments associated with it are invalidated,  and the regions of memory backing the segments are deallocated  atomically.
```
JornVernee, line 42: We don't deallocate atomically, but we invalidate atomically. So I think this should be:

```suggestion
When the arena is closed (above, this is done with the *try-with-resources* construct) the arena scope is no longer alive, all the segments associated with it are invalidated atomically, and the regions of memory backing the segments are deallocated.
```

doc/panama_memaccess.md
```diff
@@ -1,31 +1,88 @@
...
+
+### Thread-confinement
+
+Arenas provide a strong temporal safety guarantee: memory segments associated with an arena scope cannot be accessed after the arena is closed, since the arena scope is no longer alive. If an arena is opened and closed by the same thread, and all memory segments allocated with the arena's scope are accessed only  by that thread, then ensuring correctness is straightforward. However, if memory segments allocated with the arena scope are accessed by  multiple threads then ensuring correctness is complex. For example, a segment allocated with arena scope might be accessed by one thread while another thread attempts to close the arena. To guarantee temporal safety without making single-threaded clients pay undue cost, there are  two kinds of arenas: *confined* and *shared*.
```
JornVernee, line 46:
```suggestion
Arenas provide a strong temporal safety guarantee: memory segments associated with an arena scope cannot be accessed after the arena is closed, since the arena scope is no longer alive. If an arena is opened and closed by the same thread, and all memory segments allocated with the arena's scope are accessed only  by that thread, then ensuring correctness is straightforward. However, if memory segments allocated with the arena scope are accessed by  multiple threads then ensuring correctness is complex. For example, a segment allocated with an arena scope might be accessed by one thread while another thread attempts to close the arena. To guarantee temporal safety without making single-threaded clients pay undue cost, there are  two kinds of arenas: *confined* and *shared*.
```

doc/panama_memaccess.md
```diff
@@ -102,143 +145,23 @@ In the above, `xHandle` and `yHandle` are two var handle instances whose type is
...
+The attentive reader might have noted how rich the var handles obtains from memory layouts are, compared to the simple memory access var handle we have constructed here. How do we go from a simple access var handle that takes a byte offset to a var handle that can dereference a complex layout path? The answer is, by using var handle *combinators*. Developers familiar with the method handles know how simpler method handles can be combined into more complex ones using the various combinator methods in the `MethodHandles` class. These methods allow, for instance, to insert (or bind) arguments into a target method handle, filter return values, permute arguments and much more.
```
JornVernee, line 162:
```suggestion
The attentive reader might have noted how rich the var handles obtained from memory layouts are, compared to the simple memory access var handle we have constructed here. How do we go from a simple access var handle that takes a byte offset to a var handle that can dereference a complex layout path? The answer is, by using var handle *combinators*. Developers familiar with the method handles know how simpler method handles can be combined into more complex ones using the various combinator methods in the `MethodHandles` class. These methods allow, for instance, to insert (or bind) arguments into a target method handle, filter return values, permute arguments and much more.
```

doc/panama_memaccess.md
```diff
@@ -253,33 +176,33 @@ long offsetOfY = points.byteOffset(PathElement.sequenceElement(3), PathElement.g
...
+The FFM API provides basic safety guarantees for all memory segments created using the API. More specifically, a memory dereference operation should either succeed, or result in a runtime exception — but, crucially, should never result in a VM crash, or, more subtly, in memory corruption occurring *outside* the region of memory associated with a memory segment. This is indeed the case, as all memory segments feature immutable *spatial bounds*, and, as we have seen, are associated with a segment scope which make sure that segments cannot be dereferenced after their backing region of memory have been deallocated.
```
JornVernee, line 183:
```suggestion
The FFM API provides basic safety guarantees for all memory segments created using the API. More specifically, a memory dereference operation should either succeed, or result in a runtime exception — but, crucially, should never result in a VM crash, or, more subtly, in memory corruption occurring *outside* the region of memory associated with a memory segment. This is indeed the case, as all memory segments feature immutable *spatial bounds*, and, as we have seen, are associated with a segment scope which make sure that segments cannot be dereferenced after their backing regions of memory have been deallocated.
```

doc/panama_memaccess.md
```diff
@@ -253,33 +176,33 @@ long offsetOfY = points.byteOffset(PathElement.sequenceElement(3), PathElement.g
...
-* <a id="2"/>(<sup>2</sup>):<small> Clients can enforce stricter type checking when interacting with `VarHandle` instances, by obtaining an *exact* var handle, using the `VarHandle::withInvokeExactBehavior` method.</small>
-* <a id="3"/>(<sup>3</sup>):<small> Shared sessions rely on VM thread-local handshakes (JEP [312](https://openjdk.java.net/jeps/312)) to implement lock-free, safe, shared memory access; that is, when it comes to memory access, there should no difference in performance between a shared segment and a confined segment. On the other hand, `MemorySession::close` might be slower on shared sessions than on confined ones.</small>
+* <a id="1"/>(<sup>1</sup>):<small> Shared arenas rely on VM thread-local handshakes (JEP [312](https://openjdk.java.net/jeps/312)) to implement lock-free, safe, shared memory access; that is, when it comes to memory access, there should no difference in performance between a shared segment and a confined segment. On the other hand, `Arena::close` might be slower on shared arenas than on confined ones.</small>
```
JornVernee, line 205:
```suggestion
* <a id="1"/>(<sup>1</sup>):<small> Shared arenas rely on VM thread-local handshakes (JEP [312](https://openjdk.java.net/jeps/312)) to implement lock-free, safe, shared memory access; that is, when it comes to memory access, there should be no difference in performance between a shared segment and a confined segment. On the other hand, `Arena::close` might be slower on shared arenas than on confined ones.</small>
```

JornVernee (approved):

PR #760, opened 2022-12-13
8298532: Declare private constructors for FFM internal Architecture implementations

minborg:
This PR proposes declaring AArch64Architecture and X86_64Architecture final and creating a private constructor so that this redundant byte code can be eliminated:

```
 public jdk.internal.foreign.abi.x64.X86_64Architecture(); 
    Code: 
       0: aload_0 
       1: invokespecial #1 // Method java/lang/Object."<init>":()V 
       4: return 

 public jdk.internal.foreign.abi.aarch64.AArch64Architectur(); 
    Code: 
       0: aload_0 
       1: invokespecial #1 // Method java/lang/Object."<init>":()V 
       4: return 
```

and other potential optimizations can be unlocked.

---------
### Progress

### Issue
 * [JDK-8298532](https://bugs.openjdk.org/browse/JDK-8298532): Declare private constructors for FFM internal Architecture implementations

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/760/head:pull/760` \
`$ git checkout pull/760`

`$ git checkout pull/760` \
`$ git pull https://git.openjdk.org/panama-foreign pull/760/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 760`

`$ git pr show -t 760`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/760.diff">https://git.openjdk.org/panama-foreign/pull/760.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/AArch64Architecture.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/AArch64Architecture.java
@@ -40,7 +40,7 @@ public final class AArch64Architecture implements Architecture {
     private static final int VECTOR_REG_SIZE = 16;
 
     // Suppresses default constructor, ensuring non-instantiability.
-    public AArch64Architecture() {}
+    private AArch64Architecture() {}
 
     @Override
     public boolean isStackType(int cls) {
--- a/src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/AArch64Architecture.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/AArch64Architecture.java
@@ -30,7 +30,7 @@ import jdk.internal.foreign.abi.Architecture;
 import jdk.internal.foreign.abi.StubLocations;
 import jdk.internal.foreign.abi.VMStorage;
 
-public class AArch64Architecture implements Architecture {
+public final class AArch64Architecture implements Architecture {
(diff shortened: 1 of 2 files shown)


src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/AArch64Architecture.java
```diff
@@ -39,6 +39,9 @@ public class AArch64Architecture implements Architecture {
     private static final int INTEGER_REG_SIZE = 8;
     private static final int VECTOR_REG_SIZE = 16;
 
+    // Suppresses default constructor, ensuring non-instantiability.
+    public AArch64Architecture() {}
```
JornVernee, line 43:
```suggestion
    private AArch64Architecture() {}
```

JornVernee (approved):

mcimadamore:
Just a quick clarification: this change does not alter the shape of the bytecode. That is, a constructor is still emitted by javac (with the corresponding supercall to `Object::<init>`). The only difference is that the constructor is now `private`, so it doesn't show up with a simple `javap` or even with `javap -v`, because `javap` filters out non-public members by default. To show all members, `javap -p -v` should be used instead, which would then reveal the private constructor.

That said, this change makes the nature of the class (as a bag of statics) more explicit, so I still support it.

PR #762, opened 2023-01-02
Add benchmarks to MemorySegmentVsBits

minborg:
This PR proposes the addition of some benchmarks, for example using a LonBuffer and a VarHandle.

### DIFF
--- a/test/micro/org/openjdk/bench/java/lang/foreign/MemorySegmentVsBits.java
+++ b/test/micro/org/openjdk/bench/java/lang/foreign/MemorySegmentVsBits.java
@@ -38,10 +38,12 @@ import org.openjdk.jmh.annotations.TearDown;
 import org.openjdk.jmh.annotations.Warmup;
 
 import java.lang.foreign.Arena;
-import java.lang.foreign.MemoryLayout;
 import java.lang.foreign.MemorySegment;
 import java.lang.foreign.ValueLayout;
-import java.nio.ByteOrder;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.VarHandle;
+import java.nio.ByteBuffer;
+import java.nio.LongBuffer;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.TimeUnit;
 
@@ -62,12 +64,17 @@ import static java.nio.ByteOrder.LITTLE_ENDIAN;
 @Fork(value = 3, jvmArgsAppend = {"--enable-native-access=ALL-UNNAMED", "--enable-preview"})
 public class MemorySegmentVsBits {
 
+    public static final VarHandle LONG_ARRAY_VH = MethodHandles.byteArrayViewVarHandle(long[].class, LITTLE_ENDIAN);
+
     Arena arena = Arena.openConfined();
 
     @Param({"1", "2", "16", "64", "256"})
     public int size;
     private long[] longs;
     private byte[] bytes;
+
+    private ByteBuffer byteBuffer;
+    private LongBuffer longBuffer;
     private MemorySegment segment;
     private MemorySegment nativeSegment;
 
@@ -79,6 +86,8 @@ public class MemorySegmentVsBits {
     public void setup() {
         longs = ThreadLocalRandom.current().longs(size).toArray();
         bytes = new byte[size * Long.BYTES];
+        byteBuffer = ByteBuffer.wrap(bytes);
+        longBuffer = byteBuffer.asLongBuffer();
         segment = MemorySegment.ofArray(bytes);
         nativeSegment = arena.allocate(size * Long.BYTES);
     }
@@ -92,6 +101,25 @@ public class MemorySegmentVsBits {
     public void bitsEquivalent() {
         for (int i = 0; i < size; i++) {
             putLong(bytes, i * Long.BYTES, longs[i]);
+        }
+    }
+    @Benchmark
+    public void byteVarHandle() {
+        for (int i = 0; i < size; i++) {
+            LONG_ARRAY_VH.set(bytes, i * Long.BYTES, longs[i]);
+        }
+    }
+    @Benchmark
+    public void byteBuffer() {
+        for (int i = 0; i < size; i++) {
+            byteBuffer.putLong(i * Long.BYTES, longs[i]);
+        }
+    }
+
+    @Benchmark
+    public void longBuffer() {
+        for (int i = 0; i < size; i++) {
+            longBuffer.put(i, longs[i]);
         }
     }
 
@@ -126,17 +154,6 @@ public class MemorySegmentVsBits {
         b[off + 2] = (byte) (val >>> 40);
         b[off + 1] = (byte) (val >>> 48);
         b[off] = (byte) (val >>> 56);
-    }
-
-    static long getLong(byte[] b, int off) {
-        return ((b[off + 7] & 0xFFL)) +
-                ((b[off + 6] & 0xFFL) << 8) +
-                ((b[off + 5] & 0xFFL) << 16) +
-                ((b[off + 4] & 0xFFL) << 24) +
-                ((b[off + 3] & 0xFFL) << 32) +
-                ((b[off + 2] & 0xFFL) << 40) +
-                ((b[off + 1] & 0xFFL) << 48) +
-                (((long) b[off]) << 56);
     }
 
 }


uschindler:
Thanks cool. When you have results, just show them. I am interested!

minborg:
Here are the results when run on a MacBook Pro (16-inch, 2021) M1 Max and macOS 12.6.1:

```
Benchmark                                  (size)  Mode  Cnt    Score   Error  Units
MemorySegmentVsBits.bitsEquivalent              1  avgt   30    2.618 ? 0.023  ns/op
MemorySegmentVsBits.bitsEquivalent              2  avgt   30    3.595 ? 0.014  ns/op
MemorySegmentVsBits.bitsEquivalent             16  avgt   30   20.005 ? 0.092  ns/op
MemorySegmentVsBits.bitsEquivalent             64  avgt   30   81.832 ? 0.467  ns/op
MemorySegmentVsBits.bitsEquivalent            256  avgt   30  323.792 ? 5.287  ns/op

MemorySegmentVsBits.byteBuffer                  1  avgt   30    2.505 ? 0.011  ns/op
MemorySegmentVsBits.byteBuffer                  2  avgt   30    3.042 ? 0.025  ns/op
MemorySegmentVsBits.byteBuffer                 16  avgt   30    7.035 ? 0.026  ns/op
MemorySegmentVsBits.byteBuffer                 64  avgt   30   16.773 ? 0.084  ns/op
MemorySegmentVsBits.byteBuffer                256  avgt   30   51.042 ? 0.347  ns/op

MemorySegmentVsBits.byteVarHandle               1  avgt   30    1.950 ? 0.022  ns/op
MemorySegmentVsBits.byteVarHandle               2  avgt   30    2.289 ? 0.013  ns/op
MemorySegmentVsBits.byteVarHandle              16  avgt   30    5.767 ? 0.345  ns/op
MemorySegmentVsBits.byteVarHandle              64  avgt   30   12.427 ? 0.138  ns/op
MemorySegmentVsBits.byteVarHandle             256  avgt   30   27.093 ? 0.129  ns/op

MemorySegmentVsBits.longBuffer                  1  avgt   30    2.252 ? 0.018  ns/op
MemorySegmentVsBits.longBuffer                  2  avgt   30    2.633 ? 0.023  ns/op
MemorySegmentVsBits.longBuffer                 16  avgt   30    5.637 ? 0.359  ns/op
MemorySegmentVsBits.longBuffer                 64  avgt   30   10.917 ? 0.044  ns/op
MemorySegmentVsBits.longBuffer                256  avgt   30   27.627 ? 0.115  ns/op

MemorySegmentVsBits.panamaHeap                  1  avgt   30    3.464 ? 0.014  ns/op
MemorySegmentVsBits.panamaHeap                  2  avgt   30    4.554 ? 0.028  ns/op
MemorySegmentVsBits.panamaHeap                 16  avgt   30    6.963 ? 0.033  ns/op
MemorySegmentVsBits.panamaHeap                 64  avgt   30   16.057 ? 0.082  ns/op
MemorySegmentVsBits.panamaHeap                256  avgt   30   44.275 ? 0.406  ns/op

MemorySegmentVsBits.panamaNative                1  avgt   30    3.495 ? 0.044  ns/op
MemorySegmentVsBits.panamaNative                2  avgt   30    4.523 ? 0.042  ns/op
MemorySegmentVsBits.panamaNative               16  avgt   30    6.997 ? 0.025  ns/op
MemorySegmentVsBits.panamaNative               64  avgt   30   17.223 ? 0.109  ns/op
MemorySegmentVsBits.panamaNative              256  avgt   30   31.506 ? 0.149  ns/op

MemorySegmentVsBits.panamaNativeUnaligned       1  avgt   30    3.255 ? 0.019  ns/op
MemorySegmentVsBits.panamaNativeUnaligned       2  avgt   30    4.302 ? 0.018  ns/op
MemorySegmentVsBits.panamaNativeUnaligned      16  avgt   30    6.965 ? 0.046  ns/op
MemorySegmentVsBits.panamaNativeUnaligned      64  avgt   30   17.216 ? 0.086  ns/op
MemorySegmentVsBits.panamaNativeUnaligned     256  avgt   30   31.223 ? 0.130  ns/op
```

minborg:
I've run some benchmarks on Linux and Windows too and in particular, Linux seams to benefit from using `VarHandle` access:


image

uschindler (approved):

test/micro/org/openjdk/bench/java/lang/foreign/MemorySegmentVsBits.java
```diff
@@ -94,6 +103,25 @@ public void bitsEquivalent() {
             putLong(bytes, i * Long.BYTES, longs[i]);
         }
     }
+    @Benchmark
+    public void byteVarHandle() {
+        for (int i = 0; i < size; i++) {
+            LONG_ARRAY_VH.set(bytes, i * Long.BYTES, longs[i]);
```
mcimadamore, line 109: Is this offset multiplication correct? e.g. I would assume that var handles created with `byteArrayView` would take a logical index?

mcimadamore (commented):
Looks good. It seems like the memory segment implementation breaks even at size = 16?
On top of my head I can't explain why the plain var handle (`byteVarHandle`) is so much faster than anything else. 1but perhaps the benchmark is flawed there?

test/micro/org/openjdk/bench/java/lang/foreign/MemorySegmentVsBits.java
mcimadamore: Nevermind - that seems to be correct - it's equivalent to `ByteBuffer.getLong`

mcimadamore: Might be worth trying this benchmark with the VM flag: `-Djava.lang.invoke.VarHandle.VAR_HANDLE_GUARDS=false`, to see if VH guards are contributing to a major speedup here.

mcimadamore (approved):

uschindler:
> Looks good. It seems like the memory segment implementation breaks even at size = 16? On top of my head I can't explain why the plain var handle (`byteVarHandle`) is so much faster than anything else. 1but perhaps the benchmark is flawed there?

That's good! We use it everywhere in Lucene when getting long/int/short out of byte arrays :-) But I agree, the access for memory segments should be as fast as basically it is also just a VarHandle. Maybe the guards are indeed the problem.

I think, to replace `java.nio.Bits` in the JDK I would use the plain simple VarHandle instead of wrapping they bytes with a MemorySegment or a ByteBuffer. I could provide a simple PR for the JDK that changes the NIO Bits methods by delgating to the mentioned VarHandle :-). But that's unrelated to Panama, I know :-)

minborg:
> > Looks good. It seems like the memory segment implementation breaks even at size = 16? On top of my head I can't explain why the plain var handle (`byteVarHandle`) is so much faster than anything else. 1but perhaps the benchmark is flawed there?
> 
> That's good! We use it everywhere in Lucene when getting long/int/short out of byte arrays :-) But I agree, the access for memory segments should be as fast as basically it is also just a VarHandle. Maybe the guards are indeed the problem.
> 
> I think, to replace `java.nio.Bits` in the JDK I would use the plain simple VarHandle instead of wrapping they bytes with a MemorySegment or a ByteBuffer. I could provide a simple PR for the JDK that changes the NIO Bits methods by delgating to the mentioned VarHandle :-). But that's unrelated to Panama, I know :-)

I have a solution in the making of this but it would be great if you could review it @uschindler

uschindler:
B ythe way, the comment in the benchmark is wrong. The methods are in `java.io.Bits` not `java.nio.Bits`.

test/micro/org/openjdk/bench/java/lang/foreign/MemorySegmentVsBits.java
uschindler: > Is this offset multiplication correct? e.g. I would assume that var handles created with `byteArrayView` would take a logical index?

No the coordinate of those handles is always raw byte offset. It is a bit confusing in the documentation as it also tells you about unaligned accesses.

https://docs.oracle.com/en/java/javase/19/docs/api/java.base/java/lang/invoke/MethodHandles.html#byteArrayViewVarHandle(java.lang.Class,java.nio.ByteOrder) says:

"Produces a VarHandle giving access to elements of a byte[] array viewed as if it were a different primitive array type, such as int[] or long[]. The VarHandle's variable type is the component type of viewArrayClass and the list of coordinate types is (byte[], int), **where the int coordinate type corresponds to an argument that is an index into a byte[] array. The returned VarHandle accesses bytes at an index in a byte[] array, composing bytes to or from a value of the component type of viewArrayClass according to the given endianness.**"

PaulSandoz:
> Looks good. It seems like the memory segment implementation breaks even at size = 16? On top of my head I can't explain why the plain var handle (`byteVarHandle`) is so much faster than anything else. 1but perhaps the benchmark is flawed there?

I think we need to look at the generated assembler code to get some more clues as to why.

Paul.

mcimadamore:
> > Looks good. It seems like the memory segment implementation breaks even at size = 16? On top of my head I can't explain why the plain var handle (`byteVarHandle`) is so much faster than anything else. 1but perhaps the benchmark is flawed there?
> 
> I think we need to look at the generated assembler code to get some more clues as to why.
> 
> Paul.

Did some analysis - on my machine I see evident signs of vectorization in the byte var handle benchmark. Vectorization doesn't seem to be there in other cases (and the resulting code is also much bigger).

E.g. memory segment loop:

```
0x00007f9598837d80:   mov    %ebx,%r8d
  0x00007f9598837d83:   movslq %r8d,%r10
  0x00007f9598837d86:   add    (%rsp),%r10
  0x00007f9598837d8a:   mov    0x48(%r9,%r10,8),%r11
  0x00007f9598837d8f:   mov    0x40(%r9,%r10,8),%rcx
  0x00007f9598837d94:   mov    0x38(%r9,%r10,8),%rbx
  0x00007f9598837d99:   mov    0x30(%r9,%r10,8),%rdx
  0x00007f9598837d9e:   mov    0x28(%r9,%r10,8),%rsi
  0x00007f9598837da3:   mov    0x20(%r9,%r10,8),%rax
  0x00007f9598837da8:   mov    0x18(%r9,%r10,8),%rbp
  0x00007f9598837dad:   mov    0x10(%r9,%r10,8),%r13
  0x00007f9598837db2:   lea    0x0(,%r8,8),%r10d
  0x00007f9598837dba:   movslq %r10d,%r10
  0x00007f9598837dbd:   mov    %rdi,%r14
  0x00007f9598837dc0:   add    %r10,%r14
  0x00007f9598837dc3:   mov    %r13,(%r14)
  0x00007f9598837dc6:   mov    %rbp,0x8(%r14)
  0x00007f9598837dca:   mov    %rax,0x10(%r14)
  0x00007f9598837dce:   mov    %rsi,0x18(%r14)
  0x00007f9598837dd2:   mov    %rdx,0x20(%r14)
  0x00007f9598837dd6:   mov    %rbx,0x28(%r14)
  0x00007f9598837dda:   mov    %rcx,0x30(%r14)
  0x00007f9598837dde:   mov    %r11,0x38(%r14)
  0x00007f9598837de2:   lea    0x8(%r8),%ebx
  0x00007f9598837de6:   cmp    0x14(%rsp),%ebx
  0x00007f9598837dea:   jl     0x00007f9598837d80
  ```
  
  Byte var handle loop:
  
  ```
  0x00007f19a8837930:   vmovdqu 0x70(%rax,%r8,8),%ymm0
  0x00007f19a8837937:   vmovdqu 0x50(%rax,%r8,8),%ymm1
  0x00007f19a883793e:   vmovdqu 0x10(%rax,%r8,8),%ymm2
  0x00007f19a8837945:   vmovdqu 0x30(%rax,%r8,8),%ymm3
  0x00007f19a883794c:   lea    0x0(,%r8,8),%ebx
  0x00007f19a8837954:   movslq %ebx,%rbx
  0x00007f19a8837957:   vmovdqu %ymm2,0x10(%rsi,%rbx,1)
  0x00007f19a883795d:   vmovdqu %ymm3,0x30(%rsi,%rbx,1)
  0x00007f19a8837963:   vmovdqu %ymm1,0x50(%rsi,%rbx,1)
  0x00007f19a8837969:   vmovdqu %ymm0,0x70(%rsi,%rbx,1)
  0x00007f19a883796f:   add    $0x10,%r8d
  0x00007f19a8837973:   cmp    %edi,%r8d
  0x00007f19a8837976:   jl     0x00007f19a8837930
  ```

**Comment by PaulSandoz** [comment]
> All checks are hoisted out of the hot loop - so in principle the segment code should vectorize as well?

The addressing modes of loads from the long array and stores to the underlying memory are slightly different, but connected via the loop variable. I wonder if additional layers involved with memory segments means C2 is loosing that information, which makes the auto-vectorizer bail out.

**Comment by mcimadamore** [comment]
> > All checks are hoisted out of the hot loop - so in principle the segment code should vectorize as well?
> 
> The addressing modes of loads from the long array and stores to the underlying memory are slightly different, but connected via the loop variable. I wonder if additional layers involved with memory segments means C2 is loosing that information, which makes the auto-vectorizer bail out.

I thought of that - but not sure why the buffer benchmark is also slow - heap buffers are implemented using array access aren't they?

**Comment by mcimadamore** [comment]
> > > All checks are hoisted out of the hot loop - so in principle the segment code should vectorize as well?
> > 
> > 
> > The addressing modes of loads from the long array and stores to the underlying memory are slightly different, but connected via the loop variable. I wonder if additional layers involved with memory segments means C2 is loosing that information, which makes the auto-vectorizer bail out.
> 
> I thought of that - but not sure why the buffer benchmark is also slow - heap buffers are implemented using array access aren't they?

Nevermind - in both ByteBuffer and LongBuffer cases, access should occur with Unsafe::getLongUnaligned. Which makes the difference even more puzzling (since the former seems slower than the latter).

**Comment by uschindler** [comment]
> All checks are hoisted out of the hot loop - so in principle the segment code should vectorize as well?

Hah, this may be the reason for some code being slower in Lucene under some circumstances. We have many parts which easily vectorize but some code is still faster when done on standard on-heap arrays, but are consistenly slower with segments and ByteBuffer. This is one reason why we sometimes copy stuff from MMapDirectory to heap, although it would be better to do the calculations off-heap, too (especially vector calculations).

**Comment by minborg** [comment]
It should be noted that there were some errors in the byte order of some variants. These have now been fixed and I am running new performance tests.

**Comment by mcimadamore** [comment]
Tried (for fun) a version which uses bulk copy:

```
Benchmark                                      (size)  Mode  Cnt    Score   Error  Units
MemorySegmentVsBits.bitsEquivalent                  1  avgt   30    4.480 ? 0.074  ns/op
MemorySegmentVsBits.bitsEquivalent                  2  avgt   30    6.183 ? 0.066  ns/op
MemorySegmentVsBits.bitsEquivalent                 16  avgt   30   30.993 ? 0.297  ns/op
MemorySegmentVsBits.bitsEquivalent                 64  avgt   30  118.065 ? 1.362  ns/op
MemorySegmentVsBits.bitsEquivalent                256  avgt   30  469.725 ? 4.120  ns/op
MemorySegmentVsBits.byteBuffer                      1  avgt   30    3.956 ? 0.015  ns/op
MemorySegmentVsBits.byteBuffer                      2  avgt   30    4.423 ? 0.053  ns/op
MemorySegmentVsBits.byteBuffer                     16  avgt   30   11.038 ? 0.125  ns/op
MemorySegmentVsBits.byteBuffer                     64  avgt   30   23.274 ? 0.262  ns/op
MemorySegmentVsBits.byteBuffer                    256  avgt   30  117.765 ? 2.430  ns/op
MemorySegmentVsBits.byteVarHandle                   1  avgt   30    2.315 ? 0.031  ns/op
MemorySegmentVsBits.byteVarHandle                   2  avgt   30    2.679 ? 0.063  ns/op
MemorySegmentVsBits.byteVarHandle                  16  avgt   30    5.235 ? 0.032  ns/op
MemorySegmentVsBits.byteVarHandle                  64  avgt   30    8.461 ? 0.076  ns/op
MemorySegmentVsBits.byteVarHandle                 256  avgt   30   19.182 ? 0.268  ns/op
MemorySegmentVsBits.longBuffer                      1  avgt   30    2.778 ? 0.050  ns/op
MemorySegmentVsBits.longBuffer                      2  avgt   30    3.077 ? 0.046  ns/op
MemorySegmentVsBits.longBuffer                     16  avgt   30    6.950 ? 0.103  ns/op
MemorySegmentVsBits.longBuffer                     64  avgt   30   17.435 ? 0.229  ns/op
MemorySegmentVsBits.longBuffer                    256  avgt   30   60.646 ? 0.987  ns/op
MemorySegmentVsBits.panamaHeap                      1  avgt   30    4.989 ? 0.074  ns/op
MemorySegmentVsBits.panamaHeap                      2  avgt   30    6.482 ? 0.019  ns/op
MemorySegmentVsBits.panamaHeap                     16  avgt   30   10.841 ? 0.132  ns/op
MemorySegmentVsBits.panamaHeap                     64  avgt   30   21.987 ? 0.246  ns/op
MemorySegmentVsBits.panamaHeap                    256  avgt   30   66.306 ? 0.457  ns/op
MemorySegmentVsBits.panamaHeapBulk                  1  avgt   30    4.434 ? 0.011  ns/op
MemorySegmentVsBits.panamaHeapBulk                  2  avgt   30    4.855 ? 0.122  ns/op
MemorySegmentVsBits.panamaHeapBulk                 16  avgt   30    5.052 ? 0.061  ns/op
MemorySegmentVsBits.panamaHeapBulk                 64  avgt   30    8.181 ? 0.113  ns/op
MemorySegmentVsBits.panamaHeapBulk                256  avgt   30   23.111 ? 5.388  ns/op
MemorySegmentVsBits.panamaNative                    1  avgt   30    5.138 ? 0.074  ns/op
MemorySegmentVsBits.panamaNative                    2  avgt   30    6.458 ? 0.057  ns/op
MemorySegmentVsBits.panamaNative                   16  avgt   30   10.403 ? 0.033  ns/op
MemorySegmentVsBits.panamaNative                   64  avgt   30   21.548 ? 0.107  ns/op
MemorySegmentVsBits.panamaNative                  256  avgt   30   66.500 ? 0.620  ns/op
MemorySegmentVsBits.panamaNativeBulk                1  avgt   30    4.631 ? 0.066  ns/op
MemorySegmentVsBits.panamaNativeBulk                2  avgt   30    4.795 ? 0.069  ns/op
MemorySegmentVsBits.panamaNativeBulk               16  avgt   30    4.812 ? 0.064  ns/op
MemorySegmentVsBits.panamaNativeBulk               64  avgt   30    8.233 ? 0.058  ns/op
MemorySegmentVsBits.panamaNativeBulk              256  avgt   30   17.950 ? 0.148  ns/op
MemorySegmentVsBits.panamaNativeUnaligned           1  avgt   30    5.098 ? 0.080  ns/op
MemorySegmentVsBits.panamaNativeUnaligned           2  avgt   30    6.654 ? 0.093  ns/op
MemorySegmentVsBits.panamaNativeUnaligned          16  avgt   30   10.442 ? 0.161  ns/op
MemorySegmentVsBits.panamaNativeUnaligned          64  avgt   30   22.153 ? 0.280  ns/op
MemorySegmentVsBits.panamaNativeUnaligned         256  avgt   30   66.498 ? 1.033  ns/op
MemorySegmentVsBits.panamaNativeUnalignedBulk       1  avgt   30    4.410 ? 0.028  ns/op
MemorySegmentVsBits.panamaNativeUnalignedBulk       2  avgt   30    4.934 ? 0.208  ns/op
MemorySegmentVsBits.panamaNativeUnalignedBulk      16  avgt   30    5.017 ? 0.151  ns/op
MemorySegmentVsBits.panamaNativeUnalignedBulk      64  avgt   30    9.527 ? 1.611  ns/op
MemorySegmentVsBits.panamaNativeUnalignedBulk     256  avgt   30   24.086 ? 5.510  ns/op
```

It seems to outperform the others, at least for big enough sizes.

**Comment by mcimadamore** [comment]
For the records - the extra benchmarks are here:

```
@Benchmark
    public void panamaHeapBulk() {
        MemorySegment.copy(longs, 0, segment, JAVA_LONG_UNALIGNED, 0, size);
    }

    @Benchmark
    public void panamaNativeBulk() {
        MemorySegment.copy(longs, 0, nativeSegment, JAVA_LONG, 0, size);
    }

    @Benchmark
    public void panamaNativeUnalignedBulk() {
        MemorySegment.copy(longs, 0, nativeSegment, JAVA_LONG_UNALIGNED, 0, size);
    }
```

**Comment by uschindler** [comment]
Thanks @mcimadamore this confirms what we have seen in Lucene regarding the array copies. I did not implement a manual one-by-one copy for small arrays like ByteBuffer as for longs or floats the boundary would be around 2 or 3 array items. This benchmark explains what you already said.

**Comment by uschindler** [comment]
Can you add those benchmarks to the PR?

**Comment by uschindler** [comment]
Ah nevermind, PR is already closed.

**Comment by mcimadamore** [comment]
> Ah nevermind, PR is already closed.

We can always create a new one :-)

**Comment by GavinRay97** [comment]
If anyone is interested, here are the results I got. I also added the bulk ByteBuffer method, as below:

```java
    @Benchmark
    public void byteBufferBulk() {
        longBuffer.put(0, longs, 0, size);
    }
```

I'm running Fedora 38 Rawhide on WSL2, with 6.0.1 kernel and the below processor:
```sh-session
[user@MSI ~]$ uname -a
Linux MSI 6.0.1-locietta-WSL2-xanmod1 #1 SMP Wed Oct 12 22:14:49 UTC 2022 x86_64 GNU/Linux

[user@MSI ~]$ sudo lshw -short
H/W path    Device    Class      Description
============================================
/0/1                  memory     15GiB System memory
/0/2                  processor  11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz
```

Results are:
```
MemorySegmentVsBits.bitsEquivalent                  1  avgt   30      4.091 ±   0.066  ns/op
MemorySegmentVsBits.bitsEquivalent                  2  avgt   30      5.851 ±   0.086  ns/op
MemorySegmentVsBits.bitsEquivalent                 16  avgt   30     31.527 ±   0.583  ns/op
MemorySegmentVsBits.bitsEquivalent                 64  avgt   30    127.531 ±   2.979  ns/op
MemorySegmentVsBits.bitsEquivalent                256  avgt   30    484.423 ±   9.550  ns/op
MemorySegmentVsBits.bitsEquivalent                512  avgt   30   1554.069 ±  20.114  ns/op
MemorySegmentVsBits.byteBuffer                      1  avgt   30      3.717 ±   0.070  ns/op
MemorySegmentVsBits.byteBuffer                      2  avgt   30      4.775 ±   0.100  ns/op
MemorySegmentVsBits.byteBuffer                     16  avgt   30     10.616 ±   0.121  ns/op
MemorySegmentVsBits.byteBuffer                     64  avgt   30     24.960 ±   0.393  ns/op
MemorySegmentVsBits.byteBuffer                    256  avgt   30    112.617 ±   2.424  ns/op
MemorySegmentVsBits.byteBuffer                    512  avgt   30    222.570 ±   3.518  ns/op
MemorySegmentVsBits.byteBufferBulk                  1  avgt   30     20.786 ±   1.493  ns/op
MemorySegmentVsBits.byteBufferBulk                  2  avgt   30     19.371 ±   0.426  ns/op
MemorySegmentVsBits.byteBufferBulk                 16  avgt   30     23.720 ±   1.228  ns/op
MemorySegmentVsBits.byteBufferBulk                 64  avgt   30     45.268 ±   0.895  ns/op
MemorySegmentVsBits.byteBufferBulk                256  avgt   30    115.257 ±   1.508  ns/op
MemorySegmentVsBits.byteBufferBulk                512  avgt   30    271.710 ±   4.398  ns/op
MemorySegmentVsBits.byteVarHandle                   1  avgt   30      2.089 ±   0.033  ns/op
MemorySegmentVsBits.byteVarHandle                   2  avgt   30      1.848 ±   0.032  ns/op
MemorySegmentVsBits.byteVarHandle                  16  avgt   30      4.296 ±   0.136  ns/op
MemorySegmentVsBits.byteVarHandle                  64  avgt   30      7.831 ±   0.373  ns/op
MemorySegmentVsBits.byteVarHandle                 256  avgt   30     12.824 ±   0.145  ns/op
MemorySegmentVsBits.byteVarHandle                 512  avgt   30     27.028 ±   1.252  ns/op
MemorySegmentVsBits.longBuffer                      1  avgt   30      2.491 ±   0.028  ns/op
MemorySegmentVsBits.longBuffer                      2  avgt   30      2.526 ±   0.040  ns/op
MemorySegmentVsBits.longBuffer                     16  avgt   30      6.653 ±   0.132  ns/op
MemorySegmentVsBits.longBuffer                     64  avgt   30     17.858 ±   0.195  ns/op
MemorySegmentVsBits.longBuffer                    256  avgt   30     62.664 ±   1.664  ns/op
MemorySegmentVsBits.longBuffer                    512  avgt   30    123.508 ±   2.895  ns/op
MemorySegmentVsBits.panamaHeap                      1  avgt   30      4.672 ±   0.163  ns/op
MemorySegmentVsBits.panamaHeap                      2  avgt   30      5.390 ±   0.099  ns/op
MemorySegmentVsBits.panamaHeap                     16  avgt   30      8.990 ±   0.220  ns/op
MemorySegmentVsBits.panamaHeap                     64  avgt   30     18.137 ±   1.122  ns/op
MemorySegmentVsBits.panamaHeap                    256  avgt   30     62.260 ±   0.727  ns/op
MemorySegmentVsBits.panamaHeap                    512  avgt   30    123.472 ±   1.329  ns/op
MemorySegmentVsBits.panamaHeapBulk                  1  avgt   30      5.413 ±   0.092  ns/op
MemorySegmentVsBits.panamaHeapBulk                  2  avgt   30      4.274 ±   0.058  ns/op
MemorySegmentVsBits.panamaHeapBulk                 16  avgt   30      5.123 ±   0.136  ns/op
MemorySegmentVsBits.panamaHeapBulk                 64  avgt   30      9.037 ±   0.358  ns/op
MemorySegmentVsBits.panamaHeapBulk                256  avgt   30     22.320 ±   0.299  ns/op
MemorySegmentVsBits.panamaHeapBulk                512  avgt   30     35.525 ±   0.626  ns/op
MemorySegmentVsBits.panamaNative                    1  avgt   30      4.564 ±   0.124  ns/op
MemorySegmentVsBits.panamaNative                    2  avgt   30      5.540 ±   0.112  ns/op
MemorySegmentVsBits.panamaNative                   16  avgt   30      8.913 ±   0.133  ns/op
MemorySegmentVsBits.panamaNative                   64  avgt   30     18.190 ±   0.335  ns/op
MemorySegmentVsBits.panamaNative                  256  avgt   30     60.941 ±   0.893  ns/op
MemorySegmentVsBits.panamaNative                  512  avgt   30    109.495 ±   1.711  ns/op
MemorySegmentVsBits.panamaNativeBulk                1  avgt   30      4.435 ±   0.114  ns/op
MemorySegmentVsBits.panamaNativeBulk                2  avgt   30      4.367 ±   0.063  ns/op
MemorySegmentVsBits.panamaNativeBulk               16  avgt   30      5.293 ±   0.084  ns/op
MemorySegmentVsBits.panamaNativeBulk               64  avgt   30      8.886 ±   0.362  ns/op
MemorySegmentVsBits.panamaNativeBulk              256  avgt   30     22.023 ±   0.666  ns/op
MemorySegmentVsBits.panamaNativeBulk              512  avgt   30     27.262 ±   1.223  ns/op
MemorySegmentVsBits.panamaNativeUnaligned           1  avgt   30      4.388 ±   0.065  ns/op
MemorySegmentVsBits.panamaNativeUnaligned           2  avgt   30      5.471 ±   0.098  ns/op
MemorySegmentVsBits.panamaNativeUnaligned          16  avgt   30      9.021 ±   0.148  ns/op
MemorySegmentVsBits.panamaNativeUnaligned          64  avgt   30     18.251 ±   0.471  ns/op
MemorySegmentVsBits.panamaNativeUnaligned         256  avgt   30     62.630 ±   1.119  ns/op
MemorySegmentVsBits.panamaNativeUnaligned         512  avgt   30    111.240 ±   1.386  ns/op
MemorySegmentVsBits.panamaNativeUnalignedBulk       1  avgt   30      4.434 ±   0.074  ns/op
MemorySegmentVsBits.panamaNativeUnalignedBulk       2  avgt   30      4.463 ±   0.089  ns/op
MemorySegmentVsBits.panamaNativeUnalignedBulk      16  avgt   30      5.200 ±   0.089  ns/op
MemorySegmentVsBits.panamaNativeUnalignedBulk      64  avgt   30      8.568 ±   0.371  ns/op
MemorySegmentVsBits.panamaNativeUnalignedBulk     256  avgt   30     22.682 ±   3.399  ns/op
MemorySegmentVsBits.panamaNativeUnalignedBulk     512  avgt   30     27.234 ±   0.458  ns/op
MemorySegmentVsBits.panamaNativeUnalignedBulk       1  avgt   30      4.415 ±   0.096  ns/op
MemorySegmentVsBits.panamaNativeUnalignedBulk       2  avgt   30      4.377 ±   0.056  ns/op
MemorySegmentVsBits.panamaNativeUnalignedBulk      16  avgt   30      5.044 ±   0.066  ns/op
MemorySegmentVsBits.panamaNativeUnalignedBulk      64  avgt   30      9.919 ±   0.894  ns/op
MemorySegmentVsBits.panamaNativeUnalignedBulk     256  avgt   30     21.985 ±   0.722  ns/op
MemorySegmentVsBits.panamaNativeUnalignedBulk     512  avgt   30     25.142 ±   1.577  ns/op
```

mcimadamore:
Seems bulk byte buffer is slow - but I suspect that has to do with endianness - e.g. this benchmark specifically sets endianness to BIG_ENDIAN, which then causes issues when doing a bulk copy against an array (which has native order). I'd suggest to tweak the buffer to have native endianness, and see if that improves the performance of the bulk put.

mcimadamore:
I have added the byte buffer bulk benchmarks to my slightly modified benchmark that uses native order everywhere (to avoid spurious costs associated with byte swapping):

https://github.com/mcimadamore/panama-foreign/blob/bits-bench/test/micro/org/openjdk/bench/java/lang/foreign/MemorySegmentVsBits.java

I can confirm that byte buffer with bulk put is as fast as memory segments, once byte swapping is removed from the equation:

```
Benchmark                           (size)  Mode  Cnt   Score   Error  Units
MemorySegmentVsBits.byteBufferBulk     256  avgt   30  23.998 ? 5.858  ns/op
```

rwestrel:
I've been looking at some of the performance anomalies with these benchmarks. I now have a collection of c2 changes, mostly tweaks, so they perform better. I filed a number of bugs for them:
https://bugs.openjdk.org/browse/JDK-8300256
https://bugs.openjdk.org/browse/JDK-8300257
https://bugs.openjdk.org/browse/JDK-8300258
https://bugs.openjdk.org/browse/JDK-8300262
https://bugs.openjdk.org/browse/JDK-8300263

I will work on getting them fixed.

I also filed a library bug:
https://bugs.openjdk.org/browse/JDK-8299982

mcimadamore:
> I've been looking at some of the performance anomalies with these benchmarks. I now have a collection of c2 changes, mostly tweaks, so they perform better. I filed a number of bugs for them: https://bugs.openjdk.org/browse/JDK-8300256 https://bugs.openjdk.org/browse/JDK-8300257 https://bugs.openjdk.org/browse/JDK-8300258 https://bugs.openjdk.org/browse/JDK-8300262 https://bugs.openjdk.org/browse/JDK-8300263
> 
> I will work on getting them fixed.
> 
> I also filed a library bug: https://bugs.openjdk.org/browse/JDK-8299982

Thanks for the amazing work @rwestrel !

PR #765, opened 2023-01-11
8299911: Refactor struct handling in AArch64 CallArranger

JornVernee:
Refactor struct handling in AArch64 CallArranger.

Instead of having 6 very similar looking while loops, I've consolidated the storage allocation code for structs into a single helper method (`StorageCalculator::structStorages`). This should make it much easier to see which logic is being used for a particular case.

---------
### Progress

### Issue
 * [JDK-8299911](https://bugs.openjdk.org/browse/JDK-8299911): Refactor struct handling in AArch64 CallArranger

### Reviewers
 * [Maurizio Cimadamore](https://openjdk.org/census#mcimadamore) (@mcimadamore - Committer) ⚠️ Review applies to [3ac117fd](https://git.openjdk.org/panama-foreign/pull/765/files/3ac117fd66af111064c2a1912386df2baaa964c9)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/765/head:pull/765` \
`$ git checkout pull/765`

`$ git checkout pull/765` \
`$ git pull https://git.openjdk.org/panama-foreign pull/765/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 765`

`$ git pr show -t 765`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/765.diff">https://git.openjdk.org/panama-foreign/pull/765.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/CallArranger.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/CallArranger.java
@@ -298,7 +298,7 @@ public abstract class CallArranger {
             for (int i = 0; i < structStorages.length; i++) {
                 ValueLayout copyLayout;
                 if (isFieldWise) {
-                    // We only expect to see fields here, no padding (the cast would fail)
+                    // We should only get here for HFAs, which can't have padding (the cast would fail)
                     copyLayout = (ValueLayout) layout.memberLayouts().get(i);
                 } else {
                     // chunk-wise copy
--- a/src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/CallArranger.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/CallArranger.java
@@ -243,7 +243,7 @@ public abstract class CallArranger {
         The struct is split into 8-byte chunks, and those chunks are either passed in registers and/or on the stack.
 
         Homogeneous float aggregates (HFAs) can be copied in a field-wise manner, i.e. the struct is split into it's
-        fields and those fields are the chunks which are passed. The rules are more complicated and ABI based:
+        fields and those fields are the chunks which are passed. For HFAs the rules are more complicated and ABI based:
 
                 | enough registers | some registers, but not enough  | no registers
         --------+------------------+---------------------------------+-------------------------
@@ -293,9 +293,9 @@ public abstract class CallArranger {
                 VMStorage storage = nextStorage(regType, copyLayout);
                 Class<?> carrier = copyLayout.carrier();
                 if (isFieldWise && storage.type() == StorageType.STACK) {
-                    // chunkLayout is a field of an HFA
+                    // copyLayout is a field of an HFA
                     // Don't use floats on the stack
-                    carrier = adjustCarrierForStack(copyLayout.carrier());
+                    carrier = adjustCarrierForStack(carrier);
                 }
                 structStorages[i] = new StructStorage(offset, carrier, storage);
                 offset += copyLayout.byteSize();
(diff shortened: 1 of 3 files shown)


src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/CallArranger.java
```diff
@@ -217,106 +219,142 @@ public StorageCalculator(boolean forArguments, boolean forVariadicFunction) {
...
-                regAllocPartial(type, layout) :
-                regAlloc(type, requiredRegisters(layout));
-        }
+            StructStorage[] structStorages = new StructStorage[requiredStorages];
+            long offset = 0;
+            for (int i = 0; i < structStorages.length; i++) {
+                ValueLayout copyLayout;
+                if (isFieldWise) {
+                    // fields only, no padding
+                    copyLayout = (ValueLayout) layout.memberLayouts().get(i);
```
mcimadamore, line 302: Doesn't this contradict the comment? E.g. if we only copy fields, then we should skip padding layouts here (as they can be members of a group layout) ?

src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/CallArranger.java
```diff
@@ -217,106 +219,142 @@ public StorageCalculator(boolean forArguments, boolean forVariadicFunction) {
...
+                } else {
+                    // chunk-wise copy
+                    long copySize = Math.min(layout.byteSize() - offset, MAX_COPY_SIZE);
+                    boolean useFloat = false; // never use float for chunk-wise copies
```
mcimadamore, line 306: this local seems unnecessary?

src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/CallArranger.java
JornVernee: The cast would fail if we ran into a padding layout, so there is an implicit assumption that we will not see any padding. I added the comment there to emphasize that. If we could have padding at this point, the offset computation we do later would also be incorrect.

src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/CallArranger.java
JornVernee: The local is there to give a name to the magic `false` value, and it also allows putting a comment. I preferred doing this to the alternative:

    copyLayout = SharedUtils.primitiveLayoutForSize(copySize, /* useFloat = */ false); // never use float for chunk-wise copies

src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/CallArranger.java
JornVernee: I've attempted to clarify the comment a bit.

mcimadamore: > I've attempted to clarify the comment a bit.

Just to check - where do we check that these structs do not have any padding inside? Does a meaningful exception get produced if that is not the case?

JornVernee: Only HFAs are handled in this field-wise manner (see the table in the comment on this method as well). The assumption is that any struct with padding doesn't get classified as an HFA, and we can not get here (see [`jdk.internal.foreign.abi.aarch64.TypeClass::isHomogeneousFloatAggregate`](https://github.com/openjdk/panama-foreign/blob/08225e4f5e00837e974098a933f27267f3c0dbc5/src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/TypeClass.java#L61-L91). But, this is a non-local assumption, so the comment seems useful.

JornVernee: Attempted to clarify the comment once more.

mcimadamore (approved):

src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/CallArranger.java
mcimadamore: Ok, the main issue was that by the time I got there I lost track of "isFieldWise implies HFA"

minborg:
Do not forget to bump the copyright year...

PR #766, opened 2023-01-11
8299181: PaddingLayout unable to return byteAlignment value

minborg:
This PR proposes bit-aligning `PaddingLayout` to 8 by default as opposed to the current  value of 1.

More generally, the PR restricts `PaddingLayout` bit-alignment to non-negative values that are an even multiple of 8 (i.e. 0, 8, 16, ...). There are no mechanism to derive `VarHandle` objects for sub-byte access anyhow and trying to obtain a miss-aligned `VarHandle` would throw an exception. 

If integrated, all `MemoryLayout` types would be byte-aligned.

---------
### Progress

### Issue
 * [JDK-8299181](https://bugs.openjdk.org/browse/JDK-8299181): PaddingLayout unable to return byteAlignment value

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/766/head:pull/766` \
`$ git checkout pull/766`

`$ git checkout pull/766` \
`$ git pull https://git.openjdk.org/panama-foreign pull/766/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 766`

`$ git pr show -t 766`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/766.diff">https://git.openjdk.org/panama-foreign/pull/766.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
+++ b/src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
@@ -30,7 +30,6 @@ import java.lang.invoke.MethodHandles;
 import java.lang.invoke.VarHandle;
 import java.nio.ByteOrder;
 import java.util.EnumSet;
-import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
@@ -173,7 +172,7 @@ import jdk.internal.javac.PreviewFeature;
  * @since 19
  */
 @PreviewFeature(feature=PreviewFeature.Feature.FOREIGN)
-public sealed interface MemoryLayout permits GroupLayout, PaddingLayout, SequenceLayout, ValueLayout {
+public sealed interface MemoryLayout permits SequenceLayout, GroupLayout, PaddingLayout, ValueLayout {
 
     /**
      * {@return the layout size, in bits}
@@ -707,9 +706,10 @@ public sealed interface MemoryLayout permits GroupLayout, PaddingLayout, Sequenc
      */
     static StructLayout structLayout(MemoryLayout... elements) {
         Objects.requireNonNull(elements);
-        return wrapOverflow(() -> StructLayoutImpl.of(Stream.of(elements)
-                .map(Objects::requireNonNull)
-                .toList()));
+        return wrapOverflow(() ->
+                StructLayoutImpl.of(Stream.of(elements)
+                        .map(Objects::requireNonNull)
+                        .toList()));
     }
 
     /**
--- a/src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
+++ b/src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
@@ -36,6 +36,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.function.Function;
 import java.util.function.Supplier;
+import java.util.stream.Stream;
 
 import jdk.internal.foreign.LayoutPath;
 import jdk.internal.foreign.LayoutPath.PathElementImpl.PathKind;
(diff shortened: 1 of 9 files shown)


src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
```diff
@@ -611,15 +609,17 @@ static PathElement sequenceElement() {
...
+     * @param bitSize the padding size in bits.
      * @return the new selector layout.
-     * @throws IllegalArgumentException if {@code size <= 0}.
+     * @throws IllegalArgumentException if {@code bitSize < 0} or {@code bitSize % 8 != 0}
      */
-    static PaddingLayout paddingLayout(long size) {
-        MemoryLayoutUtil.checkSize(size);
-        return PaddingLayoutImpl.of(size);
+    static PaddingLayout paddingLayout(long bitSize) {
+        if (bitSize < 0 || bitSize % 8 != 0) {
```
mcimadamore, line 619: If I'm not mistaken, this is the same check that should be applied when we call `withBitAlignment`, right? It might be useful to consolidate (both impl-wise and doc-wise).

src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
```diff
@@ -709,10 +709,7 @@ static SequenceLayout sequenceLayout(MemoryLayout elementLayout) {
      */
     static StructLayout structLayout(MemoryLayout... elements) {
         Objects.requireNonNull(elements);
-        return wrapOverflow(() ->
-                StructLayoutImpl.of(Stream.of(elements)
-                        .map(Objects::requireNonNull)
-                        .toList()));
+        return wrapOverflow(() -> StructLayoutImpl.of(List.of(elements)));
```
mcimadamore, line 712: Why was Objects.requireNonNull dropped? I believe that NPE  will be thrown deeper inside the implementation - which is why we had the eager check (to make stack trace more readable).

src/java.base/share/classes/jdk/internal/foreign/layout/AbstractLayout.java
```diff
@@ -140,17 +133,17 @@ String decorateLayoutString(String s) {
...
+            s = bitAlignment() + "%" + s;
         }
         return s;
     }
 
-    private static void checkAlignment(long alignmentBitCount) {
-        if (((alignmentBitCount & (alignmentBitCount - 1)) != 0L) || //alignment must be a power of two
-                (alignmentBitCount < 8)) { //alignment must be greater than 8
-            throw new IllegalArgumentException("Invalid alignment: " + alignmentBitCount);
+    private static long requirePowerOfTwoAndGreaterOrEqualToEight(long value) {
```
mcimadamore, line 141: Should this stay here, or be merged inside MemoryLayoutUtil? It seems the same as `requireBitSizeValid`, except for the zero behavior?

test/jdk/java/foreign/TestLayouts.java
```diff
@@ -166,6 +168,51 @@ public void testStructOverflow() {
...
+    }
+
+    @Test
+    public void testPaddingZeroBitSize() {
+        MemoryLayout.paddingLayout(0);
+    }
+
+    @Test
+    public void testNullMember() {
+        var illegalLayouts = new MemoryLayout[]{JAVA_INT, null, JAVA_INT};
```
mcimadamore, line 203: this is covered by TestNulls. That test will try to pass null values to all API methods accepting arrays, or collections, and will do so in an automatic fashion.

src/java.base/share/classes/jdk/internal/foreign/layout/AbstractLayout.java
minborg: This method also checks if the parameter is an even power of two.

mcimadamore (approved):

PR #770, opened 2023-01-17
8287516: Implement fallback Linker

JornVernee:
Implement a [libffi](https://github.com/libffi/libffi) based fallback linker. The fallback linker comes with a native library that can be included in the jdk by configuring with `--enable-fallback-linker`. This requires the libffi library to be available as well (I've tested with version 3.4.2). On the zero VM configuration the fallback linker is enabled by default.

CABI is refactored a bit. It now supports explicitly setting the CABI using the `jdk.internal.foreign.CABI` system property (which is very handy for testing). Instead of falling back to `null` when the platform is not supported, we instead fall back to checking if the fallback library is supported (`FALLBACK`), and if not, default to the new `UNSUPPORTED` enum constant, which is then mapped to an `UnsupportedOperationException` in the Linker::nativeLinker implementation.

Checking whether the a full port of the foreign linker is available is done through the new jdk.internal.vm.ForeignLinkerSupport class. This follows the precedent set by Loom's jdk.internal.vm.ContinuationSupport class. The implementation simply returns `true` on platforms where we have a port.

The system lookup is changed to be selected based on the platform, rather than CABI. This allows selecting the right library implementation when running on a zero VM configuration as well. The split is now just between Windows and non-Windows platforms.

Finally, I had to slightly modify `globalDefinitions_zero.hpp` to conditionally define `FFI_GO_CLOSURES`, since I was getting a redefinition error using GCC. The macro is also defined in `ffitarget.h` which is included by `ffi.h`. The define in globalDefinitions comes from this PR: https://github.com/openjdk/jdk/pull/8195 which indicates that the define is only needed on Mac Os X. So I switched out the guard to check for `__APPLE__` instead. (the check whether it is already defined doesn't really do anything, since `FFI_GO_CLOSURES` is defined by including `ffi.h`).

The Java implementation is found in the `jdk.internal.foreign.abi.fallback` package, and consists of the following classes:
- `FallbackLinker`. The main entry point, an instance of `Linker`. Contains the code for arranging and executing calls.
- `LibFallback`. A thin wrapper around the native `libfallback` library found in the java.base module, which acts as an interface to the libffi library.
- `FFIType`. Contains code for creating native `ffi_type` struct instances from memory layouts.
- The `TypeClass`, `FFIStatus`, and `FFIABI` supporting enums, which should be pretty self-explanatory.

I've ran the `jdk_foreign` tests in the following configurations:
- Windows.
- Linux with the fallback linker lib included, but not used.
- Linux with fallback linker and running with `-Djdk.internal.foreign.CABI=FALLBACK` (this uses the fallback implementation).
- Linux zero VM (uses `FALLBACK` as CABI by default, because we don't have a port for `zero`).

---------
### Progress

### Issue
 * [JDK-8287516](https://bugs.openjdk.org/browse/JDK-8287516): Implement fallback Linker

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/770/head:pull/770` \
`$ git checkout pull/770`

`$ git checkout pull/770` \
`$ git pull https://git.openjdk.org/panama-foreign pull/770/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 770`

`$ git pr show -t 770`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/770.diff">https://git.openjdk.org/panama-foreign/pull/770.diff</a>

</details>


src/java.base/share/classes/jdk/internal/foreign/abi/fallback/FFIABI.java
```diff
@@ -0,0 +1,39 @@
...
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package jdk.internal.foreign.abi.fallback;
+
+enum FFIABI {
```
JornVernee, line 27: Adding a standalone enum for a single constant might seem a bit overkill right now, but there are many ABIs possible [1], so I expect there to be more constants added in the future (probably at least the stdcall ABI).

Plus, it's a nice bit of civilization compared to passing around `int`.

[1]: https://github.com/libffi/libffi/blob/7611bb4cfe90884b55ad225e0166136a1d2cf22b/src/x86/ffitarget.h#L83

src/java.base/share/classes/jdk/internal/foreign/abi/fallback/TypeClass.java
```diff
@@ -0,0 +1,68 @@
...
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package jdk.internal.foreign.abi.fallback;
+
+import java.lang.foreign.GroupLayout;
+import java.lang.foreign.MemoryLayout;
+import java.lang.foreign.ValueLayout;
+
+enum TypeClass {
```
mcimadamore, line 31: Is this necessary? The mapping between layouts and FFI types seems pretty 1-1. If we keep this, would it make sense to make it "smarter" by adding the write/read functions in here? (I also understand if you want to keep it as classification token).

src/java.base/share/classes/jdk/internal/foreign/Utils.java
```diff
@@ -202,4 +205,35 @@ public static void checkAllocationSizeAndAlign(long byteSize, long byteAlignment
...
+        boolean isAligned = offset == 0 || offset % align == 0;
+        if (isAligned) {
+            return 0;
+        } else {
+            long gap = offset % align;
+            return align - gap;
+        }
+    }
+
+    public static StructLayout computeStructLayout(MemoryLayout... elements) {
```
mcimadamore, line 219: Not too sure about this - at least this being here. The general spirit of Linker classification is that padding bits are inserted as required by the client. Now, it seems you need this inside the fallback FFI, so that you can create the precise layout of a struct where some of the fields might be platform-dependent - in that case I'd say it would be better to move this function over there?

mcimadamore (approved):
The code looks good and easy to follow - good job!
I'm a bit worried re. testing: given that libffi is not included in a regular JDK build, we don't have ways to test this in our CI, which seems bad (how do we detect regressions?).

src/java.base/share/classes/jdk/internal/foreign/Utils.java
JornVernee: Note that we don't use this on any of the user supplied layouts. This is only used to compute the layout of `ffi_type` in `FFIType`. On 64bit platforms this needs padding, on 32bit platforms it does not.

I can move it into `FFIType` as well, but it seemed like a more general piece of functionality (i.e. maybe we could change the CallGeneratorHelper test class to use this instead)

src/java.base/share/classes/jdk/internal/foreign/abi/fallback/TypeClass.java
JornVernee: Hmm, that's an interesting idea. We can not switch on just the carrier type, so would have to do a linear search in the write/read functions, but making them "smarter", using virtual methods to implement them, solves that.

I'll take another look here.

JornVernee: I've removed the TypeClass class. We now do a linear search over the types in the write/read functions. This could be replaced with a virtual call, but then we'd hit the bimorphic inlining limit of 2.

Eventually, this can be improved by using type switch, but let's go with this simplified version (without TypeClass) for now.

src/java.base/share/classes/jdk/internal/foreign/Utils.java
mcimadamore: Yep - I've seen that. I guess perhaps what I find confusing is the name - e.g. one might thing that MemoryLayout.structLayout should call this, but it doesn't. But I don't have any great suggestions. Perhaps I'd stress the alignment, so maybe "computePaddedStructLayout" would be clearer?

JornVernee:
> The code looks good and easy to follow - good job! I'm a bit worried re. testing: given that libffi is not included in a regular JDK build, we don't have ways to test this in our CI, which seems bad (how do we detect regressions?).

WRT testing, I think we have access to libffi internally, and should be able to request it in our task definition, which would allows running with `-Djdk.internal.foreign.CABI=FALLBACK` in order to test this.

src/java.base/share/classes/jdk/internal/foreign/abi/fallback/FallbackLinker.java
```diff
@@ -170,89 +166,105 @@ private static Object doDowncall(SegmentAllocator returnAllocator, Object[] args
...
-            case FLOAT -> argSeg.set((ValueLayout.OfFloat) layout, 0, (Float) arg);
-            case DOUBLE -> argSeg.set((ValueLayout.OfDouble) layout, 0, (Double) arg);
-            case ADDRESS -> {
-                MemorySegment addrArg = (MemorySegment) arg;
-                acquireCallback.accept(addrArg);
-                argSeg.set((ValueLayout.OfAddress) layout, 0, addrArg);
-            }
-            case SEGMENT -> argSeg.copyFrom((MemorySegment) arg); // by-value struct
-            case VOID -> {}
+        if (layout instanceof ValueLayout.OfBoolean bl) {
```
mcimadamore, line 218: Do you think we lose a lot of performance if we don't switch? E.g. use the underlying var handle (which is stored in the value layout) and access it in an inexact mode?

JornVernee: I considered that, but thought inexact access would be a lot worse in terms of performance since then we lose all inlining, and have to go through `asType` a lot of the time (since we probably won't see the same layout type all the time).

I think what I ended up doing is a nice balance between performance and simplicity, and looks kind of like a 'template' for using type switch in the future.

mcimadamore (approved):

JornVernee:
I've added the bits needed to do CI testing to the patch as well.

@erikj79 Could you maybe take a look? The files of interest are
- [`make/conf/jib-profiles.js`](https://github.com/openjdk/panama-foreign/pull/770/files#diff-378a857f27068ca53bdc8ac8d077fdffa97ca07aa5eea05a530e0df00c61aea8)
- [`make/devkit/createLibffiBundle.sh`](https://github.com/openjdk/panama-foreign/pull/770/files#diff-298e37d9aa09177f30d3432d65b45c55b0066cb2cae07c7ef04e337f5d76cbaf)

For the artifact definition for `linux-x64-zero`, I based on what was done for the gcov artifacts, changing only the name of the profile. Does that look good?

src/java.base/share/classes/jdk/internal/foreign/abi/fallback/FFIType.java
```diff
@@ -0,0 +1,144 @@
...
+        }
+        return Objects.requireNonNull(CARRIER_TO_TYPE.get(((ValueLayout) layout).carrier()));
+    }
+
+    // verify layout against what libffi sets
+    private static void verifyStructType(GroupLayout grpl, List<MemoryLayout> filteredLayouts, MemorySegment structType,
+                                         FFIABI abi) {
+        try (Arena verifyArena = Arena.openConfined()) {
+            MemorySegment offsetsOut = verifyArena.allocate(SIZE_T.byteSize() * filteredLayouts.size());
+            LibFallback.getStructOffsets(structType, offsetsOut, abi);
```
mcimadamore, line 129: Maybe I'm being paranoid, but libFFI has its own way to determine a struct layout given the struct fields, based on what the ABI types are. But the Panama layout API is more expressive than that, and it allows, theoretically, to pass arguments that have stricter or larger alignment constraints - which can end up in more or less padding being inserted (from clients), and in different struct layout sizes.

So, my question is: should we double check that what we get back from libFFI is actually compatible with what we have in the group layout before proceeding?

JornVernee: Yes, that is exactly what this code (`verifyStructType`) is trying to do.

Note that we also already reject layouts that do not have natural alignment [1](https://github.com/openjdk/panama-foreign/blob/foreign-memaccess%2Babi/src/java.base/share/classes/jdk/internal/foreign/abi/AbstractLinker.java#L111) Though, we might want to relax that constraint in the future.

mcimadamore: More specifically, if a struct has extra padding, then I wonder if libFFI could expect a struct size that is smaller than the size of the memory segment passed in by a Panama client (because that segment also has the extra padding). In our usual JDK-driven classification, we would take the struct, extract its field components and place them on registers and/or stack slots. But with libFFI, we just have to provide the struct segment in a buffer, which then libFFI will act upon - if there is a size mismatch there, I think it might have dire consequences?

mcimadamore: Anyway - I appreciate that the code verifies for mismatches. Good job. I was worried about the analysis being too "field-oriented", but you also look at offsets, so any extra padding should result in issues. Well, except maybe a trailing padding field?

src/java.base/share/classes/jdk/internal/foreign/abi/fallback/FFIType.java
```diff
@@ -127,17 +127,18 @@ private static void verifyStructType(GroupLayout grpl, List<MemoryLayout> filter
...
+                    long ffiOffset = (long) VH_SIZE_T_ARRAY.get(offsetsOut, offsetIdx++);
+                    if (ffiOffset != expectedOffset) {
                         throw new IllegalArgumentException("Invalid group layout." +
-                                " Offset of '" + name + "': " + layoutOffset + " != " + ffiOffset);
+                                " Offset of '" + element.name().orElse("<unnamed>")
+                                + "': " + expectedOffset + " != " + ffiOffset);
                     }
-                });
+                }
+                expectedOffset += element.byteSize();
```
mcimadamore, line 141: I think the offset should only be added for structs, not for unions

JornVernee: We can never encounter a union in this code. In libffi they are represented as a single element struct.

JornVernee: Support for unions is not implemented in the fallback linker at the moment [1](https://github.com/openjdk/panama-foreign/pull/770/files#diff-b2d5d47f7117501ef2f8d7c497f0e2d999ac4374ae0a17aacaf12590cb2fc111R116). I wanted to do that as part of: https://bugs.openjdk.org/browse/JDK-8300294 when we have some tests, and there might be issues on other linkers as well that need fixing.

JornVernee: I've changed the parameter type to `StructLayout` to make this more clear.

mcimadamore (approved):

PR #771, opened 2023-01-19
8299730: Add trivial call linker option

JornVernee:
Add a trivial call linker option. This option acts as a hint to the runtime which indicates that a downcall is extremely short live (comparable to calling an empty function), and does not call back into Java.

Some implementations (like ours) might use this information to apply additional optimizations that are only applicable to trivial functions (dropping thread state transitions).

Note that the VM code diff looks kind of messy, but I essentially just put 3 `if` statements around the code that deals with the thread state transitions before and after the call. This also required moving the declarations of the `Label`s around though (so they are accessible from both ifs on the return path), which is probably messing up the diff view.

---------
### Progress

### Issue
 * [JDK-8299730](https://bugs.openjdk.org/browse/JDK-8299730): Add trivial call linker option

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/771/head:pull/771` \
`$ git checkout pull/771`

`$ git checkout pull/771` \
`$ git pull https://git.openjdk.org/panama-foreign pull/771/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 771`

`$ git pr show -t 771`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/771.diff">https://git.openjdk.org/panama-foreign/pull/771.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/java/lang/foreign/Linker.java
+++ b/src/java.base/share/classes/java/lang/foreign/Linker.java
@@ -324,7 +324,7 @@ public sealed interface Linker permits AbstractLinker {
          * Marking a function using this linker option is a hint which some implementations may use to apply
          * optimizations that are only valid for trivial functions.
          * <p>
-         * Marking a function that is not trivial using this linker option will result in a loss
+         * Marking a function that is not trivial using this linker option can result in a loss
          * of performance, and has the potential to cause JVM crashes.
          * <p>
          * This method is <a href="package-summary.html#restricted"><em>restricted</em></a>.
--- a/src/java.base/share/classes/java/lang/foreign/Linker.java
+++ b/src/java.base/share/classes/java/lang/foreign/Linker.java
@@ -316,6 +316,27 @@ public sealed interface Linker permits AbstractLinker {
         }
 
         /**
+         * {@return A linker option used to mark a function as trivial}
+         * <p>
+         * A 'trivial' function is a function that has an extremely short running time
+         * in all cases (similar to calling an empty function), and does not call back into Java.
+         * <p>
+         * Marking a function using this linker option is a hint which some implementations may use to apply
+         * optimizations that are only valid for trivial functions.
+         * <p>
+         * Marking a function that is not trivial using this linker option will result in a loss
+         * of performance, and has the potential to cause JVM crashes.
+         * <p>
+         * This method is <a href="package-summary.html#restricted"><em>restricted</em></a>.
+         * Restricted methods are unsafe, and, if used incorrectly, their use might crash
+         * the JVM or, worse, silently result in memory corruption. Thus, clients should refrain from depending on
+         * restricted methods, and use safe and supported functionalities, where possible.
+         */
+        static Option isTrivial() {
+            return LinkerOptions.IsTrivial.INSTANCE;
+        }
+
+        /**
          * A linker option for saving portions of the execution state immediately
          * after calling a foreign function associated with a downcall method handle,
          * before it can be overwritten by the runtime, or read through conventional means.
--- a/src/java.base/share/classes/java/lang/foreign/Linker.java
+++ b/src/java.base/share/classes/java/lang/foreign/Linker.java
@@ -316,15 +316,15 @@ public sealed interface Linker permits AbstractLinker {
         }
 
         /**
-         * {@return A linker option used to mark a function as trivial}
+         * {@return A linker option used to mark a foreign function as <em>trivial</em>}
          * <p>
-         * A 'trivial' function is a function that has an extremely short running time
-         * in all cases (similar to calling an empty function), and does not call back into Java.
+         * A trivial function is a function that has an extremely short running time
+         * in all cases (similar to calling an empty function), and does not call back into Java (e.g. using an upcall stub).
          * <p>
-         * Marking a function using this linker option is a hint which some implementations may use to apply
+         * Using this linker option is a hint which some implementations may use to apply
          * optimizations that are only valid for trivial functions.
          * <p>
-         * Marking a function that is not trivial using this linker option is likely to have adverse effects,
+         * Using this linker option when linking non trivial functions is likely to have adverse effects,
          * such as loss of performance, or JVM crashes.
          */
         static Option isTrivial() {
--- a/src/java.base/share/classes/java/lang/foreign/Linker.java
+++ b/src/java.base/share/classes/java/lang/foreign/Linker.java
@@ -324,13 +324,8 @@ public sealed interface Linker permits AbstractLinker {
          * Marking a function using this linker option is a hint which some implementations may use to apply
          * optimizations that are only valid for trivial functions.
          * <p>
-         * Marking a function that is not trivial using this linker option can result in a loss
-         * of performance, and has the potential to cause JVM crashes.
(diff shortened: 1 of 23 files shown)


test/jdk/java/foreign/trivial/TestTrivial.java
```diff
@@ -0,0 +1,97 @@
...
+import java.lang.foreign.MemoryLayout;
+import java.lang.foreign.MemorySegment;
+import java.lang.foreign.SegmentAllocator;
+import java.lang.foreign.StructLayout;
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.VarHandle;
+
+import static org.testng.Assert.assertEquals;
+
+public class TestTrivial extends NativeTestHelper {
```
mcimadamore, line 46: Should we perhaps have another subclass of `TestDowncallBase` to test all combinations - just in case?

src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -315,6 +315,22 @@ static CaptureCallState captureCallState(String... capturedState) {
             return new LinkerOptions.CaptureCallStateImpl(set);
         }
 
+        /**
+         * {@return A linker option used to mark a function as trivial}
```
mcimadamore, line 319:
```suggestion
         * {@return A linker option used to mark a foreign function as <em>trivial</em>}
```

src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -315,6 +315,22 @@ static CaptureCallState captureCallState(String... capturedState) {
...
+         * <p>
+         * A 'trivial' function is a function that has an extremely short running time
```
mcimadamore, line 321:
```suggestion
         * A trivial function is a function that has an extremely short running time
```

src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -315,6 +315,22 @@ static CaptureCallState captureCallState(String... capturedState) {
...
+         * in all cases (similar to calling an empty function), and does not call back into Java.
```
mcimadamore, line 322:
```suggestion
         * in all cases (similar to calling an empty function), and does not call back into Java (e.g. using an upcall stub).
```

mcimadamore: We should probably mention `getpid` somewhere as the poster child for trivial calls.

src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -315,6 +315,22 @@ static CaptureCallState captureCallState(String... capturedState) {
...
+         * <p>
+         * Marking a function using this linker option is a hint which some implementations may use to apply
```
mcimadamore, line 324:
```suggestion
         * Using this linker option is a hint which some implementations may use to apply
```

src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -315,6 +315,22 @@ static CaptureCallState captureCallState(String... capturedState) {
...
+         * optimizations that are only valid for trivial functions.
+         * <p>
+         * Marking a function that is not trivial using this linker option is likely to have adverse effects,
```
mcimadamore, line 327:
```suggestion
         * Using this linker option when linking non trivial functions is likely to have adverse effects,
```

test/jdk/java/foreign/trivial/TestTrivial.java
JornVernee: Seems like overkill, tbh, since `trivial` doesn't affect any of the classification logic. I'm already testing all the cases that are handled differently by the downcall stub (where `trivial` does matter).

Also, ultimately the stuff covered by TestDowncall is limited (no unions, nested structs, array members. Limited set of primitive types). I think long term I'd rather implement some kind of fuzzing-based test, where we generate an arbitrary FunctionDescriptor/linker options/C function, and then try to link and call that. But this requires generating and compiling C code on the fly during the tests (which I expect will be problematic to get working).

JornVernee: I could maybe add a few more test cases, to the tune of what we do for the CaptureCallState tests. But my feeling is that if we go to the level of TestDowncall, we don't really get any more coverage from that.

mcimadamore (approved):

JornVernee:
Forgot to add the benchmarks before. Added them now.

Some numbers on my machine:

```
Benchmark                                  Mode  Cnt  Score   Error  Units
CallOverheadConstant.jni_blank             avgt   30  4.271 � 0.012  ns/op
CallOverheadConstant.panama_blank          avgt   30  3.300 � 0.022  ns/op
CallOverheadConstant.panama_blank_trivial  avgt   30  2.853 � 0.010  ns/op
CallOverheadVirtual.jni_blank              avgt   30  4.278 � 0.017  ns/op
CallOverheadVirtual.panama_blank           avgt   30  3.665 � 0.014  ns/op
CallOverheadVirtual.panama_blank_trivial   avgt   30  3.254 � 0.011  ns/op
```
The difference seems to be very minimal on my machine, but other have reported more significant improvements. This could still be improved by intrinsifying in C2 maybe, but not sure by how much, and if that's worth the complexity.

mcimadamore:
> Forgot to add the benchmarks before. Added them now.
> 
> Some numbers on my machine:
> 
> ```
> Benchmark                                  Mode  Cnt  Score   Error  Units
> CallOverheadConstant.jni_blank             avgt   30  4.271 � 0.012  ns/op
> CallOverheadConstant.panama_blank          avgt   30  3.300 � 0.022  ns/op
> CallOverheadConstant.panama_blank_trivial  avgt   30  2.853 � 0.010  ns/op
> CallOverheadVirtual.jni_blank              avgt   30  4.278 � 0.017  ns/op
> CallOverheadVirtual.panama_blank           avgt   30  3.665 � 0.014  ns/op
> CallOverheadVirtual.panama_blank_trivial   avgt   30  3.254 � 0.011  ns/op
> ```
> 
> The difference seems to be very minimal on my machine, but others have reported more significant improvements. This could still be improved by intrinsifying in C2 maybe, but not sure by how much, and if that's worth the complexity.

Nice!

mcimadamore (approved):

JornVernee:
The new test that tests that we can't do an upcall when a function is trivial was failing on the `zero` VM. `zero` just ignores the linker option, so it is also impossible to detect. I don't see any merit in trying to make sure that an upcall fails on `zero` though when the linker option is otherwise ignored.

mcimadamore (approved):

mcimadamore:
> The new test that tests that we can't do an upcall when a function is trivial was failing on the `zero` VM. `zero` just ignores the linker option, so it is also impossible to detect. I don't see any merit in trying to make sure that an upcall fails on `zero` though when the linker option is otherwise ignored. So, I've just disabled the test on `zero` instead.

I agree - trivial should be best effort - perhaps something worth a note in the javadoc (e.g. that some linker implementation might reserve a right to refuse?)

PR #774, opened 2023-01-23
8300201: When storing MemoryAddress.ofLong(0x0000000080000000L), MemorySegment.get is not equal to MemorySegment.set because of the expanded sign

minborg:
This PR proposes to normalize the upper 32-bit to zero on 32-bit systems.

---------
### Progress

### Issue
 * [JDK-8300201](https://bugs.openjdk.org/browse/JDK-8300201): When storing MemoryAddress.ofLong(0x0000000080000000L), MemorySegment.get is not equal to MemorySegment.set because of the expanded sign

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/774/head:pull/774` \
`$ git checkout pull/774`

`$ git checkout pull/774` \
`$ git pull https://git.openjdk.org/panama-foreign pull/774/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 774`

`$ git pr show -t 774`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/774.diff">https://git.openjdk.org/panama-foreign/pull/774.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/java/lang/foreign/MemorySegment.java
+++ b/src/java.base/share/classes/java/lang/foreign/MemorySegment.java
@@ -1068,8 +1068,8 @@ public sealed interface MemorySegment permits AbstractMemorySegmentImpl {
      * restricted methods, and use safe and supported functionalities, where possible.
      * <p>
      * On 32-bit platforms, the given address value will be normalized such that the
-     * returned memory segment's method {@link MemorySegment#address() address} will always return
-     * zeros for the upper 32 bits of the address.
+     * upper 32 bits of the {@link MemorySegment#address() address} of returned memory segment
+     * are set to zero.
      *
      * @param address the address of the returned native segment.
      * @param byteSize the size (in bytes) of the returned native segment.
@@ -1102,8 +1102,8 @@ public sealed interface MemorySegment permits AbstractMemorySegmentImpl {
      * restricted methods, and use safe and supported functionalities, where possible.
      * <p>
      * On 32-bit platforms, the given address value will be normalized such that the
-     * returned memory segment's method {@link MemorySegment#address() address} will always return
-     * zeros for the upper 32 bits of the address.
+     * upper 32 bits of the {@link MemorySegment#address() address} of returned memory segment
+     * are set to zero.
      *
      * @param address the returned segment's address.
      * @param byteSize the desired size.
@@ -1145,8 +1145,8 @@ public sealed interface MemorySegment permits AbstractMemorySegmentImpl {
      * restricted methods, and use safe and supported functionalities, where possible.
      * <p>
      * On 32-bit platforms, the given address value will be normalized such that the
-     * returned memory segment's method {@link MemorySegment#address() address} will always return
-     * zeros for the upper 32 bits of the address.
+     * upper 32 bits of the {@link MemorySegment#address() address} of returned memory segment
+     * are set to zero.
      *
      * @param address the returned segment's address.
      * @param byteSize the desired size.
--- a/src/java.base/share/classes/java/lang/foreign/MemorySegment.java
+++ b/src/java.base/share/classes/java/lang/foreign/MemorySegment.java
@@ -1042,6 +1042,11 @@ public sealed interface MemorySegment permits AbstractMemorySegmentImpl {
      * {@snippet lang = java:
      * ofAddress(address, 0);
      *}
+     * <p>
+     * On 32-bit platforms, the given address value will be normalized such that the
+     * returned memory segment's method {@link MemorySegment#address() address} will always return
+     * zeros for the upper 32-bits of the address.
+     *
      * @param address the address of the returned native segment.
      * @return a zero-length native segment with the given address.
      */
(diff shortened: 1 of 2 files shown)


src/java.base/share/classes/java/lang/foreign/MemorySegment.java
```diff
@@ -1042,6 +1042,11 @@ static MemorySegment ofArray(double[] doubleArray) {
      * {@snippet lang = java:
      * ofAddress(address, 0);
      *}
+     * <p>
+     * On 32-bit platforms, the given address value will be normalized such that the
+     * returned memory segment's method {@link MemorySegment#address() address} will always return
+     * zeros for the upper 32 bits of the address.
```
mcimadamore, line 1048:
```suggestion
     * On 32-bit platforms, the given address value will be normalized such that the
     * upper 32 bits of the {@link MemorySegment#address() address} of returned memory segment
     * are set to zero.
```

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
```diff
@@ -1044,8 +1044,8 @@ static MemorySegment ofArray(double[] doubleArray) {
      *}
      * <p>
      * On 32-bit platforms, the given address value will be normalized such that the
-     * returned memory segment's method {@link MemorySegment#address() address} will always return
-     * zeros for the upper 32 bits of the address.
+     * upper 32 bits of the {@link MemorySegment#address() address} of returned memory segment
```
mcimadamore, line 1047: Sorry, I mean to replicate this change in all the other pieces of javadoc where you talk of 32 bits.

minborg: Fixed now

GavinRay97:
Sorry to follow on here, but a question related to this -- you can use the upper 16 bits of a 64-bit pointer for pointer tagging, and also the lower 3-bits if it's on a word-aligned boundary.

This is really useful for building things like atomic tagged pointers, where you can store an ABA-like counter in the 16 bits and use the 3 bits for whatever userdata tag you want:

```
// A word-aligned, atomic tagged pointer.
// Uses both the upper 16 bits for storage, and the lower 3 bits for tagging.
//
//   64                48                32                16
// 0xXXXXXXXXXXXXXXXX  0000000000000000  0000000000000000  0000000000000XXX
//   ^                 ^                                                ^
//   |                 |                                                +-- Tag (3 bits)
//   |                 +-- Pointer (48 bits)
//   +-- Counter (16 bits)
```

Is this possible with `MemoryAddress` and bitmasks currently in Java 🤔

minborg:
> Sorry to follow on here, but a question related to this -- you can use the upper 16 bits of a 64-bit pointer for pointer tagging, and also the lower 3-bits if it's on a word-aligned boundary.
> 
> This is really useful for building things like atomic tagged pointers, where you can store an ABA-like counter in the 16 bits and use the 3 bits for whatever userdata tag you want:
> 
> ```
> // A word-aligned, atomic tagged pointer.
> // Uses both the upper 16 bits for storage, and the lower 3 bits for tagging.
> //
> //   64                48                32                16
> // 0xXXXXXXXXXXXXXXXX  0000000000000000  0000000000000000  0000000000000XXX
> //   ^                 ^                                                ^
> //   |                 |                                                +-- Tag (3 bits)
> //   |                 +-- Pointer (48 bits)
> //   +-- Counter (16 bits)
> ```
> 
> Is this possible with `MemoryAddress` and bitmasks currently in Java 🤔
> 
> (Or does an optimization like that even make sense on the JVM?)

On some systems, mapped memory segments appear as negative longs and thus, is at least using bit 63.

mcimadamore:
> Sorry to follow on here, but a question related to this -- you can use the upper 16 bits of a 64-bit pointer for pointer tagging, and also the lower 3-bits if it's on a word-aligned boundary.
> 
> This is really useful for building things like atomic tagged pointers, where you can store an ABA-like counter in the 16 bits and use the 3 bits for whatever userdata tag you want:
> 
> ```
> // A word-aligned, atomic tagged pointer.
> // Uses both the upper 16 bits for storage, and the lower 3 bits for tagging.
> //
> //   64                48                32                16
> // 0xXXXXXXXXXXXXXXXX  0000000000000000  0000000000000000  0000000000000XXX
> //   ^                 ^                                                ^
> //   |                 |                                                +-- Tag (3 bits)
> //   |                 +-- Pointer (48 bits)
> //   +-- Counter (16 bits)
> ```
> 
> Is this possible with `MemoryAddress` and bitmasks currently in Java thinking
> 
> (Or does an optimization like that even make sense on the JVM?)

On 64 bit platforms you can decide how to create your 64-bit addresses with `MemorySegment.ofAddress`. So, if you want to encode extra bit patterns on the high order bits, you can do so. But then you are also responsible for filtering these out, which require some extra case (because you have a memory segment whose base address is essentially invalid and cannot be used for access operation). Probably you will need/want to encapsulate the logic that deals with this segments. To prevent users from accessing these directly (which will likely lead to a crash).

That said, this has nothing to do with this PR, which is about when the user creates a new segment from a long address on a 32-bit system.

JornVernee (approved):

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
```diff
@@ -1068,8 +1068,8 @@ static MemorySegment ofAddress(long address) {
      * restricted methods, and use safe and supported functionalities, where possible.
      * <p>
      * On 32-bit platforms, the given address value will be normalized such that the
-     * returned memory segment's method {@link MemorySegment#address() address} will always return
-     * zeros for the upper 32 bits of the address.
+     * upper 32 bits of the {@link MemorySegment#address() address} of returned memory segment
```
mcimadamore, line 1071: Ah - late to the party. Is "upper bits" correct here? E.g. doesn't it depend on endianness?

mcimadamore: I'm partly confusing myself, but we should still draw from javadoc in other areas:
https://docs.oracle.com/en/java/javase/16/docs/api/java.base/java/lang/Long.html#highestOneBit(long)
This seems to use highest-order/leftmost for "uppermost bit" :-)

PR #775, opened 2023-01-26
Rework support for unbounded address layouts.

mcimadamore:
Initial push.

### DIFF
--- a/src/java.base/share/classes/java/lang/foreign/MemorySegment.java
+++ b/src/java.base/share/classes/java/lang/foreign/MemorySegment.java
@@ -27,6 +27,7 @@
 package java.lang.foreign;
 
 import java.io.UncheckedIOException;
+import java.lang.foreign.ValueLayout.OfAddress;
 import java.lang.invoke.MethodHandles;
 import java.nio.Buffer;
 import java.nio.ByteBuffer;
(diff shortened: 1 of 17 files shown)


src/java.base/share/classes/java/lang/foreign/MemorySegment.java
```diff
@@ -360,34 +361,48 @@
  * into the lifetime intended for said region of memory by the foreign function that allocated it. The global scope
  * ensures that the obtained segment can be passed, opaquely, to other pointer-accepting foreign functions.
  * <p>
- * To access native zero-length memory segments, clients have two options, both of which are <em>unsafe</em>. Clients
- * can {@linkplain java.lang.foreign.MemorySegment#ofAddress(long, long, SegmentScope) obtain}
- * a <em>new</em> native segment, with new spatial and temporal bounds, as follows:
+ * To access native zero-length memory segments, clients several options, all of which are <em>unsafe</em>.
```
JornVernee, line 364:
```suggestion
 * To access native zero-length memory segments, clients have several options, all of which are <em>unsafe</em>.
```

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
```diff
@@ -360,34 +361,48 @@
...
+ * <p>
+ * First, clients can unsafely resize a zero-length memory segment using an unbounded {@linkplain #asSliceUnbounded(long, long)}
```
JornVernee, line 366:
```suggestion
 * First, clients can unsafely resize a zero-length memory segment using an {@linkplain #asSliceUnbounded(long, long) unbounded}
```

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
```diff
@@ -360,34 +361,48 @@
...
- * MemorySegment foreign = someSegment.get(ValueLayout.ADDRESS, 0); // wrap address into segment (size = 0)
- * MemorySegment segment = MemorySegment.ofAddress(foreign.address(), 4, scope); // create new segment (size = 4)
- * int x = segment.get(ValueLayout.JAVA_INT, 0); //ok
+ * MemorySegment foreign = someSegment.get(ValueLayout.ADDRESS, 0); // wrap address into segment (size = Long.MAX_VALUE)
+ * foreign = foreign.asSliceUnbounded(0, JAVA_INT); // size = 4
+ * int x = foreign.get(ValueLayout.JAVA_INT, 0); //ok
+ *}
+ *
+ * Alternatively, if the size of the foreign segment is known statically, clients can associate a
+ * {@linkplain OfAddress#withTargetLayout(MemoryLayout) target layout} to the address layout used to obtain the
```
JornVernee, line 377:
```suggestion
 * {@linkplain OfAddress#withTargetLayout(MemoryLayout) target layout} with the address layout used to obtain the
```

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
```diff
@@ -360,34 +361,48 @@
...
- * MemorySegment foreign = someSegment.get(ValueLayout.ADDRESS.asUnbounded(), 0); // wrap address into segment (size = Long.MAX_VALUE)
+ * SegmentScope scope = ... // obtains a scope
+ * MemorySegment foreign = someSegment.get(ValueLayout.ADDRESS, 0); // wrap address into segment (size = 0)
+ * foreign = MemorySegment.ofAddress(foreign.address(), 4, scope); // create new segment (size = 4)
  * int x = foreign.get(ValueLayout.JAVA_INT, 0); //ok
  *}
  *
- * Both {@link #ofAddress(long, long, SegmentScope)} and {@link ValueLayout.OfAddress#asUnbounded()} are
+ * Both {@link #asSliceUnbounded(long, long)}, {@link ValueLayout.OfAddress#withTargetLayout(MemoryLayout)}
+ * and {@link #ofAddress(long, long, SegmentScope)} are
```
JornVernee, line 398: There are more than two variants now:
```suggestion
 * All of {@link #asSliceUnbounded(long, long)}, {@link ValueLayout.OfAddress#withTargetLayout(MemoryLayout)}
 * and {@link #ofAddress(long, long, SegmentScope)} are
```

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
```diff
@@ -490,6 +521,53 @@ default MemorySegment asSlice(long offset) {
         return asSlice(offset, byteSize() - offset);
     }
 
+    /**
+     * Returns a slice of this memory segment, at the given offset. The returned segment's address is the address
+     * of this segment plus the given offset; its size is specified by the given argument.
```
JornVernee, line 526:
```suggestion
     * of this segment plus the given offset; its size is specified by the given size argument.
```

src/java.base/share/classes/java/lang/foreign/ValueLayout.java
```diff
@@ -414,27 +417,29 @@ sealed interface OfAddress extends ValueLayout permits ValueLayouts.OfAddressImp
...
-         * @return an unbounded address layout with same characteristics as this layout.
+         * @param layout the target layout.
+         * @return an address layout with same characteristics as this layout, but with the provided target layout.
          * @throws IllegalCallerException If the caller is in a module that does not have native access enabled.
-         * @see #isUnbounded()
+         * @see #targetLayout()
          */
         @CallerSensitive
-        OfAddress asUnbounded();
+        OfAddress withTargetLayout(MemoryLayout layout);
```
JornVernee, line 437: I think it could still be nice to have `asUnbounded` as a shortcut for `withTargetLayout(sequenceLayout(JAVA_BYTE))`.

WDYT?

src/java.base/share/classes/jdk/internal/foreign/abi/Binding.java
```diff
@@ -348,15 +348,15 @@ static Allocate allocate(MemoryLayout layout) {
     }
 
     static BoxAddress boxAddressRaw(long size) {
-        return new BoxAddress(size, false);
+        return new BoxAddress(size, 1, false);
```
JornVernee, line 351: Now that we have the full layout, we might want to add an overload of `boxAddressRaw` that takes a pointee layout, and takes the alignment into account as well. (or just change this method to accept an alignment as well). Most of the call sites of this method are passing the result of `Utils.pointeeSize(layout)` (where layout is the address layout).

Also, the main difference with `boxAddress` seems to be the lack of scope being used. So, maybe it would be better to name this method `boxAddressNoScope`.

JornVernee (commented):
Some initial comments

test/jdk/java/foreign/StdLibTest.java
```diff
@@ -222,10 +223,21 @@ static class Tm {
...
+                    C_INT.withName("year"),
+                    C_INT.withName("wday"),
+                    C_INT.withName("yday"),
+                    C_BOOL.withName("isdst"),
+                    MemoryLayout.paddingLayout(24)
+            );
 
             Tm(MemorySegment addr) {
-                this.base = addr.asSlice(0, SIZE);
+                this.base = addr.asSlice(0, LAYOUT);
```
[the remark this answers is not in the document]
mcimadamore: true - will fix

src/java.base/share/classes/jdk/internal/foreign/abi/Binding.java
mcimadamore: I have fixed this in the latest iteration (but I left the old name) - as I saw test failures because of this

src/java.base/share/classes/java/lang/foreign/ValueLayout.java
mcimadamore: Perhaps we can leave it - but I'm on the fence, as I think having too many options can also be confusing (esp. given that the options are restricted) - should then we also add slicing methods which just remove spatial bounds? Also, it's probably a bit of a naming issue as well - e.g. if the two methods were telescopic, with same name - but one had no argument and can be explained as calling the other, I'd be more ok with it. But having both `withTargetLayout` *and* `asUnbounded` feels like you can do two very different things, as the names are unrelated.

test/jdk/java/foreign/TestAddressDereference.java
```diff
@@ -0,0 +1,156 @@
...
+        } catch (IllegalArgumentException ex) {
+            assertTrue(badAlign);
+            assertTrue(ex.getMessage().contains("alignment constraint for address"));
+        }
+    }
+
+    @Test(dataProvider = "layoutsAndAlignments")
+    public void testNativeUpcallArg(long alignment, ValueLayout layout) throws Throwable {
+        boolean badAlign = layout.byteAlignment() > alignment;
+        if (badAlign) return; // this will crash the JVM (exception occurs when going into the upcall stub)
```
mcimadamore, line 115: Any idea on this @JornVernee ? When the alignment exception is thrown (this happens when we execute the box address bindings before entering the upcall stub), the JVM exits, reporting an exception that points to where the downcall is made, which is quite confusing. I still think the check is worthwhile having (after all a layout is more than just the size), it's a bit unfortunate that in the case of upcalls a failed check ends up terminating the JVM (but then again, if you are really expecting a pointer with certain characteristics, and you are given another - you might just cross fingers and hope it would be ok).

src/java.base/share/classes/java/lang/foreign/ValueLayout.java
mcimadamore: I've decided to just focus on the primitive operations for now. New overloads can be added as required, even at a later point.

JornVernee (approved):

PR #776, opened 2023-01-30
8301360: Add dereference layout paths to memory layout API

mcimadamore:
Now that we have a more general way to attach target layouts to address layouts (see https://git.openjdk.org/panama-foreign/pull/775), it might be useful to add support for layout paths which perform dereference of a given address value. Consider this case:

```
struct b {
   int x;
};

struct a {
    struct b *b;
};
```

Here, our goal is to construct a layout path for "a->b.x".

To do this, we use a *dereference path element* (see `PathElement#dereferenceElement`). These comes in two flavors:

* no argument factory - this assumes that the address layout being dereferenced has some target layout set;
* layout argument factory - this assumes that the address layout being dereferenced does *not* have a target layout set (e.g. it is not always possible to associate a target layout, especially in cyclic cases like a -> b -> a).

When a dereference element is applied to an existing path, a var handle that fetches the address pointed to by the current path is pushed on the stack, and a new layout path is started. This way, dereference path can be used to "chain" together multiple layout paths. All the intermediate memory read operations occur using a *plain* get. For more complex access operation (e.g. if atomic access is required for the whole dereference chain), multiple var handles have to be obtained.

It is an error to pass a dereference path to other layout operations such as `select`, `byte/bitOffset[Handle]`.

---------
### Progress

### Issue
 * [JDK-8301360](https://bugs.openjdk.org/browse/JDK-8301360): Add dereference layout paths to memory layout API

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/776/head:pull/776` \
`$ git checkout pull/776`

`$ git checkout pull/776` \
`$ git pull https://git.openjdk.org/panama-foreign pull/776/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 776`

`$ git pr show -t 776`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/776.diff">https://git.openjdk.org/panama-foreign/pull/776.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
+++ b/src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
@@ -259,7 +259,7 @@ public sealed interface MemoryLayout permits SequenceLayout, GroupLayout, Paddin
      * layout path contains one or more path elements that select multiple sequence element indices
      * (see {@link PathElement#sequenceElement()} and {@link PathElement#sequenceElement(long, long)}).
      * @throws IllegalArgumentException if the layout path contains one or more dereference path elements
-     * (see {@link PathElement#dereferenceElement()} and {@link PathElement#dereferenceElement(MemoryLayout)}).
+     * (see {@link PathElement#dereferenceElement()}).
      * @throws NullPointerException if either {@code elements == null}, or if any of the elements
      * in {@code elements} is {@code null}.
      */
@@ -295,7 +295,7 @@ public sealed interface MemoryLayout permits SequenceLayout, GroupLayout, Paddin
      * @throws IllegalArgumentException if the layout path contains one or more path elements that select
      * multiple sequence element indices (see {@link PathElement#sequenceElement(long, long)}).
      * @throws IllegalArgumentException if the layout path contains one or more dereference path elements
-     * (see {@link PathElement#dereferenceElement()} and {@link PathElement#dereferenceElement(MemoryLayout)}).
+     * (see {@link PathElement#dereferenceElement()}).
      */
     default MethodHandle bitOffsetHandle(PathElement... elements) {
         return computePathOp(LayoutPath.rootPath(this), LayoutPath::offsetHandle,
@@ -312,7 +312,7 @@ public sealed interface MemoryLayout permits SequenceLayout, GroupLayout, Paddin
      * layout path contains one or more path elements that select multiple sequence element indices
      * (see {@link PathElement#sequenceElement()} and {@link PathElement#sequenceElement(long, long)}).
      * @throws IllegalArgumentException if the layout path contains one or more dereference path elements
-     * (see {@link PathElement#dereferenceElement()} and {@link PathElement#dereferenceElement(MemoryLayout)}).
+     * (see {@link PathElement#dereferenceElement()}).
      * @throws UnsupportedOperationException if {@code bitOffset(elements)} is not a multiple of 8.
      * @throws NullPointerException if either {@code elements == null}, or if any of the elements
      * in {@code elements} is {@code null}.
@@ -352,7 +352,7 @@ public sealed interface MemoryLayout permits SequenceLayout, GroupLayout, Paddin
      * @throws IllegalArgumentException if the layout path contains one or more path elements that select
      * multiple sequence element indices (see {@link PathElement#sequenceElement(long, long)}).
      * @throws IllegalArgumentException if the layout path contains one or more dereference path elements
-     * (see {@link PathElement#dereferenceElement()} and {@link PathElement#dereferenceElement(MemoryLayout)}).
+     * (see {@link PathElement#dereferenceElement()}).
      */
     default MethodHandle byteOffsetHandle(PathElement... elements) {
         MethodHandle mh = bitOffsetHandle(elements);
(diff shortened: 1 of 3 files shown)


src/java.base/share/classes/jdk/internal/foreign/LayoutPath.java
```diff
@@ -222,11 +266,18 @@ public MemoryLayout layout() {
...
+
+    private static LayoutPath nestedPath(MemoryLayout layout, long offset, long[] strides, long[] bounds, MethodHandle[] derefAdapters, LayoutPath encl) {
+        return new LayoutPath(layout, offset, strides, bounds, derefAdapters, encl);
     }
 
-    private static LayoutPath nestedPath(MemoryLayout layout, long offset, long[] strides, long[] bounds, LayoutPath encl) {
-        return new LayoutPath(layout, offset, strides, bounds, encl);
+    private static LayoutPath derefPath(MemoryLayout layout, MethodHandle handle, LayoutPath encl) {
+        MethodHandle[] handles = new MethodHandle[encl.derefAdapters.length + 1];
+        System.arraycopy(encl.derefAdapters, 0, handles, 0, encl.derefAdapters.length);
```
JornVernee, line 278: This could be simplified using Arrays.copyOf:

```suggestion
        MethodHandle[] handles = Arrays.copyOf(encl.derefAdapters, encl.derefAdapter.length + 1);
```

(no need for separate call to System.arraycopy)

src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
```diff
@@ -594,6 +629,36 @@ static PathElement sequenceElement() {
             return new LayoutPath.PathElementImpl(PathKind.SEQUENCE_ELEMENT,
                                                   LayoutPath::sequenceElement);
         }
+
+        /**
+         * Returns a path element which dereferences an address layout as its
+         * {@linkplain OfAddress#targetLayout() target layout} (where set).
+         * The path element returned by this method does not alter the number of free dimensions of any path
+         * that is combined with such element. Using this path layout to dereference an address layout
+         * that has no target layout results in an {@link IllegalArgumentException} (e.g. when
```
JornVernee, line 638: This IAE should be specified on `varHandle` as well.

src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
```diff
@@ -594,6 +629,36 @@ static PathElement sequenceElement() {
...
+        static PathElement dereferenceElement() {
+            return new LayoutPath.PathElementImpl(PathKind.DEREF_ELEMENT,
+                    LayoutPath::derefElement);
+        }
+
+        /**
+         * Returns a path element which dereferences an address layout as the given layout.
+         * The path element returned by this method does not alter the number of free dimensions of any path
+         * that is combined with such element. Using this path layout to dereference an address layout
+         * that already has a target layout results in an {@link IllegalArgumentException} (e.g. when
```
JornVernee, line 652: Same here

src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
```diff
@@ -594,6 +629,36 @@ static PathElement sequenceElement() {
...
+         * a var handle is {@linkplain #varHandle(PathElement...) obtained}).
+         *
+         * @param layout the
+         * @return a path element which dereferences an address layout.
+         */
+        static PathElement dereferenceElement(MemoryLayout layout) {
```
JornVernee, line 658: This method needs to be restricted, I believe. Since it performs a resizing of a MemorySegment. The other overload of `dereferenceElement` is already covered because attaching a pointee layout to an address layout is restricted already.

Consider the following:

```
StructLayout hasAddr = MemoryLayout.structLayout(ADDRESS.withName("addr"));

MemorySegment rawSeg = MemorySegment.allocateNative(0, SegmentScope.auto()); // inaccessible segment
// rawSeg.get(JAVA_INT, 0); // not allowed
MemorySegment addrHolder = MemorySegment.allocateNative(hasAddr, SegmentScope.auto());
addrHolder.set(ADDRESS, 0, rawSeg); // fine
VarHandle derefHandle = hasAddr.varHandle(groupElement("addr"), dereferenceElement(JAVA_INT));
int x = derefHandle.get(addrHolder); // oops!
```

Assuming I eye-balled this code correctly, I think this is problematic?

JornVernee (commented):
Implementation looks good. I think there's some issue with resizing the segment for the layout accepting overload though (see inline comment).

src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
mcimadamore: Sure - the layout accepting one needs to be restricted - I started off with only the other one (which doesn't need to be restricted), then added the layout accepting one, and forgot to add the restricted check. Any way - as a validation, do we agree that it is handy to have both forms? If you have a layout like A -> B -> A, there is no way to attach target layouts statically - which is the main reason why I added that variant.

JornVernee: Yes, it seems good to have it for that case.

mcimadamore: In the end, I've decided to drop the overload, at least for now. While it could be useful in some pathological situation, the other method should be good for the vast majority of cases. It occurred to me that there are other possible ways to attack the problem of cyclic references - for instance, we could still require a target layout on the address - but then just perform a "layout cast" when dereferencing (which then got me thinking as to whether a more general "cast element" would be required). Overall, I believe it would be better to think a bit more on this one.

JornVernee (approved):

PR #778, opened 2023-02-02
8301262: Simplify the CaptureCallState support

JornVernee:
Simplify capture call state support based on Maurizio's suggestion.

We currently compute a layout to write the captured state into based on the values that the user wants to capture. However, we could simplify the API by using the same layout every time, that can hold every value that someone might want to capture. This is only 12 bytes, so the memory usage is not a big deal. The important part is to not read every thread local value every time, which works just as it does today.

The second issue in today's implementation, is that it is still possible to create a capture state linker option that captures `GetLastError` on Linux. But, this doesn't make any sense. In the new implementation the GetLastError and WSAGetLastError options are only available on Windows.

---------
### Progress

### Issue
 * [JDK-8301262](https://bugs.openjdk.org/browse/JDK-8301262): Simplify the CaptureCallState support

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/778/head:pull/778` \
`$ git checkout pull/778`

`$ git checkout pull/778` \
`$ git pull https://git.openjdk.org/panama-foreign pull/778/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 778`

`$ git pr show -t 778`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/778.diff">https://git.openjdk.org/panama-foreign/pull/778.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/java/lang/foreign/Linker.java
+++ b/src/java.base/share/classes/java/lang/foreign/Linker.java
@@ -35,7 +35,6 @@ import jdk.internal.reflect.Reflection;
 
 import java.lang.foreign.ValueLayout.OfAddress;
 import java.lang.invoke.MethodHandle;
-import java.util.Arrays;
 import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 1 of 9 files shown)


src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -303,18 +301,51 @@ static Option firstVariadicArg(int index) {
...
+         *     handle.invoke(capturedState);
+         *     int errno = errnoHandle.get(capturedState);
+         *     // use errno
+         * }
+         * }
          *
          * @param capturedState the names of the values to save.
-         * @see CaptureCallState#supported()
+         * @see #capturedStateLayout()
          */
```
JornVernee, line 335: I've copied here most of the doc that was on the class before.

mcimadamore (approved):
Thanks for fixing this - looks really good!

PR #780, opened 2023-02-03
8300294: Add tests for by-value unions and structs with nested fixed-length arrays

JornVernee:
I've added some tests for by-value unions, structs/unions nested into other structs, and aggregates with nested inline arrays.

These test cases are taken from a fuzzer I wrote a while ago. So, they are a bit random in terms of the generated layouts, but should nonetheless  serve as a basic test for the cases that we miss with TestDowncall/TestUpcall today. This mostly helps give some coverage of the linker's classification code for these cases.

I've also pulled in some test value generation code, which can generate random test values when given a MemoryLayout, allocator and random generator. (if it seems useful, I can re-write the existing tests to use this generation code as well, but I'd like to save that for a followup PR).

Unfortunately, the fallback linker doesn't seem to be able to reliably handle by-value unions. It's a know issue with libffi. I've filed: https://bugs.openjdk.org/browse/JDK-8301800 for now. I've changed the implementation to throw an IAE, which was already specified for `Linker::downcallHandle/upcallStub`: 

https://github.com/openjdk/panama-foreign/blob/f61f3a31af4976d0e64d3bfa72cda95b501e2a7d/src/java.base/share/classes/java/lang/foreign/Linker.java#L232-L235

There was also a bug in the classification of HFAs on AArch64 which I've fixed. The issue was that structs with float/double fields that are nested into arrays or other structs/unions should also be counted as HFAs. But, the correct classification logic only accounted for the 'flat' case, where a struct only contains ValueLayouts.

---------
### Progress

### Issue
 * [JDK-8300294](https://bugs.openjdk.org/browse/JDK-8300294): Add tests for by-value unions and structs with nested fixed-length arrays

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/780/head:pull/780` \
`$ git checkout pull/780`

`$ git checkout pull/780` \
`$ git pull https://git.openjdk.org/panama-foreign pull/780/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 780`

`$ git pr show -t 780`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/780.diff">https://git.openjdk.org/panama-foreign/pull/780.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/CallArranger.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/CallArranger.java
@@ -264,10 +264,17 @@ public abstract class CallArranger {
         Windows, VF     | CW in regs       | CW split between regs and stack | CW on the stack
          */
         StructStorage[] structStorages(GroupLayout layout, boolean forHFA) {
-            int regType = forHFA ? StorageType.VECTOR : StorageType.INTEGER;
             int numChunks = (int)Utils.alignUp(layout.byteSize(), MAX_COPY_SIZE) / MAX_COPY_SIZE;
-            List<MemoryLayout> scalarLayouts = forHFA ? TypeClass.scalarLayouts(layout) : null;
-            int requiredStorages = forHFA ? scalarLayouts.size() : numChunks;
+
+            int regType = StorageType.INTEGER;
+            List<MemoryLayout> scalarLayouts = null;
+            int requiredStorages = numChunks;
+            if (forHFA) {
+                regType = StorageType.VECTOR;
+                scalarLayouts = TypeClass.scalarLayouts(layout);
+                requiredStorages = scalarLayouts.size();
+            }
+
             boolean hasEnoughRegisters = hasEnoughRegisters(regType, requiredStorages);
 
             // For the ABI variants that pack arguments spilled to the
--- a/src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/CallArranger.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/CallArranger.java
@@ -266,7 +266,8 @@ public abstract class CallArranger {
         StructStorage[] structStorages(GroupLayout layout, boolean forHFA) {
             int regType = forHFA ? StorageType.VECTOR : StorageType.INTEGER;
             int numChunks = (int)Utils.alignUp(layout.byteSize(), MAX_COPY_SIZE) / MAX_COPY_SIZE;
-            int requiredStorages = forHFA ? layout.memberLayouts().size() : numChunks;
+            List<MemoryLayout> scalarLayouts = forHFA ? TypeClass.scalarLayouts(layout) : null;
+            int requiredStorages = forHFA ? scalarLayouts.size() : numChunks;
             boolean hasEnoughRegisters = hasEnoughRegisters(regType, requiredStorages);
 
             // For the ABI variants that pack arguments spilled to the
@@ -299,7 +300,7 @@ public abstract class CallArranger {
                 ValueLayout copyLayout;
                 if (isFieldWise) {
                     // We should only get here for HFAs, which can't have padding
-                    copyLayout = (ValueLayout) layout.memberLayouts().get(i);
+                    copyLayout = (ValueLayout) scalarLayouts.get(i);
                 } else {
                     // chunk-wise copy
                     long copySize = Math.min(layout.byteSize() - offset, MAX_COPY_SIZE);
--- a/src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/TypeClass.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/TypeClass.java
@@ -63,14 +63,14 @@ public enum TypeClass {
 
     static List<MemoryLayout> scalarLayouts(GroupLayout gl) {
         List<MemoryLayout> out = new ArrayList<>();
-        scalarLayoutsR(out, gl);
+        scalarLayoutsInternal(out, gl);
         return out;
     }
 
-    private static void scalarLayoutsR(List<MemoryLayout> out, GroupLayout gl) {
+    private static void scalarLayoutsInternal(List<MemoryLayout> out, GroupLayout gl) {
         for (MemoryLayout member : gl.memberLayouts()) {
             if (member instanceof GroupLayout memberGl) {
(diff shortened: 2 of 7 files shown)


src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/TypeClass.java
```diff
@@ -58,15 +61,38 @@ static boolean isRegisterAggregate(MemoryLayout type) {
...
+        for (MemoryLayout member : gl.memberLayouts()) {
+            if (member instanceof GroupLayout memberGl) {
+                scalarLayoutsR(out, memberGl);
+            } else if (member instanceof SequenceLayout memberSl) {
+                for (long i = 0; i < memberSl.elementCount(); i++) {
+                    out.add(memberSl.elementLayout());
+                }
+            } else {
+                // padding or value layouts
+                out.add(member);
```
JornVernee, line 80: Note here that groups with padding will be rejected later when each field is checked in `isHomogeneousFloatAggregate`

src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/CallArranger.java
```diff
@@ -266,7 +266,8 @@ Homogeneous float aggregates (HFAs) can be copied in a field-wise manner, i.e. t
         StructStorage[] structStorages(GroupLayout layout, boolean forHFA) {
             int regType = forHFA ? StorageType.VECTOR : StorageType.INTEGER;
             int numChunks = (int)Utils.alignUp(layout.byteSize(), MAX_COPY_SIZE) / MAX_COPY_SIZE;
-            int requiredStorages = forHFA ? layout.memberLayouts().size() : numChunks;
+            List<MemoryLayout> scalarLayouts = forHFA ? TypeClass.scalarLayouts(layout) : null;
```
mcimadamore, line 269: Perhaps, instead of two conditionals, it would be better to use a simple `if` ? I'm a bit worried about one path being `null` which then works in the subsequent conditional because the leading condition is the same. Seems a bit brittle.

src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/TypeClass.java
```diff
@@ -58,15 +61,38 @@ static boolean isRegisterAggregate(MemoryLayout type) {
         return type.bitSize() <= MAX_AGGREGATE_REGS_SIZE * 64;
     }
 
+    static List<MemoryLayout> scalarLayouts(GroupLayout gl) {
+        List<MemoryLayout> out = new ArrayList<>();
+        scalarLayoutsR(out, gl);
+        return out;
+    }
+
+    private static void scalarLayoutsR(List<MemoryLayout> out, GroupLayout gl) {
```
mcimadamore, line 70: where does the `R` in the name comes from?

src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/TypeClass.java
mcimadamore: Does the order in which you collect the layouts matter? E.g. this will do a depth-first visit of the layout tree - I see that the code above does an indexed get on the scalar layout list, so I believe the order does matter.

My impression is that it's ok: the depth-first visit returns layouts in the order in which they are laid out in memory, which seems to make sense for an ABI.

test/jdk/java/foreign/nested/TestNested.java
```diff
@@ -0,0 +1,260 @@
...
+
+    @DataProvider
+    public static Object[][] nestedLayouts() {
+        List<GroupLayout> layouts = List.of(
+                S1, U1, U17, S2, S3, S4, S5, S6, U2, S7, U3, U4, U5, U6, U7, S8, S9, U8, U9, U10, S10,
+                U11, S11, U12, S12, U13, U14, U15, S13, S14, U16, S15);
+        return layouts.stream().map(l -> new Object[]{l}).toArray(Object[][]::new);
+    }
+
+    static final StructLayout S1 = MemoryLayout.structLayout(
```
mcimadamore, line 86: Wouldn't it be better to have these structs defined inline inside the `List.of` call in the provider? That should make it a little easier to just add a new case.

mcimadamore (commented):
Nice test!

src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/CallArranger.java
JornVernee: That's also a possibility, I can put the initialization of `regType` into the `if` as well. I'll do that.

src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/TypeClass.java
JornVernee: R stands for 'recursive'. I'll type it out. (or, maybe another name like `scalarLayoutsInternal` is better. That seems to be more common in JDK code at least)

src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/TypeClass.java
JornVernee: It doesn't matter for the classification, all the ValueLayouts just need to be the same. Later when computing the struct storages in CallArranger, it also doesn't really matter. There we look at the carrier, byte size, and maybe alignment. But since all the layouts are again the same, they could technically be processed in any order.

test/jdk/java/foreign/nested/TestNested.java
JornVernee: I don't think so, because there are struct and union layouts that are nested inside other layouts, but then _also_ are a test case themselves. So, I think we need some intermediate variables either way (or copy paste), and then it seems cleaner to do that for all the layouts.

src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/TypeClass.java
mcimadamore: I was about to suggest `internal` - but then I wasn't sure if it was derived from an ABI specific term.

test/jdk/java/foreign/nested/TestNested.java
mcimadamore: I see - missed that, thanks.

mcimadamore (approved):

PR #782, opened 2023-02-07
Fix test failures following 8300201

mcimadamore:
The fix for JDK-8300201 has introduced an initialization issue involving `NativeMemorySegmentImpl`. The constructor of that class is accessing the `UNSAFE` static field (declared in same class), but that field seems to be set to `null` in some cases (probably because of a static initializer loop). This is something I also have observed when working on https://git.openjdk.org/panama-foreign/pull/781

The fix is to just use the Unsafe static field for getting the address size, and avoid the `UNSAFE` field in the constructor. More investigation might follow, to pinpoint exactly where the initialization problem is.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/782/head:pull/782` \
`$ git checkout pull/782`

`$ git checkout pull/782` \
`$ git pull https://git.openjdk.org/panama-foreign pull/782/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 782`

`$ git pr show -t 782`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/782.diff">https://git.openjdk.org/panama-foreign/pull/782.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/jdk/internal/foreign/NativeMemorySegmentImpl.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/NativeMemorySegmentImpl.java
@@ -54,7 +54,7 @@ public sealed class NativeMemorySegmentImpl extends AbstractMemorySegmentImpl pe
     @ForceInline
     NativeMemorySegmentImpl(long min, long length, boolean readOnly, SegmentScope scope) {
         super(length, readOnly, scope);
-        this.min = (UNSAFE.addressSize() == 4)
+        this.min = (Unsafe.ADDRESS_SIZE == 4)
                 // On 32-bit systems, normalize the upper unused 32-bits to zero
                 ? min & 0x0000_0000_FFFF_FFFFL
                 // On 64-bit systems, all the bits are used


JornVernee (approved):
Thanks for fixing. 

I think the virtual call might be triggering to earlier class loading during verification, to see whether the receiver argument on the stack is actually compatible with the receiver type of the declared called method.

Unsafe depends on ByteBuffer, which then leads back to MemorySegment, and from there to NativeMemorySegmentImpl I'm guessing.

mcimadamore:
> Thanks for fixing.
> 
> I think the virtual call might be triggering to earlier class loading during verification, to see whether the receiver argument on the stack is actually compatible with the receiver type of the declared called method.
> 
> Unsafe depends on ByteBuffer, which then leads back to MemorySegment, and from there to NativeMemorySegmentImpl I'm guessing.

I don't think that's it - with the other PR (with the API changes) doing `Arena.ofAuto().allocate(...)` was enough to reproduce. No ByteBuffers.

PR #791, opened 2023-02-14
8302346: Lift upcall sharing mechanism to AbstractLinker

JornVernee:
Lift the upcall sharing mechanism to AbstractLinker, where it can live next to the similar mechanism for down calls. This also allows the fallback linker to use this mechanism.

Instead of an upcall stub, AbstractLinker::arrangeDowncall now return an UpcallStubFactory, which is a callback accepting a target MethodHandle and Arena, which are then used to construct the actual upcall stub.

Perhaps the trickiest part of this patch is that we have to simulate the effects of `SharedUtils::adaptUpcallForIMR` on the target method type. I've added a new method in SharedUtils, called `computeUpcallIMRType` for that, which mimics the shape of the `adaptUpcallForIMR` method.

Since the shape of most of the arrangeUpcall methods was very similar, I factored them out into a helper method in SharedUtils as well.

---------
### Progress

### Issue
 * [JDK-8302346](https://bugs.openjdk.org/browse/JDK-8302346): Lift upcall sharing mechanism to AbstractLinker

### Reviewers
 * [Maurizio Cimadamore](https://openjdk.org/census#mcimadamore) (@mcimadamore - Committer) ⚠️ Review applies to [6bd745f8](https://git.openjdk.org/panama-foreign/pull/791/files/6bd745f83cf18340bf3554ce39d2da168522e13c)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/791/head:pull/791` \
`$ git checkout pull/791`

`$ git checkout pull/791` \
`$ git pull https://git.openjdk.org/panama-foreign pull/791/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 791`

`$ git pr show -t 791`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/791.diff">https://git.openjdk.org/panama-foreign/pull/791.diff</a>

</details>


src/java.base/share/classes/jdk/internal/foreign/abi/SharedUtils.java
```diff
@@ -153,6 +154,51 @@ public static MethodHandle adaptUpcallForIMR(MethodHandle target, boolean dropRe
...
+
+    public static UpcallStubFactory arrangeUpcallHelper(MethodType targetType, boolean isInMemoryReturn, boolean dropReturn,
+                                                        ABIDescriptor abi, CallingSequence callingSequence) {
+        if (isInMemoryReturn) {
+            targetType = computeUpcallIMRType(targetType, dropReturn);
+        }
+
+        UpcallStubFactory factory = UpcallLinker.makeFactory(targetType, abi, callingSequence);
+
+        if (isInMemoryReturn) {
```
mcimadamore, line 191: Question: does this logic belong here? E.g. we have a method that is used to create an upcall factory, namely `UpcallLinker.makeFactory`, but it turns out that if we have `isInMemoryReturn` some other factory should be used instead. Shouldn't we tweak the code so that `makeFactory` always does the right thing?

src/java.base/share/classes/jdk/internal/foreign/abi/SharedUtils.java
```diff
@@ -153,6 +154,51 @@ public static MethodHandle adaptUpcallForIMR(MethodHandle target, boolean dropRe
...
+    private static MethodType computeCollectArgumentsType(MethodType target, int insertIndex, MethodType collector) {
+        if (collector.returnType() != void.class){
+            // for non-void collector, the returned value is inserted into the argument list passed to the target
+            target = target.dropParameterTypes(insertIndex, insertIndex + 1);
+        }
+        return target.insertParameterTypes(insertIndex, collector.parameterArray()); // insert targetType parameters instead
+    }
+
+    // this just computes the adjusted type
+    private static MethodType computeUpcallIMRType(MethodType targetType, boolean dropReturn) {
```
mcimadamore, line 166: Crazy idea - what if we created a fake method handle of the given target type - then just called adaptUpcallForIMR on it, and got the resulting type? I'm a bit worried for the type logic and the MH adaptation logic to go out of sync, and require changing things in two places.

src/java.base/share/classes/jdk/internal/foreign/abi/SharedUtils.java
JornVernee: It's a matter of organization. I think we should keep `UpcallLinker` ABI agnostic, and in memory returns are figments of an ABI. (note e.g. that the `CallingSequence` we pass to `makeFactory` has no notion of this in memory return. Not to be confused with the return buffer)

I see the additional step for in memory return not as using another factory, but as providing some pre-processing of the MethodHandle, before it's passed to `makeFactory`. The pre-processing is ABI-specific, so should live outside of UpcallLinker, IMO.

src/java.base/share/classes/jdk/internal/foreign/abi/SharedUtils.java
JornVernee: Nice idea! This is trivial to implement as well using `MethodHandles::empty`.

mcimadamore (approved):
Nice simplification!

PR #794, opened 2023-02-14
Remove PlatformLayout

minborg:
This PR suggests removing the class `PlatformLayout`.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/794/head:pull/794` \
`$ git checkout pull/794`

`$ git checkout pull/794` \
`$ git pull https://git.openjdk.org/panama-foreign pull/794/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 794`

`$ git pr show -t 794`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/794.diff">https://git.openjdk.org/panama-foreign/pull/794.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/jdk/internal/foreign/PlatformLayouts.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/PlatformLayouts.java
@@ -30,17 +30,16 @@ import java.lang.foreign.ValueLayout;
 
 public final class PlatformLayouts {
 
-    private PlatformLayouts() {
-        //just the one
-    }
+    // Suppresses default constructor, ensuring non-instantiability.
+    private PlatformLayouts() {}
 
     /**
      * This class defines layout constants modelling standard primitive types supported by the x64 SystemV ABI.
      */
     public static final class SysV {
-        private SysV() {
-            //just the one
-        }
+
+        // Suppresses default constructor, ensuring non-instantiability.
+        private SysV() {}
 
         /**
          * The {@code bool} native type.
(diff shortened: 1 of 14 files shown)


JornVernee:
I've been thinking for a while now that `PlatformLayout` should just be removed: https://bugs.openjdk.org/browse/JDK-8291499 It is a remnant of a time where we exposed layout constant for each platform.

In the implementation we only use one or two layouts for each platform. I think `PlatformLayout` should just be removed, and the needed layouts moved to the particular `CallArranger` class where they are needed. So, if you want to work on `PlatformLayouts` I suggest removing it instead :)

Last time I looked into this, the tricky part was that there was a test that used these layouts as test values. That test will have to do something else.

src/java.base/share/classes/jdk/internal/foreign/PlatformLayouts.java
```diff
@@ -233,42 +215,44 @@ private RISCV64() {
...
-                .withTargetLayout(MemoryLayout.sequenceLayout(C_CHAR));
+        public static final ValueLayout.OfAddress C_POINTER = cPointer(C_CHAR);
 
-        /**
-         * The {@code va_list} native type, as it is passed to a function.
-         */
-        public static final ValueLayout.OfAddress C_VA_LIST = RISCV64.C_POINTER;
     }
+
+    private static ValueLayout.OfAddress cPointer(ValueLayout targetLayoutElementLayout) {
```
minborg, line 252: Maybe there is a better name? `elementLayout` perhaps?

mcimadamore: I think you can just omit the parameter here - and use JAVA_BYTE. We're not interested in what the layout is - the main thing is that it's unbounded (and that it has minimum alignment). Even using a padding layout of Long.MAX_VALUE would work :-)

test/jdk/java/foreign/callarranger/platform/PlatformLayouts.java
```diff
@@ -55,53 +54,47 @@ private SysV() {
...
          */
-        public static final ValueLayout.OfDouble C_DOUBLE = ValueLayout.JAVA_DOUBLE.withBitAlignment(64);
+        public static final ValueLayout.OfDouble C_DOUBLE = ValueLayout.JAVA_DOUBLE;
 
         /**
          * The {@code T*} native type.
          */
-        public static final ValueLayout.OfAddress C_POINTER = ValueLayout.ADDRESS.withBitAlignment(64)
-                .withTargetLayout(MemoryLayout.sequenceLayout(C_CHAR));
+        public static final ValueLayout.OfAddress C_POINTER = cPointer(C_CHAR);
```
mcimadamore, line 87: Can you use SharedUtils.C_POINTER here? It's the same thing.

mcimadamore (approved):

JornVernee (approved):

mcimadamore:
Btw, why is this PR called "Remove PlatformLayout" given that it does not remove it?

PR #797, opened 2023-02-15
8302556: Find better way to create unsafe native segments

mcimadamore:
Consider the following code, which is quite common when trying to retroactively secure memory segments originating in native code:

```java
MemorySegment malloc(long size, Arena arena) {
     MemorySegment raw = <MALLOC_HANDLE>.invokeExact(size);
     return MemorySegment.ofAddress(raw.address(), size, arena, () -> StdLib.free(raw));
}
```

There are few issues with this code:

* Clients need a throwaway local variable (`raw`) where to store the raw segment returned from the linker API;
* to create a new segment with `MemorySegment::ofAddress`, one has to retrieve the address from the old segment and pass it to the factory (see call to `raw.address()`);
* the cleanup action has to carefully refer to the `raw` segment - that's because the new segment will be invalidated when the arena is closed, so only `raw` is guaranteed to still be alive at that point.

This patch introduces an API change to cope with the issues above. More specifically, this patch introduces a new *instance* method, namely `MemorySegment::reinterpret` which can be used to customize an existing native segment by giving it a new size and scope. `MemorySegment::ofAddress` is still there, but this patch removes all the overloads: effectively now `ofAddress` is only used to turn a `long` into a (zero-length) native memory segment.

With the changes in this patch the above code becomes:

```java
MemorySegment malloc(long size, Arena arena) {
     return <MALLOC_HANDLE>.invokeExact(size);
                           .reinterpret(size, arena.scope(), StdLib::free);
}
```

Not only the new code is more "fluent" than the old one (no extra local variable declaration required), but the story for attaching custom cleanup action has been improved significantly: instead of just taking a `Runnable` the new API takes a `Consumer<MemorySegment>` which is passed a _fresh_ zero-length memory segment whose scope is always alive (in other words, an _alias_ of the memory segment to be cleaned up). This seems acceptable for a restricted method that is meant to be used in very specific situations, and leads to much cleaner code.

There are several overloads for `reinterpret`, to reset size, scope and both size *and* scope. Note that, since `reinterpret` accepts a scope (and not just an `Arena`), it is possible to use this method to rescue the co-allocation use case (albeit at a lower level) - that is, when reading a pointer from another memory segment, a client could set the scope on the read segment to be the same as that of the containing segment.


mcimadamore:
Sample javadoc:
http://cr.openjdk.java.net/~mcimadamore/panama/segment_reinterpret_v2/

src/java.base/share/classes/jdk/internal/foreign/abi/UpcallStubs.java
```diff
@@ -50,13 +50,13 @@ private static void freeUpcallStub(long stubAddress) {
...
-        MemorySessionImpl.toMemorySession(scope).addOrCleanupIfFail(new MemorySessionImpl.ResourceList.ResourceCleanup() {
+    static MemorySegment makeUpcall(long entry, Arena arena) {
+        MemorySessionImpl.toMemorySession(arena).addOrCleanupIfFail(new MemorySessionImpl.ResourceList.ResourceCleanup() {
             @Override
             public void cleanup() {
                 freeUpcallStub(entry);
             }
         });
-        return MemorySegment.ofAddress(entry, 0, scope);
+        return MemorySegment.ofAddress(entry).reinterpret(arena.scope(), null);
```
mcimadamore, line 60: In principle we could just express this using public API if we tweak the semantics for cleanup action to always run in case the scope is found to be already closed.

src/java.base/share/classes/jdk/internal/foreign/SystemLookup.java
```diff
@@ -87,14 +89,17 @@ private static SymbolLookup makeWindowsLookup() {
                     libLookup(libs -> libs.load(jdkLibraryPath("syslookup")));
 
             int numSymbols = WindowsFallbackSymbols.values().length;
-            MemorySegment funcs = MemorySegment.ofAddress(fallbackLibLookup.find("funcs").orElseThrow().address(),
-                ADDRESS.byteSize() * numSymbols, Arena.global());
+            MemorySegment funcs = fallbackLibLookup.find("funcs").orElseThrow()
+                    .reinterpret(ADDRESS.byteSize() * numSymbols);
```
JornVernee, line 93: I realize now that this might be better expressed as a static final `SequenceLayout` in `WindowsFallbackSymbols`, since the number of symbols doesn't change. WDYT?

src/java.base/share/classes/jdk/internal/foreign/abi/UpcallStubs.java
JornVernee: Yeah, I think that might be better as semantics in general. If the attach fails, don't leak, and run the cleanup action.

test/jdk/java/foreign/TestSlices.java
```diff
@@ -128,7 +128,7 @@ public void testUnboundedSlice() {
             MemorySegment segment = arena.allocate(MemoryLayout.sequenceLayout(2, ValueLayout.JAVA_INT)); // size = 8
             MemorySegment slice = segment.asSlice(0, ValueLayout.JAVA_INT); // size = 4
             assertThrows(IndexOutOfBoundsException.class, () -> slice.getAtIndex(ValueLayout.JAVA_INT, 1));
-            MemorySegment unbounded = slice.asUnbounded();
+            MemorySegment unbounded = slice.reinterpret(Long.MAX_VALUE, arena.scope(), null);
```
JornVernee, line 131: Looking at this, I'm wondering if the whole test `testUnboundedSlice` is still needed at all, now that `asUnbounded` is removed. I'd say just remove the test.

test/micro/org/openjdk/bench/java/lang/foreign/JavaLayouts.java
```diff
@@ -23,6 +23,7 @@
 
 package org.openjdk.bench.java.lang.foreign;
 
+import java.lang.foreign.MemoryLayout;
```
JornVernee, line 26: Spurious import?
```suggestion
```

src/java.base/share/classes/jdk/internal/foreign/abi/UpcallStubs.java
mcimadamore: Not super sure how to specify this though - e.g. should we say that the cleanup is always run when the method completes abnormally? Possible failure causes are:
* the scope is no longer alive (this is what we care about)
* the size is negative
* the method is called on a segment that is not native
* the method is called and restricted access is not enabled.

Note 100% sure *when* we should say we would attempt to run the cleanup action. Kind of tempted leave it as is.

src/java.base/share/classes/jdk/internal/foreign/SystemLookup.java
```diff
@@ -86,15 +86,17 @@ private static SymbolLookup makeWindowsLookup() {
             SymbolLookup fallbackLibLookup =
                     libLookup(libs -> libs.load(jdkLibraryPath("syslookup")));
 
-            int numSymbols = WindowsFallbackSymbols.values().length;
-            MemorySegment funcs = MemorySegment.ofAddress(fallbackLibLookup.find("funcs").orElseThrow().address(),
-                ADDRESS.byteSize() * numSymbols, Arena.global());
+            MemorySegment funcs = fallbackLibLookup.find("funcs").orElseThrow()
+                    .reinterpret(WindowsFallbackSymbols.LAYOUT.byteSize());
```
JornVernee, line 90: Nice! Thanks.

src/java.base/share/classes/jdk/internal/foreign/SystemLookup.java
```diff
@@ -201,5 +203,8 @@ static WindowsFallbackSymbols valueOfOrNull(String name) {
                 return null;
             }
         }
+
+        static SequenceLayout LAYOUT = MemoryLayout.sequenceLayout(
```
JornVernee, line 207: Could be `final` as well.

JornVernee (approved):

src/java.base/share/classes/jdk/internal/foreign/abi/UpcallStubs.java
JornVernee: Ok, probably better to leave it as is for now.

PR #798, opened 2023-02-15
Simplify ValueLayouts

minborg:
This PR proposes a more shortened and simplified scheme when implementing value layouts. 

This provides a significant reduction of the code size. A new threeary duplicator is introduced. In the solution below, I sugest using a function (`ValueLayoutDuplicator<V>`) which reduces code but arguably makes it a bit more complicated. Another alternative would be to simply mandate a corresponding abstract method and Implement an explicit delegator to the constructor for each type. 

Let me know your thoughts around how to balance this.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/798/head:pull/798` \
`$ git checkout pull/798`

`$ git checkout pull/798` \
`$ git pull https://git.openjdk.org/panama-foreign pull/798/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 798`

`$ git pr show -t 798`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/798.diff">https://git.openjdk.org/panama-foreign/pull/798.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/jdk/internal/foreign/layout/ValueLayouts.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/layout/ValueLayouts.java
@@ -93,7 +93,7 @@ public final class ValueLayouts {
          */
         public final V withOrder(ByteOrder order) {
             Objects.requireNonNull(order);
-            return duplicator().dup(order, bitAlignment(), name());
+            return dup(order, bitAlignment(), name());
         }
 
         @Override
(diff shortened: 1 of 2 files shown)


src/java.base/share/classes/jdk/internal/foreign/layout/ValueLayouts.java
```diff
@@ -143,7 +142,16 @@ public int hashCode() {
         }
 
         @Override
-        abstract V dup(long bitAlignment, Optional<String> name);
+        final V dup(long bitAlignment, Optional<String> name) {
```
mcimadamore, line 145: question: instead of using a duplicator object, can't we have a new `dup` overload here with the required arguments, rewire the old `dup` to the new one, and then override as appropriate in all the subclasses? That would still allow you to define `withOrder` in a single place (which, AFAICS appears to be the main simplification?)

minborg: > question: instead of using a duplicator object, can't we have a new `dup` overload here with the required arguments, rewire the old `dup` to the new one, and then override as appropriate in all the subclasses? That would still allow you to define `withOrder` in a single place (which, AFAICS appears to be the main simplification?)

This is exactly what I am proposing as an alternative in the PR header and perhaps that is easier to read. I will update the proposal accordingly.

mcimadamore: Sorry - I did miss that comment.

mcimadamore (approved):
Looks very nice - thanks!

PR #806, opened 2023-02-24
8303017: Passing by-value structs whose size is not power of 2 doesn't work on all platforms

JornVernee:
Fix passing of by-value structs whose size is not a power of 2.

The issue is that currently we try to do loads using a ValueLayout that is the size of the nearest power of two that can fit the struct (or part of it, if it is passed in multiple registers). For instance, for a struct of size 6, we try to load its value using `ValueLayout.OfLong`, which is size 8, and thus produce an out of bounds error. A similar issue applies to writes.

For the solution I've implemented in this patch, I've attached an explicit byte width to BufferLoads/Stores, which indicates the size of the value we want to load/store. The type that is produced by the binding is still the same. For example, loading a struct of size 6 is implemented as an `int` load and a `short` load, which are then combined into a `long`, instead of attempting to do a single `long` load (and failing). This allows us to avoid doing an out of bounds access.

I've added a new test that tests a bunch of structs with varying byte sizes being passed in registers and on the stack. Using a nested `char[]` to precisely tweak the byte size of each struct.

---------
### Progress

### Issue
 * [JDK-8303017](https://bugs.openjdk.org/browse/JDK-8303017): Passing by-value structs whose size is not power of 2 doesn't work on all platforms

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/806/head:pull/806` \
`$ git checkout pull/806`

`$ git checkout pull/806` \
`$ git pull https://git.openjdk.org/panama-foreign pull/806/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 806`

`$ git pr show -t 806`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/806.diff">https://git.openjdk.org/panama-foreign/pull/806.diff</a>

</details>


src/java.base/share/classes/jdk/internal/foreign/abi/Binding.java
```diff
@@ -553,9 +581,39 @@ public void verify(Deque<Class<?>> stack) {
...
+                SharedUtils.write(writeAddress, offset(), type(), value);
+            } else {
+                // non-exact match, need to do chunked load
+                long longValue = ((Number) value).longValue();
+                int remaining = byteWidth();
+                int chunkOffset = 0;
+                do {
+                    int chunkSize = Integer.highestOneBit(remaining); // next power of 2, in bytes
+                    long writeOffset = offset() + SharedUtils.pickChunkOffset(chunkOffset, byteWidth(), chunkSize);
+                    int shiftAmount = chunkOffset * Byte.SIZE;
```
mcimadamore, line 596: Is this correct? E.g. assume we have to write 3 shorts { 13, 14, 15 }, and we do that in two chunks. If we are big endian, I think we first write { 14, 15 } then { 13 }. When we write { 14, 15 }, Isn't the offset of the mask 2 bytes (e.g. 16 bits) ? In other words shouldn't `chunkOffset` here also be decorated with `pinkChunkOffset` to take endianness into account? (it seems to me that the shift amount should be equal to `writeOffset - offset()`).

src/java.base/share/classes/jdk/internal/foreign/abi/Binding.java
```diff
@@ -582,9 +640,38 @@ public void verify(Deque<Class<?>> stack) {
...
+                    int chunkSize = Integer.highestOneBit(remaining); // next power of 2
+                    long readOffset = offset() + SharedUtils.pickChunkOffset(chunkOffset, byteWidth(), chunkSize);
+                    long readChunk = switch (chunkSize) {
+                        case 4 -> Integer.toUnsignedLong(readAddress.get(JAVA_INT_UNALIGNED, readOffset));
+                        case 2 -> Short.toUnsignedLong(readAddress.get(JAVA_SHORT_UNALIGNED, readOffset));
+                        case 1 -> Byte.toUnsignedLong(readAddress.get(JAVA_BYTE, readOffset));
+                        default ->
+                            throw new IllegalStateException("Unexpected chunk size for chunked write: " + chunkSize);
+                    };
+                    result |= readChunk << (chunkOffset * Byte.SIZE);
```
mcimadamore, line 662: same here - is the naked reference to `chunkOffset` ok (w.r.t. endianness) ?

src/java.base/share/classes/jdk/internal/foreign/abi/Binding.java
JornVernee: I don't think so. But I will try to verify BE as well in the test (somehow).

My understanding is this:

For LE, we have a struct with 3 shorts in memory like so:

    lo1, hi1, lo2, hi2, lo3, hi3
    0         2         4         6

The native compiler does an oversized load of eight bytes, so we also load the trailing 2 bytes (i.e. byte 6 - 8). And the resulting register looks like this (using `000` to represent junk bytes):

    000, 000, hi3, lo3, hi2, lo2, hi1, lo1
    8         6         4         2         0

So, we first load an `int` from bytes 0 - 4, and get `000, 000, 000, 000, hi2, lo2, hi1, lo1`, then we load bytes 4 - 6 with shift amount = 4: `000, 000, hi3, lo3, 000, 000, 000, 000` and combine the two to get: `000, 000, hi3, lo3, hi2, lo2, hi1, lo1` (just like we would get from an oversized load).

---

For BE, we have in memory:
   
    hi1, lo1, hi2, lo2, hi3, lo3
    0         2         4         6

We load byte 2 - 6 which using an `int` and get: `000, 000, 000, 000, hi2, lo2, hi3, lo3` (no flipping this time), then we load byte 0 - 2 using a `short` and shift amount = 4 and get `000, 000, hi1, lo1, 000, 000, 000, 000`. Then combined: `000, 000, hi1, lo1, hi2, lo2, hi3, lo3`.

i.e., the order of the shorts within the register is flipped on BE, compare to LE. I think this is correct since when I have code like this:

```
struct Foo {
    short f0;
    short f1;
    short f2;
};

short func(struct Foo f) {
    return f.f2;
}
```

The compiler generates: [`extsh 3, 3`](https://www.ibm.com/docs/en/aix/7.3?topic=set-extsh-exts-extend-sign-halfword-instruction) which grabs to lower 16 bits of the register (which contains `hi3, lo3`). (for returning `f0` and `f1` the compiler generates 32 and 16 bit shifts respectively. Well, assuming I'm reading PPC disassembly correctly :))

But, if we flipped the shift amount as well, we'd get  `000, 000, hi2, lo2, hi3, lo3, 000, 000` on the first load, and combined we'd get `000, 000, hi2, lo2, hi3, lo3, hi1, lo1`, which doesn't seem right, since the order of the shorts is mixed.

JornVernee: I also realized that I could beef up the test by letting the native code (which we assume to be correctly compiled) pass each individual byte to the callback as well. That way we can also test whether the order of the elements is correct.

mcimadamore: I've been trying to read the PPC ABI [1], which has some useful diagrams. I found this interesting:

```
struct {
  char  c;
  char  d;
  short s;
  int   n;
};

word aligned, sizeof is 8

little endian:

+-------+-------+-------+-------+
|              2|      1|      0|
|       s       |   d   |   c   |
+-------+-------+-------+-------+
|                              4|
|               n               |
+-------+-------+-------+-------+

big endian:

+-------+-------+-------+-------+
|0      |1      |2              |
|   c   |   d   |       s       |
+-------+-------+-------+-------+
|4                              |
|               n               |
+-------+-------+-------+-------+
```

(I think in these diagrams, on the left you always have MSB and on the right the LSB - as explained in an earlier section. The numbers indicate the byte offsets - I believe as laid out in memory - of the struct fields).

This seems to confirm your expectations - after a load, struct fields are "flipped" in little-endian, and are not flipped in big-endian. Your implementation takes that into account and avoids a "double flip" (which is what I was proposing as I did not immediately get the difference in the register layout).


[1] - https://refspecs.linuxfoundation.org/ELF/ppc64/PPC-elf64abi-1.4.1.txt

src/java.base/share/classes/jdk/internal/foreign/abi/BindingSpecializer.java
```diff
@@ -701,13 +769,82 @@ private void emitUnboxAddress() {
...
-        String descriptor = methodType(loadType, valueLayoutType, long.class).descriptorString();
-        emitInvokeInterface(MemorySegment.class, "get", descriptor);
+        if (SharedUtils.isPowerOfTwo(byteWidth)) {
+            Class<?> valueLayoutType = emitLoadLayoutConstant(loadType);
+            emitConst(offset);
+            String descriptor = methodType(loadType, valueLayoutType, long.class).descriptorString();
+            emitInvokeInterface(MemorySegment.class, "get", descriptor);
+        } else {
+            // chunked
+            int readAddrIdx = newLocal(MemorySegment.class);
```
mcimadamore, line 783: It's nice that, when we specialize we can get rid of the loop we have when interpreting - and just emit a sequence of load/stores (with some extra shifts).

src/java.base/share/classes/jdk/internal/foreign/abi/SharedUtils.java
```diff
@@ -339,6 +340,36 @@ static void checkType(Class<?> actualType, Class<?> expectedType) {
         }
     }
 
+    public static boolean isPowerOfTwo(int width) {
+        return Integer.bitCount(width) == 1;
+    }
+
+    public static int byteWidthOfPrimitive(Class<?> type) {
```
mcimadamore, line 347: When we create memory access var handles, we use this to compute the byte size of a primitive (`VarHandles::memorySegmentViewHandle`):

```
Wrapper.forPrimitiveType(carrier).bitWidth() / 8
```

Perhaps we should put this into Utils and then reuse in both places?

test/jdk/java/foreign/arraystructs/TestArrayStructs.java
```diff
@@ -0,0 +1,138 @@
...
+import static java.lang.foreign.MemoryLayout.structLayout;
+
+public class TestArrayStructs extends NativeTestHelper {
+    static {
+        System.loadLibrary("ArrayStructs");
+    }
+
+    // Test if structs of various different sizes, including non-powers of two, work correctly
+    @Test(dataProvider = "arrayStructs")
+    public void testArrayStruct(String functionName, FunctionDescriptor upcallDesc, int numPrefixArgs) throws Throwable {
```
mcimadamore, line 74: Nice test!

mcimadamore (approved):

src/java.base/share/classes/jdk/internal/foreign/abi/Binding.java
```diff
@@ -553,9 +581,39 @@ public void verify(Deque<Class<?>> stack) {
...
-            MemorySegment writeAddress = operand.asSlice(offset());
-            SharedUtils.write(writeAddress, type(), value);
+            MemorySegment writeAddress = (MemorySegment) stack.pop();
+            if (SharedUtils.isPowerOfTwo(byteWidth())) {
+                // exact size match
+                SharedUtils.write(writeAddress, offset(), type(), value);
+            } else {
+                // non-exact match, need to do chunked load
+                long longValue = ((Number) value).longValue();
+                int remaining = byteWidth();
```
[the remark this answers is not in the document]
JornVernee: Done. I also added the missing check to the `bufferStore` factory.

JornVernee:
I've address review comments (https://github.com/openjdk/panama-foreign/pull/806/commits/b1ccd916913065d2a146b0d3c328cdc4401546c2)

I don't think there's a good way to test the BE behavior on LE platforms, unfortunately. It also made me realize that our linkers are porbably tied to a particular endianess, and we should/could check that the layouts found in a FunctionDescriptor have the expected byte order as well.

src/java.base/share/classes/jdk/internal/foreign/abi/SharedUtils.java
JornVernee: Good idea. Done.

mcimadamore (approved):
Looks good. I agree on validating endianness assumptions.

PR #807, opened 2023-03-02
8303519: Replace Binding.Context with an arena

mcimadamore:
This patch contains a cleanup I wanted to do for some time. Basically, the linker uses an internal class (`Binding.Context`) to perform allocation needed during a native call/upcall. This context class used to provide both a scope (e.g. `ResourceScope`) and an allocator (`SegmentAllocator`).

Given that in recent iteratons of the API, `Arena` *is* an allocator, we can tweak bindings to accept a segment allocator parameter (instead of a context object). For bindings that only need to allocate, they can just call one of the `SegmentAllocator::allocate`) overloads. This simplifies the case where we need to use an external allocation for structs returned by value - as we can now just pass the external allocator - no need to wrap it in a "context".

If, on the other hand, the binding needs access to the scope, we can still implement that, by downcasting the segment allocator to `Arena` and then get the scope from there. This is always sound, because this arena is always internally created by the linker.

I moved some helper methods/fields to provide ready-made arena flavors in `SharedUtils`, so as to keep `Bindings.java` all about bindings. I think the result is rather nice, and in principle it should perform a little better, given that we don't need to wrap everything into a separate context object.

---------
### Progress

### Issue
 * [JDK-8303519](https://bugs.openjdk.org/browse/JDK-8303519): Replace Binding.Context with an arena

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/807/head:pull/807` \
`$ git checkout pull/807`

`$ git checkout pull/807` \
`$ git pull https://git.openjdk.org/panama-foreign pull/807/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 807`

`$ git pr show -t 807`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/807.diff">https://git.openjdk.org/panama-foreign/pull/807.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/jdk/internal/foreign/abi/Binding.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/abi/Binding.java
@@ -24,9 +24,9 @@
  */
 package jdk.internal.foreign.abi;
 
-import jdk.internal.foreign.MemorySessionImpl;
-import jdk.internal.foreign.NativeMemorySegmentImpl;
 import jdk.internal.foreign.Utils;
+import jdk.internal.foreign.abi.BindingInterpreter.LoadFunc;
+import jdk.internal.foreign.abi.BindingInterpreter.StoreFunc;
 
 import java.lang.foreign.*;
 import java.lang.invoke.MethodHandle;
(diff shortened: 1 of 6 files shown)


JornVernee (approved):
Nice cleanup!

PR #811, opened 2023-03-06
8303666: Move address layout outside of ValueLayout namespace

mcimadamore:
This patch moves ValueLayout.OfAddress into a toplevel class. This seems desirable since address layouts are very important in the FFM API, and also provides seadditional methods on top of `ValueLayout` to get/set a *target layout*.

---------
### Progress

### Issue
 * [JDK-8303666](https://bugs.openjdk.org/browse/JDK-8303666): Move address layout outside of ValueLayout namespace

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/811/head:pull/811` \
`$ git checkout pull/811`

`$ git checkout pull/811` \
`$ git pull https://git.openjdk.org/panama-foreign pull/811/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 811`

`$ git pr show -t 811`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/811.diff">https://git.openjdk.org/panama-foreign/pull/811.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/java/lang/foreign/AddressLayout.java
+++ b/src/java.base/share/classes/java/lang/foreign/AddressLayout.java
@@ -1,0 +1,87 @@
+package java.lang.foreign;
+
+import jdk.internal.foreign.layout.ValueLayouts;
+import jdk.internal.javac.PreviewFeature;
+import jdk.internal.reflect.CallerSensitive;
+
+import java.lang.foreign.Linker.Option;
+import java.lang.invoke.MethodHandle;
+import java.nio.ByteOrder;
+import java.util.Optional;
+
+/**
+ * A value layout used to model the address of some region of memory. The carrier associated with an address layout is
+ * {@code MemorySegment.class}. The size and alignment of an address layout are platform dependent
+ * (e.g. on a 64-bit platform, the size and alignment of an address layout are set to 64 bits).
+ * <p>
+ * An address layout may optionally feature a {@linkplain #targetLayout() target layout}. The target layout of an address
+ * layout is used to model the layout of the region of memory whose address is described by that address layout.
+ * For instance, if an address layout has target layout {@link ValueLayout#JAVA_INT}, the region of memory pointed to by the address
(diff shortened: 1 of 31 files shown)


src/java.base/share/classes/java/lang/foreign/AddressLayout.java
```diff
@@ -0,0 +1,113 @@
...
+
+/**
+ * A value layout used to model the address of some region of memory. The carrier associated with an address layout is
+ * {@code MemorySegment.class}. The size and alignment of an address layout are platform dependent
+ * (e.g. on a 64-bit platform, the size and alignment of an address layout are set to 64 bits).
+ * <p>
+ * An address layout may optionally feature a {@linkplain #targetLayout() target layout}. The target layout of an address
+ * layout is used to model the layout of the region of memory whose address is described by that address layout.
+ * For instance, if an address layout has target layout {@link ValueLayout#JAVA_INT}, the region of memory pointed to by the address
+ * described by the address layout is 4 bytes long. Specifying a target layout can be useful in the following situations:
```
JornVernee, line 46: Some redundant text here it seems
```suggestion
 * For instance, if an address layout has target layout {@link ValueLayout#JAVA_INT}, the region of memory 
 * described by the address layout is 4 bytes long. Specifying a target layout can be useful in the following situations:
```

JornVernee (approved):

src/java.base/share/classes/java/lang/foreign/AddressLayout.java
mcimadamore: Not sure about this - (I went a bit back and forth with the javadoc text). An address layout describe an address value, not a memory region. E.g. it describes 64 bits. With a target layout you can say that those 64 bits point to a memory region that has a given layout. It is a bit convoluted to describe.

mcimadamore: Gave it another stab - let me know what you think.

JornVernee (approved):
Nice!

PR #820, opened 2023-03-16
8298096: Refine javadoc for Linker

mcimadamore:
The javadoc for the `Linker` interface is very abstract: it defines the general characteristics of downcall method handles and upcall stubs, and it does so in a very general way, so as not to bias the discussion towards a specific kind of linker implementation.

The result is that, looking at the `Linker` javadoc, one has very little clues on how this interface is meant to be used e.g. to write native interop code. While the toplevel package javadoc provides some examples, these examples are relatively simple and do not cover all the features provided by the FFM API.

To rectify this, I did another pass on the `Linker` javadoc. I retained the very good general intro (the first few paras). But then, instead of talking about "downcalls" and "upcalls" - we immediately start talking about "Calling native functions" and we ground the discussion on the native linker. This gives us the opportunity to discuss native calls, function pointers, variadic calls, and even how to deal with functions returning pointers.

The text of the sections on upcalls/downcall (which is normative text) has been moved in the javadoc of the relevant methods.

Finally, since with the new examples there was some overlapping between the toplevel package javadoc and the `Linker` javadoc, I also decided to take another pass at the package javadoc, and simplify it further: now there are only 3 sections: foreign memory, foreign function, restricted methods. For each of the first two sections there's a quick summary (which describes the main abstractions and their roles), followed by an idiomatic example. Restricted methods also belong here, as they are a cross-cutting concern of the FFM API.

Overall, I believe this iteration is much better than what we had, and more useful. While covering all the details of native interop would be outside the scope of the FFM javadoc, I think the new `Linker` javadoc strikes a good balance.

---------
### Progress

### Issue
 * [JDK-8298096](https://bugs.openjdk.org/browse/JDK-8298096): Refine javadoc for Linker

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign pull/820/head:pull/820` \
`$ git checkout pull/820`

`$ git checkout pull/820` \
`$ git pull https://git.openjdk.org/panama-foreign pull/820/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 820`

`$ git pr show -t 820`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/820.diff">https://git.openjdk.org/panama-foreign/pull/820.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/java/lang/foreign/Linker.java
+++ b/src/java.base/share/classes/java/lang/foreign/Linker.java
@@ -415,6 +415,11 @@ public sealed interface Linker permits AbstractLinker {
     /**
      * Returns a linker for the ABI associated with the underlying native platform. The underlying native platform
      * is the combination of OS and processor where the Java runtime is currently executing.
+     * <p>
+     * This method is <a href="package-summary.html#restricted"><em>restricted</em></a>.
+     * Restricted methods are unsafe, and, if used incorrectly, their use might crash
+     * the JVM or, worse, silently result in memory corruption. Thus, clients should refrain from depending on
+     * restricted methods, and use safe and supported functionalities, where possible.
      *
      * @apiNote It is not currently possible to obtain a linker for a different combination of OS and processor.
      * @implNote The libraries exposed by the {@linkplain #defaultLookup() default lookup} associated with the returned
--- a/src/java.base/share/classes/java/lang/foreign/Linker.java
+++ b/src/java.base/share/classes/java/lang/foreign/Linker.java
@@ -109,8 +109,10 @@ import java.util.stream.Stream;
  * <p>
  * Composite types are modelled as {@linkplain GroupLayout group layouts}. More specifically, a C {@code struct} type
  * maps to a {@linkplain StructLayout struct layout}, whereas a C {@code union} type maps to a {@link UnionLayout union
- * layout}. Depending on the ABI implemented by the native linker, additional {@linkplain MemoryLayout#paddingLayout(long) padding}
- * member layouts might be required to conform to the size and alignment constraint of a composite type definition in C.
+ * layout}. When defining a struct or union layout, clients must pay attention to the size and alignment constraint
+ * of the corresponding composite type definition in C. For instance, padding between two struct fields
+ * must be modelled explicitly, by adding an adequately sized {@linkplain PaddingLayout padding layout} member
+ * to the resulting struct layout.
  * <p>
  * Finally, pointer types such as {@code int**} and {@code int(*)(size_t*, size_t*)} are modelled as
  * {@linkplain AddressLayout address layouts}. When the spatial bounds of the pointer type are known statically,
--- a/src/java.base/share/classes/java/lang/foreign/Linker.java
+++ b/src/java.base/share/classes/java/lang/foreign/Linker.java
@@ -1,5 +1,5 @@
 /*
- *  Copyright (c) 2020, 2022, Oracle and/or its affiliates. All rights reserved.
+ *  Copyright (c) 2020, 2023, Oracle and/or its affiliates. All rights reserved.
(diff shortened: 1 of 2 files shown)


src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -60,64 +61,340 @@
...
+ * of a suitable carrier. The mapping between a scalar type and its corresponding layout is dependent on the ABI
+ * implemented by the native linker. For instance, the C type {@code long} maps to the layout constant
+ * {@link ValueLayout#JAVA_LONG} on Linux/x64, but maps to the layout constant {@link ValueLayout#JAVA_INT} on
+ * Windows/x64. Similarly, the C type {@code size_t} maps to the layout constant {@link ValueLayout#JAVA_LONG}
+ * on 64-bit platforms, but maps to the layout constant {@link ValueLayout#JAVA_INT} on 32-bit platforms.
+ * <p>
+ * Composite types are modelled as {@linkplain GroupLayout group layouts}. More specifically, a C {@code struct} type
+ * maps to a {@linkplain StructLayout struct layout}, whereas a C {@code union} type maps to a {@link UnionLayout union
+ * layout}. Depending on the ABI implemented by the native linker, additional {@linkplain MemoryLayout#paddingLayout(long) padding}
+ * member layouts might be required to conform to the size and alignment constraint of a composite type definition in C.
```
JornVernee, line 113: I think the comment on padding here is maybe a bit too vague. I think it would be better to say explicitly that padding is not added to group layouts automatically, and has to be manually specified.

(On that note, maybe we want to add a factory in MemoryLayout that basically exposes `Utils.computePaddedStruct`)

src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -60,64 +61,340 @@
...
+ * {@snippet lang = c:
+ * void qsort(void *base, size_t nmemb, size_t size,
+ *            int (*compar)(const void *, const void *));
+ * }
  *
- * {@linkplain #downcallHandle(FunctionDescriptor, Option...) Linking a foreign function} is a process which requires a function descriptor,
- * a set of memory layouts which, together, specify the signature of the foreign function to be linked, and returns,
- * when complete, a downcall method handle, that is, a method handle that can be used to invoke the target foreign function.
+ * The {@code qsort} function can be used to sort the contents of an array, using a custom comparator function which is
+ * passed as a function pointer (the {@code compar} parameter). To be able to call {@code qsort} function from Java,
```
JornVernee, line 208:
```suggestion
 * passed as a function pointer (the {@code compar} parameter). To be able to call the {@code qsort} function from Java,
```

src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -60,64 +61,340 @@
...
+ * returned segment is a fresh scope that is always alive. To provide safe access to the segment, we must,
+ * unsafely, resize the segment to the desired size (100, in this case). It might also be desirable to
+ * attach the segment to some existing {@linkplain Arena arena}, so that the lifetime of the region of memory
+ * backing the segment can be managed automatically, as for any other native segment created directly from Java code.
+ * Both these operations are accomplished using the restricted {@link MemorySegment#reinterpret(long, Arena, Consumer)}
+ * method, as follows:
+ *
+ * {@snippet lang = java:
+ * MemorySegment allocateMemory(long byteSize, Arena arena) {
+ *     MemorySegment segment = (MemorySegment)malloc.invokeExact(byteSize);   // size = byteSize, scope = always alive
```
JornVernee, line 323:
```suggestion
 *     MemorySegment segment = (MemorySegment)malloc.invokeExact(byteSize);   // size = 0, scope = always alive
```

JornVernee (approved):

PR #824, opened 2023-03-28
8305087: MemoryLayout API checks should be more eager

mcimadamore:
This patch adds eager check on construction of sequence and group layouts. More specifically:

* sequenceLayout(L) is well-formed iff L.bitAlignment() <= L.bitSize()
* groupLayout(L1, … Ln) is well-formed iff for-each 1..n, offset(Li) % Li.bitAlignment() == 0

Moreover, this patch also validates the alignment parameter to calls of `MemoryLayout::withBitAlignment` on sequence and group layouts:

* sequenceLayout(L).withBitAlignment(a) is well-formed iff L.bitAlignment() <= a
* groupLayout(L1, ..., Ln).withBitAlignment(a) is well-formed iff max(L1.bitAlignment(), ...,  Ln.bitAlignment()) <= a

These checks prevent bad layouts from being constructed (either with factories, or with bad calls to `withBitAlignment`), w/o restricting the expressiveness of the memory layout API.
Moreover, since all layouts are now well-formed by construction, we no longer need lazy checks when creating a deferefence var handle.

---------
### Progress

### Issue
 * [JDK-8305087](https://bugs.openjdk.org/browse/JDK-8305087): MemoryLayout API checks should be more eager

### Reviewers
 * [Per Minborg](https://openjdk.org/census#pminborg) (@minborg - Committer) ⚠️ Review applies to [49b83016](https://git.openjdk.org/panama-foreign/pull/824/files/49b8301666135298b6aa894177be6a3d273dc756)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign.git pull/824/head:pull/824` \
`$ git checkout pull/824`

`$ git checkout pull/824` \
`$ git pull https://git.openjdk.org/panama-foreign.git pull/824/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 824`

`$ git pr show -t 824`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/824.diff">https://git.openjdk.org/panama-foreign/pull/824.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/java/lang/foreign/GroupLayout.java
+++ b/src/java.base/share/classes/java/lang/foreign/GroupLayout.java
@@ -68,7 +68,7 @@ public sealed interface GroupLayout extends MemoryLayout permits StructLayout, U
     /**
      * {@inheritDoc}
      * @throws IllegalArgumentException {@inheritDoc}
-     * @throws IllegalArgumentException if {@code bitAlignment} is {@code M}, where {@code M} is the maximum alignment
+     * @throws IllegalArgumentException if {@code bitAlignment} is less than {@code M}, where {@code M} is the maximum alignment
      * constraint in any of the member layouts associated with this group layout.
      */
     @Override
--- a/src/java.base/share/classes/java/lang/foreign/GroupLayout.java
+++ b/src/java.base/share/classes/java/lang/foreign/GroupLayout.java
@@ -67,6 +67,9 @@ public sealed interface GroupLayout extends MemoryLayout permits StructLayout, U
 
     /**
      * {@inheritDoc}
+     * @throws IllegalArgumentException {@inheritDoc}
+     * @throws IllegalArgumentException if {@code bitAlignment} is {@code M}, where {@code M} is the maximum alignment
+     * constraint in any of the member layouts associated with this group layout.
      */
     @Override
     GroupLayout withBitAlignment(long bitAlignment);
--- a/src/java.base/share/classes/jdk/internal/foreign/layout/AbstractGroupLayout.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/layout/AbstractGroupLayout.java
@@ -49,15 +49,13 @@ public sealed abstract class AbstractGroupLayout<L extends AbstractGroupLayout<L
 
     private final Kind kind;
     private final List<MemoryLayout> elements;
+    final long minBitAlignment;
 
-    AbstractGroupLayout(Kind kind, List<MemoryLayout> elements) {
-        this(kind, elements, kind.alignof(elements), Optional.empty());
-    }
-
-    AbstractGroupLayout(Kind kind, List<MemoryLayout> elements, long bitAlignment, Optional<String> name) {
-        super(kind.sizeof(elements), bitAlignment, name); // Subclassing creates toctou problems here
+    AbstractGroupLayout(Kind kind, List<MemoryLayout> elements, long bitSize, long bitAlignment, long minBitAlignment, Optional<String> name) {
+        super(bitSize, bitAlignment, name); // Subclassing creates toctou problems here
         this.kind = kind;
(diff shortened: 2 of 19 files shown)


test/jdk/java/foreign/callarranger/TestSysVCallArranger.java
```diff
@@ -145,66 +145,6 @@ public void testNestedUnion() {
         assertEquals(bindings.nVectorArgs(), 0);
     }
 
-    @Test
-    public void testNestedStructsUnaligned() {
```
mcimadamore, line 149: These tests are no longer required, since they refer to a situation where the one of the layout involved would have an alignment different from the natural alignment (which is now enforced at link time).

src/java.base/share/classes/java/lang/foreign/GroupLayout.java
```diff
@@ -67,6 +67,9 @@ public sealed interface GroupLayout extends MemoryLayout permits StructLayout, U
 
     /**
      * {@inheritDoc}
+     * @throws IllegalArgumentException {@inheritDoc}
+     * @throws IllegalArgumentException if {@code bitAlignment} is {@code M}, where {@code M} is the maximum alignment
```
JornVernee, line 71: Typo?
```suggestion
     * @throws IllegalArgumentException if {@code bitAlignment} is less than {@code M}, where {@code M} is the maximum alignment
```

mcimadamore: Good catch

src/java.base/share/classes/jdk/internal/foreign/layout/AbstractGroupLayout.java
```diff
@@ -132,6 +140,9 @@ enum Kind {
         long sizeof(List<MemoryLayout> elems) {
             long size = 0;
             for (MemoryLayout elem : elems) {
+                if (this == STRUCT && (size % elem.bitAlignment() != 0)) {
```
minborg, line 143: We might drop `sizeOp` and instead have two different `sizeof` implementations in a more OO style.

test/jdk/java/foreign/TestLayoutPaths.java
```diff
@@ -252,9 +214,9 @@ public void testStructPaths(IntFunction<PathElement> groupSelector) {
         long[] offsets = { 0, 8, 24, 56 };
         GroupLayout g = MemoryLayout.structLayout(
                 ValueLayout.JAVA_BYTE.withName("0"),
-                ValueLayout.JAVA_CHAR.withName("1"),
-                ValueLayout.JAVA_FLOAT.withName("2"),
-                ValueLayout.JAVA_LONG.withName("3")
+                ValueLayout.JAVA_CHAR.withBitAlignment(8).withName("1"),
+                ValueLayout.JAVA_FLOAT.withBitAlignment(8).withName("2"),
+                ValueLayout.JAVA_LONG.withBitAlignment(8).withName("3")
```
minborg, line 219: Maybe there should be an alias for `withBitAlignment(8)`? E.g. `JAVA_LONG.withoutAlignment()`.

minborg (approved):

src/java.base/share/classes/jdk/internal/foreign/layout/AbstractGroupLayout.java
mcimadamore: I've did another pass on the impl, where I've greatly simplified AbstractGroupLayoutImpl, StructLayoutImpl and UnionLayoutImpl. Now all the computation occurs in their factories (which is better, as the stack trace in case of an issue is more readable), rather than inside the enum methods. Also, the new classes avoid any kind of re-computation: for each group, we compute size and alignment upfront. We then store the original alignment into a new field (`minBitAlignment`) so that we can use that to determine validity of subsequent `withBitAlignment` operations.

PR #827, opened 2023-04-13
8303831: TestSegmentCopy times out

JornVernee:
The underlying issue turned out to be that this test generates a lot of test cases, 165890, which TestNG then spends forever generating an XML report for.

This patch refactors the test to generate less test cases. Instead of generating all the combinations of slices up front in the data provider, it is done on the fly in the test methods themselves.

Additionally, this patch improves the space that is covered. The current code tests copies between disjoint slices of the same segment. The new code uses two separate segments, which also allows testing mixes of segment types (heap/off-heap).

The current code also uses `asSlice` to generate the slices, but it seems more interesting to pass the offset of a slice directly to MemorySegment::copy, so that we test with different offsets as well (currently only offset 0 is used). I think we already have enough coverage of `asSlice` in other tests, such as TestSlices.

Lastly, I've removed the zeroing of the target area/slice. If there's garbage in target slice, we can test that the copy properly overwrites the garbage as well.

---------
### Progress

### Issue
 * [JDK-8303831](https://bugs.openjdk.org/browse/JDK-8303831): TestSegmentCopy times out

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign.git pull/827/head:pull/827` \
`$ git checkout pull/827`

`$ git checkout pull/827` \
`$ git pull https://git.openjdk.org/panama-foreign.git pull/827/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 827`

`$ git pr show -t 827`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/827.diff">https://git.openjdk.org/panama-foreign/pull/827.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/panama-foreign/pull/827#issuecomment-1507260248)

### DIFF
--- a/test/jdk/java/foreign/TestSegmentCopy.java
+++ b/test/jdk/java/foreign/TestSegmentCopy.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019, 2023, Oracle and/or its affiliates. All rights reserved.
  *  DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  *  This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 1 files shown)


mcimadamore (approved):
Looks great!

PR #828, opened 2023-04-21
8306668: Some foreign tests fail on x86

mcimadamore:
Some foreign tests fail on x86, as a result of the changes in https://github.com/openjdk/panama-foreign/pull/775.
This patch reverts the `long` -> `int` conversion logic in `Utils`, and it also fixes the following:
* a new microbenchmark fails compilation on x86 because of a bad cast from pointer to long
* exclude WindowsCallArrangerTest on x86

I've tested this by building (and testing) a 32-bit JDK on my Linux x64 using `--with-target-bits=32`.

---------
### Progress

### Issue
 * [JDK-8306668](https://bugs.openjdk.org/browse/JDK-8306668): Some foreign tests fail on x86

### Reviewers
 * [Jorn Vernee](https://openjdk.org/census#jvernee) (@JornVernee - Committer) ⚠️ Review applies to [b609e5e4](https://git.openjdk.org/panama-foreign/pull/828/files/b609e5e463fa3d8cc63a6b066775a727976bc424)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign.git pull/828/head:pull/828` \
`$ git checkout pull/828`

`$ git checkout pull/828` \
`$ git pull https://git.openjdk.org/panama-foreign.git pull/828/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 828`

`$ git pr show -t 828`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/828.diff">https://git.openjdk.org/panama-foreign/pull/828.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/panama-foreign/pull/828#issuecomment-1517610020)

### DIFF
--- a/test/jdk/java/foreign/callarranger/TestWindowsCallArranger.java
+++ b/test/jdk/java/foreign/callarranger/TestWindowsCallArranger.java
@@ -25,7 +25,7 @@
 /*
  * @test
  * @enablePreview
- * @requires ((os.arch == "amd64" | os.arch == "x86_64") & sun.arch.data.model == "64") | os.arch == "aarch64"
+ * @requires sun.arch.data.model == "64"
  * @compile platform/PlatformLayouts.java
  * @modules java.base/jdk.internal.foreign
  *          java.base/jdk.internal.foreign.abi
--- a/test/jdk/java/foreign/callarranger/TestWindowsCallArranger.java
+++ b/test/jdk/java/foreign/callarranger/TestWindowsCallArranger.java
@@ -25,6 +25,7 @@
 /*
  * @test
  * @enablePreview
+ * @requires ((os.arch == "amd64" | os.arch == "x86_64") & sun.arch.data.model == "64") | os.arch == "aarch64"
  * @compile platform/PlatformLayouts.java
  * @modules java.base/jdk.internal.foreign
  *          java.base/jdk.internal.foreign.abi
--- a/test/micro/org/openjdk/bench/java/lang/foreign/libPtr.c
+++ b/test/micro/org/openjdk/bench/java/lang/foreign/libPtr.c
@@ -29,16 +29,18 @@
 #define EXPORT
 #endif
 
+#include <stddef.h>
+
 EXPORT long long id_long_long(long long value) {
   return value;
 }
 
 EXPORT long long id_ptr_long(void* ptr) {
-  return (long long)ptr;
+  return (long long)(size_t)ptr;
 }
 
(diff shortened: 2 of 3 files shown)


test/micro/org/openjdk/bench/java/lang/foreign/libPtr.c
```diff
@@ -29,16 +29,18 @@
...
 
+#include <stddef.h>
+
 EXPORT long long id_long_long(long long value) {
   return value;
 }
 
 EXPORT long long id_ptr_long(void* ptr) {
-  return (long long)ptr;
+  return (long long)(size_t)ptr;
```
mcimadamore, line 39: Note - these benchmarks do not care about the fact that some bits in the original value could be dropped to the floor. The benchmark aims at comparing the performance of passing/returning longs with that of passing/returning segments (e.g. to make sure that escape analysis works its magic). As such, lossy conversions are immaterial here.

mcimadamore: I guess a different approach could be to avoid using `void*` and use `size_t` instead, which would still allow FFM to link correctly. If that is preferred I can do that instead.

test/jdk/java/foreign/callarranger/TestWindowsCallArranger.java
```diff
@@ -25,6 +25,7 @@
 /*
  * @test
  * @enablePreview
+ * @requires ((os.arch == "amd64" | os.arch == "x86_64") & sun.arch.data.model == "64") | os.arch == "aarch64"
```
JornVernee, line 28: Some of the other call arranger tests just use `@requires sun.arch.data.model == "64"`. Please do that here as well. It should be fine to run this test as long as the platform is 64 bit.
```suggestion
 * @requires sun.arch.data.model == "64"
```

test/micro/org/openjdk/bench/java/lang/foreign/libPtr.c
JornVernee: I think this is a good solution.

JornVernee (approved):

test/jdk/java/foreign/callarranger/TestWindowsCallArranger.java
JornVernee: FWIW, the underlying issue with this test is that the size of the address layout changes across platforms.

PR #830, opened 2023-05-02
8307253: Make FunctionDescriptor::toMethodType an instance method of Linker

mcimadamore:
This patch moves `FunctionDescriptor::toMethodType` as an instance method of `Linker`. This allows linker implementations to be more opinionated on how the function descriptor -> method type conversion should behave.

---------
### Progress

### Issue
 * [JDK-8307253](https://bugs.openjdk.org/browse/JDK-8307253): Make FunctionDescriptor::toMethodType an instance method of Linker

### Reviewers
 * [Jorn Vernee](https://openjdk.org/census#jvernee) (@JornVernee - Committer) ⚠️ Review applies to [c63efe57](https://git.openjdk.org/panama-foreign/pull/830/files/c63efe57e7808d095e6235df126913f25d642f20)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign.git pull/830/head:pull/830` \
`$ git checkout pull/830`

`$ git checkout pull/830` \
`$ git pull https://git.openjdk.org/panama-foreign.git pull/830/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 830`

`$ git pr show -t 830`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/830.diff">https://git.openjdk.org/panama-foreign/pull/830.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/panama-foreign/pull/830#issuecomment-1531558007)


src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/CallArranger.java
```diff
@@ -480,12 +480,10 @@ List<Binding> getBindings(Class<?> carrier, MemoryLayout layout) {
                     StorageCalculator.StructStorage[] structStorages
                             = storageCalculator.structStorages((GroupLayout) layout, forHFA);
 
-                    for (StorageCalculator.StructStorage(
-                            long offset, Class<?> ca, int byteWidth, VMStorage storage
-                    ) : structStorages) {
+                    for (StorageCalculator.StructStorage structStorage : structStorages) {
```
mcimadamore, line 483: This change is not strictly required, but I was getting weird errors when building javadoc from IntelliJ w/o it. Since patterns in for-each are going away anyways, I thought perhaps we could just take care of this.

mcimadamore:
After some offline discussion, we realized that the current API is actually good enough: FunctionDescriptor::toMethodType defines the "lowest common denominator" for turning a function descriptor into a MethodType. In doing this conversion, we follow the carrier types indicated by value layouts. For structs/unions/sequences we assume the carrier is MemorySegment (it is possible that in the future we will also add a carrier() accessor on composite layouts that can be overridden). Finally, for padding layout we just throw (after all, a padding argument/return doesn't seem to make much sense). Capturing this conversion at the level of FunctionDescription still seems like the right thing to do.

However, there is a small detail which this patch fixes: FunctionDescriptor::toMethodType currently rejects sequence layouts. While it is certainly the case that the C ABI does not allow to pass arrays by value. And even if this is the most common choice, this is not the only possible choice (for instance, Pascal has special convention for passing arrays to functions where the size of the array is known). As such, I think we should not bias the conversion with what we know the native linker will and will not support.

For this reason, I've tweaked the checks in the conversion, and then made some checks stricter in the linker - so that now sequence layouts are rejected at the level of the linker.

src/java.base/share/classes/jdk/internal/foreign/abi/AbstractLinker.java
```diff
@@ -101,12 +102,19 @@ public SystemLookup defaultLookup() {
...
-        descriptor.argumentLayouts().forEach(AbstractLinker::checkHasNaturalAlignmentRecursive);
+    // Native linkers do not support sequence or padding layouts. They also do not support "packed" struct layouts, or
+    // struct layouts containing additional padding. Such illegal layouts are ruled out here.
+    private static void checkIsSupportedDescriptor(FunctionDescriptor descriptor) {
+        descriptor.returnLayout().ifPresent(AbstractLinker::checkIsSupportedLayout);
+        descriptor.argumentLayouts().forEach(AbstractLinker::checkIsSupportedLayout);
+    }
+
+    private static void checkIsSupportedLayout(MemoryLayout layout) {
+        if (layout instanceof SequenceLayout || layout instanceof PaddingLayout) {
```
JornVernee, line 113: Maybe we even want to check for padding layouts when creating the function descriptor. Does padding ever make sense as an argument/return layout?

JornVernee (approved):

src/java.base/share/classes/jdk/internal/foreign/abi/AbstractLinker.java
mcimadamore: I believe FunctionDescriptor::toMethodType is specified to throw if it sees padding?

JornVernee: Yes, I mean reject padding upon construction of the FunctionDescriptor.

mcimadamore: > Yes, I mean reject padding upon construction of the FunctionDescriptor (e.g. using `of` or `ofVoid`).

Ah... good point! I'll think some more.

src/java.base/share/classes/jdk/internal/foreign/FunctionDescriptorImpl.java
```diff
@@ -48,6 +49,13 @@ public final class FunctionDescriptorImpl implements FunctionDescriptor {
     private final List<MemoryLayout> argLayouts;
 
     private FunctionDescriptorImpl(MemoryLayout resLayout, List<MemoryLayout> argLayouts) {
+        if (resLayout instanceof PaddingLayout) {
+            throw new IllegalArgumentException("Unsupported padding layout return in function descriptor: " + resLayout);
+        }
+        Optional<MemoryLayout> paddingLayout = argLayouts.stream().filter(l -> l instanceof PaddingLayout).findAny();
```
JornVernee, line 55: This could use `Stream::anyMatch`.

test/jdk/java/foreign/TestFunctionDescriptor.java
```diff
@@ -141,4 +141,29 @@ public void testIllegalInsertArgOutOfBounds() {
...
 
+    @Test(expectedExceptions = IllegalArgumentException.class)
+    public void testBadPaddingInVoidFunction() {
+        FunctionDescriptor.ofVoid(MemoryLayout.paddingLayout(8));
+    }
+
+    @Test(expectedExceptions = IllegalArgumentException.class)
+    public void testBadPaddingInNonVoidFunction() {
+        FunctionDescriptor.ofVoid(MemoryLayout.paddingLayout(8));
+    }
```
JornVernee, line 152: This test is the same as the one above. Looks like you meant to call `of` here? (instead of `ofVoid`)

JornVernee (approved):
Nice! Couple of minor comments.

src/java.base/share/classes/jdk/internal/foreign/FunctionDescriptorImpl.java
JornVernee: Ok.

PR #831, opened 2023-05-04
8307461: Linker::nativeLinker should not be restricted

mcimadamore:
Since `Linker` is a sealed interface, we can now move restricted-ness from the static linker factory, down to the methods which are really problematic, namely `Linker::downcallHandle` and `Linker::upcallStub`.

With this change, the list of restricted method looks as follows:

* Linker::downcallHandle
* Linker::upcallStub
* MemorySegment::reinterpret
* SymbolLookup::libraryLookup

That is, crucially, restricted methods are methods that have a true risk of compromising the integrity of the Java platform. Making Linker::nativeLinker restricted casts too broad of a net, as obtaining the linker is not, per se, a dangerous or unsafe operation.

---------
### Progress

### Issue
 * [JDK-8307461](https://bugs.openjdk.org/browse/JDK-8307461): Linker::nativeLinker should not be restricted

### Reviewers
 * [Jorn Vernee](https://openjdk.org/census#jvernee) (@JornVernee - Committer) ⚠️ Review applies to [0e192ba8](https://git.openjdk.org/panama-foreign/pull/831/files/0e192ba88fedea9156ef34c28dc8671505b73ccb)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign.git pull/831/head:pull/831` \
`$ git checkout pull/831`

`$ git checkout pull/831` \
`$ git pull https://git.openjdk.org/panama-foreign.git pull/831/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 831`

`$ git pr show -t 831`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/831.diff">https://git.openjdk.org/panama-foreign/pull/831.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/panama-foreign/pull/831#issuecomment-1534909022)


test/jdk/java/foreign/handles/invoker_module/handle/invoker/MethodHandleInvoker.java
```diff
@@ -88,6 +88,8 @@ static <Z> void addDefaultMapping(Class<Z> carrier, Z value) {
         addDefaultMapping(AddressLayout.class, ValueLayout.ADDRESS);
         addDefaultMapping(SymbolLookup.class, SymbolLookup.loaderLookup());
         addDefaultMapping(Consumer.class, (Consumer<Object>)(Object o) -> {});
+        addDefaultMapping(FunctionDescriptor.class, FunctionDescriptor.ofVoid());
+        addDefaultMapping(Linker.Option[].class, null);
```
mcimadamore, line 92: This is a pragmatic change (not too proud of it). Since the linker methods are varargs, invoking these methods with `invokeWithArguments` will expect separate `Linker.Option` parameters instead of a single array. So I pass null instead.

JornVernee (approved):

src/java.base/share/classes/jdk/internal/foreign/abi/AbstractLinker.java
```diff
@@ -68,7 +70,17 @@ private record LinkRequest(FunctionDescriptor descriptor, LinkerOptions options)
...
+    public final MethodHandle downcallHandle(MemorySegment symbol, FunctionDescriptor function, Option... options) {
+        Reflection.ensureNativeAccess(Reflection.getCallerClass(), Linker.class, "downcallHandle");
+        SharedUtils.checkSymbol(symbol);
+        return downcallHandle(function, options).bindTo(symbol);
+    }
+
+    @Override
+    @CallerSensitive
+    public final MethodHandle downcallHandle(FunctionDescriptor function, Option... options) {
+        Reflection.ensureNativeAccess(Reflection.getCallerClass(), Linker.class, "downcallHandle");
```
ExE-Boss, line 83: This needs to use `downcallHandle0` so as to not perform 2 consecutive `Reflection::ensureNativeAccess` calls with the second having a caller class of `jdk.internal.foreign.abi.AbstractLinker`.
```suggestion
        return downcallHandle0(function, options).bindTo(symbol);
    }

    @Override
    @CallerSensitive
    public final MethodHandle downcallHandle(FunctionDescriptor function, Option... options) {
        Reflection.ensureNativeAccess(Reflection.getCallerClass(), Linker.class, "downcallHandle");
        return downcallHandle0(function, options);
    }

    private MethodHandle downcallHandle0(FunctionDescriptor function, Option[] options) {
```

PR #833, opened 2023-05-26
8308967: Add a record mapper for mapping MemorySegments to Record instances

minborg:
This PR proposes to add a "Record Mapper" that allows `MemorySegment` instances to be projected onto `Record` types given `GroupLayout` instances.

Here is a simple example of how to use the mapper:

```
    private static final GroupLayout POINT_LAYOUT = MemoryLayout.structLayout(
            JAVA_INT.withName("x"),
            JAVA_INT.withName("y"));

    MemorySegment segment =  MemorySegment.ofArray(new int[]{3, 4});

    Point point = POINT_LAYOUT.recordMapper(Point.class)
          .apply(segment); // Point[x=3, y=4]
```

I think the implementation can be improved later, for example by de-duplicating handling of arrays and maybe add recursive "un-pealing" of multidimensional arrays.

---------
### Progress
- [ ] Change requires CSR request [JDK-8308968](https://bugs.openjdk.org/browse/JDK-8308968) to be approved

### Issues
 * [JDK-8308967](https://bugs.openjdk.org/browse/JDK-8308967): Add a record mapper for mapping MemorySegments to Record instances (**Enhancement** - P3)
 * [JDK-8308968](https://bugs.openjdk.org/browse/JDK-8308968): Add a record mapper for mapping MemorySegments to Record instances (**CSR**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign.git pull/833/head:pull/833` \
`$ git checkout pull/833`

`$ git checkout pull/833` \
`$ git pull https://git.openjdk.org/panama-foreign.git pull/833/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 833`

`$ git pr show -t 833`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/833.diff">https://git.openjdk.org/panama-foreign/pull/833.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/panama-foreign/pull/833#issuecomment-1564331329)


src/java.base/share/classes/java/lang/foreign/GroupLayout.java
```diff
@@ -73,4 +76,136 @@ public sealed interface GroupLayout extends MemoryLayout permits StructLayout, U
...
+     *
+     * @param <R> record type
+     * @param type the type (Class) of the record
+     * @throws IllegalArgumentException if the provided record {@code type} contains components for which
+     *                                  there are no exact mapping (of names and types) in this group layout
+     *                                  or if the provided {@code type} is not public or if the method is otherwise
+     *                                  unable to create a record mapper as specified above.
+     * @since 21
+     */
+    <R extends Record> Function<MemorySegment, R> recordMapper(Class<R> type);
```
forax, line 209: you also need to throw an IAE if type is Record.class, java.lang.Record is the only subtype of java.lang.Record which is not a record

src/java.base/share/classes/jdk/internal/foreign/LayoutRecordMapper.java
```diff
@@ -0,0 +1,641 @@
...
+
+        var missingComponents = recordComponents.stream()
+                .map(RecordComponent::getName)
+                .filter(l -> !nameToLayoutMap.containsKey(l))
+                .toList();
+
+        if (!missingComponents.isEmpty()) {
+            throw new IllegalArgumentException("There is no mapping for " +
+                    missingComponents + " in " + type.getName() +
+                    "(" + String.join(", ", recordComponents.stream().map(RecordComponent::getName).collect(Collectors.joining(", "))) + ")" +
```
forax, line 108: Two joins, String.join and Collectors.joining() ??

src/java.base/share/classes/jdk/internal/foreign/LayoutRecordMapper.java
```diff
@@ -0,0 +1,641 @@
...
+                .toList();
+
+        Class<?>[] ctorParameterTypes = recordComponents.stream()
+                .map(RecordComponent::getType)
+                .toArray(Class<?>[]::new);
+
+        // An array of the record component MethodHandle extractors, each of type (MemorySegment)X
+        // where X is the component type.
+        MethodHandle[] handles = componentAndLayoutList.stream()
+                .map(cl -> {
```
forax, line 123: map() + filter(Object::isNull) should be a flatMap() or a mapMulti()

src/java.base/share/classes/jdk/internal/foreign/LayoutRecordMapper.java
```diff
@@ -0,0 +1,641 @@
...
+            case ValueLayout.OfDouble __ -> ValueLayout.OfDouble.class;
+            case AddressLayout __ -> AddressLayout.class;
+        };
+    }
+
+    static MethodHandle findStaticToArray(Class<?> rType,
+                                          Class<?> layoutType,
+                                          Class<?> extra) throws NoSuchMethodException, IllegalAccessException {
+
+        var pTypes = Stream.of(MemorySegment.class, layoutType, long.class, long.class, extra)
```
forax, line 373: seems a weird way to test if extra is null

src/java.base/share/classes/java/lang/foreign/GroupLayout.java
```diff
@@ -73,4 +76,136 @@ public sealed interface GroupLayout extends MemoryLayout permits StructLayout, U
...
+
+    /**
+     * {@return a {@link Function} that can project {@linkplain MemorySegment MemorySegments} onto new instances
+     * of the provided {@link Record} {@code type} by means of matching the names of the
+     * record components with the names of the elements in this group layout}
+     * <p>
+     * The mapping between the record type and this memory layout is defined as follows;
+     * <p>
+     * Let <em>R</em> be the provided record {@code type} with the constituent components
+     * <em>C<sub>0</sub></em>, ..., <em>C<sub>N-1</sub></em>, where <em>N</em> is non-negative.
```
mcimadamore, line 88: general stylistic note - you some fancy html here to get close to some math notation. This is nice, but is also inconsistent with what has been done in other parts of the memory layout API - see e.g. MemoryLayout::var handle, which just uses `A_0`, `A_1` etc.

src/java.base/share/classes/java/lang/foreign/GroupLayout.java
```diff
@@ -73,4 +76,136 @@ public sealed interface GroupLayout extends MemoryLayout permits StructLayout, U
...
+     * <p>
+     * Let <em>L</em> be this group layout with the elements <em>E<sub>0</sub></em>, ..., <em>E<sub>M-1</sub></em>
+     * , where <em>M</em> {@code >=} <em>N</em>.
+     * <p>
+     * Then, for each <em>C<sub>a</sub></em> (<em>a</em> {@code <} <em>N</em>), there must be a corresponding distinct
```
mcimadamore, line 93: As discussed offline, I think the documentation would be clearer (and more precise) if we:

* Defined some helper function `offsetOf(L, LL)` which gives offset of layout LL in L (no need to specify how it's done, just use it as an explanation device (perhaps hinting at `MemoryLayout::byteOffset`
* Defined a mapping from (offset, layout, class<T>) to a value of type T. This mapping needs to be defined for _all_ layouts
* This allow us to define the sequence case as just using recursion: e.g. given the triple (O, Sequence(E), T[]), for each element E at offset OO = O + offsetOf(Sequence(E), E), we compute the value associated with (OO, E, T). Then we collect all such values into a new T[] array (whose size is the same as that of the sequence layout). I believe this does the right thing even in the case of multi-dimensional arrays.
* Then , recordMapper(G, R) is defined simply as the value associated with the triple (0, G, R).

src/java.base/share/classes/java/lang/foreign/GroupLayout.java
mcimadamore: I wonder if an overload which takes a MethodHandles.Lookup might help with the accessibility issues (but we can add that later)

src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
```diff
@@ -163,6 +163,30 @@
  * long offset2 = (long) offsetHandle.invokeExact(2L); // 16
  * }
  *
+ * <h2 id="record-mapping">Record mapping</h2>
+ * <p>A {@linkplain GroupLayout group layout} can provide mapping capabilities from memory segments to Java
+ * {@linkplain Record Records} by means of matching named elements in the group layout with component names
```
mcimadamore, line 168: This should be a link to the JLS section for records - we do that in other parts of the javadoc - e.g.:

```
Note the use of the <em>try-with-resources</em> construct: this idiom ensures that the off-heap region
 * of memory backing the native segment will be released at the end of the block, according to the semantics described
 * in Section {@jls 14.20.3} of <cite>The Java Language Specification</cite>.
 ```

File: src/java.base/share/classes/jdk/internal/foreign/LayoutRecordMapper.java (hunk 2 above)
Comment (mcimadamore) [reply src/java.base/share/classes/jdk/internal/foreign/LayoutRecordMapper.java:108]: Also, I think this can be simplified - e.g. "there is no mapping for [list of component names] in [group layout]". E.g. just printing the group layout will be enough, I think.

File: src/java.base/share/classes/jdk/internal/foreign/LayoutRecordMapper.java [hunk 8]
```diff
@@ -0,0 +1,641 @@
...
+
+                                if (dimensionOf(componentType) != info.sequences().size()) {
+                                    throw new IllegalArgumentException("Unable to map '" + sl + "'" +
+                                            " of dimension " + info.sequences().size() +
+                                            " because the component '" + componentType.getName() + " " + name + "'" +
+                                            " has a dimension of " + dimensionOf(componentType));
+                                }
+
+                                // Handle multi-dimensional arrays
+                                if (info.sequences().size() > 1) {
```
mcimadamore, line 179: I have a feeling that this could be simplified if we embraced recursion, and defined the mapping as a function taking three arguments: offset, layout and carrier. The output of that function is a MethodHandle. Record mapper is not special - I mean, it is special as that's the only entry point we provide in the API, but that is based on a more general way to map layouts and carriers to value, so we should probably write that general function first, and then map that to the recordMapper API point.

src/java.base/share/classes/jdk/internal/foreign/LayoutRecordMapper.java
```diff
@@ -0,0 +1,641 @@
...
+
+        // An array of the record component MethodHandle extractors, each of type (MemorySegment)X
+        // where X is the component type.
+        MethodHandle[] handles = componentAndLayoutList.stream()
+                .map(cl -> {
+                    var name = cl.layout().name().orElseThrow();
+                    var pathElement = MemoryLayout.PathElement.groupElement(name);
+                    long byteOffset = layout.byteOffset(pathElement) + offset;
+
+                    return switch (cl.layout()) {
```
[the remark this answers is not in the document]
mcimadamore: I agree - ideally, each case should be split into its own method. That should help recursion too.

src/java.base/share/classes/jdk/internal/foreign/LayoutRecordMapper.java
```diff
@@ -0,0 +1,641 @@
...
+
+/**
+ * A record mapper that is matching a GroupLayout to match the components of a record.
+ *
+ * @param <T> the Record type
+ */
+public final class LayoutRecordMapper<T extends Record>
+        implements Function<MemorySegment, T> {
+
+    enum Allow {EXACT, BOXING, BOXING_NARROWING_AND_WIDENING}
```
mcimadamore, line 66: I'd prefer if this implementation class (which is already complex) started off simple, and provided only mappings for doing exact match (which is what the API point says anyway). There is some noise that is introduced later on to deal with wrapper classes, and I'd like us to focus on the basics first.

src/java.base/share/classes/jdk/internal/foreign/LayoutRecordMapper.java
```diff
@@ -0,0 +1,641 @@
...
+    }
+
+    private LayoutRecordMapper(Class<T> type,
+                               GroupLayout layout,
+                               long offset) {
+        this.type = type;
+        this.layout = layout;
+
+        var recordComponents = Arrays.asList(type.getRecordComponents());
+        var recordComponentNames = recordComponents.stream()
```
mcimadamore, line 89: I have a feeling that the use of streams here is hiding a much simpler loop where we loop over the record components, we check that each of them have a corresponding layout in the Group. For instance:

```
Map<String, MemoryLayout> nameToLayoutMap = ...
for (RecordComponent rc : type.getRecordComponents()) {
    try {
         MemoryLayout memberLayout = layout.select(PathElement.groupElement(rc.name());
         nameToLayoutMap = nameToLayoutMap(memberLayout.name().get(), memberLayout);
    } catch (IllegalArgumentException ex) {
         throw new IllegalArgumentException("There is no mapping...");
    }
}
```

src/java.base/share/classes/jdk/internal/foreign/LayoutRecordMapper.java
mcimadamore: > I have a feeling that this could be simplified if we embraced recursion, and defined the mapping as a function taking three arguments: offset, layout and carrier. The output of that function is a MethodHandle. Record mapper is not special - I mean, it is special as that's the only entry point we provide in the API, but that is based on a more general way to map layouts and carriers to value, so we should probably write that general function first, and then map that to the recordMapper API point.

That said, I think your optimization of the one-dimensional case also makes sense, as in those cases we know we can just bulk copy into the target array. So perhaps, the split between one-dimensional and multi-dimensional is gonna remain (but I think in terms of the javadoc, the split shouldn't be there).

Glavo:
Is it possible that we will get something like this in the future:

C Header:
```c
struct Point {
    int x, y;
};

extern void f1(struct Point p);
extern void f2(struct Point *pp);
```

Java Code:
```java
record Point(int x, int y) {}

GroupLayout pointLayout = MemoryLayout.structLayout(
            JAVA_INT.withName("x"),
            JAVA_INT.withName("y"));

ValueLayout pointValueLayout = pointLayout.withJavaType(Point.class);
ValueLayout pointPtrLayout = ADDRESS.withTargetLayout(pointLayout).withJavaType(Point.class);

MethodHandle f1 = nativeLinker.downcallHandle(lookup.find("f1").get(), FunctionDescriptor.ofVoid(pointValueLayout));
MethodHandle f2 = nativeLinker.downcallHandle(lookup.find("f2").get(), FunctionDescriptor.ofVoid(pointPtrLayout));

f1.invoke(new Point(10, 20)); // ok
f2.invoke(new Point(10, 20)); // ok
```

src/java.base/share/classes/jdk/internal/foreign/LayoutRecordMapper.java
```diff
@@ -0,0 +1,641 @@
...
+                                                        ms -> ms.toArray(ofDouble);
+                                                case AddressLayout addressLayout ->
+                                                        ms -> ms.elements(addressLayout)
+                                                                .map(s -> s.get(addressLayout, 0))
+                                                                .toArray(MemorySegment[]::new);
+                                            };
+                                            // (MemorySegment, Function mapper) ->
+                                            // (MemorySegment)
+                                            mh = MethodHandles.insertArguments(mh, 1, leafArrayMapper);
+                                            yield castReturnType(mh, cl.component().getType());
```
mcimadamore, line 220: I'm not sure about this and other casts in this branch of the code. isn't the method handle here returning some T[] ? If so, how can we cast it to the component type of cl? (btw, we already have `cl.componentType` stored in a variable, I think).

mcimadamore: > I'm not sure about this and other casts in this branch of the code

Never mind, I confused myself between record components and Class::componentType (for the array element type) :-)

src/java.base/share/classes/jdk/internal/foreign/LayoutRecordMapper.java
```diff
@@ -0,0 +1,641 @@
...
+                                        }
+                                        case GroupLayout gl -> {
+                                            @SuppressWarnings("unchecked")
+                                            var arrayComponentType = (Class<? extends Record>) deepArrayComponentType(cl.component.getType());
+                                            // The "local" byteOffset for the record component mapper is zero
+                                            var componentMapper = recordMapper(arrayComponentType, gl, 0);
+                                            Function<MemorySegment, Object> leafArrayMapper = ms -> {
+                                                Object leafArray = Array.newInstance(arrayComponentType, info.lastDimension());
+
+                                                int[] i = new int[]{0};
```
mcimadamore, line 230: probably another instance where a loop would be simpler

src/java.base/share/classes/java/lang/foreign/GroupLayout.java
forax: Yes ! and i'm not even sure that having a version that does not take a Lookup make sense.

src/java.base/share/classes/jdk/internal/foreign/LayoutRecordMapper.java
mcimadamore: Using this method for both the primitive and the record case leads to issues. Better use a different way for the recursive record case. Also, you can create a method type from a List::of, which is probably simpler.

src/java.base/share/classes/jdk/internal/foreign/LayoutRecordMapper.java
minborg: The filter operation should be removed.

Glavo:
@mcimadamore I realized that establishing a mapping between record and MemorySegment was an overly high-level operation, so I later modified the above example to focus on mapping between record and struct.



> ```java
> record Point(int x, int y) {}
> 
> ValueLayout pointLayout = MemoryLayout.structLayout(
>                 JAVA_INT.withName("x"),
>                 JAVA_INT.withName("y"))
>         .withCarrier(Point.class);
> 
> MethodHandle f = nativeLinker.downcallHandle(lookup.find("f").get(), FunctionDescriptor.ofVoid(pointLayout));
> f.invoke(new Point(10, 20)); // ok
> ```

Now we do not have any type that directly corresponds to the struct type of C, so we have to convert the value of the structure to MemorySegment before passing it.

I think this status quo is unsatisfactory. Because of this, I think providing the ability to establish a mapping between record and struct is meaningful.

mcimadamore:
> @mcimadamore I realized that establishing a mapping between record and MemorySegment was an overly high-level operation, so I later modified the above example to focus on mapping between record and struct.
> 
> > ```java
> > record Point(int x, int y) {}
> > 
> > ValueLayout pointLayout = MemoryLayout.structLayout(
> >                 JAVA_INT.withName("x"),
> >                 JAVA_INT.withName("y"))
> >         .withCarrier(Point.class);
> > 
> > MethodHandle f = nativeLinker.downcallHandle(lookup.find("f").get(), FunctionDescriptor.ofVoid(pointLayout));
> > f.invoke(new Point(10, 20)); // ok
> > ```
> 
> Now we do not have any type that directly corresponds to the struct type of C, so we have to convert the value of the structure to MemorySegment before passing it.
> 
> I think this status quo is unsatisfactory. Because of this, I think providing the ability to establish a mapping between record and struct is meaningful.

Having bi-directional conversions from memory segments to records is attractive, and it is something that we have explored (as mentioned above), but abandoned because, as stated above, it cannot be generalized to all by-value arguments. While structs are the most common case, it is smelly that we can only support "half" of the composite types with the system you envision.

The other issue is that you say that it is suboptimal that we need to go from the record to the segment when calling a native function. Well, yes, but it is similarly suboptimal to just use records directly - as you can see from the RecordMapper implementation, there's quite a lot of "plumbing" that needs to occur so that fields can be extracted. And, in many cases, the Linker will have to bulk-copy the struct or union somewhere else (e.g. in the stack). So, a record class with indirections is not in a form that is "easy" to use for the linker (unlike a flat memory segment). This is the main reason as to why the Linker API likes its by-value composite arguments to be passed as memory segments.

If a tool or a client would like to provide different mappings for structs, they are certainly free to do so (and the Linker API should not block them). But if clients are after performance, a record is a a handy, but lousy carrier for passing arguments to the linker (or to dereference APIs). A much better carrier is (as stated before) a class that wraps a memory segment, so that the segment can be extracted immediately, and then used by the linker to do what is required.

Ultmately, I think you are reading too much into what this API is trying to provide (but I understand the angle you came from, as we were tempted in similar directions). This is a way for clients to see what's inside a memory segment. It replaces other experimental APIs that have been provided in the past to - e.g. generate an hex dump from a segment, or generate a string representation of the contents of a segment. We realized that, by mapping a segment into a record, we could provide a much better description of "what's in the segment", in a way that is transparent to Java code, can be pattern-matched, etc.

Glavo:
@minborg Thank you very much for your answer. Despite this, I am still a bit frustrated.

> The other issue is that you say that it is suboptimal that we need to go from the record to the segment when calling a native function. Well, yes, but it is similarly suboptimal to just use records directly - as you can see from the RecordMapper implementation, there's quite a lot of "plumbing" that needs to occur so that fields can be extracted. And, in many cases, the Linker will have to bulk-copy the struct or union somewhere else (e.g. in the stack). So, a record class with indirections is not in a form that is "easy" to use for the linker (unlike a flat memory segment). This is the main reason as to why the Linker API likes its by-value composite arguments to be passed as memory segments.

Maybe my thinking is too idealistic. My thought is that in theory the JIT compiler also understands the calling convention, so it can optimize in conjunction with scalar replacement (or Valhalla), use registers to pass parameters and accept return values ​​as much as possible, so that the call cost is close to native.

This may be me being too picky, I just hope that Panama performance has more room for optimization. It would be great if there is a better solution to these problems in the future.

mcimadamore:
> Maybe my thinking is too idealistic. My thought is that in theory the JIT compiler also understands the calling convention, so it can optimize in conjunction with scalar replacement (or Valhalla), use registers to pass parameters and accept return values ​​as much as possible, so that the call cost is close to native.

Heh - believe me, I get where you are coming from. But my feeling is that what you are thinking about a world where instances of value classes are "flat" and have a layout which overlaps exactly with the layout of their corresponding struct in C. This is **not** the case, at least to date, with value types. E.g. the JVM will reserve the right of insert extra padding, align fields, pretty much in every way it wants. So there's no guarantee that a record (even if a value record) is going to be "bulk-copy compatible" with a memory segment.

What Valhalla is experimenting with, in the context of the vector API [1] is an approach by which some "selected" carriers within the JDK will be known to the VM to have a layout that is "bulk-copy compatible". For such carrier types, we might be able to support multi-byte reads and writes (e.g. using vector operations) and Unsafe might well provide support for such a thing. But these types will be only usable by the JDK - user-defined value types will not, in general, obey the same properties, and will not provide enough guarantees to the FFM API.

What Valhalla might provide is:
* better Java carriers for e.g. Long128 (and also vector types) - this is useful, as some C types are bigger than 64 bits
* proper ways to express types such as Complex<T> and Unsigned<T>
* a much more lightweight way to map structs to segment wrappers (e.g. a value class which wraps a segment with getters/setters)

All these things are all "pieces of the puzzle" to define native interop that is both efficient and easy to use for developers. But we would like to get there in a layered fashion. Memory segments seem to provide the "lingua franca" that allows the linker to go about its business. Once we have all these other useful (and cheap!) types at our disposal, surely jextract can use a combination of tactics to make its bindings more user-friendly. And, if some of the jextract idioms sediments and seems useful enough, it would certainly be possible to capture it at the level of the FFM API itself, as something that sits immediately on top of Linker.

[1] - https://github.com/openjdk/valhalla/pull/790

src/java.base/share/classes/jdk/internal/foreign/LayoutRecordMapper.java
```diff
@@ -0,0 +1,641 @@
...
+        try {
+            var ctor = PUBLIC_LOOKUP.findConstructor(type, MethodType.methodType(void.class, ctorParameterTypes));
+            for (int i = 0; i < handles.length; i++) {
+                // Insert the respective handler for the constructor
+                ctor = MethodHandles.filterArguments(ctor, i, handles[i]);
+            }
+
+            var mt = MethodType.methodType(type, MemorySegment.class);
+            // Fold the many identical MemorySegment arguments into a single argument
+            ctor = MethodHandles.permuteArguments(ctor, mt, IntStream.range(0, handles.length)
```
Glavo, line 312: What is the special meaning of using `IntStream`? Why not `new int[handles.length]`?

Glavo:
> Heh - believe me, I get where you are coming from. But my feeling is that what you are thinking about a world where instances of value classes are "flat" and have a layout which overlaps exactly with the layout of their corresponding struct in C. This is not the case, at least to date, with value types. E.g. the JVM will reserve the right of insert extra padding, align fields, pretty much in every way it wants. So there's no guarantee that a record (even if a value record) is going to be "bulk-copy compatible" with a memory segment.

I think you misunderstood my meaning. I understand that we cannot assume the actual memory layout of valhalla value objects, so we cannot truly equate a Java object with a C struct value. 

But these are all for objects stored on the heap. For local variables, the JVM can handle them in a completely different way, scalar replacement is an example.

So, what I hope to do is not treat the struct value as a record **object**, but rather treat the struct value as a record **variable**.

Glavo:
> And, in many cases, the Linker will have to bulk-copy the struct or union somewhere else (e.g. in the stack). 

Because I only want to map the struct value to the record local variable, I don't really want to allocate a record object unless it escapes. So, I don't think this will cause any trouble on this issue.

Well, specifically...Okay, let's use this PR as an example.

I don't want to map a `struct Point *` to `Point[]`. What I want is:

```java
private static final ValueLayout.OfRecord<Point> POINT_LAYOUT = MemoryLayout.structLayout(
                JAVA_INT.withName("x"),
                JAVA_INT.withName("y"))
        .withCarrier(Point.class);

MemorySegment segment =  MemorySegment.ofArray(new int[]{3, 4});

Point p = segment.get(POINT_LAYOUT, 0); // Point[x=3, y=4]
segment.set(POINT_LAYOUT, 0, new Point(5, 6));
```

I prefer this kind of thing to this PR.

Similar to the example I wrote earlier that called the C function, all records are only used locally in the method. I hope the JIT compiler can think that passing a record to an foreign MethodHandle is not an escape, and store the record object freely on the stack, passing it through registers to the C function, thereby avoiding additional overhead.

mcimadamore:
> 
> I don't want to map a `struct Point *` to `Point[]`. What I want is:
> 
This PR doesn't map a `struct Point*` to `Point[]`. This PR allows (_if you want and/or find it useful_) to view the bytes of a memory segment as components of a record class.

You keep bringing up the topic of having a deeper connection between structs and records, using this for dereference and for linking. And we keep looking at extremely simplistic struct declaration. What if the struct contains an array? Or a union? Those are the bits where the mapping becomes not as clearly defined as you make it to be. Should we add the API points you describe only for the case of a simple struct with X and Y fields? And, what if a developer realizes that some other kind of mapping, not involving records, might be more useful? How would that scenario be plugged into the API (which would, at that point be already biased towards records?).

I'm not trying to minimize your suggestions - but I think if you think through all the consequences of what you are proposing, you end up pulling in a fairly general custom serialization/deserialization strategy for segments, which, arguably, is exactly what we'd like the core API **not** to do.

(Your example is not clear as you use ValueLayout.OfRecord<X> (as you need generics for what you want to do) but then use a `structLayout` on the other side)

mcimadamore:
> So, what I hope to do is not treat the struct value as a record **object**, but rather treat the struct value as a record **variable**.

I have no problem believing that a simple record such as point can scalarize well, and that access to components can be fast. But again, a struct can contain other structs, unions and arrays. And once you pull that stuff in, I think the kind of optimizations you have in mind don't scale that well. As I said, in case of a struct bigger than 64 or 128, most ABIs will just want to copy all its bytes onto the stack (or on a combination on registers and stack). Now, if I have a segment that is easy to do. If you have a record that means pulling every single record components, recursively (chasing arrays and records in records). There's lots of indirections to follow in order to be able to "put back" all the bytes that needs to be copied. I don't think something like this is likely to scale in the general case.

So, while "returning a record" might provide acceptable performance, in the case where everything is inlined and the all the components are scalarized (and maybe the JIT figures out you are only going to use a couple out of the many ones you have). But if there's no inlining, scalarization, things will become a lot worse than just wrapping in a memory segment.
And, for parameter passing, there's no smart speculation possible, since we have to copy all the record components in places dictated by the ABI - meaning we have to read everything (and follow the indirections and take the memory hits).

src/java.base/share/classes/jdk/internal/foreign/LayoutRecordMapper.java
```diff
@@ -0,0 +1,634 @@
...
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Objects;
+import java.util.Set;
+import java.util.function.BinaryOperator;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
```
Glavo, line 53: redundant import

minborg:
With the composed method handles, performance now looks excellent. Here is a benchmark where we materialize a `record Point(int x, int y)` (benchmarks "point*Mapper") and an array of three `int` components (benchmarks "array*Mapper"):

```
Benchmark                         Mode  Cnt  Score   Error  Units
RecordMapper.arrayExplicitMapper  avgt   30  9.361 ± 0.060  ns/op
RecordMapper.arrayMapper          avgt   30  7.697 ± 0.035  ns/op
RecordMapper.pointExplicitMapper  avgt   30  2.235 ± 0.016  ns/op
RecordMapper.pointMapper          avgt   30  2.299 ± 0.007  ns/op
```

The explicit mappers are hand-crafted mappers like this one:

```
private static final Function<MemorySegment, Point> POINT_EXPLICIT_MAPPER =
        ms -> new Point(ms.get(JAVA_INT, 0L), ms.get(JAVA_INT, 4));
```

src/java.base/share/classes/java/lang/foreign/GroupLayout.java
```diff
@@ -82,73 +82,81 @@ public sealed interface GroupLayout extends MemoryLayout permits StructLayout, U
      * of the provided {@link Record} {@code type} by means of matching the names of the
      * record components with the names of the elements in this group layout}
      * <p>
-     * The mapping between the record type and this memory layout is defined as follows;
+     * A mapper {@code M} between a record type {@code R}, a memory layout {@code L} and an {@code offset}
+     * (which is zero for the initial invocation of this method) is defined as follows {@code M(type, layout, offset)};
      * <p>
-     * Let <em>R</em> be the provided record {@code type} with the constituent components
-     * <em>C<sub>0</sub></em>, ..., <em>C<sub>N-1</sub></em>, where <em>N</em> is non-negative.
+     * Let {@code R} comprise its the constituent components
```
mcimadamore, line 88: Maybe simplify: "Let C1, C2 ... Cn" be the components of R (with N non-negative)"

src/java.base/share/classes/java/lang/foreign/GroupLayout.java
```diff
@@ -82,73 +82,81 @@ public sealed interface GroupLayout extends MemoryLayout permits StructLayout, U
...
-     * , where <em>M</em> {@code >=} <em>N</em>.
+     * Let {@code f_a(MemorySegment ms)} be a function that takes
+     * a {@link MemorySegment} {@code ms} and produces a value for {@code c_a}.
      * <p>
-     * Then, for each <em>C<sub>a</sub></em> (<em>a</em> {@code <} <em>N</em>), there must be a corresponding distinct
-     * <em>E<sub>b</sub></em> such that the {@link MemoryLayout#name() name} of <em>C<sub>a</sub></em>
-     * {@link Object#equals(Object) equals} the {@link RecordComponent#getName() name} of the <em>E<sub>b</sub></em> and:
+     * Let {@code L} be a group layout with the elements {@code e_1, e_2, ..., e_M}, where {@code M >= N}.
+     * <p>
+     * Let {@code offsetBetween(GroupLayout layout, MemoryLayout target)} be a function that, via
```
mcimadamore, line 96: I would use two letters as the argument of `offsetBetween`. E.g.
"Let offsetOf(L, LL) be a function that, via ... , can compute the offset from the layout L to a sub-layout LL"
Note the change of name. I don't think "between" is a good choice of word.

src/java.base/share/classes/java/lang/foreign/GroupLayout.java
```diff
@@ -82,73 +82,81 @@ public sealed interface GroupLayout extends MemoryLayout permits StructLayout, U
...
+     * the method {@link MemoryLayout#byteOffset(PathElement...)}, can compute the offset from the
+     * {@code layout L} to a sub-layout of {@code L}.
+     * <p>
+     * Then, for each {@code c_a, a <= N}, there must be a corresponding distinct
```
mcimadamore, line 100: I thought letters `c` were for record components? - but you say "name of c_a" ?

src/java.base/share/classes/java/lang/foreign/GroupLayout.java
```diff
@@ -82,73 +82,81 @@ public sealed interface GroupLayout extends MemoryLayout permits StructLayout, U
...
+     * {@code e_b} such that the {@link MemoryLayout#name() name} of {@code c_a}
+     * {@link Object#equals(Object) equals} the {@link RecordComponent#getName() name} of {@code e_b} and:
```
mcimadamore, line 102: such that the name ... and the name ... are the same

src/java.base/share/classes/java/lang/foreign/GroupLayout.java
```diff
@@ -82,73 +82,81 @@ public sealed interface GroupLayout extends MemoryLayout permits StructLayout, U
...
      * <ul>
      *    <li>
-     *        <h4>If <em>E<sub>b</sub></em> is a {@link ValueLayout };</h4>
-     *        then <em>C<sub>a</sub></em> must be of the exact type of <em>E<sub>b</sub>'s</em> {@link ValueLayout#carrier() carrier()}<br>
-     *        whereby C<sub>a</sub> = f<sub>a</sub>(MemorySegment ms) = ms.get(E<sub>b</sub>, offset) at the appropriate offset.<br>
+     *        <h4>If {@code e_b} is a {@link ValueLayout };</h4>
+     *        then {@code c_a} must be of the exact type of {@code e_b}'s {@link ValueLayout#carrier() carrier()}<br>
+     *        whereby {@code c_a = f_a(MemorySegment ms) = ms.get(e_b, offset + offsetBetween(layout, e_b))}.<br>
```
mcimadamore, line 107: It is not defined what "offset" is. I believe that has to be a parameter to the f_a - e.g. the function takes a segment *and* and offset.

src/java.base/share/classes/java/lang/foreign/GroupLayout.java
```diff
@@ -82,73 +82,81 @@ public sealed interface GroupLayout extends MemoryLayout permits StructLayout, U
...
-     *        <h4>If <em>E<sub>b</sub></em> is a {@link SequenceLayout };</h4>
-     *        then <em>C<sub>a</sub></em> must be an array <em>C[]<sup>D</sup></em> (of depth <em>D</em>
-     *        and with an array component type <em>C</em>) that can be mapped to <em>E<sub>b</sub></em> via a resulting
-     *        "array mapper" <em>A2</em> obtained via recursively pealing off nested sequence layouts in <em>E<sub>b</sub></em>
-     *        and then (after <em>D</em> pealing operations)
-     *        finally determining the leaf element layout <em>LL</em> = {@link SequenceLayout#elementLayout() elementLayout()}
-     *        and subsequently obtaining a leaf record mapper:
+     *        <h4>If {@code e_b} is a {@link SequenceLayout };</h4>
+     *        then {@code c_a} must be an array {@code C[]^D} (an array of depth {@code D}
+     *        and with an array component type {@code C}) that can be mapped to {@code e_b} via a resulting
```
mcimadamore, line 119: Can't all this now be expressed more simply using recursion? E.g. given the element layout E and the component type C (which could be another array), we compute, for each element in the sequence layout, the mapping between E, C and the corresponding element offset. Then we return new C[] with all the elements so computed. If you propagate the offsets correctly it should "just work" ?

src/java.base/share/classes/java/lang/foreign/GroupLayout.java
```diff
@@ -82,73 +82,81 @@ public sealed interface GroupLayout extends MemoryLayout permits StructLayout, U
...
-     *        applying {@code A2} which, in turn, will apply {@code LM} recursively at the appropriate offset(s).<br>
+     *        whereby {@code c_a} = {@code f_a(MemorySegment ms)} will be extracted by
+     *        applying {@code AM} which, in turn, will apply {@code LM} recursively at
+     *        a memory segment slice at the applicable multidimensional array offset(s).<br>
      *        Note: boolean arrays are not supported despite the above and if an attempt is made to map
      *        a boolean array, an {@link IllegalArgumentException} will be thrown.
      *    </li>
      *    <li>
-     *        <h4>If <em>E<sub>b</sub></em> is a {@link PaddingLayout };</h4>
+     *        <h4>If {@code e_b} is a {@link PaddingLayout };</h4>
```
mcimadamore, line 143: It feels like these two last bullets can be merged into "otherwise throw". Maybe that text can make an example of when that could happen (e.g. padding layout or String carrier).

src/java.base/share/classes/jdk/internal/foreign/LayoutRecordMapper.java
```diff
@@ -42,301 +42,289 @@
...
     private final GroupLayout layout;
+    private final long offset;
+    private final int depth;
+    private final MethodHandles.Lookup lookup;
     private final MethodHandle ctor;
 
     public LayoutRecordMapper(Class<T> type,
                               GroupLayout layout) {
-        this(type, layout, 0);
+        this(type, layout, 0L, 0, PUBLIC_LOOKUP);
```
mcimadamore, line 72: I tend to agree with Remi that passing an explicit lookup to the mapper factory is unavoidable here.

src/java.base/share/classes/jdk/internal/foreign/LayoutRecordMapper.java
```diff
@@ -42,301 +42,289 @@
...
 
-        var componentAndLayoutList = recordComponents.stream()
-                .map(c -> new ComponentAndLayout(c, nameToLayoutMap.get(c.getName())))
+        // For each component, find an f(a) = MethodHandle(MemorySegment) that returns the component type
+        var handles = Arrays.stream(type.getRecordComponents())
+                .map(this::methodHandle)
                 .toList();
 
-        Class<?>[] ctorParameterTypes = recordComponents.stream()
+        System.out.println("handles for " + type + " are " + handles);
```
mcimadamore: This is a leftover from some debugging code? There's also some others.

src/java.base/share/classes/jdk/internal/foreign/LayoutRecordMapper.java
```diff
@@ -42,301 +42,289 @@
...
+                                      long byteOffset) throws NoSuchMethodException, IllegalAccessException {
+
+        String name = component.getName();
+        var componentType = component.getType();
+        if (!componentType.isArray()) {
+            throw new IllegalArgumentException("Unable to map '" + sl +
+                    "' because the component '" + componentType.getName() + " " + name + "' is not an array");
+        }
+
+        MultidimensionalSequenceLayoutInfo info = MultidimensionalSequenceLayoutInfo.of(sl, componentType);
```
mcimadamore, line 174: Now that we have offsets in the right places, I feel like this can be simplified using recursion (similarly to what I proposed to do in the javadoc). Although here we need to special case element layouts for which a faster option is available. So, something like:

* if element layout is a value layout V, use segment.toArray(V)
* otherwise, retrieve the objects corresponding to the array elements (using the mapper recursively) and return an array of them (it is likely that this part is more imperative, and harder to express with method handles).

The key difference is that the second step above only concerns about going "one dimension deeper".

mcimadamore (commented):
Great progress. The new code structure in the mapper implementation is much clearer and the javadoc has improved a lot.

src/java.base/share/classes/jdk/internal/foreign/LayoutRecordMapper.java
```diff
@@ -477,24 +448,35 @@ static int dimensionOf(Class<?> arrayClass) {
...
+
+    @SuppressWarnings("unchecked")
+    static <R> R[] toArray(MemorySegment segment,
+                           GroupLayout elementLayout,
+                           long count,
+                           Class<R> type,
+                           MethodHandle mapper) {
+
+        R[] result = (R[]) Array.newInstance(type, Math.toIntExact(count));
+        Spliterator<MemorySegment> spliterator = segment.spliterator(elementLayout);
```
mcimadamore, line 470: Using MemorySegment::elements allows you to avoid interacting with the spliterator directly (which I don't think you need here?)

src/java.base/share/classes/java/lang/foreign/GroupLayout.java
```diff
@@ -73,4 +76,144 @@ public sealed interface GroupLayout extends MemoryLayout permits StructLayout, U
...
+     *
+     * @param <R> record type
+     * @param type the type (Class) of the record
+     * @throws IllegalArgumentException if the provided record {@code type} is the class {@link Record} or contains
+     *                                  components for which there are no exact mapping (of names and types) in
+     *                                  this group layout or if the provided {@code type} is not public or
+     *                                  if the method is otherwise unable to create a record mapper as specified above.
+     * @since 21
+     */
+    <R extends Record> Function<MemorySegment, R> recordMapper(Class<R> type);
```
JornVernee, line 217: I feel like the documentation is a bit _too_ formal, and will go over some people's heads. In general, I suggest using plain English as the main structure, and then weaving in the formal definitions instead.

I think by re-ordering some of the information, you can end up with simpler/smaller paragraphs. I think there are 3 main steps:
1. The `recordMapper` function itself finds, for each record component, a corresponding member layout with the same name in this group layout. There are restrictions on the layout type of each corresponding member layout based on the type of the record component (e.g. a java primitive maps to a ValueLayout, MemorySegment maps to an AddressLayout, etc.)
2. An extraction function is determined based on the component/layout type combination (and then have a list with the process for each type combination). In this part you can lean on the information defined in 1, and use language like "the corresponding member layout"/"the corresponding record component", rather than `c_a` and `e_b`.
3. The canonical constructor of the record type is called with the result of the extraction functions.

I think 1 & 2 are too mixed up right now in the big formal definition of the extraction functions.

src/java.base/share/classes/java/lang/foreign/GroupLayout.java
```diff
@@ -73,4 +76,144 @@ public sealed interface GroupLayout extends MemoryLayout permits StructLayout, U
      */
     @Override
     GroupLayout withByteAlignment(long byteAlignment);
+
+    /**
+     * {@return a {@link Function} that can project {@linkplain MemorySegment MemorySegments} onto new instances
```
JornVernee, line 81: Since a new record is returned, 'into' seems more correct. ('onto' sounds like there is an existing record instance).
```suggestion
     * {@return a {@link Function} that can project {@linkplain MemorySegment MemorySegments} into new instances
```

minborg: "project ... onto" makes sense I think. But maybe we can reword here.

src/java.base/share/classes/java/lang/foreign/GroupLayout.java
minborg: I am happy to simplify. What do you think @mcimadamore ?

src/java.base/share/classes/jdk/internal/foreign/LayoutRecordMapper.java
minborg: This was an attempt to optimize but I agree we should use a Stream instead.

test/jdk/java/foreign/TestRecordMapper.java
```diff
@@ -0,0 +1,1001 @@
...
+        String view = POINT_LAYOUT
+                .recordMapper(Point.class)
+                .apply(POINT_SEGMENT)
+                .toString();
+
+        assertEquals("Point[x=3, y=4]", view);
+    }
+
+    @Test
+    public void inspectMemory() {
```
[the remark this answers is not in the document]
minborg: As the original bug mentions it is desirable to be able to convert to hex, this test shows how.

GavinRay97:
I am so unbelievably hyped for this 🙌

PR #839, opened 2023-06-15
8308293: A linker should expose the layouts it supports

mcimadamore:
This is another stab at https://github.com/openjdk/jdk/pull/14037

I believe, after some offline discussion, that we have found a more satisfying solution to the problem of JAVA_CHAR being exposed. Jorn suggested that linkers should also provide mappings for JNI types such as `jint`, `jshort` and such (which are aliases for our layout constants anyway). I think that's a great way to bring `JAVA_CHAR` back into the fold.

For now, I decided not to specify support for JNI canonical layouts (but I could do so, if that's preferred). I think the highest priority is to provide some stable mappings for C builtin types (+ `size_t`) as that's what 99% of developers will be struggling with.

API-wise, we just expose a map. In the preovious PR there were questions as to whether the map should be split into two methods. In general I see the following options:

1. Just expose a map (that's the primitive, other things can derived from it)
2. Expose a map, plus a method to get a canonical layout from a type name (that's the `Charset` approach, which has both `availableCharsets` *and* `forName`)
3. Expose a method to get canonical layout from name, plus a method that returns the set of supported canonical layout names

My (not so strong) preference would be for either (1) or (2).

---------
### Progress

### Issues
 * [JDK-8308293](https://bugs.openjdk.org/browse/JDK-8308293): A linker should expose the layouts it supports (**Bug** - P3)
 * [JDK-8308304](https://bugs.openjdk.org/browse/JDK-8308304): A linker should expose the layouts it supports (**CSR**) (Withdrawn)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign.git pull/839/head:pull/839` \
`$ git checkout pull/839`

`$ git checkout pull/839` \
`$ git pull https://git.openjdk.org/panama-foreign.git pull/839/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 839`

`$ git pr show -t 839`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/839.diff">https://git.openjdk.org/panama-foreign/pull/839.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/panama-foreign/pull/839#issuecomment-1592791309)


src/java.base/share/classes/jdk/internal/foreign/abi/AbstractLinker.java
```diff
@@ -270,15 +278,33 @@ private static FunctionDescriptor stripNames(FunctionDescriptor function) {
                 .orElseGet(() -> FunctionDescriptor.ofVoid(stripNames(function.argumentLayouts())));
     }
```
minborg, line 280: Would it make sense to add a "hook" that exotic platforms can use to modify the initialization of the Map?

minborg:
Maybe we could add a test that ensures basic mapping is in place?

src/java.base/share/classes/jdk/internal/foreign/abi/AbstractLinker.java
minborg: Yes, I meant an internal hook only for "porters".

src/java.base/share/classes/jdk/internal/foreign/abi/AbstractLinker.java
```diff
@@ -270,15 +278,33 @@ private static FunctionDescriptor stripNames(FunctionDescriptor function) {
...
-            ValueLayout.JAVA_DOUBLE,
-            ValueLayout.ADDRESS
+    private static final Map<String, MemoryLayout> CANONICAL_LAYOUTS_MAP = Map.ofEntries(
+            // specified canonical layouts
+            Map.entry("bool", ValueLayout.JAVA_BOOLEAN),
+            Map.entry("char", ValueLayout.JAVA_BYTE),
+            Map.entry("short", ValueLayout.JAVA_SHORT),
+            Map.entry("int", ValueLayout.JAVA_INT),
+            Map.entry("float", ValueLayout.JAVA_FLOAT),
+            Map.entry("long", CABI.current() == CABI.WIN_64 ? ValueLayout.JAVA_INT : ValueLayout.JAVA_LONG),
```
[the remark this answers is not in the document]
mcimadamore: > ? Zero supports JNI (we use it to implement the fallback linker)

Sorry. I confused myself.

src/java.base/share/classes/jdk/internal/foreign/abi/AbstractLinker.java
```diff
@@ -134,6 +137,11 @@ public SystemLookup defaultLookup() {
         return SystemLookup.getInstance();
     }
 
+    @Override
+    public Map<String, MemoryLayout> canonicalLayouts() {
+        return CANONICAL_LAYOUTS_MAP;
```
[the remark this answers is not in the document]
mcimadamore: I'm not too sure on what this shared factories would look like. In the general case (e.g. fallback) it would have at least to specify the size of short/int/long/long long. Plus possibly, size_t. Basically everything besides float/double/char/bool. I'm not sure that would end up simpler than just having duplicated code with the layout creation?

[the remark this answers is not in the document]
mcimadamore: ok - I can try that

test/jdk/java/foreign/TestLinker.java
```diff
@@ -24,7 +24,6 @@
 /*
  * @test
  * @enablePreview
- * @requires jdk.foreign.linker != "UNSUPPORTED"
```
JornVernee, line 27: This shouldn't be removed. (Note that the fallback linker is `FALLBACK`, so this test was already enabled for it).

mcimadamore: Apologies - I was fighting with jtreg to get the fallback linker enabled (so I could test). This line got in between :-)

JornVernee (approved):

mcimadamore:
Following some offline discussion, I've also decided to add support for `wchar_t` which is a type with platform-dependent size that can be useful when working with wide strings (esp. given the new, more general string methods).

src/java.base/share/classes/jdk/internal/foreign/abi/fallback/LibFallback.java
```diff
@@ -233,6 +243,11 @@ private NativeConstants() {}
         static final MemorySegment FLOAT_TYPE = MemorySegment.ofAddress(ffi_type_float());
         static final MemorySegment DOUBLE_TYPE = MemorySegment.ofAddress(ffi_type_double());
         static final MemorySegment POINTER_TYPE = MemorySegment.ofAddress(ffi_type_pointer());
+        static final int SIZEOF_SHORT = ffi_sizeof_short();
```
mcimadamore, line 246: Decided to change this a bit - e.g. instead of returning FFI type (which we don't use), I return sizeof instead, which seems more directly usable. This leads to some code simplifications (see definition of SIZE_T in FFIType).

JornVernee (approved):

JornVernee:
I'm in favor of using a map as well. I don't think there are any issue with doing that?

src/java.base/share/classes/jdk/internal/foreign/abi/fallback/FallbackLinker.java
```diff
@@ -283,4 +293,41 @@ private static Object readValue(MemorySegment seg, MemoryLayout layout) {
         assert layout == null;
         return null;
     }
+
+    @Override
+    public Map<String, MemoryLayout> canonicalLayouts() {
+        return CANONICAL_LAYOUTS;
```
ExE-Boss, line 299: `CANONICAL_LAYOUTS` is set to a `HashMap`, which is not unmodifiable.
```suggestion
        return Collections.unmodifiableMap(CANONICAL_LAYOUTS);
```

ExE-Boss (changes requested):

PR #840, opened 2023-06-20
8310362: Improve composability of handle derived from layouts

JornVernee:
Me and Maurizio have been discussing how to better address certain use cases like accessing structs with variable length array members, or matrices that have a dynamic row and column size.

The solution we came up with is captured in this patch. It adds an additional base offset parameter to the handles produced by `MemoryLayout::varHandle`, `MemoryLayout::sliceHandle` and `MemoryLayout::byteOffsetHandle`. This allows these handles to be composed with other offset computations that are not necessarily derived from layouts.

This patch also adds a `scale` method to MemoryLayout, which can be used to scale and index by the size of a layout. Essentially: `offset + layout.byteSize() * index`. This is useful for expressing ad-hoc offset computations. A `scaleHandle` method is also added for convience, which returns a MethodHandle for `scale` bound to the receiver layout.

Both these changes allowed for several simplifications:
- `MethodHandles::memorySegmentViewVarHandle` can be removed, as it's now exactly the same as calling ML::varHandle on a ValueLayout with an empty layout path. (the former method also didn't take advantage of the access handle cache found on ValueLayouts).
- `ValueLayout::arrayHandle` is removed, as most use cases can now simply be expressed using ML::varHandle + ML::scaleHandle. e.g.:
```
MethodHandles.collectCoordinates(layout.varHandle(),
            1, MethodHandles.insertArguments(layout.scaleHandle(), 0, 0L));
```

---------
### Progress

### Issue
 * [JDK-8310362](https://bugs.openjdk.org/browse/JDK-8310362): Improve composability of handle derived from layouts (**Enhancement** - P2)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign.git pull/840/head:pull/840` \
`$ git checkout pull/840`

`$ git checkout pull/840` \
`$ git pull https://git.openjdk.org/panama-foreign.git pull/840/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 840`

`$ git pr show -t 840`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/840.diff">https://git.openjdk.org/panama-foreign/pull/840.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/panama-foreign/pull/840#issuecomment-1600722385)


src/java.base/share/classes/jdk/internal/foreign/Utils.java
```diff
@@ -96,7 +96,19 @@ static VarHandle put(ValueLayout layout, VarHandle handle) {
...
+                return HANDLE_MAP.get(layout);
+            }
         }
+        layout = layout.withoutName(); // name doesn't matter
+        // keep the addressee layout as it's used below
+
+        VarHandle handle = VarHandleCache.get(layout);
+        if (handle != null) {
+            return handle;
+        }
```
JornVernee, line 110: Added a cache lookup here so that we don't end up re-creating the var handle if one is already in the cache. Also, I noticed that previously the cache was sensitive to layout names, which it shouldn't I think, since the names are not of consequence to the produced var handle.

src/java.base/share/classes/jdk/internal/foreign/layout/AbstractLayout.java
```diff
@@ -53,8 +53,9 @@ public final L withName(String name) {
         return dup(byteAlignment(), Optional.of(name));
     }
 
+    @SuppressWarnings("unchecked")
     public final L withoutName() {
-        return dup(byteAlignment(), Optional.empty());
+        return name.isPresent() ? dup(byteAlignment(), Optional.empty()) : (L) this;
```
JornVernee, line 58: Made this idempotent to make the `withoutName()` call in `makeSegmentViewVarHandle` cheaper.

mcimadamore (commented):
Overall looks very good. I like how it unifies access to memory segments (now VH access is similar to e.g. MS::get), and how there's less types of var handles to keep track of. The fact that var handles obtained with the MethodHandles combinator were not 100% similar to the ones obtained from layouts has always been a bit of a smell, and this patch neatly unifies the two. Similarly, it was a bit sad that the implementation of MS::get was using "super powers" that were not available to regular developers (ValueLayout::accessHandle). All that is gone now, which is very good.

mcimadamore:
I might have missed something but I did not see tests for the new scale/scaleHandle methods.

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
```diff
@@ -130,33 +127,44 @@
  * int value = segment.get(ValueLayout.JAVA_INT.withOrder(BIG_ENDIAN), 0);
  * }
  *
- * For more complex access operations (e.g. structured memory access), clients can obtain a
- * {@linkplain MethodHandles#memorySegmentViewVarHandle(ValueLayout) var handle}
- * that accepts a segment and a {@code long} offset. More complex var handles
- * can be obtained by adapting a segment var handle view using the var handle combinator functions defined in the
- * {@link java.lang.invoke.MethodHandles} class:
+ * More complex access operations can be expressed using var handles. The {@link ValueLayout#varHandle()}
```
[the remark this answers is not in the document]
JornVernee: I think either one works. i.e. we're expressing a idea in a programming language. I'll switch to 'implemented', it feels a bit more colloquial.

src/java.base/share/classes/java/lang/foreign/MemorySegment.java
```diff
@@ -130,33 +127,44 @@
...
- * int value = (int) intHandle.get(segment, 3L); // get int element at offset 3 * 4 = 12
+ * MemoryLayout segmentLayout = MemoryLayout.sequenceLayout(4, ValueLayout.JAVA_INT); // array of 4 elements
+ * VarHandle intHandle = segmentLayout.varHandle(MemoryLayout.PathElement.sequenceElement());
+ * int value = (int) intHandle.get(segment, 0L, 3L); // get int element at offset 0 + 3 * 4 = 12
  * }
  *
+ * Both the var handle returned by {@link ValueLayout#varHandle()} and
+ * {@link MemoryLayout#varHandle(MemoryLayout.PathElement...)}, as well as the method handle returned by
+ * {@link MemoryLayout#byteOffsetHandle(MemoryLayout.PathElement...)} and {@link MemoryLayout#sliceHandle(MemoryLayout.PathElement...)}
+ * feature a <em>base offset</em> parameter. This parameter represents a base offset for the offset computation. This extra
```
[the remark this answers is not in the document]
JornVernee: Yeah, good point

src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
```diff
@@ -231,6 +233,45 @@
  * Any attempt to provide a layout path {@code P} that is not well-formed for an initial layout {@code C_0} will result
  * in an {@link IllegalArgumentException}.
  *
+ * <h2 id="access-mode-restrictions">Access mode restrictions</h2>
+ *
+ * The var handles returned by {@link #varHandle(PathElement...)} and {@link ValueLayout#varHandle()} feature certain
+ * <i>access mode restrictions</i>. A var handle is associated with an access size {@code S}, derived from the
+ * {@linkplain ValueLayout#byteSize() byte size} of this layout, and an alignment constraint {@code B}, derived from the
```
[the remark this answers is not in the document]
mcimadamore: "of the receiver layout" seems more correct, even if more verbose.

src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
```diff
@@ -351,40 +423,41 @@ default MethodHandle byteOffsetHandle(PathElement... elements) {
...
+     * <p>
+     * The accessed address must be <a href="MemorySegment.html#segment-alignment">aligned</a> according to the {@linkplain
+     * #byteAlignment() alignment constraint} of the root layout (this layout). Note that this can be more strict
+     * (but not less) than the alignment constraint of the selected value layout.
      * <p>
      * Additionally, the provided dynamic values must conform to bounds which are derived from the layout path, that is,
      * {@code 0 <= x_i < b_i}, where {@code 1 <= i <= n}, or {@link IndexOutOfBoundsException} is thrown.
      * <p>
-     * Multiple paths can be chained, with <a href=#deref-path-elements>dereference path elements</a>.
+     * Multiple path sections can be chained, with <a href=#deref-path-elements>dereference path elements</a>.
```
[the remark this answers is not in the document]
JornVernee: I've uploaded a new version which attempts to improve this by merging the two paragraphs about offset computation. I'm not sure which version is better, though... Given the length of the text, I feel like we need _some_ kind of concept that refer to sections of a path which don't contain dereference path elements.

src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
```diff
@@ -292,6 +333,35 @@ public sealed interface MemoryLayout permits SequenceLayout, GroupLayout, Paddin
...
     MemoryLayout withByteAlignment(long byteAlignment);
 
+    /**
+     * {@return {@code offset + (byteSize() * index)}}
+     *
+     * @param offset the base offset
+     * @param index the index to be scaled by the byte size of this layout
+     */
+    @ForceInline
+    default long scale(long offset, long index) {
```
mcimadamore, line 343: was thinking... do we want the output of this method to ever be negative? (since it's used as an offset to a segment). Also, should we check for overflows with multiplyExact?

src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
```diff
@@ -314,21 +384,23 @@ default long byteOffset(PathElement... elements) {
      * The returned method handle has the following characteristics:
      * <ul>
      *     <li>its return type is {@code long};</li>
-     *     <li>it has as zero or more parameters of type {@code long}, one for each <a href=#open-path-elements>open path element</a>
+     *     <li>it has at least one {@code long} leading parameter representing the base offset;</li>
```
mcimadamore, line 387: should be "it has one" e.g. not "at least one" right?

src/java.base/share/classes/java/lang/foreign/ValueLayout.java
```diff
@@ -95,7 +94,7 @@ public sealed interface ValueLayout extends MemoryLayout permits
     ValueLayout withByteAlignment(long byteAlignment);
 
     /**
-     * {@return a var handle which can be used to access values represented by this value layout, in a given memory segment.}
+     * {@return a var handle which can be used to access values describe by this value layout, in a given memory segment.}
```
mcimadamore, line 97:
```suggestion
     * {@return a var handle which can be used to access values described by this value layout, in a given memory segment.}
```

src/java.base/share/classes/java/lang/foreign/ValueLayout.java
```diff
@@ -105,6 +104,9 @@ public sealed interface ValueLayout extends MemoryLayout permits
      * The returned var handle checks that accesses are aligned according to this value layout's
      * {@linkplain MemoryLayout#byteAlignment() alignment constraint}.
      *
+     * @apiNote This method is similar, but more efficient, than calling {@code MemoryLayout#varHandle(PathElement...)}
+     * with zero arguments, as it avoids the creation of the var args array.
```
mcimadamore, line 108: perhaps we should say "with an empty path element array" otherwise one might confuse it with this method?

src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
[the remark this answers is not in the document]
mcimadamore: I agree that the text is still not optimal. I'd like to think about it some more, and maybe I can come up with some more concrete suggestions.

src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
JornVernee: I put at least one as a hint there might be more (dynamic) coordinates. But maybe it's not important. I'll simplify

src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
JornVernee: I initially consider negative inputs, I wasn't sure, but thought they might be useful in some cases, so I left them in.

multiplyExact/addExtact seems like a good idea to use.

mcimadamore: If the output of `scale` is always used as an offset parameter in some VH, or memory segment accessor, then I think there's a very strong gravitational pull for making it positive. I can't think of any methods in the MemorySegment API that would be happy with a negative offset, so it seems like a place where bugs might hide, or be reported in unexpected places.

JornVernee: That's a good point

JornVernee:
Addressed the review comments, added some tests for the updated `scale` as well.

src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
```diff
@@ -338,10 +338,15 @@ public sealed interface MemoryLayout permits SequenceLayout, GroupLayout, Paddin
      *
      * @param offset the base offset
      * @param index the index to be scaled by the byte size of this layout
+     * @throws IllegalArgumentException if {@code offset} is negative
```
mcimadamore, line 341: also if `index` is negative?

JornVernee: Ah, right :)

src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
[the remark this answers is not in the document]
mcimadamore: Looking at this some more, here are some things which, I believe, can be improved:

* The text is unnecessarily speaking about _addresses_. This forces us to define an abstract function to compute the base address of a segment (we need to, since heap segments have a virtualized base). In reality, all this is unnecessary - there's just segments and offsets, we don't really care about the address.
* Ignoring dereference handles, the behavior of the var handle is straightforward: we just derive an offset into the segment by calling `offsetHandle` (all the indices are passed and validated there). Then it's as if we called MS::get with the target (value) layout and provided the offset obtained as above. Perhaps, relying on MS::get is suspicious here, as we're on a more primitive level - but I think just saying "accessing a value whose type is the carrier of the target layout, at the offset computed as XYZ in the provided segment" is ok (and that is what happens). I don't think there is a need to provide formulas.
* I realize that we're hand-waving a bit for when the target layout is an AddressLayout - in that case the resulting segment has certain size characteristics, which depend on the target layout of the address - and we should say that.
* Ad for dereference path elements (assuming we want to keep them, which I'm less sure about), we could use the existing structure where first we explain what happens when there's no dereference path element. Then, we could explain what happens when there's dereference path elements by saying that we split the path into multiple "chunks" (this is similar to what you have done earlier), and that we compute the var handle for each chunk, and show how things are wired together (at this point we should have explained all the concepts we needed for this).

JornVernee:
Fixed the merge, should be all good now

mcimadamore (approved):
Looks good!

PR #843, opened 2023-06-26
8310659: The jar tool should support allowing access to restricted methods from executable jars

JornVernee:
Add `Enable-Native-Access` jar attribute. `Enable-Native-Access: true` is equal to passing `--enable-native-access=ALL-UNNAMED` on the command line when running an executable jar (using `-jar <jar file>`).

Question: right now I've implemented this to check if the attribute value is exactly `true`, otherwise it is ignored. Should we only allow `true` or `false`, and throw an error on other values?

Reviewer note: I've factored some of a test code for analyzing output to `TestEnableNativeAccessBase` for the enable native access tests.

---------
### Progress

### Issue
 * [JDK-8310659](https://bugs.openjdk.org/browse/JDK-8310659): The jar tool should support allowing access to restricted methods from executable jars (**Enhancement** - P3)

### Reviewers
 * [Maurizio Cimadamore](https://openjdk.org/census#mcimadamore) (@mcimadamore - Committer) ⚠️ Review applies to [387205af](https://git.openjdk.org/panama-foreign/pull/843/files/387205afde68df6a283bce6b4f595083948a1fd4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign.git pull/843/head:pull/843` \
`$ git checkout pull/843`

`$ git checkout pull/843` \
`$ git pull https://git.openjdk.org/panama-foreign.git pull/843/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 843`

`$ git pr show -t 843`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/843.diff">https://git.openjdk.org/panama-foreign/pull/843.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/panama-foreign/pull/843#issuecomment-1608114711)

### DIFF
--- a/test/jdk/java/foreign/enablenativeaccess/TestEnableNativeAccessBase.java
+++ b/test/jdk/java/foreign/enablenativeaccess/TestEnableNativeAccessBase.java
@@ -41,7 +41,6 @@ public class TestEnableNativeAccessBase {
     static final String PANAMA_JNI_CLS = "org.openjdk.foreigntest.PanamaMainJNI";
     static final String PANAMA_JNI = "panama_module/" + PANAMA_JNI_CLS;
     static final String UNNAMED = "org.openjdk.foreigntest.unnamed.PanamaMainUnnamedModule";
-    static final String REINVOKER = "TestEnableNativeAccessJarManifest$Reinvoker";
 
     /**
      * Represents the expected result of a test.
--- a/test/jdk/java/foreign/enablenativeaccess/TestEnableNativeAccessJarManifest.java
+++ b/test/jdk/java/foreign/enablenativeaccess/TestEnableNativeAccessJarManifest.java
@@ -53,6 +53,8 @@ import org.testng.annotations.DataProvider;
 
 public class TestEnableNativeAccessJarManifest extends TestEnableNativeAccessBase {
 
+    private static final String REINVOKER = "TestEnableNativeAccessJarManifest$Reinvoker";
+
     static record Attribute(String name, String value) {}
 
     @Test(dataProvider = "cases")


mcimadamore:
> Question: right now I've implemented this to check if the attribute value is exactly `true`, otherwise it is ignored. Should we only allow `true` or `false`, and throw an error on other values?

This document:
https://docs.oracle.com/en/java/javase/20/docs/specs/jar/jar.html

Seems to point out to other attributes that can be boolean. For each it says that "true" and "false" are the possible values (which makes me think other values aren't possible), and that the case doesn't matter.

Perhaps we should just follow what the code for parsing these standard attribute does.

src/java.base/share/classes/sun/launcher/LauncherHelper.java
```diff
@@ -596,6 +597,10 @@ static String getMainClassFromJar(String jarname) {
             if (opens != null) {
                 addExportsOrOpens(opens, true);
             }
+            String enablesNativeAccess = mainAttrs.getValue(ENABLE_NATIVE_ACCESS);
+            if ("true".equals(enablesNativeAccess)) {
```
mcimadamore, line 601: What I see done elsewhere is something like this:
```
isMultiRelease = Boolean.parseBoolean(
                            attr.getValue(Attributes.Name.MULTI_RELEASE));
```

test/jdk/java/foreign/enablenativeaccess/TestEnableNativeAccessJarManifest.java
```diff
@@ -0,0 +1,97 @@
...
+                    "-jar", jarfile.toString())
+                .outputTo(System.out)
+                .errorTo(System.out);
+        checkResult(expectedResult, outputAnalyzer);
+        return outputAnalyzer;
+    }
+
+    @Test
+    public void testSucceed() throws Exception {
+        run("panama_no_unnamed_module_native_access", UNNAMED, successWithWarning("ALL-UNNAMED"), new String[]{});
```
mcimadamore, line 92: Maybe using a record to capture attribute name/value would avoid the need to split the string?
Also, other tests use a data provider, should we do that here too?

src/java.base/share/classes/sun/launcher/LauncherHelper.java
JornVernee: Ah, can do. I thought that it would throw an exception if the input wasn't `true` or `false`, just like e.g. [`Integer::parseInt`](https://docs.oracle.com/en/java/javase/20/docs/api/java.base/java/lang/Integer.html#parseInt(java.lang.String)) TIL.

JornVernee:
> > Question: right now I've implemented this to check if the attribute value is exactly `true`, otherwise it is ignored. Should we only allow `true` or `false`, and throw an error on other values?
> 
> This document: https://docs.oracle.com/en/java/javase/20/docs/specs/jar/jar.html
> 
> Seems to point out to other attributes that can be boolean. For each it says that "true" and "false" are the possible values (which makes me think other values aren't possible), and that the case doesn't matter.
> 
> Perhaps we should just follow what the code for parsing these boolean standard attributes (`Java-Bean` and `sealed`) does.

I can't find the `Java-Bean` attribute in the source code. Maybe it's been removed, the `Sealed` attribute seems to be tolerant to values outside of true/false, e.g. https://github.com/openjdk/jdk/blob/05e9c41eddf8961d1384c88ccedf993d86822a6b/src/java.base/share/classes/java/net/URLClassLoader.java#L611 & https://github.com/openjdk/jdk/blob/05e9c41eddf8961d1384c88ccedf993d86822a6b/src/java.base/share/classes/jdk/internal/loader/BuiltinClassLoader.java#L998

JornVernee:
I've addressed the review comments:
- Use data provider and records in test.
- Switch to `Boolean.parseBoolean` for parsing.

I've also removed the useless `String cls` parameter, since it's value was always the same. And reformatted the jtreg test tags per the advice of netbeans.

JornVernee:
Ah, I missed some changes. Hold on.

src/java.base/share/classes/sun/launcher/LauncherHelper.java
```diff
@@ -596,6 +597,9 @@ static String getMainClassFromJar(String jarname) {
             if (opens != null) {
                 addExportsOrOpens(opens, true);
             }
+            if (Boolean.parseBoolean(mainAttrs.getValue(ENABLE_NATIVE_ACCESS))) {
```
mcimadamore, line 600: From a design perspective there are two options here:
* `false` means "ignore" (which is what is being implemented here)
* `false` means `no native access`.

The latter seems a bit odd - but I'd like to remind the behavior of what happens when `--enable-native-access foo` is specified, and access to restricted method occurs in a module other than `foo`. In that case we throw an exception (e.g. we validate strictly, because a command line option was found).

Jar could do a similar thing:
* if no manifest attribute is specified, do the current warning behavior
* if attribute is set to `true`, do not warn
* if attribute is set to `false`, access to restricted methods in this jar are simply not allowed and must fail with exception

mcimadamore: I had a chat with @AlanBateman. He agrees that seeing "Enable-Native-Access:false" strongly suggests of a deny-like behavior (e.g. access should fail). That said, if we just did that, then it would be possible to have an `--enable-native-access` command line option and a manifest attribute pulling in different directions (e.g. enable vs. disable).

Ideally, we'd like a value-less attribute (e.g. `EnableNativeAccess:`) so that there's no confusion re. true/false. But that is not allowed by the manifest attribute grammar.

The next best thing seems to be `EnableNativeAccess:ALL-UNNAMED`. E.g. we only allow the name of the module we're allowed to enable native access for (the unnamed module). If, in the future, jars will start supporting multiple modules, we can enhance the attribute to accept a comma separated list.

This is slightly less confusing. There's no "deny" state, and no possibilities for the attribute to conflict with the command line option.

JornVernee: > but I'd like to remind the behavior of what happens when --enable-native-access foo is specified
> if attribute is set to false, access to restricted methods in this jar are simply not allowed and must fail with exception

Right... I agree here.

I've gone ahead and implemented this in the latest commit. This is a bit tricky since we initialize ModuleBootstrap, which holds the 'has enable native access' flag before we process the manifest. I've implemented it as a mutable static that gets lazily set in LauncherHelper.

WRT the tests. In order to test this I'm launching an executable jar from the class path, which then calls into a module that tries to do the native access.

But, I'm now starting to wonder whether the `true`/`false` value is really good enough... Does it make sense that a class path jar referencing several modules might want to enable native access for them on a per-module basis? Is an executable jar referencing a module even a case we want to care about at all?

src/java.base/share/classes/jdk/internal/module/ModuleBootstrap.java
```diff
@@ -773,9 +773,14 @@ private static void addExtraExportsOrOpens(ModuleLayer bootLayer,
 
     private static final boolean HAS_ENABLE_NATIVE_ACCESS_FLAG;
     private static final Set<String> NATIVE_ACCESS_MODULES;
+    private static boolean ENABLE_NATIVE_ACCESS_SET_IN_MANIFEST = false;
```
JornVernee, line 776: We initialize this class before the manifest is read, so I made this flag mutable.

test/jdk/java/foreign/enablenativeaccess/TestEnableNativeAccess.java
```diff
@@ -30,7 +30,7 @@
  * @library /test/lib
  * @build TestEnableNativeAccess
  *        panama_module/*
- *        org.openjdk.foreigntest.PanamaMainUnnamedModule
+ *        org.openjdk.foreigntest.unnamed.PanamaMainUnnamedModule
```
JornVernee, line 33: Had to move these classes to a different package, since when we put the same package on the class and module path, the package in the module hides the one on the class path.

test/jdk/java/foreign/enablenativeaccess/panama_module/org/openjdk/foreigntest/libLinkerInvokerModule.cpp
```diff
@@ -52,8 +55,12 @@ extern "C" {
         context.desc = env->NewGlobalRef(desc);
         context.opts = env->NewGlobalRef(opts);
         run_in_new_thread_and_join(call, &context);
+        if (context.exception != nullptr) {
+            env->Throw(context.exception); // transfer exception to this thread
+        }
```
JornVernee, line 60: Since we invoke the target method on a different thread, the exception wasn't being propagated properly, and as a result, the exit code was just 0, even when we had an error occur.

I've added the exception transfer here.

test/jdk/java/foreign/enablenativeaccess/panama_module/org/openjdk/foreigntest/PanamaMain.java
```diff
@@ -1,34 +0,0 @@
-/*
- * Copyright (c) 2021, Oracle and/or its affiliates. All rights reserved.
```
JornVernee, line 2: I noticed this file was unused, so I've remove it. It is superseded by `PanamaMainDirect`, but I guess we got both files through a merge somehow.

src/java.base/share/classes/sun/launcher/LauncherHelper.java
JornVernee: > The next best thing seems to be EnableNativeAccess:ALL-UNNAMED. E.g. we only allow the name of the module we're allowed to enable native access for (the unnamed module). If, in the future, jars will start supporting multiple modules, we can enhance the attribute to accept a comma separated list.

> This is slightly less confusing. There's no "deny" state, and no possibilities for the attribute to conflict with the command line option.

Didn't see these comments. This sounds like a good plan. Will work on that next.

JornVernee: > do we want to allow module names for modules outside of the executable jar file? i.e. should it just be a supplement to the --enable-native-access command line option?

After thinking about this for a bit, I think we should allow multiple module names. Consider for instance an executable jar which depends on a library, inside another modular jar file. The library jar is put on the module path, the executable jar passed to `-jar`. I think there should be a way in this case in the jar manifest to enable native access for the library if it needs it. If we just allow `EnableNativeAccess: ALL-UNNAMED` than that doesn't seem possible?

mcimadamore: > > do we want to allow module names for modules outside of the executable jar file? i.e. should it just be a supplement to the --enable-native-access command line option?
> 
> After thinking about this for a bit, I think we should allow multiple module names. Consider for instance an executable jar which depends on a library, inside another modular jar file. The library jar is put on the module path, the executable jar passed to `-jar`. I think there should be a way in this case in the jar manifest to enable native access for the library if it needs it. If we just allow `EnableNativeAccess: ALL-UNNAMED` than that doesn't seem possible?

The intent in my previous comment was to only allow ALL-UNNAMED as possible value. I believe this is consistent with the Add-Opens attribute which also targets the unnamed module:

> A <module>/<package> pair in the value of an Add-Opens attribute has the same meaning as the command-line option --add-opens <module>/<package>=ALL-UNNAMED.

So, there is a strong precedent here for the attribute in the manifest of a jar **not** to affect other modules. If other modules are present, a command line option needs to be added too.

src/java.base/share/classes/jdk/internal/module/ModuleBootstrap.java
AlanBateman: I don't think ModuleBootstrap is the right place for this mutable state. The module system is initialised early in the startup, it's not meant to deal with state changes that happen after startup. 

If the launcher invokes Modules.addEnableNativeAccessToAllUnnamed() when it will update ALL_UNNAMED_MODULE to enable native access, does it need any more than that?

AlanBateman: ModuleBootstrap is for initialising the module system early in the startup. It's not meant to have mutable state that is changed post VM initialization by the launcher. So I'd prefer not introduce this coupling.

If the manifest has `Enable-Native-Access: ALL-UNNAMED` then I would just expect to see the launcher use Modules.addEnableNativeAccessToAllUnnamed(), nothing more. So I think I'd like to understand how setEnableNativeAccessSetInManifest came about. Is there an ordering issue where somehow ModuleBootstrap.hasEnableNativeAccessFlag is called before EnableNativeAccess.isNativeAccessEnabled on the target module?

src/java.base/share/classes/sun/launcher/LauncherHelper.java
AlanBateman: > So, there is a strong precedent here for the attribute in the manifest of a jar **not** to affect other modules. If other modules are present, a command line option needs to be added too.

Yes, I think limit it for now to the value "ALL-UNNAMED". Maybe in the future there will be OOTB support for having multiple modules in the same JAR file or executable modular JARs, at which point this might have to looked at again.

src/java.base/share/classes/jdk/internal/module/ModuleBootstrap.java
JornVernee: There are 2 parts of the puzzle:
1. we need to enable native access for a module. This is what `addEnableNativeAccessToAllUnnamed` does
2. we need to track whether an enable native access flag has been set 'externally', through either the command line or the jar manifest. This dictates whether we should produce an error or a warning.

Right now the latter is implemented by decoding a bunch of system properties that are initialized in arguments.cpp [1], [2]. This decoding than produces a set of module names that was specified on the command line. And this is used to initialize the `HAS_ENABLE_NATIVE_ACCESS_FLAG` variable [3].

I just hooked into the latter process by adding the `ENABLE_NATIVE_ACCESS_SET_IN_MANIFEST` flag. This flag is mutable since `ModuleBootstrap` is intialized before we read the jar manifest. i.e. the variable is mutable because it is lazily initialized.  This works out because we only ever call `Module::ensureNativeAccess` after reading the jar manifest.

`ModuleBoostrap::hasEnableNativeAccessFlag` is really `hasEnableNativeAccessFlagOnCommandLine`, and we need another `hasEnableNativeAccessFlagInJarManifest`, and then the call to `hasEnableNativeAccessFlag` in `Module::ensureNativeAccess` would need to be replaced by `hasEnableNativeAccessFlagOnTheCommandLine() || hasEnableNativeAccessFlagInJarManifest()`.

But, I think we would still have a chicken and egg problem, since we start using modules before reading the jar manifest, so before we really know whether we should be throwing an error or warning for native access. So, at least in theory, if someone called `ensureNativeAccess` before the jar manifest is read, then they could observe the incorrect behavior.

So, I'm not sure if we could avoid this issue, other than not letting the flag set in the manifest factor into the decision whether to throw an error or just issue a warning.

Another option which I tried earlier, is for `LauncherHelper` to set a system property, and then initialize a `static final` flag in a holder class based on that property the first time `ModuleBootsrap::hasEnableNativeAccessFlag` is called. But, that seemed like it was just hiding the mutable state in the system property. Though, maybe it's preferable? It would at least show that we only initialize the state once, and it is initialized somewhere before we call `hasEnableNativeAccessFlag`.

[1]: https://github.com/openjdk/panama-foreign/blob/e326dd172e988532102519abe0ba46a9a4e7fca7/src/hotspot/share/runtime/arguments.cpp#L2341-L2344
[2]: https://github.com/openjdk/panama-foreign/blob/e326dd172e988532102519abe0ba46a9a4e7fca7/src/java.base/share/classes/jdk/internal/module/ModuleBootstrap.java#L807
[3]: https://github.com/openjdk/panama-foreign/blob/e326dd172e988532102519abe0ba46a9a4e7fca7/src/java.base/share/classes/jdk/internal/module/ModuleBootstrap.java#L783

AlanBateman: The launcher code is is java.base.  If `java -jar app.jar` is used then there shouldn't be application code running, and trying to do native access, before the application main class is loaded. In other words, there shouldn't be code in an unnamed module trying to do native access before any application or library code has been loaded. It's similar to Add-Exports and Add-Opens. They are used to reflectively export or open packages before the application code executes.

The same applies to an agent that may be packaged with the application and configured with Launcher-Agent-Class. We probably need to move the handling of Add-Exports, Add-Opens, and Enable-Native-Access to before Launcher-Agent-Class.

mcimadamore: The problem here is that the current command line option for native access has a rather sophisticated behavior:
* if the option is not present, a warning is emitted
* if the option is present, names module M, and access occurs within M, nothing is emitted
* if the option is present, names module M, and access occurs outside M, an exception is thrown

This makes enable-native-access non-additive (unlike add-opens). That is, if `Enable-Native-Access:ALL-UNNAMED` is translated into `--enable-native-access=ALL-UNNAMED` well, that means that:

* access to restricted methods from unnamed module is granted
* access to restricted methods from a module M that is not unnamed is denied

Now, this patch tries to implement this approach, but I do have worries: effectively, once a manifest specifies the Enable-Native-Access attribute, can the command line still add other modules in the trusted bubble? If yes, that's ok. If no, then we have a problem.

mcimadamore: Perhaps, setting an implicit system property in the launcher with the enable-native-access policy (e.g. either warn or deny) would be cleaner. This property will be set by the launcher and read by module bootstrap (like many other properties).

mcimadamore: (That said, we still have a problem, in that, by the time we get to LauncherHelper, ModuleBootstrap has already ran. But perhaps we can delay initialization of `HAS_ENABLE_NATIVE_ACCESS_FLAG` on first use, and make the field stable).

JornVernee: > once a manifest specifies the Enable-Native-Access attribute, can the command line still add other modules in the trusted bubble? If yes, that's ok. If no, then we have a problem.

The answer is "yes", and I've added some extra tests that verify that as well.

I've pushed the new implementation that only allows `ALL-UNNAMED` as a value for `Enable-Native-Access`

JornVernee: After some offline discussion, we decided to go back to not throwing an error when Enable-Native-Access is set in the manifest, and another module attempts a native access. Based on the facts that, the warning behavior is temporary any way, and it is unlikely for an executable jar to depend on another (external) module.

mcimadamore:
After quick discussion with @AlanBateman and @pron, it seems like the most straightforward path is to just treat the manifest attribute as a "dynamic" addition of the unnamed module to the set of modules allowed to do restricted operations, and one that should not alter the deny vs. warn policy (which is still determined by the command line option being used). While this is mildly inconsistent, the fact that (a) executable jars are unlikely to work with named modules and (b) the warning treatment is transitional anyway, it is preferable to keep things simple.

test/jdk/java/foreign/enablenativeaccess/TestEnableNativeAccessBase.java
```diff
@@ -0,0 +1,125 @@
...
+    static final String PANAMA_MAIN_CLS = "org.openjdk.foreigntest.PanamaMainDirect";
+    static final String PANAMA_MAIN = "panama_module/" + PANAMA_MAIN_CLS;
+    static final String PANAMA_REFLECTION_CLS = "org.openjdk.foreigntest.PanamaMainReflection";
+    static final String PANAMA_REFLECTION = "panama_module/" + PANAMA_REFLECTION_CLS;
+    static final String PANAMA_INVOKE_CLS = "org.openjdk.foreigntest.PanamaMainInvoke";
+    static final String PANAMA_INVOKE = "panama_module/" + PANAMA_INVOKE_CLS;
+    static final String PANAMA_JNI_CLS = "org.openjdk.foreigntest.PanamaMainJNI";
+    static final String PANAMA_JNI = "panama_module/" + PANAMA_JNI_CLS;
+    static final String UNNAMED = "org.openjdk.foreigntest.unnamed.PanamaMainUnnamedModule";
+    static final String REINVOKER = "TestEnableNativeAccessJarManifest$Reinvoker";
```
mcimadamore, line 44: Why is this constant defined here?

mcimadamore (approved):
Looks good

PR #844, opened 2023-07-07
8311594: Avoid GlobalSession liveness check

JornVernee:
This patch overrides `checkValidStateRaw` in GlobalSession, to do nothing. The current `checkValidStateRaw` in MemorySessionImpl checks the owner thread, and liveness, which are both not needed for the global session.

Since the state field is mutable, I found that the JIT can not fold away the liveness check. This has a very marginal effect on performance, but since the fix is also very non-intrusive, I think it's okay to do.

I've added 2 new benchmarks that I've used to measure the difference. One measures the performance of MS::copy, and the other measures MS::get. Both use a special `ALL` segment which spans `0` to `Long.MAX_VALUE`, which means that the target address can be used as an offset when doing the accesses. This helps the bounds checks being eliminated, and makes the result of overriding `checkValidStateRaw` in GlobalSession easier to see.

Here are some of the interesting numbers:

Before:

```
Benchmark                                  Mode  Cnt  Score   Error  Units
MemorySegmentCopyALL.panama_ALL_to_ALL     avgt   30  7.788 ± 0.029  ns/op
MemorySegmentCopyALL.unsafe_array_to_addr  avgt   30  7.543 ± 0.018  ns/op

Benchmark                               Mode  Cnt  Score   Error  Units
MemorySegmentGet.segment_ALL_unaligned  avgt   30  0.486 ± 0.002  ns/op
MemorySegmentGet.unsafe                 avgt   30  0.398 ± 0.003  ns/op
```

After:

```
Benchmark                                  Mode  Cnt  Score   Error  Units
MemorySegmentCopyALL.panama_ALL_to_ALL     avgt   30  7.571 ± 0.017  ns/op
MemorySegmentCopyALL.unsafe_array_to_addr  avgt   30  7.550 ± 0.019  ns/op

Benchmark                               Mode  Cnt  Score   Error  Units
MemorySegmentGet.segment_ALL_unaligned  avgt   30  0.409 ± 0.002  ns/op
MemorySegmentGet.unsafe                 avgt   30  0.397 ± 0.003  ns/op
```

---------
### Progress

### Issue
 * [JDK-8311594](https://bugs.openjdk.org/browse/JDK-8311594): Avoid GlobalSession liveness check (**Enhancement** - P3)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign.git pull/844/head:pull/844` \
`$ git checkout pull/844`

`$ git checkout pull/844` \
`$ git pull https://git.openjdk.org/panama-foreign.git pull/844/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 844`

`$ git pr show -t 844`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/844.diff">https://git.openjdk.org/panama-foreign/pull/844.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/panama-foreign/pull/844#issuecomment-1626007188)

### DIFF
--- a/src/java.base/share/classes/jdk/internal/foreign/GlobalSession.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/GlobalSession.java
@@ -67,4 +67,10 @@ final class GlobalSession extends MemorySessionImpl {
     public void justClose() {
         throw nonCloseable();
     }
+
+    @Override
+    @ForceInline
+    public void checkValidStateRaw() {
+        // do nothing, avoid liveness check
+    }
 }
--- a/test/micro/org/openjdk/bench/java/lang/foreign/LoopOverPollutedSegments.java
+++ b/test/micro/org/openjdk/bench/java/lang/foreign/LoopOverPollutedSegments.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2020, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -46,7 +46,7 @@ import static java.lang.foreign.ValueLayout.*;
 @Measurement(iterations = 10, time = 500, timeUnit = TimeUnit.MILLISECONDS)
 @State(org.openjdk.jmh.annotations.Scope.Thread)
 @OutputTimeUnit(TimeUnit.MILLISECONDS)
-@Fork(value = 3, jvmArgsAppend = "--enable-preview")
+@Fork(value = 3, jvmArgsAppend = {"--enable-preview", "--enable-native-access=ALL-UNNAMED"})
 public class LoopOverPollutedSegments extends JavaLayouts {
 
     static final int ELEM_SIZE = 1_000_000;
@@ -57,7 +57,7 @@ public class LoopOverPollutedSegments extends JavaLayouts {
 
 
     Arena confinedArena, sharedArena;
-    MemorySegment nativeSegment, nativeSharedSegment, heapSegmentBytes, heapSegmentFloats;
+    MemorySegment nativeSegment, nativeSharedSegment, nativeGlobal, heapSegmentBytes, heapSegmentFloats;
     byte[] arr;
     long addr;
 
@@ -74,6 +74,7 @@ public class LoopOverPollutedSegments extends JavaLayouts {
         nativeSegment = scope1.allocate(ALLOC_SIZE, 4);
         Arena scope = sharedArena;
         nativeSharedSegment = scope.allocate(ALLOC_SIZE, 4);
+        nativeGlobal = nativeSegment.reinterpret(Arena.global(), null);
         heapSegmentBytes = MemorySegment.ofArray(new byte[ALLOC_SIZE]);
         heapSegmentFloats = MemorySegment.ofArray(new float[ELEM_SIZE]);
 
@@ -84,6 +85,8 @@ public class LoopOverPollutedSegments extends JavaLayouts {
                 nativeSegment.setAtIndex(JAVA_FLOAT_UNALIGNED, i, i);
                 nativeSharedSegment.setAtIndex(JAVA_INT_UNALIGNED, i, i);
                 nativeSharedSegment.setAtIndex(JAVA_FLOAT_UNALIGNED, i, i);
+                nativeGlobal.setAtIndex(JAVA_INT_UNALIGNED, i, i);
+                nativeGlobal.setAtIndex(JAVA_FLOAT_UNALIGNED, i, i);
                 VH_INT_UNALIGNED.set(nativeSegment, (long)i, i);
                 heapSegmentBytes.setAtIndex(JAVA_INT_UNALIGNED, i, i);
                 heapSegmentBytes.setAtIndex(JAVA_FLOAT_UNALIGNED, i, i);
--- a/test/micro/org/openjdk/bench/java/lang/foreign/MemorySegmentCopyALL.java
+++ b/test/micro/org/openjdk/bench/java/lang/foreign/MemorySegmentCopyALL.java
@@ -1,0 +1,112 @@
+/*
+ * Copyright (c) 2023, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package org.openjdk.bench.java.lang.foreign;
+
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.CompilerControl;
+import org.openjdk.jmh.annotations.Fork;
+import org.openjdk.jmh.annotations.Measurement;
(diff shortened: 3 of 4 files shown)


src/java.base/share/classes/jdk/internal/foreign/GlobalSession.java
```diff
@@ -67,4 +67,10 @@ void addInternal(ResourceList.ResourceCleanup resource) {
     public void justClose() {
         throw nonCloseable();
     }
+
+    @Override
+    @ForceInline
+    public void checkValidStateRaw() {
+        // do nothing, avoid liveness check
```
mcimadamore, line 74: This looks ok, although (see comments on MemorySessionImpl::checkValidStateRaw) sometimes having too many versions of this method can cause profile pollution issues (which is why we consolidated the impl at some point). I suggest trying other benchmarks, such as LoopOverPollutedXYZ

JornVernee: Good point. Something I also realized over the weekend. I had thought bimorphic inlining would save us (and we overload other methods such as acquire/release already). But, we have 3 concrete implementation classes of MemorySessionImpl: confined, shared, and global. This blows up inlining it seems, causes a big regression when all three session kinds are used (see latest push I did):

```
Before:

Benchmark                                              Mode  Cnt  Score    Error  Units
LoopOverPollutedSegments.heap_segment_floats_VH        avgt   30  0.311 ±  0.001  ms/op
LoopOverPollutedSegments.heap_segment_floats_instance  avgt   30  0.311 ±  0.001  ms/op
LoopOverPollutedSegments.heap_segment_ints_VH          avgt   30  0.311 ±  0.001  ms/op
LoopOverPollutedSegments.heap_segment_ints_instance    avgt   30  0.311 ±  0.001  ms/op
LoopOverPollutedSegments.heap_unsafe                   avgt   30  0.217 ±  0.001  ms/op
LoopOverPollutedSegments.native_segment_VH             avgt   30  0.237 ±  0.005  ms/op
LoopOverPollutedSegments.native_segment_instance       avgt   30  0.245 ±  0.005  ms/op
LoopOverPollutedSegments.native_unsafe                 avgt   30  0.221 ±  0.007  ms/op

After:

Benchmark                                              Mode  Cnt  Score    Error  Units
LoopOverPollutedSegments.heap_segment_floats_VH        avgt   30  3.310 ±  0.005  ms/op
LoopOverPollutedSegments.heap_segment_floats_instance  avgt   30  3.333 ±  0.004  ms/op
LoopOverPollutedSegments.heap_segment_ints_VH          avgt   30  3.305 ±  0.007  ms/op
LoopOverPollutedSegments.heap_segment_ints_instance    avgt   30  3.439 ±  0.106  ms/op
LoopOverPollutedSegments.heap_unsafe                   avgt   30  0.217 ±  0.001  ms/op
LoopOverPollutedSegments.native_segment_VH             avgt   30  3.595 ±  0.043  ms/op
LoopOverPollutedSegments.native_segment_instance       avgt   30  3.696 ±  0.006  ms/op
LoopOverPollutedSegments.native_unsafe                 avgt   30  0.204 ±  0.001  ms/op
```

JornVernee: I've also tried another solution which puts an `instanceof` check in the existing `checkValidStateRaw`. This does improve the case with the special `ALL` segment, but it also regresses the polluted case by about just as much it seems.

I think this is a dead end.

JornVernee:
I'd still like to upstream the benchmark changes. I'll create a separate PR for that.

src/java.base/share/classes/jdk/internal/foreign/GlobalSession.java
mcimadamore: IMHO, the only sane way to add support for this is to have a well-known constant - e.g. MemorySegment::allNative (which should be a restricted factory of course). Then special case this constant w.r.t. bound checks and temporal checks (since there's a well-known constant to compare against, I'm confident that pollution will not rear its ugly head). But that brings in a lot of API questions: for instance, if this segment is unbounded, what happens if you slice (of course one possible answer might be to throw). While we *might* add something like that in the future, I don't think we have enough evidence yet that something like this is really required.

JornVernee: I suppose that is not too much different from the instanceof check I tried. The issue I see is that, while when the segment/scope are constant, everything folds away, everybody else now has to do an additional identity check on every access.

mcimadamore: > So what's the next step?

I believe the next best step is to try and determine exactly where the 2-3% performance regression comes from. We have many benchmarks that show no extra cost compared with unsafe if access occurs in a counted loop (because most we made sure that all checks can be hoisted outside the loop). But we need to be able to isolate the access pattern that is causing issue: does your code have a lot of "lone" reads/writes scattered through the code not surrounded by any loop? Or, maybe the loop is there, but for some reason the JIT is failing to hoist some of the checks? Or maybe a failure to vectorize the loop accordingly? These are very different problems - with the latter two probably more fixable (either with some changes in your code, on in our VM optimization logic) than the first (a **single** memory access using MemorySegment, ByteBuffer or any other API which enforces some extra checks is always going to be slower than same access using Unsafe). But loop-access is, typically, where the big wins are performance-wise, with potential for check hoisting, vectorization, and more.

broneill: There's a ton of different access patterns, and most of them cannot be implemented as simple loops. This statement concerns me the most: "a single memory access using MemorySegment, ByteBuffer or any other API which enforces some extra checks is always going to be slower than same access using Unsafe"  If the MemorySegment is global and encompasses the entire address space, why should there be any extra checks at all compared to Unsafe?

mcimadamore: > There's a ton of different access patterns, and most of them cannot be implemented as simple loops. This statement concerns me the most: "a single memory access using MemorySegment, ByteBuffer or any other API which enforces some extra checks is always going to be slower than same access using Unsafe" If the MemorySegment is global and encompasses the entire address space, why should there be any extra checks at all compared to Unsafe?

The main point I'm trying to make here is that, so far at least, we're speculating as to what the cause of the regression might be. What we need to do is (a) be able to reproduce the issue you are experiencing (there is no evidence so far that the fix in this PR does anything to ameliorate your situation and I wouldn't be too surprised if it did not help at all) and (b) find some ways to mitigate it (where and if possible). For (a) we need help, as we are essentially discussing code without being able to see it, which is not optimal.

mcimadamore: Stuff like this can be very helpful, thanks!

broneill: Panama vs. Unsafe: https://gist.github.com/broneill/3a39051635e3cb758d0cca5a963c685e

JornVernee: Thanks for the benchmark. We've been able to reproduce your result. We are investigating further (that will take some time).

mcimadamore: Filed this:
https://bugs.openjdk.org/browse/JDK-8311932

merykitty:
For liveness check, a solution on the top of my head is to push the branch to the slow path. Instead of having one `state` field, we can have 2, which can take values according to the state of the session, (0, x): Open, (1, 0): Closing, (1, 1): Closed. These can be marked as stable, since they do not transit from value 1 to value 0. Then, we can special case the global session to have the `openState` field being 1, and we do type branching when encountering a 1 when reading this state. This have the advantage that it does not affect the execution of other sessions, as well as being able to fold the check for the global ones.

For ownership check, maybe we can tell the compiler to trust this particular field. Stable does not work here, though maybe we can add a dummy `Thread` object to use as the pivot value instead of `null`.

For range check, maybe we can do the same as liveness one, instead of failing on out of bounds, we branch on a constant field to see if it is a global one, and only throw if it is not.

Thanks a lot.

mcimadamore:
> For liveness check, a solution on the top of my head is to push the branch to the slow path. Instead of having one `state` field, we can have 2, which can take values according to the state of the session, (0, x): Open, (1, 0): Closing, (1, 1): Closed. These can be marked as stable, since they do not transit from value 1 to value 0. Then, we can special case the global session to have the `openState` field being 1, and we do type branching when encountering a 1 when reading this state. This have the advantage that it does not affect the execution of other sessions, as well as being able to fold the check for the global ones.
> 
> For ownership check, maybe we can tell the compiler to trust this particular field. Stable does not work here, though maybe we can add a dummy `Thread` object to use as the pivot value instead of `null`. But comparing to another pointer is a bit tad more expensive than comparing to `null`, so changing the compiler seems more preferable.
> 
> For range check, maybe we can do the same as liveness one, instead of failing on out of bounds, we branch on a constant field to see if it is a global one, and only throw if it is not.
> 
> Thanks a lot.

What you suggest are sensible optimization tactics. However, it is unclear how much this path needs to be optimized further. In most cases, when we see a significant performance difference between MS and Unsafe, the liveness check is never the culprit, the bound check is.

And, when using tactics to e.g. generate wrapper segments on the fly to avoid bound checks [1], we can observe that liveness check is also gone from the corresponding assembly. So, this optimization seems to be in YAGNI territory, at least for now.

[1] - https://mail.openjdk.org/pipermail/panama-dev/2023-July/019487.html

JornVernee:
No

PR #848, opened 2023-07-20
8312473: Return value corrupted when using CCS + isTrivial

JornVernee:
Due to a bug in the downcall linker stub generation, we don't save the return value when capturing call state for trivial functions, and the return value gets corrupted.

We try not to save the return register around calls on the return path of a downcall stub, if it is not needed. Currently we don't save the return register when we're using a return buffer, since we write the return value to the return buffer before the calls on the return path, which means it is safe for those calls to overwrite the return register. But, the current logic also says we don't need to save the return register if the function is trivial (`_needs_transition == false`). The logic behind this was initially that, since we don't have _any_ calls on the return path, we don't need to save the return register. But, after adding support for capturing call state, we now also have a call on the return path for trivial functions that capture call state, and around that call, we might need to save the return register.

The fix is to simply save the return register when capturing call state, regardless of whether the function is trivial or not. In the case of just a trivial function that doesn't capture call state, we still don't save the return register around the return path calls for the thread state transition (which is not needed), since we don't generate those thread state transitions in the first first place.

I've updated the capture call state test to also test all the cases in trivial mode.

---------
### Progress

### Issue
 * [JDK-8312473](https://bugs.openjdk.org/browse/JDK-8312473): Return value corrupted when using CCS + isTrivial (**Bug** - P2)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign.git pull/848/head:pull/848` \
`$ git checkout pull/848`

`$ git checkout pull/848` \
`$ git pull https://git.openjdk.org/panama-foreign.git pull/848/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 848`

`$ git pr show -t 848`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/848.diff">https://git.openjdk.org/panama-foreign/pull/848.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/panama-foreign/pull/848#issuecomment-1644523931)

### DIFF
--- a/src/hotspot/cpu/aarch64/downcallLinker_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/downcallLinker_aarch64.cpp
@@ -168,7 +168,7 @@ void DowncallStubGenerator::generate() {
   assert(_abi._shadow_space_bytes == 0, "not expecting shadow space on AArch64");
   allocated_frame_size += arg_shuffle.out_arg_bytes();
 
-  bool should_save_return_value = !_needs_return_buffer && _needs_transition;
+  bool should_save_return_value = !_needs_return_buffer;
   RegSpiller out_reg_spiller(_output_registers);
   int spill_offset = -1;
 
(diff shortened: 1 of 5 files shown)


mcimadamore (approved):
Well spotted!

PR #855, opened 2023-08-01
Initial push

mcimadamore:


mcimadamore:
This chart shows the improvements associated with this patch.

allocateFrom

mcimadamore:
Note that MallocArena is still significantly faster than UnsafeArena. We are currently investigating as to why that would be the case, but that is not something that has to do with FFM API code (after all, both arenas use FFM API). For some reason, a downcall handle to `malloc` seems much faster than a call to `Unsafe::allocateMemory`. We have already ruled out NativeMemoryTracking (commenting out all the code doesn't show any benefits) as well as thread state transitions (removing them for Unsafe only gives a marginal gain). Perfasm seems to point at the JNI stub (for the native Unsafe::allocateMemory0) as the place where ~50% of execution time is being spent, which doesn't make a lot of sense, given that FFM API also has to generate stubs and given that the shape of such stubs is not too different.

merykitty:
The benchmark methods seem to include arena creation, memory allocation and memory deallocation, is it intended? I think this benchmark should only take into consideration memory allocation, no?

> One caveat: a custom arena does NOT inherit this special behavior. That is, it is the responsibility of the custom arena to define "shortcut" for the `allocateFrom` methods.

This means that the custom arenas have to use another way (e.g `malloc`) to allocate memory without initialization themselves, right? The following sentences give me the impression that there are some kinds of protected methods that can be delegated to.

Thanks a lot.

mcimadamore:
> The benchmark methods seem to include arena creation, memory allocation and memory deallocation, is it intended? I think this benchmark should only take into consideration memory allocation, no?

I mean it was intended, as I wanted to measure that too - but you are right that doing so will add noise. 

> 
> > One caveat: a custom arena does NOT inherit this special behavior. That is, it is the responsibility of the custom arena to define "shortcut" for the `allocateFrom` methods.
> 
> This means that the custom arenas have to use another way (e.g `malloc`) to allocate memory without initialization themselves, right? The following sentences give me the impression that there are some kinds of protected methods that can be delegated to.

Say you want to wrap a confined arena in another custom arena. A custom arena is only required to implement a single allocation method (`allocate(long, long)`). But if you do that, then _all_ the default methods the custom arena inherits from `SegmentAllocator` will delegate back to that. Which means no distinction between `allocate` and `allocateFrom`.

So, if a custom arena wants to delegate to the underlying arena, and keep the distinction between zeroing and non-zeroing allocation (e.g. because it doesn't want to use malloc, or something else), it _must_ override all the `allocateFrom` methods, and delegate them to the underlying arena.

Of course if a custom arena implements its single allocation via malloc `malloc`, or if the custom arena recycles memory, then there's no need to override any other methods, as the primary allocation method (the one all the other delegate to) is already avoiding zeroing.

> 
> Thanks a lot.

src/java.base/share/classes/jdk/internal/foreign/NativeMemorySegmentImpl.java
```diff
@@ -147,6 +150,14 @@ public void cleanup() {
         return segment;
     }
 
+    private static long allocateMemoryWrapper(long size) {
```
mcimadamore, line 153: This is required as we no longer call Bits::reserveMemory, so it is possible when passing `Long.MAX_VALUE` as a size parameter, to get a IAE from Unsafe. Note that I think this was probably a latent issue - while the behavior changed for `Long.MAX_VALUE` with this patch, I believe using `Long.MAX_VALUE -1` would have created similar issues with the original code. Point is: `Unsafe` might fail because the size is too big - and we should handle that.

mrserb:
>Note that MallocArena is still significantly faster than UnsafeArena

This is interesting, is it possible that memset via panama could be faster than "UNSAFE.setMemory"?

src/java.base/share/classes/jdk/internal/foreign/abi/SharedUtils.java
```diff
@@ -90,7 +90,7 @@ public Scope scope() {
         }
 
         @Override
-        public MemorySegment allocate(long byteSize) {
+        public MemorySegment allocate(long byteSize, long byteAlignment) {
             throw new UnsupportedOperationException();
         }
```
JornVernee, line 95: Nice, so making the allocate method abstract caught this bug.

test/jdk/java/foreign/TestSegmentAllocators.java
```diff
@@ -157,7 +157,7 @@ public void testBadAllocationArrayOverflow(SegmentAllocator allocator) {
     @Test(expectedExceptions = OutOfMemoryError.class)
     public void testBadArenaNullReturn() {
         try (Arena arena = Arena.ofConfined()) {
-            arena.allocate(Long.MAX_VALUE, 2);
+             arena.allocate(Long.MAX_VALUE, 2);
```
JornVernee, line 160: This change seems spurious? (I count 5 space indentation now).

test/micro/org/openjdk/bench/java/lang/foreign/AllocFromTest.java
```diff
@@ -0,0 +1,178 @@
...
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
```
JornVernee, line 24: Wrong copyright header (year and CE)

test/micro/org/openjdk/bench/java/lang/foreign/AllocFromTest.java
```diff
@@ -0,0 +1,178 @@
...
+    SlicingPool pool = new SlicingPool();
+
+    @Param({"5", "20", "100", "500", "1000"})
+    public int size;
+    public byte[] arr;
+
+    @Setup
+    public void setup() {
+        arr = new byte[size];
+        Random random = new Random();
```
JornVernee, line 67: Please use an explicit seed here (e.g. `0`), so that the data is always the same between benchmark runs.

test/micro/org/openjdk/bench/java/lang/foreign/AllocFromTest.java
```diff
@@ -0,0 +1,178 @@
...
+        arena.close();
+        return segment;
+    }
+
+    @Benchmark
+    public long alloc_malloc_arena() {
+        MallocArena arena = new MallocArena();
+        MemorySegment segment = arena.allocateFrom(ValueLayout.JAVA_BYTE, arr);
+        arena.close();
+        return segment.address();
```
JornVernee, line 84: Why not return the `segment` here as well? (like in `alloc`). Seems like an unfair comparison?

src/java.base/share/classes/jdk/internal/foreign/ArenaImpl.java
```diff
@@ -0,0 +1,80 @@
...
+
+package jdk.internal.foreign;
+
+import java.lang.foreign.Arena;
+import java.lang.foreign.MemoryLayout;
+import java.lang.foreign.MemorySegment;
+import java.lang.foreign.MemorySegment.Scope;
+import java.util.Objects;
+
+public class ArenaImpl implements Arena {
```
ExE-Boss, line 34: This class should probably be `final`:
```suggestion
public final class ArenaImpl implements Arena {
```

src/java.base/share/classes/jdk/internal/foreign/NativeMemorySegmentImpl.java
```diff
@@ -147,6 +150,14 @@ public void cleanup() {
...
+        try {
+            return UNSAFE.allocateMemory(size);
+        } catch (IllegalArgumentException ex) {
+            throw new OutOfMemoryError();
```
ExE-Boss, line 157: Maybe include the cause?
```suggestion
            throw new OutOfMemoryError(ex);
```

test/micro/org/openjdk/bench/java/lang/foreign/AllocFromTest.java
overheadhunter: Shouldn't we also explicitly choose the RNG algorithm via `RandomGenerator.of("...").create(0L)` to track regressions, should the default algorithm ever change?

Or are tests only required to be reproducible within the same JDK?

mrserb:
Why not use malloc/memset/free by default instead of Unsafe?

mcimadamore:
> Why not use malloc/memset/free by default instead of Unsafe?

We could do that. The main question is what developers think re. NativeMemoryTracking, which is enabled in Unsafe - see:

https://mail.openjdk.org/pipermail/panama-dev/2023-August/019623.html

src/java.base/share/classes/jdk/internal/foreign/NativeMemorySegmentImpl.java
mcimadamore: OutOfMemoryError doesn't have a constructor that takes an exception. I'll leave it as is for the time being.

test/micro/org/openjdk/bench/java/lang/foreign/AllocFromTest.java
JornVernee: > Or are tests only required to be reproducible within the same JDK?

Yes, that is the main requirement. Reproducible between forks and different runs of these benchmarks. We don't do build to build comparisons of these benchmarks at the moment.

JornVernee (approved):

mcimadamore:
Updated benchmark

image

mcimadamore:
I've updated the patch. I've realized that all the indirection when calling the SegmentAllocator::allocate methods were costing us big time. For instance, calling `allocate(JAVA_INT, array)` calls `allocate(JAVA_INT, array.length)` which calls `allocate(MemoryLayout.sequenceLayout(JAVA_INT, array.length))` which finally calls `allocate(long, long)`. This indirection (esp. the layout allocation, ends up being expensive on the faster allocators such as PoolArena (which is now 2x faster!).

Now all the SegmentAllocator methods jump straight into `SegmentAllocator::allocate(long, long)` (I needed to change the various implSpec clauses for this). This simplifies the patch a bit, as `ArenaImpl` only needs two different allocation methods.

I have also rearranged `NativeMemorySegmentImpl.makeNativeSegment` so that there's no `shouldInit` flag. Instead, the regular allocation method in `ArenaImpl` takes care of zeroing memory (by calling MS::fill on the allocated segment). This seems to be enough to take care of some bimodality across benchmark runs.

I've also marked `VM::pageAlignDirectMemory` as stable, as that field is set to the value of some JDK property and never touched again (I'm not too sure whether this value, which was meant for byte buffer, should also be honored by the memory segment API, but for now we can leave it as is).

JornVernee (approved):

PR #856, opened 2023-08-03
8313680: Disallow combining caputreCallState with isTrivial

JornVernee:
Native functions that modify call state (errno) are by definition not 'trivial'. Capturing call state would also prevent intrinsification of a trivial call. I think to be on the safe side, we just want to disallow combining these two for now.

---------
### Progress

### Issue
 * [JDK-8313680](https://bugs.openjdk.org/browse/JDK-8313680): Disallow combining caputreCallState with isTrivial (**Bug** - P3)

### Reviewers
 * [Maurizio Cimadamore](https://openjdk.org/census#mcimadamore) (@mcimadamore - Committer) ⚠️ Review applies to [0f243ad0](https://git.openjdk.org/panama-foreign/pull/856/files/0f243ad03b12af818839206708fb19e5bd520102)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign.git pull/856/head:pull/856` \
`$ git checkout pull/856`

`$ git checkout pull/856` \
`$ git pull https://git.openjdk.org/panama-foreign.git pull/856/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 856`

`$ git pr show -t 856`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/856.diff">https://git.openjdk.org/panama-foreign/pull/856.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/panama-foreign/pull/856#issuecomment-1665435208)

### DIFF
--- a/src/java.base/share/classes/java/lang/foreign/Linker.java
+++ b/src/java.base/share/classes/java/lang/foreign/Linker.java
@@ -729,6 +729,8 @@ public sealed interface Linker permits AbstractLinker {
          * }
          * }
          *
+         * @apiNote This linker option can not be combined with {@link #isTrivial}.
+         *
          * @param capturedState the names of the values to save.
          * @throws IllegalArgumentException if at least one of the provided {@code capturedState} names
          *                                  is unsupported on the current platform.
--- a/src/java.base/share/classes/jdk/internal/foreign/abi/LinkerOptions.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/abi/LinkerOptions.java
@@ -63,7 +63,11 @@ public class LinkerOptions {
             optionMap.put(option.getClass(), opImpl);
         }
 
-        return new LinkerOptions(optionMap);
+        LinkerOptions linkerOptions = new LinkerOptions(optionMap);
+        if (linkerOptions.hasCapturedCallState() && linkerOptions.isTrivial()) {
+            throw new IllegalArgumentException("captureCallState and isTrivial are mutually exclusive");
+        }
+        return linkerOptions;
     }
 
     public static LinkerOptions empty() {
--- a/src/java.base/share/classes/jdk/internal/foreign/abi/LinkerOptions.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/abi/LinkerOptions.java
@@ -65,7 +65,7 @@ public class LinkerOptions {
 
         LinkerOptions linkerOptions = new LinkerOptions(optionMap);
         if (linkerOptions.hasCapturedCallState() && linkerOptions.isTrivial()) {
-            throw new IllegalArgumentException("captureCallState and isTrivial are mutually exclusive");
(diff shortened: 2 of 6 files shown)


src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -729,6 +729,8 @@ static Option firstVariadicArg(int index) {
          * }
          * }
          *
+         * @apiNote This linker option can not be combined with {@link #isTrivial}.
+         *
```
JornVernee, line 733: I'm not sure if we should a similar API note on `isTrivial`. Just having one here seems enough?

test/jdk/java/foreign/capturecallstate/TestCaptureCallState.java
```diff
@@ -61,16 +61,12 @@ public class TestCaptureCallState extends NativeTestHelper {
         }
     }
 
-    private record SaveValuesCase(String nativeTarget, FunctionDescriptor nativeDesc, boolean trivial, String threadLocalName, Consumer<Object> resultCheck) {}
```
JornVernee, line 64: For the changes in this file, I've just reverting things to how the were before: https://github.com/openjdk/panama-foreign/pull/848

src/java.base/share/classes/java/lang/foreign/Linker.java
mcimadamore: I think it's enough to have it here

src/java.base/share/classes/jdk/internal/foreign/abi/LinkerOptions.java
```diff
@@ -63,7 +63,11 @@ private static LinkerOptions forShared(BiConsumer<LinkerOptionImpl, FunctionDesc
             optionMap.put(option.getClass(), opImpl);
         }
 
-        return new LinkerOptions(optionMap);
+        LinkerOptions linkerOptions = new LinkerOptions(optionMap);
+        if (linkerOptions.hasCapturedCallState() && linkerOptions.isTrivial()) {
+            throw new IllegalArgumentException("captureCallState and isTrivial are mutually exclusive");
```
mcimadamore, line 68: The message is good. I wonder if we could get it to scale better with something like "incompatible linker options: <option list>" which is sometimes what we use e.g. in the Java compiler (e.g. for flags). I wonder e.g. if there might be other set of incompatible options we might need to worry about in the future - seems a possibility. (In any case, we can always change the message later, so this is not a big priority).

mcimadamore (approved):

src/java.base/share/classes/jdk/internal/foreign/abi/LinkerOptions.java
JornVernee: I like that suggestion.

PR #857, opened 2023-08-03
8313238: Improve the performance of strlen for null-terminated strings

minborg:
This PR proposes to use native calls for determining the string length for C-type strings. 

Smaller segments (<1024 bytes) is using a trivial call whereas all other segments are using a normal call. We might consider always using regular calls if we think most segments are unbound anyhow.

The PR also contains a number of new tests, one of which requires a large heap size to run. It is likely that this latter test will not run in most test environments. In such cases, it is silently ignored.

---------
### Progress

### Issue
 * [JDK-8313238](https://bugs.openjdk.org/browse/JDK-8313238): Improve the performance of strlen for null-terminated strings (**Enhancement** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign.git pull/857/head:pull/857` \
`$ git checkout pull/857`

`$ git checkout pull/857` \
`$ git pull https://git.openjdk.org/panama-foreign.git pull/857/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 857`

`$ git pr show -t 857`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/857.diff">https://git.openjdk.org/panama-foreign/pull/857.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/panama-foreign/pull/857#issuecomment-1664036186)

### DIFF
--- a/src/java.base/share/classes/jdk/internal/foreign/StringSupport.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/StringSupport.java
@@ -34,24 +34,29 @@ import java.lang.invoke.MethodHandle;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 
-import static java.lang.foreign.ValueLayout.*;
+import static java.lang.foreign.ValueLayout.ADDRESS;
+import static java.lang.foreign.ValueLayout.JAVA_BYTE;
+import static java.lang.foreign.ValueLayout.JAVA_INT;
+import static java.lang.foreign.ValueLayout.JAVA_LONG;
+import static java.lang.foreign.ValueLayout.JAVA_SHORT;
 
 /**
  * Miscellaneous functions to read and write strings, in various charsets.
  */
 public class StringSupport {
 
+    // Maximum segment byte size for which a trivial method will be invoked.
+    private static final long MAX_TRIVIAL_SIZE = 1024L;
+    private static final MethodHandle STRLEN_TRIVIAL;
     private static final MethodHandle STRLEN;
-    private static final MethodHandle STRLEN_CRITICAL;
-    private static final long MAX_CRITICAL_SIZE = 1024;
 
     static {
         Linker linker = Linker.nativeLinker();
         var strlen = linker.defaultLookup().find("strlen").orElseThrow();
         var description = FunctionDescriptor.of(JAVA_LONG, ADDRESS);
 
+        STRLEN_TRIVIAL = linker.downcallHandle(strlen, description, Linker.Option.isTrivial());
         STRLEN = linker.downcallHandle(strlen, description);
-        STRLEN_CRITICAL = linker.downcallHandle(strlen, description, Linker.Option.isTrivial());
     }
 
     public static String read(MemorySegment segment, long offset, Charset charset) {
@@ -115,8 +120,8 @@ public class StringSupport {
             if (start > 0) {
                 segment = segment.asSlice(start);
             }
-            long len = segment.byteSize() < MAX_CRITICAL_SIZE
-                    ? (long)STRLEN_CRITICAL.invokeExact(segment)
+            long len = segment.byteSize() < MAX_TRIVIAL_SIZE
+                    ? (long)STRLEN_TRIVIAL.invokeExact(segment)
                     : (long)STRLEN.invokeExact(segment);
 
             // On platforms where `size_t` maps to an `int`, we must check if `len < 0`
--- a/src/java.base/share/classes/jdk/internal/foreign/StringSupport.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/StringSupport.java
@@ -33,11 +33,11 @@ import java.lang.foreign.MemorySegment;
 import java.lang.invoke.MethodHandle;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
+import java.util.Objects;
 
 import static java.lang.foreign.ValueLayout.ADDRESS;
 import static java.lang.foreign.ValueLayout.JAVA_BYTE;
 import static java.lang.foreign.ValueLayout.JAVA_INT;
-import static java.lang.foreign.ValueLayout.JAVA_LONG;
 import static java.lang.foreign.ValueLayout.JAVA_SHORT;
 
 /**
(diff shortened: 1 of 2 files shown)


minborg:
Here are some benchmarks for various string lengths:

Before:

```
Benchmark                         (size)  Mode  Cnt    Score   Error  Units
InternalStrLen.currentSingleByte       4  avgt   30    2.989 ? 0.013  ns/op
InternalStrLen.currentSingleByte     251  avgt   30  124.791 ? 0.210  ns/op
```

After:

```
Benchmark                         (size)  Mode  Cnt    Score   Error  Units
InternalStrLen.nativeSingleByte        4  avgt   30    3.345 ? 0.020  ns/op
InternalStrLen.nativeSingleByte      251  avgt   30    8.753 ? 0.022  ns/op
```

src/java.base/share/classes/jdk/internal/foreign/StringSupport.java
```diff
@@ -93,15 +115,23 @@ private static void writeFast_int(MemorySegment segment, long offset, Charset ch
...
-        for (int offset = 0; offset >= 0; offset++) {
-            byte curr = segment.get(JAVA_BYTE, start + offset);
-            if (curr == 0) {
-                return offset;
+    private static int native_strlen_byte(MemorySegment segment, long start) {
+        try {
+            if (start > 0) {
+                segment = segment.asSlice(start);
             }
+            long len = segment.byteSize() < MAX_TRIVIAL_SIZE
```
rsmogura, line 123: When calling `strlen` should it be considered if segments is heap or off-heap?

minborg:
Here are some more variants of sizes:

Before:
```
Benchmark                         (size)  Mode  Cnt    Score   Error  Units
InternalStrLen.currentSingleByte       1  avgt   30    1.599 ? 0.031  ns/op
InternalStrLen.currentSingleByte       4  avgt   30    2.969 ? 0.009  ns/op
InternalStrLen.currentSingleByte      16  avgt   30    8.586 ? 0.013  ns/op
InternalStrLen.currentSingleByte     251  avgt   30  124.995 ? 0.832  ns/op
InternalStrLen.currentSingleByte    1024  avgt   30  489.230 ? 2.585  ns/op
```

After:
```
Benchmark                         (size)  Mode  Cnt    Score   Error  Units
InternalStrLen.nativeSingleByte        1  avgt   30    3.313 ? 0.008  ns/op
InternalStrLen.nativeSingleByte        4  avgt   30    3.310 ? 0.006  ns/op
InternalStrLen.nativeSingleByte       16  avgt   30    3.479 ? 0.017  ns/op
InternalStrLen.nativeSingleByte      251  avgt   30    8.817 ? 0.102  ns/op
InternalStrLen.nativeSingleByte     1024  avgt   30   24.257 ? 0.031  ns/op
```

image

src/java.base/share/classes/jdk/internal/foreign/StringSupport.java
```diff
@@ -93,15 +115,23 @@ private static void writeFast_int(MemorySegment segment, long offset, Charset ch
...
+                    ? (long)STRLEN_TRIVIAL.invokeExact(segment)
+                    : (long)STRLEN.invokeExact(segment);
+
+            // On platforms where `size_t` maps to an `int`, we must check if `len < 0`
+            if (len < 0 || len > ArraysSupport.SOFT_MAX_ARRAY_LENGTH) {
+                throw new IllegalArgumentException("String too large");
+            }
+            return (int)len;
+        } catch (Throwable e) {
```
ExE-Boss, line 132:
```suggestion
        } catch (RuntimeException | Error e) {
            throw e;
        } catch (Throwable e) {
```

src/java.base/share/classes/jdk/internal/foreign/StringSupport.java
```diff
@@ -120,20 +123,47 @@ private static int native_strlen_byte(MemorySegment segment, long start) {
...
             }
-            long len = segment.byteSize() < MAX_TRIVIAL_SIZE
-                    ? (long)STRLEN_TRIVIAL.invokeExact(segment)
-                    : (long)STRLEN.invokeExact(segment);
+            long segmentSize = segment.byteSize();
+            final long len;
+            if (SIZE_T_IS_INT) {
+                if (segmentSize < MAX_TRIVIAL_SIZE) {
+                    len = (int)STRNLEN_TRIVIAL.invokeExact(segment, (int) segmentSize);
+                } else if (segmentSize < Integer.MAX_VALUE) {
```
JornVernee, line 131: size_t is unsigned. So it should be fine to use the native variant as long as the upper 32 bits are not set.

src/java.base/share/classes/jdk/internal/foreign/StringSupport.java
```diff
@@ -120,20 +123,47 @@ private static int native_strlen_byte(MemorySegment segment, long start) {
...
+                    len = (int)STRNLEN.invokeExact(segment, (int) segmentSize);
+                } else {
+                    // There is no way to express the max size in the native method using an int so, revert
+                    // to a Java method. It is possible to use a reduction of several STRNLEN invocations
+                    // in a future optimization.
+                    len = strlen_byte(segment);
```
JornVernee, line 137: Can't we just always throw here? `len` is not allowed to be larger than an `int` any ways.

src/java.base/share/classes/jdk/internal/foreign/StringSupport.java
```diff
@@ -93,10 +118,45 @@ private static void writeFast_int(MemorySegment segment, long offset, Charset ch
         segment.set(JAVA_INT, offset + bytes.length, 0);
     }
 
-    private static int strlen_byte(MemorySegment segment, long start) {
+    private static int native_strlen_byte(MemorySegment segment, long start) {
+        try {
```
JornVernee, line 122: I think this `try` block is only meant to deal with a potential exception from `invokeExact`? But, since it covers many lines of code, it's hard to tell what it's doing.

I suggest trying to limit the covered lines, for example by creating out-of-line wrapper methods for `STRNLEN` and `STRNLEN_TRIVIAL` and then putting the try/catch only in there.

src/java.base/share/classes/jdk/internal/foreign/StringSupport.java
minborg: It might be the case that we have a very large segment but there is a zero byte early in the segment.

JornVernee: Ok, thanks. That makes sense.

JornVernee (approved):

src/java.base/share/classes/jdk/internal/foreign/StringSupport.java
minborg: The linker can call native methods where the segment is heap-based too.

minborg: Or, no. You are right. We have to handle this separately.

src/java.base/share/classes/jdk/internal/foreign/StringSupport.java
```diff
@@ -93,15 +118,83 @@ private static void writeFast_int(MemorySegment segment, long offset, Charset ch
...
+            throw new IllegalArgumentException(e);
+        }
+    }
+
+    private static int strlen_byte(MemorySegment segment) {
         // iterate until overflow (String can only hold a byte[], whose length can be expressed as an int)
-        for (int offset = 0; offset >= 0; offset++) {
-            byte curr = segment.get(JAVA_BYTE, start + offset);
+        for (int offset = 0; offset >= 0; offset += 2) {
+            short curr = segment.get(JAVA_SHORT, offset);
```
ExE-Boss, line 192: The old version intentionally used `JAVA_BYTE` here, as `strlen` and `strnlen` is for byte‑encoded strings (e.g.: **UTF‑8**):
```suggestion
        for (int offset = 0; offset >= 0; offset++) {
            short curr = segment.get(JAVA_BYTE, offset);
```

ExE-Boss (changes requested):

src/java.base/share/classes/jdk/internal/foreign/StringSupport.java
minborg: We are planning on a new PR that is using another algorithm.

PR #862, opened 2023-08-14
Improve strlen performance

minborg:
This PR suggests removing the use of native calls for strlen and instead use Java implementations.

The PR also suggest performance improvements for quad word strlen.

Here are some benchmarks that compares the performance of the new proposed methods with the performance of the JDK 21 variants (called "legacy" methods):

image

---------
### Progress

### Reviewers
 * [Maurizio Cimadamore](https://openjdk.org/census#mcimadamore) (@mcimadamore - Committer) ⚠️ Review applies to [03f5bfe6](https://git.openjdk.org/panama-foreign/pull/862/files/03f5bfe6cd65ecc420bab3f155d134adb937049a)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign.git pull/862/head:pull/862` \
`$ git checkout pull/862`

`$ git checkout pull/862` \
`$ git pull https://git.openjdk.org/panama-foreign.git pull/862/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 862`

`$ git pr show -t 862`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/862.diff">https://git.openjdk.org/panama-foreign/pull/862.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/panama-foreign/pull/862#issuecomment-1677043145)


minborg:
In the benchmark `InternalStrLen.java` it would be better to `--add-exports=java.base/jdk.internal.foreign=ALL-UNNAMED` and run the real methods. For some reason, that does not work so suggestions are welcome on how to fix that.

test/jdk/java/foreign/TestStringEncoding.java
```diff
@@ -191,7 +192,7 @@ public void testOffset(String testString) {
         if (testString.length() < 3 || !containsOnlyRegularCharacters(testString)) {
             return;
         }
-        for(var charset:singleByteCharsets()) {
+        for (var charset:singleByteCharsets()) {
```
mcimadamore, line 195: need spaces around `:`

src/java.base/share/classes/jdk/internal/foreign/StringSupport.java
```diff
@@ -46,23 +40,6 @@ public final class StringSupport {
...
-        var description = FunctionDescriptor.of(size_t, ADDRESS, size_t);
-
-        STRNLEN_CRITICAL = linker.downcallHandle(strnlen, description, Linker.Option.critical());
-        STRNLEN = linker.downcallHandle(strnlen, description);
-        SIZE_T_IS_INT = (size_t.byteSize() == Integer.BYTES);
-    }
-
     public static String read(MemorySegment segment, long offset, Charset charset) {
         return switch (CharsetKind.of(charset)) {
             case SINGLE_BYTE -> readFast_byte(segment, offset, charset);
```
mcimadamore, line 45: I believe that, in the new code, there's no more distinction between slow and fast path? So we could simplify names to just "read/write_byte".

src/java.base/share/classes/jdk/internal/foreign/StringSupport.java
```diff
@@ -117,83 +95,103 @@ private static void writeFast_int(MemorySegment segment, long offset, Charset ch
...
+     *
+     * @param segment to examine
+     * @param start   from where examination shall begin
+     * @throws IllegalArgumentException if the examined region contains no zero bytes
+     *                                  within a length that can be accepted by a String
+     */
+    public static int chunked_strlen_byte(MemorySegment segment, long start) {
+
+        // Handle the first unaligned "head" bytes separately
+        int headCount = (int)SharedUtils.remainsToAlignment(segment.address() + start, Long.BYTES);
```
mcimadamore, line 123: This logic still seems to be tied to native segments. Heap segments have virtualized addresses, so this logic seems invalid. The best we can do, for heap segments, is to only use the better algo if we're using a long[] as an heap base (in which case the logic here would work reliably, and turn into aligned reads).

src/java.base/share/classes/jdk/internal/foreign/StringSupport.java
```diff
@@ -202,21 +200,77 @@ private static int strlen_byte(MemorySegment segment, long start) {
...
 
-    private static int strlen_int(MemorySegment segment, long start) {
-        // iterate until overflow (String can only hold a byte[], whose length can be expressed as an int)
-        for (int offset = 0; offset >= 0; offset += 4) {
-            int curr = segment.get(JAVA_INT, start + offset);
+    // The gain of using `long` wide operations for `int` is lower than for the two other `byte` and `short` variants
+    public static int strlen_int(MemorySegment segment, long start) {
+        // Do an initial read using aligned semantics.
+        // If this succeeds, we know that all other subsequent reads will be aligned
+        if (segment.get(JAVA_INT, start) == 0) {
```
mcimadamore, line 268: I would expect a simple loop here? What's the deal with avoiding alignment? If that's to speed up by avoiding alignment checks, it should not be required, as the segment being accessed is always at the same base address and the alignment check should be hoisted.

minborg: The proposed solution retains the same semantics as the old method where we would get an alignment error if presented with an unaligned MS (e.g. `alignedSegment.asSlice(1)`) but with improved performance.

mcimadamore: Ok, I understand you might see improved numbers. What I'm saying is that in principle this trick should not be necessary: the loop is a counted loop, and we know we're incrementing by 8 on each iteration. So the alignment check should be hoist-able as once we check the base address there's nothing more to check.

mcimadamore: So, I'd suggest to capture this in a small benchmark, and perhaps file a C2 compiler issue.

rsmogura: I wonder if optimisation issues can be related to using int counted loops instead of long, and using `ArraysSupport.SOFT_MAX_ARRAY_LENGTH`, maybe limit should be memory segment size. Adding large int to start can theoretically overflow long.

src/java.base/share/classes/jdk/internal/foreign/StringSupport.java
```diff
@@ -42,54 +42,62 @@ private StringSupport() {}
...
+            case SINGLE_BYTE -> writeByte(segment, offset, charset, string);
+            case DOUBLE_BYTE -> writeShort(segment, offset, charset, string);
+            case QUAD_BYTE -> writeInt(segment, offset, charset, string);
         }
     }
 
-    private static String readFast_byte(MemorySegment segment, long offset, Charset charset) {
-        long len = chunked_strlen_byte(segment, offset);
+    private static String readByte(MemorySegment segment, long offset, Charset charset) {
+        long len = segment.isNative() ||
```
mcimadamore, line 60: Maybe this test should be captured into its own predicate, to avoid repetition?

test/jdk/java/foreign/TestStringEncoding.java
```diff
@@ -447,4 +457,61 @@ private static int strlen_byte(MemorySegment segment, long start) {
...
+            if (!type.isPrimitive()) {
+                throw new IllegalArgumentException(type.toString());
+            }
+            this.type = type;
+            this.delegate = Arena.ofConfined();
+        }
+
+        @Override
+        public MemorySegment allocate(long byteSize, long byteAlignment) {
+            var seg = switch (type) {
```
mcimadamore, line 484: Technically speaking, this is not an arena - as all returned segments have disjoint scopes. That said, you can try to create an heap-based arena using a slicing allocator? E.g. have HeapArena take a Java array - then wrap the array into a segment in the constructor, and wrap that segment into a SegmentAllocator. The Arena scope now is the scope of the heap segment. The close operation can be a no-op.

JornVernee:
> In the benchmark `InternalStrLen.java` it would be better to `--add-exports=java.base/jdk.internal.foreign=ALL-UNNAMED` and run the real methods. For some reason, that does not work so suggestions are welcome on how to fix that.

What exactly does not work in that case?

Note that you have to modify both the build flags in make/test/BuildMicrobenchmark.gmk, and then pass the required --add-exports as extra JVM arguments through the `@Fork` annotation.

test/jdk/java/foreign/TestStringEncoding.java
```diff
@@ -62,26 +63,29 @@ public class TestStringEncoding {
...
-                            testString.getBytes(charset).length +
-                            terminatorSize;
+                        assertEquals(text.byteSize(), expectedByteLength);
 
-                    assertEquals(text.byteSize(), expectedByteLength);
-
-                    String roundTrip = text.getString(0, charset);
-                    if (charset.newEncoder().canEncode(testString)) {
-                        assertEquals(roundTrip, testString);
+                        String roundTrip = text.getString(0, charset);
```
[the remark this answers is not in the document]
minborg: The unaligned access and the use of heap segments in the tests were added at the same time.

minborg:
Here are some up-to-date benchmarks:

```
Benchmark                                  (size)  Mode  Cnt    Score   Error  Units
InternalStrLen.legacyDoubleByte                 1  avgt   30    1.884 ? 0.012  ns/op
InternalStrLen.legacyDoubleByte                 4  avgt   30    3.748 ? 0.014  ns/op
InternalStrLen.legacyDoubleByte                16  avgt   30   11.264 ? 0.059  ns/op
InternalStrLen.legacyDoubleByte               251  avgt   30  164.827 ? 0.514  ns/op
InternalStrLen.legacyDoubleByte              1024  avgt   30  647.327 ? 2.600  ns/op
InternalStrLen.legacyQuadByte                   1  avgt   30    1.961 ? 0.082  ns/op
InternalStrLen.legacyQuadByte                   4  avgt   30    3.746 ? 0.012  ns/op
InternalStrLen.legacyQuadByte                  16  avgt   30   11.253 ? 0.048  ns/op
InternalStrLen.legacyQuadByte                 251  avgt   30  165.220 ? 0.809  ns/op
InternalStrLen.legacyQuadByte                1024  avgt   30  647.814 ? 2.576  ns/op
InternalStrLen.legacySingleByte                 1  avgt   30    1.586 ? 0.016  ns/op
InternalStrLen.legacySingleByte                 4  avgt   30    2.995 ? 0.056  ns/op
InternalStrLen.legacySingleByte                16  avgt   30   10.032 ? 1.354  ns/op
InternalStrLen.legacySingleByte               251  avgt   30  125.963 ? 0.636  ns/op
InternalStrLen.legacySingleByte              1024  avgt   30  489.120 ? 2.170  ns/op
InternalStrLen.legacySingleByteMisaligned       1  avgt   30    1.607 ? 0.035  ns/op
InternalStrLen.legacySingleByteMisaligned       4  avgt   30    3.004 ? 0.024  ns/op
InternalStrLen.legacySingleByteMisaligned      16  avgt   30    8.779 ? 0.113  ns/op
InternalStrLen.legacySingleByteMisaligned     251  avgt   30  127.541 ? 1.196  ns/op
InternalStrLen.legacySingleByteMisaligned    1024  avgt   30  494.867 ? 5.129  ns/op
InternalStrLen.newDoubleByte                    1  avgt   30    1.967 ? 0.204  ns/op
InternalStrLen.newDoubleByte                    4  avgt   30    4.002 ? 0.014  ns/op
InternalStrLen.newDoubleByte                   16  avgt   30    6.025 ? 0.050  ns/op
InternalStrLen.newDoubleByte                  251  avgt   30   28.835 ? 0.185  ns/op
InternalStrLen.newDoubleByte                 1024  avgt   30   72.941 ? 1.307  ns/op
InternalStrLen.newQuadByte                      1  avgt   30    1.430 ? 0.020  ns/op
InternalStrLen.newQuadByte                      4  avgt   30    3.844 ? 0.017  ns/op
InternalStrLen.newQuadByte                     16  avgt   30    6.822 ? 0.052  ns/op
InternalStrLen.newQuadByte                    251  avgt   30   48.940 ? 1.078  ns/op
InternalStrLen.newQuadByte                   1024  avgt   30  196.276 ? 1.722  ns/op
InternalStrLen.newSingleByte                    1  avgt   30    1.920 ? 0.016  ns/op
InternalStrLen.newSingleByte                    4  avgt   30    5.717 ? 0.748  ns/op
InternalStrLen.newSingleByte                   16  avgt   30    5.450 ? 0.047  ns/op
InternalStrLen.newSingleByte                  251  avgt   30   22.148 ? 0.155  ns/op
InternalStrLen.newSingleByte                 1024  avgt   30   39.291 ? 0.220  ns/op
InternalStrLen.newSingleByteMisaligned          1  avgt   30    1.761 ? 0.010  ns/op
InternalStrLen.newSingleByteMisaligned          4  avgt   30    4.937 ? 0.029  ns/op
InternalStrLen.newSingleByteMisaligned         16  avgt   30    9.741 ? 0.041  ns/op
InternalStrLen.newSingleByteMisaligned        251  avgt   30   28.152 ? 0.346  ns/op
InternalStrLen.newSingleByteMisaligned       1024  avgt   30   44.796 ? 0.805  ns/op
```

test/jdk/java/foreign/TestStringEncoding.java
```diff
@@ -368,4 +454,60 @@ private static int strlen_byte(MemorySegment segment, long start) {
...
+        private static final int ELEMENT_SIZE = 1_000;
+
+        private final MemorySegment backingSegment;
+        private final SegmentAllocator allocator;
+
+        public HeapArena(Class<?> type) {
+            if (!type.isPrimitive()) {
+                throw new IllegalArgumentException(type.toString());
+            }
+            backingSegment = switch (type) {
```
mcimadamore, line 477: Is this better than if/else? :-)

test/micro/org/openjdk/bench/java/lang/foreign/InternalStrLen.java
```diff
@@ -0,0 +1,161 @@
...
+        singleByteSegment.set(ValueLayout.JAVA_BYTE, singleByteSegment.byteSize() - Byte.BYTES, (byte) 0);
+        doubleByteSegment.set(ValueLayout.JAVA_SHORT, doubleByteSegment.byteSize() - Short.BYTES, (short) 0);
+        quadByteSegment.set(ValueLayout.JAVA_INT, quadByteSegment.byteSize() - Integer.BYTES, 0);
+        singleByteSegmentMisaligned = arena.allocate(singleByteSegment.byteSize() + 1).
+                asSlice(1);
+        MemorySegment.copy(singleByteSegment, 0, singleByteSegmentMisaligned, 0, singleByteSegment.byteSize());
+    }
+
+    @Benchmark
+    public int legacySingleByte() {
```
mcimadamore, line 87: I think we should rename `legacy` and `new` to something else. E.g. `byte` vs. `bulk`. You might rename the various methods as `byte_2` (meaning byte-by-byte, with two bytes chars).

src/java.base/share/classes/jdk/internal/foreign/StringSupport.java
```diff
@@ -46,154 +40,158 @@ public final class StringSupport {
...
-        if (start > 0) {
-            segment = segment.asSlice(start);
+    /**
+     * {@return the shortest distance beginning at the provided {@code start}
+     *  to the encountering of a zero byte in the provided {@code segment}}
+     * <p>
+     * The method divides the region of interest into three distinct regions:
+     * <ul>
+     *     <li>head (un-aligned access handling on a byte-by-byte basis) (if any)</li>
+     *     <li>body (long aligned access handling eight bytes at a time) (if any)</li>
```
mcimadamore, line 105: This and other comments in this class refer to alignment properties which I don't think are strictly valid. You might want to search for them and adjust them/remove them.

mcimadamore (approved):
Looks good - added few nitpick comments

PR #865, opened 2023-08-15
8314260: Unable to load system libraries on Windows when using a SecurityManager

minborg:
This PR proposes to read the system environment variable "SystemRoot" using a privileged operation so it will work in the event a default `SecurityManager` is in place.

### DIFF
--- a/src/java.base/share/classes/jdk/internal/foreign/SystemLookup.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/SystemLookup.java
@@ -29,6 +29,8 @@ import java.lang.foreign.*;
 import java.lang.invoke.MethodHandles;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.function.Function;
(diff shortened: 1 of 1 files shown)


src/java.base/share/classes/jdk/internal/foreign/SystemLookup.java
```diff
@@ -72,7 +74,15 @@ private static SymbolLookup makeSystemLookup() {
...
 
     private static SymbolLookup makeWindowsLookup() {
-        Path system32 = Path.of(System.getenv("SystemRoot"), "System32");
+        @SuppressWarnings("removal")
+        String systemRoot = AccessController.doPrivileged(new PrivilegedAction<String>() {
+            @Override
+            public String run() {
+                return System.getenv("SystemRoot");
+            }
+        });
```
JornVernee, line 83: We can use `sun.security.action.GetPropertyAction.privilegedGetProperty` here for simplicity. See e.g. https://github.com/openjdk/panama-foreign/blob/foreign-memaccess%2Babi/src/java.base/share/classes/jdk/internal/foreign/CABI.java#L50 where it's used as well.

JornVernee (commented):
Nice find!

src/java.base/share/classes/jdk/internal/foreign/SystemLookup.java
JornVernee: Never mind, that is for System properties. Here we're reading an environment variable.

ExE-Boss: I mean create such a class by copying what the other `sun.security.action.Get*Action` classes do.

mcimadamore:
> We will convert this PR into a PR on the mainline after the JEP PR is merged.

Please close this PR if this is not meant to be integrated.

minborg:
These proposals will be targeted to the mainline and thus, I am closing this PR.

PR #868, opened 2023-08-17
Revert SegmentAllocator API changes

mcimadamore:
This patch reverts the changes to the `implSpec` of the various default methods in `SegmentAllocator`.
We have realized that the changes there are irrelevant. What matters in order to get to optimal peak performance is that `SegmentAllocator::copyArrayWithSwapIfNeeded` is force-inlined.

The segment allocator test is also reverted to its former self.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign.git pull/868/head:pull/868` \
`$ git checkout pull/868`

`$ git checkout pull/868` \
`$ git pull https://git.openjdk.org/panama-foreign.git pull/868/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 868`

`$ git pr show -t 868`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/868.diff">https://git.openjdk.org/panama-foreign/pull/868.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/panama-foreign/pull/868#issuecomment-1682663662)

### DIFF
--- a/src/java.base/share/classes/java/lang/foreign/SegmentAllocator.java
+++ b/src/java.base/share/classes/java/lang/foreign/SegmentAllocator.java
@@ -112,7 +112,7 @@ public interface SegmentAllocator {
      * @return a new native segment containing the converted C string.
      * @throws UnsupportedOperationException if {@code charset} is not a {@linkplain StandardCharsets standard charset}.
      * @implSpec the default implementation for this method copies the contents of the provided Java string
-     * into a new memory segment obtained by calling {@code this.allocate(B + N, 1)}, where:
+     * into a new memory segment obtained by calling {@code this.allocate(B + N)}, where:
      * <ul>
      *     <li>{@code B} is the size, in bytes, of the string encoded using the provided charset
      *     (e.g. {@code str.getBytes(charset).length});</li>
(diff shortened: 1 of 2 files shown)


test/jdk/java/foreign/TestSegmentAllocators.java
```diff
@@ -181,9 +189,18 @@ public void testArrayAllocateDelegation() {
     @Test
     public void testStringAllocateDelegation() {
         AtomicInteger calls = new AtomicInteger();
-        SegmentAllocator allocator = (byteSize, byteAlignment) -> {
-            calls.incrementAndGet();
-            return Arena.ofAuto().allocate(byteSize, byteAlignment);
+        SegmentAllocator allocator = new SegmentAllocator() {
+            @Override
+
```
minborg, line 194: Extraneous line

src/java.base/share/classes/java/lang/foreign/SegmentAllocator.java
```diff
@@ -340,10 +325,10 @@ default MemorySegment allocateFrom(ValueLayout.OfDouble elementLayout, double...
         return copyArrayWithSwapIfNeeded(elements, elementLayout);
     }
 
+    @ForceInline
     private MemorySegment copyArrayWithSwapIfNeeded(Object array, ValueLayout elementLayout) {
         int size = Array.getLength(Objects.requireNonNull(array));
-        Objects.requireNonNull(elementLayout);
-        MemorySegment segment = allocateNoInit(elementLayout.byteSize() * size, elementLayout.byteAlignment());
+        MemorySegment segment = allocateNoInit(Objects.requireNonNull(elementLayout), size);
         if (size > 0) {
```
minborg, line 332: Do we need this check? Is it not so that a vast majority of provided arrays would have a size > 0 and in the rare occasion there is a zero-length array, it will still work without this check?

mcimadamore: The check was needed to preserve semantics of the method (there was one test failure because of that - but I think we can fix the test instead).

JornVernee (approved):

PR #871, opened 2023-08-18
8314592: Add shortcut to SymbolLookup::find

mcimadamore:
This patch adds a default method in `SymbolLookup`, namely `SymbolLookup::get`. Its implementation is very simple: it calls `SymbolLookup::find` and then `Optional::get` on the result. This allows to simplify many clients that end up calling `Optional::get` or `Optional::orElseThrow` after a symbol lookup, but still retains the compositional advantage of the optional-returning lookup primitive.

---------
### Progress

### Issue
 * [JDK-8314592](https://bugs.openjdk.org/browse/JDK-8314592): Add shortcut to SymbolLookup::find (**Enhancement** - P3)

### Reviewers
 * [Per Minborg](https://openjdk.org/census#pminborg) (@minborg - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign.git pull/871/head:pull/871` \
`$ git checkout pull/871`

`$ git checkout pull/871` \
`$ git pull https://git.openjdk.org/panama-foreign.git pull/871/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 871`

`$ git pr show -t 871`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/871.diff">https://git.openjdk.org/panama-foreign/pull/871.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/panama-foreign/pull/871#issuecomment-1684206067)

### DIFF
--- a/src/java.base/share/classes/java/lang/foreign/SymbolLookup.java
+++ b/src/java.base/share/classes/java/lang/foreign/SymbolLookup.java
@@ -69,7 +69,7 @@ import java.util.function.BiFunction;
  * {@snippet lang = java:
  * try (Arena arena = Arena.ofConfined()) {
  *     SymbolLookup libGL = SymbolLookup.libraryLookup("libGL.so", arena); // libGL.so loaded here
- *     MemorySegment glGetString = libGL.find("glGetString").orElseThrow();
+ *     MemorySegment glGetString = libGL.get("glGetString");
  *     ...
  * } //  libGL.so unloaded here
  *}
@@ -82,7 +82,7 @@ import java.util.function.BiFunction;
  * System.loadLibrary("GL"); // libGL.so loaded here
  * ...
  * SymbolLookup libGL = SymbolLookup.loaderLookup();
- * MemorySegment glGetString = libGL.find("glGetString").orElseThrow();
+ * MemorySegment glGetString = libGL.get("glGetString");
  * }
  *
  * This symbol lookup, which is known as a <em>loader lookup</em>, is dynamic with respect to the libraries associated
@@ -115,7 +115,7 @@ import java.util.function.BiFunction;
  * {@snippet lang = java:
  * Linker nativeLinker = Linker.nativeLinker();
  * SymbolLookup stdlib = nativeLinker.defaultLookup();
- * MemorySegment malloc = stdlib.find("malloc").orElseThrow();
+ * MemorySegment malloc = stdlib.get("malloc");
  *}
  *
  * @since 22
@@ -129,6 +129,17 @@ public interface SymbolLookup {
      * @return a zero-length memory segment whose address indicates the address of the symbol, if found.
      */
     Optional<MemorySegment> find(String name);
+
+    /**
+     * Returns the address of the symbol with the given name.
+     * @implSpec the default implementation for this method calls {@code this.find(name).get()}.
+     * @param name the symbol name.
+     * @return a zero-length memory segment whose address indicates the address of the symbol.
+     * @throws java.util.NoSuchElementException if no symbol with the given name exists.
+     */
+    default MemorySegment get(String name) {
+        return find(name).get();
+    }
 
     /**
      * {@return a composed symbol lookup that returns result of finding the symbol with this lookup if found,
--- a/src/java.base/share/classes/java/lang/foreign/SymbolLookup.java
+++ b/src/java.base/share/classes/java/lang/foreign/SymbolLookup.java
@@ -37,6 +37,7 @@ import jdk.internal.reflect.Reflection;
 
 import java.lang.invoke.MethodHandles;
 import java.nio.file.Path;
+import java.util.NoSuchElementException;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.function.BiFunction;
@@ -138,7 +139,13 @@ public interface SymbolLookup {
      * @throws java.util.NoSuchElementException if no symbol with the given name exists.
      */
     default MemorySegment get(String name) {
-        return find(name).get();
+        var address = this.find(name);
+        // Note: the code below could use orElseThrow, but that would require a capturing lambda
+        if (address.isPresent()) {
+            return address.get();
+        } else {
+            throw new NoSuchElementException("Unable to find a symbol with name '" + name + "'");
+        }
     }
 
     /**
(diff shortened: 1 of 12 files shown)


JornVernee (commented):
This looks like a good move, but I'm not super sure about the name `get`. It seems very close to `find` to me. I'd prefer something like `findOrThrow`.

mcimadamore:
> This looks like a good move, but I'm not super sure about the name `get`. It seems very close to `find` to me. I'd prefer something like `findOrThrow`.

The naming scheme was inspired by this:
https://blog.joda.org/2015/09/naming-optional-query-methods.html

src/java.base/share/classes/java/lang/foreign/SymbolLookup.java
```diff
@@ -130,6 +130,17 @@ public interface SymbolLookup {
...
+    /**
+     * Returns the address of the symbol with the given name.
+     * @implSpec the default implementation for this method calls {@code this.find(name).get()}.
+     * @param name the symbol name.
+     * @return a zero-length memory segment whose address indicates the address of the symbol.
+     * @throws java.util.NoSuchElementException if no symbol with the given name exists.
+     */
+    default MemorySegment get(String name) {
+        return find(name).get();
+    }
```
ExE-Boss, line 142: Maybe provide a better error message:
```suggestion
    /**
     * Returns the address of the symbol with the given name.
     * @implSpec the default implementation for this method calls {@code this.find(name).orElseThrow()}.
     * @param name the symbol name.
     * @return a zero-length memory segment whose address indicates the address of the symbol.
     * @throws java.util.NoSuchElementException if no symbol with the given name exists.
     */
    default MemorySegment get(String name) {
        var address = this.find(name);

        // Avoid using a capturing lambda
        if (address.isPresent()) {
            return address.orElseThrow();
        }
        throw new NoSuchElementException("Unable to find a native symbol named '" + name + "'");
    }
```

See also: https://github.com/openjdk/jdk/pull/15286

https://github.com/openjdk/panama-foreign/blob/c4e05f5c25910e5e23d3f6af585d066f671fb9b7/src/java.base/share/classes/jdk/internal/foreign/support/DefaultNativeLookupUtil.java#L47-L57

minborg:
I think the idea of having such a method is good (especially if the exception provides more information as to what went wrong) but I was also expecting a name indicating something would be thrown if the symbol was not found. The meaning of `get` is different in the JDK with `Map::get` returning `null` and `Optional::get` throwing for example.

mcimadamore:
I've made the exception message clearer. My general feeling on naming is that if we make it too verbose (e.g. `findOrThrow`), then it's not very different from `find(...).orElseThrow()` ?

If there's no consensus on the naming scheme, I'm also ok leaving it out for now, as this is something that can always be added at a later point.

minborg (approved):

mcimadamore:
It occurred to me that another possible way to reduce boilerplate could be to leave `SymbolLookup` alone, and act on `Linker` instead, by adding two additional `default` overloads of `downcallHandle`:

```
default downcallHandle(String funcName, FunctionDescriptor desc, Linker.Options... options) {
     return downcallHandle(defaultLookup(), funcName, desc, options);
}

default downcallHandle(SymbolLookup lookup, String funcName, FunctionDescriptor desc, Linker.Options... options) {
     return downcallHandle(lookup.find(funcName).orElseThrow(), desc, options);
}
```

Now, let's look at some code creating a downcall handle. This is with 21 API:

```
MethodHandle strlen = linker.downcallHandle(
      linker.defaultLookup().find("strlen").orElseThrow(),
      FunctionDescriptor.of(JAVA_LONG, ADDRESS)
  );
```
This is with the changes proposed in this PR:

```
MethodHandle strlen = linker.downcallHandle(
     linker.defaultLookup().get("strlen"),
     FunctionDescriptor.of(JAVA_LONG, ADDRESS)
 );
 ```
 
 And this is with the provided `Linker` overloads:
 
 ```
 MethodHandle strlen = linker.downcallHandle(
     "strlen",
     FunctionDescriptor.of(JAVA_LONG, ADDRESS)
 );  
 ```

The latter is certainly the most succinct. And if we were to lookup a symbol in some other library:

```
 MethodHandle clangVersion = linker.downcallHandle(
     clangLookup,
     "clang_getClangVersion",
     FunctionDescriptor.of(JAVA_LONG, ADDRESS)
 );  
```

Which seems to scale well. Of course this is a bit more ad-hoc as it would only work when interacting with `Linker::downcallHandle`, but perhaps that's where most of the verbosity comes from (at least looking at this patch) ?

minborg:
> It occurred to me that another possible way to reduce boilerplate could be to leave `SymbolLookup` alone, and act on `Linker` instead, by adding two additional `default` overloads of `downcallHandle`:
> 
> ```
> default downcallHandle(String funcName, FunctionDescriptor desc, Linker.Options... options) {
>      return downcallHandle(defaultLookup(), funcName, desc, options);
> }
> 
> default downcallHandle(SymbolLookup lookup, String funcName, FunctionDescriptor desc, Linker.Options... options) {
>      return downcallHandle(lookup.find(funcName).orElseThrow(), desc, options);
> }
> ```
> 
> Now, let's look at some code creating a downcall handle. This is with 21 API:
> 
> ```
> MethodHandle strlen = linker.downcallHandle(
>       linker.defaultLookup().find("strlen").orElseThrow(),
>       FunctionDescriptor.of(JAVA_LONG, ADDRESS)
>   );
> ```
> 
> This is with the changes proposed in this PR:
> 
> ```
> MethodHandle strlen = linker.downcallHandle(
>      linker.defaultLookup().get("strlen"),
>      FunctionDescriptor.of(JAVA_LONG, ADDRESS)
>  );
> ```
> 
> And this is with the provided `Linker` overloads:
> 
> ```
> MethodHandle strlen = linker.downcallHandle(
>     "strlen",
>     FunctionDescriptor.of(JAVA_LONG, ADDRESS)
> );  
> ```
> 
> The latter is certainly the most succinct. And if we were to lookup a symbol in some other library:
> 
> ```
>  MethodHandle clangVersion = linker.downcallHandle(
>      clangLookup,
>      "clang_getClangVersion",
>      FunctionDescriptor.of(JAVA_LONG, ADDRESS)
>  );  
> ```
> 
> Which seems to scale well. Of course this is a bit more ad-hoc as it would only work when interacting with `Linker::downcallHandle`, but perhaps that's where most of the verbosity comes from (at least looking at this patch) ?

In my panamization prototype, I have added utility methods that do something similar. I think this would be a welcomed addition, especially if the the default method is refined slightly to provide guidance as to what symbol could not be looked up, should that be the case.

mcimadamore:
I'm closing this for now. We can always revisit at a later point.

PR #876, opened 2023-08-28
8315096: Allowed access modes in memory segment should depend on layout alignment

mcimadamore:
There are some issues with the way memory segment var handle are specified and implemented: if an access mode is not supported, then using such access mode should result in an `UnsupportedOperationException`. But in our implementation, all alignment issues result in `IllegalArgumentException`.

This means that clients can get the same exception if:

* they try to read a 4-byte aligned int at a physical address that is not multiple of 4 (use-site error, IAE is good)
* they try to do a compareAndSwap on a var handle obtained using an unaligned int layout (decl-site error, IAE is not good, and UOE should be used instead).

This PR rectifies this issue: now a var handle is either *aligned* (meaning it is constructed with a value layout which respects the *size* of the accessed carrier type), or *unaligned*. Aligned var handles enjoy more access modes. Unaligned var handle only support `get`/`set`. Trying to use any other access mode on an unaligned var handle will now result in `UOE` regardless of the address being accessed - which makes the code more robust and portable.

This PR also clarifies the javadoc, by avoiding using the term "atomic access" to mean "absence of tearing". This is confusing because we then refer to terms such as "atomic updates" also to refer to operations such as CAS.

The proposed javadoc now uses the term "Word Tearing" which is taken from the memory model section of the JLS (and even has a link to it).

---------
### Progress

### Issue
 * [JDK-8315096](https://bugs.openjdk.org/browse/JDK-8315096): Allowed access modes in memory segment should depend on layout alignment (**Bug** - P3)

### Reviewers
 * [Paul Sandoz](https://openjdk.org/census#psandoz) (@PaulSandoz - Committer) ⚠️ Review applies to [b86a6860](https://git.openjdk.org/panama-foreign/pull/876/files/b86a686076f1f303cb6d538bbd0a5c6ef24aa9d6)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign.git pull/876/head:pull/876` \
`$ git checkout pull/876`

`$ git checkout pull/876` \
`$ git pull https://git.openjdk.org/panama-foreign.git pull/876/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 876`

`$ git pr show -t 876`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/876.diff">https://git.openjdk.org/panama-foreign/pull/876.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/panama-foreign/pull/876#issuecomment-1695564048)

### DIFF
--- a/src/java.base/share/classes/java/lang/invoke/X-VarHandleSegmentView.java.template
+++ b/src/java.base/share/classes/java/lang/invoke/X-VarHandleSegmentView.java.template
@@ -43,7 +43,7 @@ final class VarHandleSegmentAs$Type$s extends VarHandleSegmentViewBase {
 
     static final ScopedMemoryAccess SCOPED_MEMORY_ACCESS = ScopedMemoryAccess.getScopedMemoryAccess();
 
-    static final int VM_ALIGN = $BoxType$.BYTES - 1;
+    static final int NON_PLAIN_ACCESS_MIN_ALIGN_MASK = $BoxType$.BYTES - 1;
 
     static final VarForm FORM = new VarForm(VarHandleSegmentAs$Type$s.class, MemorySegment.class, $type$.class, long.class);
 
@@ -104,16 +104,15 @@ final class VarHandleSegmentAs$Type$s extends VarHandleSegmentViewBase {
     }
 
     @ForceInline
-    static long offset(AbstractMemorySegmentImpl bb, long offset, long alignmentMask) {
-        long address = offsetNoVMAlignCheck(bb, offset, alignmentMask);
-        if ((address & VM_ALIGN) != 0) {
-            throw VarHandleSegmentViewBase.newIllegalArgumentExceptionForMisalignedAccess(address);
+    static long offsetNonPlain(AbstractMemorySegmentImpl bb, long offset, long alignmentMask) {
+        if ((alignmentMask & NON_PLAIN_ACCESS_MIN_ALIGN_MASK) != NON_PLAIN_ACCESS_MIN_ALIGN_MASK) {
+            throw VarHandleSegmentViewBase.newUnsupportedAccessModeForAlignment(alignmentMask + 1);
         }
-        return address;
+        return offsetPlain(bb, offset, alignmentMask);
     }
 
     @ForceInline
-    static long offsetNoVMAlignCheck(AbstractMemorySegmentImpl bb, long offset, long alignmentMask) {
+    static long offsetPlain(AbstractMemorySegmentImpl bb, long offset, long alignmentMask) {
         long base = bb.unsafeGetOffset();
         long address = base + offset;
         long maxAlignMask = bb.maxAlignMask();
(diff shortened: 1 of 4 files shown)


PaulSandoz (commented):
Looks good.

I notice a subtle difference between this VH behavior and that for arrays and `ByteBuffer`, as specified by the methods on `MethodHandles`, which state:
```
     * <ul>
     * <li>read write access modes for all {@code T}, with the exception of
     *     access modes {@code get} and {@code set} for {@code long} and
     *     {@code double} on 32-bit platforms.
```
I don't exactly recall why we restricted on 32-bit platforms, but it might have been because of some `Unsafe` restriction/complexity in supporting the atomicity that is required for volatile (or more generally non-plain) access (e.g., see https://shipilev.net/blog/2014/all-accesses-are-atomic/#_implementation_specifics). So we probably said well we won't support it at all, but i doubt we assert on that given 32-bit is not a system we generally focus on.

mcimadamore:
> Looks good.
> 
> I notice a subtle difference between this VH behavior and that for arrays and `ByteBuffer`, as specified by the methods on `MethodHandles`, which state:
> 
> ```
>      * <ul>
>      * <li>read write access modes for all {@code T}, with the exception of
>      *     access modes {@code get} and {@code set} for {@code long} and
>      *     {@code double} on 32-bit platforms.
> ```
> 
> I don't exactly recall why we restricted on 32-bit platforms, but it might have been because of some `Unsafe` restriction/complexity in supporting the atomicity that is required for volatile (or more generally non-plain) access (e.g., see https://shipilev.net/blog/2014/all-accesses-are-atomic/#_implementation_specifics). So we probably said well we won't support it at all, but i doubt we assert on that given 32-bit is not a system we generally focus on.

Note: I believe this text is really meant to mean: "these access modes are supported, but access is not atomic" (as in: might tear).

Which is caused by the slightly overloaded use of the letters `atomic` in the var handle javadoc.

mcimadamore:
> Note: I believe this text is really meant to mean: "these access modes are supported, but access is not atomic" (as in: might tear).

There is no check in any of the VH implementation that I know of that makes it throw UOE if get/set is called in 32 bits, so to speak.

PaulSandoz:
> > I don't exactly recall why we restricted on 32-bit platforms, but it might have been because of some `Unsafe` restriction/complexity in supporting the atomicity that is required for volatile (or more generally non-plain) access (e.g., see https://shipilev.net/blog/2014/all-accesses-are-atomic/#_implementation_specifics). So we probably said well we won't support it at all, but i doubt we assert on that given 32-bit is not a system we generally focus on.
> 
> Note: I believe this text is really meant to mean: "these access modes are supported, but access is not atomic" (as in: might tear).
> 
> Which is caused by the slightly overloaded use of the letters `atomic` in the var handle javadoc.

Yes, i agree.. I will log an issue to update the text VarHandle text.

src/java.base/share/classes/java/lang/invoke/X-VarHandleSegmentView.java.template
```diff
@@ -180,7 +179,7 @@ final class VarHandleSegmentAs$Type$s extends VarHandleSegmentViewBase {
         return convEndian(handle.be,
                           SCOPED_MEMORY_ACCESS.get$RawType$Volatile(bb.sessionImpl(),
                                   bb.unsafeGetBase(),
-                                  offset(bb, base, handle.alignmentMask)));
+                                   offsetNonPlain(bb, base, handle.alignmentMask)));
```
ExE-Boss, line 182: The `offsetNonPlain` calls have an extra space:
```suggestion
                                  offsetNonPlain(bb, base, handle.alignmentMask)));
```

(this would’ve been avoidable if JDK used the more accessible tabs[^1] for indentation)

[^1]: https://github.com/prettier/prettier/issues/7475

mcimadamore: > The `offsetNonPlain` calls have an extra space:
> 
> (this would’ve been avoidable if JDK used the more accessible tabs[1](#user-content-fn-1-a99cebfdb5a6913c73e9a68c7c0e7c46) for indentation)
> ## Footnotes
> 
>     1. [Change `useTabs` to `true` by default prettier/prettier#7475](https://github.com/prettier/prettier/issues/7475) [↩](#user-content-fnref-1-a99cebfdb5a6913c73e9a68c7c0e7c46)

Part of the issue here is that this file is a template - so it's not a ".java" file, and as such, the IDE doesn't help with this stuff.

junaga: Prettier is used on the CLI and in git hooks.

PR #877, opened 2023-09-01
Add missing @implSpec to AddressLayout

mcimadamore:
As the title of this PR says, this simple patch adds a missing `@implSpec` doclet on `AddressLayout`, to bring it in sync with other layout classes.


minborg (approved):
Nice catch.

PR #878, opened 2023-09-06
8315769: Add support for sliced allocation

mcimadamore:
This PR adds a new method in `SegmentAllocator`:

```
default MemorySegment allocateFrom(ValueLayout elementLayout, MemorySegment source,
                                       ValueLayout sourceElementLayout, long sourceOffset, long srcElementCount) {
```

This method allows clients to allocate a new memory segment and copy the contents of a portion of an existing segment into the newly allocated region of memory. As such it can be used to address the following use cases:

* allocate from a `ByteBuffer`
* allocate from another memory segment
* allocate from a Java array slice

All these cases were not covered by the existing API points, which meant that developers had to use a more general allocation request (such as `allocate(long, long)`) and then pay a performance cost (because of memory zeroing). In other words, the new method in this PR completes the allocation API, by providing a flexible way to allocate a new segment from an existing source (another segment) with given offset and length.

Given that the new method is more general than the existing array-accepting `allocateFrom`, this PR rewires the existing array-accepting method to be rewritten on top of the new overload (and tweaks the javadoc of such methods accordingly).

One detail to note is that the new method takes _two_ element layouts - one is the layout of the newly allocated segment, whereas the other is the layout of the source segment. Such layouts must have same alignment and same carrier - but they can have different endianness (in which case a bulk copy with swap is performed). This is not too different from the most general `MemorySegment::copy` static overload.

---------
### Progress

### Issue
 * [JDK-8315769](https://bugs.openjdk.org/browse/JDK-8315769): Add support for sliced allocation (**Enhancement** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign.git pull/878/head:pull/878` \
`$ git checkout pull/878`

`$ git checkout pull/878` \
`$ git pull https://git.openjdk.org/panama-foreign.git pull/878/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 878`

`$ git pr show -t 878`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/878.diff">https://git.openjdk.org/panama-foreign/pull/878.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/panama-foreign/pull/878#issuecomment-1708144558)

### DIFF
--- a/src/java.base/share/classes/java/lang/foreign/SegmentAllocator.java
+++ b/src/java.base/share/classes/java/lang/foreign/SegmentAllocator.java
@@ -285,7 +285,7 @@ public interface SegmentAllocator {
         Objects.requireNonNull(sourceElementLayout);
         Objects.requireNonNull(elementLayout);
         MemorySegment dest = allocateNoInit(elementLayout, srcElementCount);
-        MemorySegment.copy(source, sourceElementLayout, sourceOffset, dest, elementLayout, 0, dest.byteSize());
+        MemorySegment.copy(source, sourceElementLayout, sourceOffset, dest, elementLayout, 0, srcElementCount);
         return dest;
     }
 
--- a/src/java.base/share/classes/java/lang/foreign/SegmentAllocator.java
+++ b/src/java.base/share/classes/java/lang/foreign/SegmentAllocator.java
@@ -296,8 +296,7 @@ public interface SegmentAllocator {
      * @implSpec the default implementation for this method is equivalent to the following code:
      * {@snippet lang = java:
      * this.allocateFrom(layout, MemorySegment.ofArray(array),
-     *                   layout.withOrder(ByteOrder.nativeOrder()),
-     *                   0, array.length)
+     *                   ValueLayout.JAVA_BYTE, 0, array.length)
      *}
      * @param elementLayout the element layout of the array to be allocated.
      * @param elements the byte elements to be copied to the newly allocated memory block.
@@ -307,7 +306,7 @@ public interface SegmentAllocator {
     @ForceInline
     default MemorySegment allocateFrom(ValueLayout.OfByte elementLayout, byte... elements) {
         return allocateFrom(elementLayout, MemorySegment.ofArray(elements),
-                elementLayout.withOrder(ByteOrder.nativeOrder()), 0, elements.length);
+                ValueLayout.JAVA_BYTE, 0, elements.length);
     }
 
     /**
@@ -315,8 +314,7 @@ public interface SegmentAllocator {
      * @implSpec the default implementation for this method is equivalent to the following code:
      * {@snippet lang = java:
      * this.allocateFrom(layout, MemorySegment.ofArray(array),
-     *                   layout.withOrder(ByteOrder.nativeOrder()),
-     *                   0, array.length)
+     *                   ValueLayout.JAVA_SHORT, 0, array.length)
      *}
      * @param elementLayout the element layout of the array to be allocated.
      * @param elements the short elements to be copied to the newly allocated memory block.
@@ -326,7 +324,7 @@ public interface SegmentAllocator {
     @ForceInline
     default MemorySegment allocateFrom(ValueLayout.OfShort elementLayout, short... elements) {
         return allocateFrom(elementLayout, MemorySegment.ofArray(elements),
-                elementLayout.withOrder(ByteOrder.nativeOrder()), 0, elements.length);
+                ValueLayout.JAVA_SHORT, 0, elements.length);
     }
 
     /**
@@ -334,8 +332,7 @@ public interface SegmentAllocator {
      * @implSpec the default implementation for this method is equivalent to the following code:
      * {@snippet lang = java:
      * this.allocateFrom(layout, MemorySegment.ofArray(array),
-     *                   layout.withOrder(ByteOrder.nativeOrder()),
-     *                   0, array.length)
+     *                   ValueLayout.JAVA_CHAR, 0, array.length)
      *}
      * @param elementLayout the element layout of the array to be allocated.
      * @param elements the char elements to be copied to the newly allocated memory block.
@@ -345,7 +342,7 @@ public interface SegmentAllocator {
     @ForceInline
     default MemorySegment allocateFrom(ValueLayout.OfChar elementLayout, char... elements) {
         return allocateFrom(elementLayout, MemorySegment.ofArray(elements),
-                elementLayout.withOrder(ByteOrder.nativeOrder()), 0, elements.length);
+                ValueLayout.JAVA_CHAR, 0, elements.length);
     }
 
     /**
@@ -353,8 +350,7 @@ public interface SegmentAllocator {
      * @implSpec the default implementation for this method is equivalent to the following code:
      * {@snippet lang = java:
      * this.allocateFrom(layout, MemorySegment.ofArray(array),
-     *                   layout.withOrder(ByteOrder.nativeOrder()),
-     *                   0, array.length)
+     *                   ValueLayout.JAVA_INT, 0, array.length)
      *}
      * @param elementLayout the element layout of the array to be allocated.
      * @param elements the int elements to be copied to the newly allocated memory block.
@@ -364,7 +360,7 @@ public interface SegmentAllocator {
     @ForceInline
     default MemorySegment allocateFrom(ValueLayout.OfInt elementLayout, int... elements) {
         return allocateFrom(elementLayout, MemorySegment.ofArray(elements),
-                elementLayout.withOrder(ByteOrder.nativeOrder()), 0, elements.length);
+                ValueLayout.JAVA_INT, 0, elements.length);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


mcimadamore:
Here's the benchmark results I got:

```
Benchmark                                (size)  Mode  Cnt   Score   Error  Units
AllocFromSliceTest.alloc_confined             5  avgt   30  69.853 ? 1.834  ns/op
AllocFromSliceTest.alloc_confined            20  avgt   30  65.872 ? 3.113  ns/op
AllocFromSliceTest.alloc_confined           100  avgt   30  61.102 ? 2.832  ns/op
AllocFromSliceTest.alloc_confined           500  avgt   30  68.474 ? 1.573  ns/op
AllocFromSliceTest.alloc_confined          1000  avgt   30  82.487 ? 1.597  ns/op
AllocFromSliceTest.alloc_confined_slice       5  avgt   30  41.945 ? 0.666  ns/op
AllocFromSliceTest.alloc_confined_slice      20  avgt   30  43.159 ? 1.294  ns/op
AllocFromSliceTest.alloc_confined_slice     100  avgt   30  42.603 ? 0.439  ns/op
AllocFromSliceTest.alloc_confined_slice     500  avgt   30  46.771 ? 1.099  ns/op
AllocFromSliceTest.alloc_confined_slice    1000  avgt   30  58.050 ? 1.484  ns/op
```

As it can be seen, absence of any memory zeroing helps the numbers to remain more "flat" with growing sizes.

src/java.base/share/classes/java/lang/foreign/SegmentAllocator.java
```diff
@@ -257,89 +257,162 @@ default MemorySegment allocateFrom(AddressLayout layout, MemorySegment value) {
...
+     *}
      * @param elementLayout the element layout of the array to be allocated.
      * @param elements the byte elements to be copied to the newly allocated memory block.
      * @return a segment for the newly allocated memory block.
      */
+    @ForceInline
     default MemorySegment allocateFrom(ValueLayout.OfByte elementLayout, byte... elements) {
-        return copyArrayWithSwapIfNeeded(elements, elementLayout);
+        return allocateFrom(elementLayout, MemorySegment.ofArray(elements),
+                elementLayout.withOrder(ByteOrder.nativeOrder()), 0, elements.length);
```
JornVernee, line 307: Can't the source layout here just be `JAVA_BYTE`? Since we're copying from a java byte array. (similar for the other overloads).

```suggestion
                JAVA_BYTE, 0, elements.length);
```

src/java.base/share/classes/java/lang/foreign/SegmentAllocator.java
```diff
@@ -257,89 +257,162 @@ default MemorySegment allocateFrom(AddressLayout layout, MemorySegment value) {
         return segment;
     }
 
+    /**
+     * Allocates a memory segment with the given layout and initializes it with the bytes in the provided
+     * source memory segment.
+     * @implSpec the default implementation for this method calls {@code this.allocate(elementLayout, elementCount)}.
+     * @param elementLayout the value to be set on the newly allocated memory block.
```
JornVernee, line 264: This reads a bit strange. How is the value set 'on' the newly allocated memory block? I'd suggest following the flavor of the other allocateFrom overloads, e.g.

```suggestion
     * @param elementLayout the element layout of the array to be allocated.
```

Given that we have a pair of element layouts, I think it's reasonable to say that the result is an 'array'. (not a Java array though)

src/java.base/share/classes/java/lang/foreign/SegmentAllocator.java
```diff
@@ -257,89 +257,162 @@ default MemorySegment allocateFrom(AddressLayout layout, MemorySegment value) {
...
+     * @param source the source segment.
+     * @param sourceElementLayout the element layout of the source segment.
+     * @param sourceOffset the starting offset, in bytes, of the source segment.
+     * @param srcElementCount the number of elements in the source segment to be copied.
```
JornVernee, line 268: Maybe just `elementCount` here. It doesn't necessarily have any connection to the source? (`elementCount` is also used in the IOOBE `@throws` clauses

test/micro/org/openjdk/bench/java/lang/foreign/AllocFromSliceTest.java
```diff
@@ -0,0 +1,57 @@
+package org.openjdk.bench.java.lang.foreign;
```
JornVernee, line 1: Missing copyright header

JornVernee (approved):
Very nice improvement!

The implementation feels easier to follow as well.

src/java.base/share/classes/java/lang/foreign/SegmentAllocator.java
```diff
@@ -257,89 +257,171 @@ default MemorySegment allocateFrom(AddressLayout layout, MemorySegment value) {
...
+     * @param source the source segment.
+     * @param sourceElementLayout the element layout of the source segment.
+     * @param sourceOffset the starting offset, in bytes, of the source segment.
+     * @param elementCount the number of elements in the source segment to be copied.
+     * @return a segment for the newly allocated memory block.
+     * @throws IllegalArgumentException if {@code elementLayout.byteSize() != sourceElementLayout.byteSize()}.
+     * @throws IllegalArgumentException if the source segment/offset are <a href="MemorySegment.html#segment-alignment">incompatible with the alignment constraint</a>
+     * in the source element layout.
+     * @throws IllegalArgumentException if {@code elementLayout.byteAlignment() > elementLayout.byteSize()}.
+     * @throws IllegalArgumentException if {@code sourceElementLayout.byteAlignment() > sourceElementLayout.byteSize()}.
```
mcimadamore, line 274: Note: I want to rewrite and split similar assertions in the javadoc of some methods in MemorySegment too - but I'll do that as a separate patch (as it's going to be a bit noisy).

src/java.base/share/classes/java/lang/foreign/SegmentAllocator.java
```diff
@@ -257,89 +257,171 @@ default MemorySegment allocateFrom(AddressLayout layout, MemorySegment value) {
...
+     * @implSpec the default implementation for this method is equivalent to the following code:
+     * {@snippet lang = java:
+     * this.allocateFrom(layout, MemorySegment.ofArray(array),
+     *                   layout.withOrder(ByteOrder.nativeOrder()),
+     *                   0, array.length)
+     *}
      * @param elementLayout the element layout of the array to be allocated.
      * @param elements the short elements to be copied to the newly allocated memory block.
      * @return a segment for the newly allocated memory block.
+     * @throws IllegalArgumentException if {@code elementLayout.byteAlignment() > elementLayout.byteSize()}.
```
mcimadamore, line 324: I've added this to all the array-accepting allocateFrom, as I think this will be thrown by the underlying copy op. There are a lot of other exceptions that memory copy could throw, in the following conditions:
1. layout size * element count is too big
2. offset/elementCount < 0
3. the interval (offset, size) falls outside the spatial bounds of the segment
4. access occurs from a thread other than the thread owning segment
5. access occurs when segment scope is no longer alive
6. start of segment is not aligned according to provided layout

Now, I think 1-5 are impossible because:
1. number of bytes is always derived from an array size, no overflow is possible
2. default impl guarantees positive lengths/offsets
3. access is provably in bounds (from 0 to array size)
4. array segment can be accessed from any thread
5. array segment is always alive

As for (6), I think right now it can happen, but I think we should probably rectify the implementation of the array-accepting `allocateFrom` to pass `JAVA_SHORT`, `JAVA_CHAR`, ... as the "source layout element" rather than just passing a modified version of the input destination layout. If we do that, then access would be aligned by construction.

src/java.base/share/classes/java/lang/foreign/SegmentAllocator.java
JornVernee: > No, if we do not propagate the original layout, some of the alignment checks are going to be missing.

I don't get this. I'm not saying we don't propagate the original layout, I'm just saying that the source layout should be `JAVA_BYTE` here.

The alignment checks of the destination layout only apply to the destination segment, right? Or are you saying that if the destination layout is hyper-aligned (> 1 byte), we should fail because the byte[] elements are only aligned to 1 byte?

src/java.base/share/classes/java/lang/foreign/SegmentAllocator.java
JornVernee: Okay, I agree with your last statement. We seem to be on the same page there.

src/java.base/share/classes/java/lang/foreign/SegmentAllocator.java
mcimadamore: > > No, if we do not propagate the original layout, some of the alignment checks are going to be missing.
> 
> I don't get this. I'm not saying we don't propagate the original layout, I'm just saying that the source layout should be `JAVA_BYTE` here.
> 
> The alignment checks of the destination layout only apply to the destination segment, right? Or are you saying that if the destination layout is hyper-aligned (> 1 byte), we should fail because the byte[] elements are only guaranteed to be aligned to 1 byte?

I believe I just misinterpreted what you were saying - as I made essentially the same point below :-)

PR #881, opened 2023-09-11
8315917: Passing struct by values seems under specified

JornVernee:
As discussed in the JBS issue, the exceptions thrown by a downcall method handle for invalid by-value struct arguments are currently not specified.

There are two cases in which we access a memory segment passed as an argument when unboxing it as part of a downcall handle invocation:
1. In the impl of the `BufferLoad` binding, where we call `MemorySegment::get`. This is on platforms that decompose the struct into registers, such as SysV. 
2. In the impl of `Copy`, where we call `MemorySegment::copy`. This is on Windows where we need to copy a struct to implement by-value semantics.

Since these by-value struct memory segments are given to us by the user, there can be various reasons why these segments can not be accessed, such as the scope already being closed, or the segment belonging to a different thread. The following exceptions may occur: `IllegalStateException`, `WrongThreadException`, `IndexOutOfBoundsException`, `UnsupportedOperationException`, and `IllegalArgumentException` (as specified by MS::get and MS::copy).

Solution: update the specification of `Linker::downcallHandle` to specify that an exception thrown by MS::get or MS::copy might be thrown when invoking the downcall handle.

### DIFF
--- a/src/java.base/share/classes/java/lang/foreign/Linker.java
+++ b/src/java.base/share/classes/java/lang/foreign/Linker.java
@@ -568,8 +568,13 @@ public sealed interface Linker permits AbstractLinker {
      * {@code T}, then the size of the returned segment is set to {@code T.byteSize()}.
      * <p>
      * The returned method handle will throw an {@link IllegalArgumentException} if the {@link MemorySegment}
-     * representing the target address of the foreign function is the {@link MemorySegment#NULL} address.
-     * The returned method handle will additionally throw {@link NullPointerException} if any argument passed to it is {@code null}.
+     * representing the target address of the foreign function is the {@link MemorySegment#NULL} address. If an argument
+     * is a {@link MemorySegment}, whose corresponding layout is an {@linkplain GroupLayout group layout}, that needs to
+     * be accessed by the linker implementation, one of the exceptions specified by the
+     * {@link MemorySegment#get(ValueLayout.OfByte, long)} or the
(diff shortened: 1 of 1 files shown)


src/java.base/share/classes/java/lang/foreign/Linker.java
```diff
@@ -568,8 +568,13 @@ static Linker nativeLinker() {
      * {@code T}, then the size of the returned segment is set to {@code T.byteSize()}.
      * <p>
      * The returned method handle will throw an {@link IllegalArgumentException} if the {@link MemorySegment}
-     * representing the target address of the foreign function is the {@link MemorySegment#NULL} address.
-     * The returned method handle will additionally throw {@link NullPointerException} if any argument passed to it is {@code null}.
+     * representing the target address of the foreign function is the {@link MemorySegment#NULL} address. If an argument
+     * is a {@link MemorySegment}, whose corresponding layout is an {@linkplain GroupLayout group layout}, that needs to
+     * be accessed by the linker implementation, one of the exceptions specified by the
```
mcimadamore, line 573:
```suggestion
     * representing the target address of the foreign function is the {@link MemorySegment#NULL} address. If an argument
     * is a {@link MemorySegment},whose corresponding layout is an {@linkplain GroupLayout group layout}, the linker might attempt to access the contents of the segment. As such, one of the exceptions specified by the
```

mcimadamore: I've tried to simplify a little by making the sentence more direct. Also, removed "linker implementation" which we have tried to stay clear from in other places.

PR #886, opened 2023-09-18
Improve exception messages and simplify LayoutPath

minborg:
This PR proposes to improve the exception messages in the internal `LayoutPath` class and also do some code simplifications.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign.git pull/886/head:pull/886` \
`$ git checkout pull/886`

`$ git checkout pull/886` \
`$ git pull https://git.openjdk.org/panama-foreign.git pull/886/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 886`

`$ git pr show -t 886`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/886.diff">https://git.openjdk.org/panama-foreign/pull/886.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/panama-foreign/pull/886#issuecomment-1722975849)

### DIFF
--- a/src/java.base/share/classes/jdk/internal/foreign/LayoutPath.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/LayoutPath.java
@@ -282,7 +282,7 @@ public class LayoutPath {
     private static void checkAlign(MemorySegment segment, long offset, MemoryLayout constraint) {
         if (!((AbstractMemorySegmentImpl) segment).isAlignedForElement(offset, constraint)) {
             throw new IllegalArgumentException(String.format(
-                    "Target offset %d is incompatible with byteAlignment %d (of %s) for segment %s"
+                    "Target offset %d is incompatible with alignment constraint %d (of %s) for segment %s"
                     , offset, constraint.byteAlignment(), constraint, segment));
         }
     }
--- a/src/java.base/share/classes/jdk/internal/foreign/LayoutPath.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/LayoutPath.java
@@ -105,15 +105,13 @@ public class LayoutPath {
     // Layout path selector methods
 
     public LayoutPath sequenceElement() {
-        check(SequenceLayout.class, "attempting to select a sequence element from a non-sequence layout");
-        SequenceLayout seq = (SequenceLayout)layout;
+        SequenceLayout seq = requireLayoutType(SequenceLayout.class);
         MemoryLayout elem = seq.elementLayout();
         return LayoutPath.nestedPath(elem, offset, addStride(elem.byteSize()), addBound(seq.elementCount()), derefAdapters, this);
     }
 
     public LayoutPath sequenceElement(long start, long step) {
-        check(SequenceLayout.class, "attempting to select a sequence element from a non-sequence layout");
-        SequenceLayout seq = (SequenceLayout)layout;
+        SequenceLayout seq = requireLayoutType(SequenceLayout.class);
         checkSequenceBounds(seq, start);
         MemoryLayout elem = seq.elementLayout();
         long elemSize = elem.byteSize();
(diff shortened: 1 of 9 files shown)


JornVernee (commented):
I like the overall idea. I think improving the error messages can help tell a user from _where_ in a layout path an error occurs.

But, currently you are only printing out the inner-most enclosing layout. Another idea would be to print a kind of back trace, where each `enclosing` layout path is listed in the exception message, until we get to the root. That should, I think, give the full image of where in the layout path the error occurs.

src/java.base/share/classes/jdk/internal/foreign/LayoutPath.java
```diff
@@ -314,15 +320,24 @@ private static LayoutPath derefPath(MemoryLayout layout, MethodHandle handle, La
...
 
-    private void check(Class<?> layoutClass, String msg) {
+    private <T extends MemoryLayout> T requireLayoutType(Class<T> layoutClass) {
         if (!layoutClass.isAssignableFrom(layout.getClass())) {
-            throw badLayoutPath(msg);
+            var name = layoutClass.getSimpleName();
+            var type = name
+                    // Take what is before "Layout"
+                    .substring(0, name.indexOf("Layout"))
+                    .toLowerCase(Locale.ROOT);
```
JornVernee, line 329: I'd much prefer the simplicity of just passing in the name of the type as a `String` parameter.

Also, it seems like this wouldn't work for `AddressLayout` ?: `"attempting to select a address element from a non-address layout ..."`

Maybe a couple of simple `checkSequenceLayout`, `checkGroupLayout` would be better to avoid the existing duplication. I kinda feel like the new solution ends up introducing more complexity than it resolves.

JornVernee (commented):
Thanks for adding the breadcrumbs, I think they could be useful for debugging.

src/java.base/share/classes/jdk/internal/foreign/LayoutPath.java
```diff
@@ -314,15 +319,27 @@ private static LayoutPath derefPath(MemoryLayout layout, MethodHandle handle, La
...
+
+    private GroupLayout requireGroupLayout() {
+        return requireLayoutType(GroupLayout.class, "group");
+    }
+
+    private <T extends MemoryLayout> T requireLayoutType(Class<T> layoutClass, String name) {
         if (!layoutClass.isAssignableFrom(layout.getClass())) {
-            throw badLayoutPath(msg);
+            throw badLayoutPath(
+                    String.format("attempting to select a %s element from a non-%s layout: %s (%s)",
```
minborg, line 333: While it is good to see the breadcrumbs, having them enclosed in parentheses creates confusion as the layouts themselves contain parentheses as shown in the example below.

Is there a better way (e.g. using `{}`)?

```
Bad layout path: attempting to select a sequence element from a non-sequence layout: [i4(3a)i4(3b)](2) ([[2:[i4(3a)i4(3b)](2)](1)](0) -> [2:[i4(3a)i4(3b)](2)](1) -> [i4(3a)i4(3b)](2))
```

JornVernee: Maybe instead of parenthesis, you could have something like : `path =` in the message? Or maybe something like
```
Bad layout path: attempting to select a sequence element from a non-sequence layout: [i4(3a)i4(3b)](2), selected from: [2:[i4(3a)i4(3b)](2)](1), selected from: [[2:[i4(3a)i4(3b)](2)](1)](0)
```

JornVernee (approved):

src/java.base/share/classes/jdk/internal/foreign/LayoutPath.java
```diff
@@ -26,6 +26,7 @@
 package jdk.internal.foreign;
 
 import jdk.internal.vm.annotation.ForceInline;
+import jdk.internal.vm.annotation.Stable;
```
mcimadamore, line 29: This doesn't seem used?

src/java.base/share/classes/jdk/internal/foreign/LayoutPath.java
```diff
@@ -40,9 +41,13 @@
 import java.lang.invoke.VarHandle;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Locale;
```
mcimadamore, line 44: This also seems unused?

GavinRay97:
Could I make a suggestion?

For the following `MemoryLayout`:

```java
        MemoryLayout struct = MemoryLayout.structLayout(
                MemoryLayout.sequenceLayout(2, MemoryLayout.structLayout(
                                JAVA_INT.withOrder(ByteOrder.LITTLE_ENDIAN).withName("3a"),
                                JAVA_INT.withOrder(ByteOrder.LITTLE_ENDIAN).withName("3b")
                        ).withName("2")
                ).withName("1")
        ).withName("0");
```

I find the notation like:
```
                "[i4(3a)i4(3b)](2), selected from: " +
                "[2:[i4(3a)i4(3b)](2)](1), selected from: " +
                "[[2:[i4(3a)i4(3b)](2)](1)](0)";
```

Difficult to interpret. Maybe something like the below could be clearer/easier-to-read?
```
(0: struct)
├── (1: sequence) x 2
│   ├── (2: struct)      ERROR: Invalid sequence selection!
│   │   ├── (3a: i4 LE)
│   │   └── (3b: i4 LE)
```

mcimadamore:
> Could I make a suggestion?
> 
> For the following `MemoryLayout`:
> 
> ```java
>         MemoryLayout struct = MemoryLayout.structLayout(
>                 MemoryLayout.sequenceLayout(2, MemoryLayout.structLayout(
>                                 JAVA_INT.withOrder(ByteOrder.LITTLE_ENDIAN).withName("3a"),
>                                 JAVA_INT.withOrder(ByteOrder.LITTLE_ENDIAN).withName("3b")
>                         ).withName("2")
>                 ).withName("1")
>         ).withName("0");
> ```
> 
> I find the notation like:
> 
> ```
>                 "[i4(3a)i4(3b)](2), selected from: " +
>                 "[2:[i4(3a)i4(3b)](2)](1), selected from: " +
>                 "[[2:[i4(3a)i4(3b)](2)](1)](0)";
> ```
> 
> Difficult to interpret. Maybe something like the below could be clearer/easier-to-read?
> 
> ```
> (0: struct)
> ├── (1: sequence) x 2
> │   ├── (2: struct)      ERROR: Invalid sequence selection!
> │   │   ├── (3a: i4 LE)
> │   │   └── (3b: i4 LE)
> ```

Honestly I find the notation hard to parse too - but, I have to admit, I wonder if a structural representation, such as the one in this PR, or the one in the comment above really "fit" an exception message - they seem more akin to compiler messages to me.

PR #891, opened 2023-09-22
Fix visibility issues

minborg:
This PR proposes to fix two visibility issues where package private interfaces are exposed via another class' public metods.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign.git pull/891/head:pull/891` \
`$ git checkout pull/891`

`$ git checkout pull/891` \
`$ git pull https://git.openjdk.org/panama-foreign.git pull/891/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 891`

`$ git pr show -t 891`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/891.diff">https://git.openjdk.org/panama-foreign/pull/891.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/panama-foreign/pull/891#issuecomment-1731316682)

### DIFF
--- a/src/java.base/share/classes/jdk/internal/foreign/abi/BindingInterpreter.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/abi/BindingInterpreter.java
@@ -49,12 +49,12 @@ public class BindingInterpreter {
     }
 
     @FunctionalInterface
-    interface StoreFunc {
+    public interface StoreFunc {
         void store(VMStorage storage, Class<?> type, Object o);
     }
 
     @FunctionalInterface
-    interface LoadFunc {
+    public interface LoadFunc {
         Object load(VMStorage storage, Class<?> type);
     }
 }


mcimadamore (commented):
Not super sure about this?

mcimadamore:
> Not super sure about this?

E.g. in the sense - is the fix to make these interface public - or is the fix to make the overriding classes non-public?

minborg:
> Not super sure about this?
It appears difficult to contain the use of the classes that are changed to `public` as they are used from several packages. It might be possible but, I am not sure if it is worth the effort?

mcimadamore:
> > Not super sure about this?
> 
> It appears difficult to contain the use of the classes that are changed to `public` as they are used from several packages. It might be possible but, I am not sure if it is worth the effort?

I think that in this particular case I don't see a lot of benefit changing the code one way or another as all the classes are in non-exported packages anyway. So one could argue that "public" doesn't really mean public, or, on the other hand, also argue that exposing a non-public interface in a public method (which isn't really public) isn't a big deal (even though the IDE might mark that with a squiggly line).

minborg:
Getting rid of warnings reduces the probability of missing more important warnings down the line. But if the consensus is to drop this PR, we could certainly do that.

mcimadamore (approved):

PR #894, opened 2023-09-22
Modernize code

minborg:
This PR proposes to modernize some parts of the code with constructs from newer Java versions.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign.git pull/894/head:pull/894` \
`$ git checkout pull/894`

`$ git checkout pull/894` \
`$ git pull https://git.openjdk.org/panama-foreign.git pull/894/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 894`

`$ git pr show -t 894`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/894.diff">https://git.openjdk.org/panama-foreign/pull/894.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/panama-foreign/pull/894#issuecomment-1731452183)

### DIFF
--- a/src/java.base/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
@@ -614,7 +614,7 @@ public abstract sealed class AbstractMemorySegmentImpl
                             Object dstArray, int dstIndex,
                             int elementCount) {
 
-        var dstInfo = baseAndScale(dstArray);
+        var dstInfo = BaseAndScale.of(dstArray);
         if (dstArray.getClass().componentType() != srcLayout.carrier()) {
             throw new IllegalArgumentException("Incompatible value layout: " + srcLayout);
         }
@@ -641,7 +641,7 @@ public abstract sealed class AbstractMemorySegmentImpl
                             MemorySegment dstSegment, ValueLayout dstLayout, long dstOffset,
                             int elementCount) {
 
-        var srcInfo = baseAndScale(srcArray);
+        var srcInfo = BaseAndScale.of(srcArray);
         if (srcArray.getClass().componentType() != dstLayout.carrier()) {
             throw new IllegalArgumentException("Incompatible value layout: " + dstLayout);
         }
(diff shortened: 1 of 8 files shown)


src/java.base/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
```diff
@@ -714,22 +706,27 @@ public static long mismatch(MemorySegment srcSegment, long srcFromOffset, long s
...
+                    make(Unsafe.ARRAY_FLOAT_BASE_OFFSET, Unsafe.ARRAY_FLOAT_INDEX_SCALE);
+            case Class<?> c when c.equals(long[].class) ->
+                    make(Unsafe.ARRAY_LONG_BASE_OFFSET,  Unsafe.ARRAY_LONG_INDEX_SCALE);
+            case Class<?> c when c.equals(double[].class) ->
+                    make(Unsafe.ARRAY_DOUBLE_BASE_OFFSET, Unsafe.ARRAY_DOUBLE_INDEX_SCALE);
+            default -> throw new IllegalArgumentException("Not a supported array class: " + arrayType.getSimpleName());
+        };
+    }
+
+    static long make(int offset, int scale) {
```
mcimadamore, line 728: Maybe let's just use a record for this? Escape analysis should work fine when it matters. Or, even better, the record could declare a bunch of static constants (e.g. for char, int, long, ...) and just return the appropriate constant given the switch type (e.g. so that there's no allocation).

mcimadamore: Also, we call `.getClass()` on the `Object` array passed in, which then makes the switch more complex (e.g. as it needs `when`). Would it be better to just pass the object "as is", and then switch on the object type instead?

src/java.base/share/classes/jdk/internal/foreign/abi/x64/sysv/TypeClass.java
```diff
@@ -209,34 +209,37 @@ private static List<ArgumentClassImpl>[] groupByEightBytes(GroupLayout group) {
     }
 
     private static void groupByEightBytes(MemoryLayout l, long offset, List<ArgumentClassImpl>[] groups) {
```
mcimadamore, line 211: please rename `l` to `layout`

src/java.base/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
```diff
@@ -566,23 +566,15 @@ public static AbstractMemorySegmentImpl ofBuffer(Buffer bb) {
...
-        } else if (buffer instanceof FloatBuffer) {
-            return 2;
-        } else if (buffer instanceof LongBuffer) {
-            return 3;
-        } else if (buffer instanceof DoubleBuffer) {
-            return 3;
-        } else {
-            throw new AssertionError("Cannot get here");
-        }
+        return switch (buffer) {
```
mcimadamore, line 569: sorry, this is, I think, buggy. Note that you can have an IntBuffer wrapped around a byte[], in which case the calculation will misfire. I think the factory using this method should just call BaseAndScale::of (on the heap array) and then get the scale from there.

mcimadamore: We should probably add a test to detect regressions such as these

src/java.base/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
```diff
@@ -713,23 +701,35 @@ public static long mismatch(MemorySegment srcSegment, long srcFromOffset, long s
...
-            return (long)Unsafe.ARRAY_INT_BASE_OFFSET | ((long) Unsafe.ARRAY_INT_INDEX_SCALE << 32);
-        } else if (arrayType.equals(float[].class)) {
-            return (long)Unsafe.ARRAY_FLOAT_BASE_OFFSET | ((long)Unsafe.ARRAY_FLOAT_INDEX_SCALE << 32);
-        } else if (arrayType.equals(long[].class)) {
-            return (long)Unsafe.ARRAY_LONG_BASE_OFFSET | ((long)Unsafe.ARRAY_LONG_INDEX_SCALE << 32);
-        } else if (arrayType.equals(double[].class)) {
-            return (long)Unsafe.ARRAY_DOUBLE_BASE_OFFSET | ((long)Unsafe.ARRAY_DOUBLE_INDEX_SCALE << 32);
-        } else {
-            throw new IllegalArgumentException("Not a supported array class: " + arrayType.getSimpleName());
+    record BaseAndScale(int base, long scale) {
```
mcimadamore, line 704: This is very nice. Perhaps we could consider moving this to Util

src/java.base/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
minborg: Perhaps we can do that under a separate issue as it is a broader effort than just modernizing the code?

src/java.base/share/classes/jdk/internal/foreign/HeapMemorySegmentImpl.java
```diff
@@ -89,7 +86,7 @@ ByteBuffer makeByteBuffer() {
             throw new UnsupportedOperationException("Not an address to an heap-allocated byte array");
         }
         JavaNioAccess nioAccess = SharedSecrets.getJavaNioAccess();
-        return nioAccess.newHeapByteBuffer(baseByte, (int)offset - BYTE_ARR_BASE, (int) byteSize(), null);
+        return nioAccess.newHeapByteBuffer(baseByte, (int)offset - Utils.BaseAndScale.BYTE.base(), (int) byteSize(), null);
```
minborg, line 89: As records instance variables are trusted to be constant, this still allows for inlining and constant folding.

src/java.base/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
mcimadamore: Please, ignore this comment. I looked again at the code, and the scale factor is used to scale the buffer's position. Of course, the position needs to be scaled depending on the buffer kind (e.g. an int buffer has a scaling factor of 4, a short buffer has a scaling factor of 2). The underlying array type doesn't matter when scaling the buffer position: the position is a logical index expressed in elements, so only the buffer type matters (which is what the code does).

src/java.base/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
```diff
@@ -540,23 +540,23 @@ public static AbstractMemorySegmentImpl ofBuffer(Buffer bb) {
...
-            } else if (base instanceof long[]) {
-                return new HeapMemorySegmentImpl.OfLong(bbAddress + (pos << scaleFactor), base, size << scaleFactor, readOnly, bufferScope);
-            } else if (base instanceof double[]) {
-                return new HeapMemorySegmentImpl.OfDouble(bbAddress + (pos << scaleFactor), base, size << scaleFactor, readOnly, bufferScope);
-            } else {
-                throw new AssertionError("Cannot get here");
-            }
+            return switch (base) {
+                case byte[] __ ->
+                        new HeapMemorySegmentImpl.OfByte(bbAddress + (pos << scaleFactor), base, size << scaleFactor, readOnly, bufferScope);
```
mcimadamore, line 545: Actually, I think this logic is not correct...

mcimadamore (changes requested):

src/java.base/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
mcimadamore: NVM, got confused by the diff - getScaleFactor is still there (I thought it was removed)

PR #898, opened 2023-09-27
8317050: Scope should reflect lifetime of underying resource

mcimadamore:
This patch addresses some issues with `MemorySegment.Scope::equals`. More specifically, when two segments are created from the same array/buffer, their underlying scope should be equal, but currently it is not, which is confusing.

This patch addesses that by adding a new subclass of `GlobalSession` (`HeapSession`) whose `equals`/`hashcode` methods are defined in terms of the heap reference it wraps. Note that the reference is deduped on construction - that is, we always try to find the "bottom" refrence that we should try to keep alive with the session, and we use that for comparisons (which avoid issues when comparing buffer slices and views).

As part of the changes, this patch also adds more javadoc on `MemorySegment.Scope` to document how equality works. And, this patch also makes all zero-length memory segments associate with the same "native" scope. This is a scope that is always alive, but is not the global scope (not to confuse with segments created by the global arena). While what the implementation does today is technically correct (a new fresh scope is returned with each new ZLMS), it seems also overkill given that the scope associated with segments originated outside Java code is not very interesting - if users really care about lifetime of such segments they should override the scope with `MemorySegment::reinterpret`. This leads to some changes in the javadoc of other methods where I replaced `fresh scope that is always alive` with just `scope that is always alive`, to avoid suggesting that a new scope should be created each time.

---------
### Progress

### Issue
 * [JDK-8317050](https://bugs.openjdk.org/browse/JDK-8317050): Scope should reflect lifetime of underying resource (**Enhancement** - P2)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign.git pull/898/head:pull/898` \
`$ git checkout pull/898`

`$ git checkout pull/898` \
`$ git pull https://git.openjdk.org/panama-foreign.git pull/898/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 898`

`$ git pr show -t 898`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/898.diff">https://git.openjdk.org/panama-foreign/pull/898.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/panama-foreign/pull/898#issuecomment-1737582076)

### DIFF
--- a/src/java.base/share/classes/jdk/internal/foreign/GlobalSession.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/GlobalSession.java
@@ -25,20 +25,23 @@
 
 package jdk.internal.foreign;
 
+import jdk.internal.access.JavaNioAccess;
+import jdk.internal.access.SharedSecrets;
 import jdk.internal.vm.annotation.ForceInline;
+import sun.nio.ch.DirectBuffer;
+
+import java.nio.Buffer;
+import java.util.Objects;
 
 /**
  * The global, non-closeable, shared session. Similar to a shared session, but its {@link #close()} method throws unconditionally.
  * Adding new resources to the global session, does nothing: as the session can never become not-alive, there is nothing to track.
  * Acquiring and or releasing a memory session similarly does nothing.
  */
-final class GlobalSession extends MemorySessionImpl {
+non-sealed class GlobalSession extends MemorySessionImpl {
 
-    final Object ref;
-
-    public GlobalSession(Object ref) {
+    public GlobalSession() {
         super(null, null);
-        this.ref = ref;
     }
 
     @Override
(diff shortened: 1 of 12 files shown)


test/jdk/java/foreign/TestScope.java
```diff
@@ -0,0 +1,122 @@
...
+
+import java.lang.foreign.Arena;
+import java.lang.foreign.MemorySegment;
+import java.lang.foreign.SymbolLookup;
+import java.nio.ByteBuffer;
+import java.nio.IntBuffer;
+
+import static org.testng.Assert.*;
+
+public class TestScope {
```
JornVernee, line 39: We already have `TestMemorySession`. Maybe these two test classes can be merged.

src/java.base/share/classes/jdk/internal/foreign/GlobalSession.java
```diff
@@ -67,4 +70,52 @@ void addInternal(ResourceList.ResourceCleanup resource) {
...
+        private static Object base(Object o) {
+            if (o instanceof DirectBuffer directBuffer) {
+                return directBuffer.attachment() != null ?
+                        directBuffer.attachment() : directBuffer;
+            } else if (o instanceof Buffer heapBuffer) {
+                return NIO_ACCESS.getBufferBase(heapBuffer);
+            } else {
+                return o;
+            }
+        }
```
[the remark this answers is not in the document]
mcimadamore: good point

test/jdk/java/foreign/TestScope.java
mcimadamore: Uhmmm - not sure. TestMemorySession is really about close(). In that respect, it is more of an Arena test than it is about a scope test. This new test is about asserting properties on the scope of some segments.

PR #900, opened 2023-09-29
8317324: Add an overload to bind a memory segment var handle at an offset

mcimadamore:
This PR adds an overload which binds a memory segment var handle offset coordinate to a provided constant value.

This is useful when the client knows that the accessed offset will be e.g. zero, or some other known constant value, thus reducing the number of coordinates that var handles clients will have to provide on each access.

---------
### Progress

### Issue
 * [JDK-8317324](https://bugs.openjdk.org/browse/JDK-8317324): Add an overload to bind a memory segment var handle at an offset (**Enhancement** - P4)

### Reviewers
 * [Jorn Vernee](https://openjdk.org/census#jvernee) (@JornVernee - Committer) ⚠️ Review applies to [54c931c6](https://git.openjdk.org/panama-foreign/pull/900/files/54c931c6ce8c58f3460e5b8ca281db954f7516dd)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign.git pull/900/head:pull/900` \
`$ git checkout pull/900`

`$ git checkout pull/900` \
`$ git pull https://git.openjdk.org/panama-foreign.git pull/900/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 900`

`$ git pr show -t 900`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/900.diff">https://git.openjdk.org/panama-foreign/pull/900.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/panama-foreign/pull/900#issuecomment-1741197703)

### DIFF
--- a/src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
+++ b/src/java.base/share/classes/java/lang/foreign/MemoryLayout.java
@@ -151,8 +151,8 @@ import jdk.internal.vm.annotation.ForceInline;
  * the open elements in the path:
  *
  * {@snippet lang=java :
- * VarHandle valueHandle = taggedValues.varHandle(PathElement.sequenceElement(),
- *                                                PathElement.groupElement("value"));
+ * VarHandle valueHandle = taggedValues.varHandle(0L, PathElement.sequenceElement(),
+ *                                                    PathElement.groupElement("value"));
  * MemorySegment valuesSegment = ...
  * int val = (int) valueHandle.get(valuesSegment, 2); // reads the "value" field of the third struct in the array
  * }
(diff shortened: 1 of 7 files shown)


JornVernee (approved):
Do we need an overload for `ValueLayout::varHandle` as well? This overload allows accessing the var handle cache on a ValueLayout, but the overload would always have to create a new VarHandle with the bound coordinate (unless we want to specialize for `0` maybe...). So, I don't think having on overload there is super important.

mcimadamore:
> Do we need an overload for `ValueLayout::varHandle` as well? That method allows accessing the var handle cache on a ValueLayout, but the overload would always have to create a new VarHandle with the bound coordinate (unless we want a specialized cache for `0` offset maybe...). So, I don't think having on overload there is super important.

Good point. As you mention, that overload would have to spin a new MH (unless we add another cached stable field somewhere - but that would have to special case `0L` somehow). So, I think the current combination of overloads is fine.

mcimadamore:
Looking more (and after some offline discussions), it we go down this path, we'll need to do the same for several other methods:

* scaleHandle
* sliceHandle
* offsetHandle

And, we will probably need to add a new method like:

* arrayElementVarHandle

Which composes the var handle with the scale handle. And that would also be required in two variants. So, that's a total of 5 new methods, which while not a big deal (after all, the new methods are rather straightforwardly obtained from calling combinators on plain var handles), it makes me wonder a bit how much value there really is in making the MemoryLayout that much bigger.

mcimadamore:
I've added offset-accepting overloads for several methods. In the process it occurred to me that the javadoc of `MemoryLayout` contains some incorrect samples where the additional offset parameter is not passed to the resulting var handle.

Still on the fence on whether doing these change is worth it. But at the very least, the javadoc should be fixed.

mcimadamore:
Javadoc fixes were pushed as part of https://git.openjdk.org/panama-foreign/pull/901

Closing this PR.

PR #902, opened 2023-10-09
8317514: Ensure MemorySegment is initialized before touching NativeMemorySegmentImpl 

JornVernee:
Please review this fix for a class initializer deadlock.

The issue is that `MemorySegment` references `NativeMemorySegmentImpl` in its initializer, through the `NULL` constant. `NativeMemorySegmentImpl` also depends on `MemorySegment` to be initialized, since it's a super interface with default methods. The VM does not drop the initialization monitor of `NativeMemorySegmentImpl` while trying to initialize super classes/interfaces (as mandated by the JVM spec).

When 2 threads try to intialize `MemorySegment` (e.g. through `ofAddress`) and `NativeMemorySegmentImpl` (e.g. through Arena::allocateFrom), at the same time, a deadlock can occur.

This PR fixes this issue by ensuring `MemorySegment` is initialized in all cases where we directly access `NativeMemorySegmentImpl` or `MappedMemorySegmentImpl`.

We might consider a fix that removes the cycle in the future as well.

---------
### Progress

### Issue
 * [JDK-8317514](https://bugs.openjdk.org/browse/JDK-8317514): Ensure MemorySegment is initialized before touching NativeMemorySegmentImpl (**Bug** - P3) ⚠️ Issue is already resolved. Consider making this a "backport pull request" by setting the PR title to `Backport <hash>` with the hash of the original commit. See [Backports](https://wiki.openjdk.org/display/SKARA/Backports).

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/panama-foreign.git pull/902/head:pull/902` \
`$ git checkout pull/902`

`$ git checkout pull/902` \
`$ git pull https://git.openjdk.org/panama-foreign.git pull/902/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 902`

`$ git pr show -t 902`

</details>

<a href="https://git.openjdk.org/panama-foreign/pull/902.diff">https://git.openjdk.org/panama-foreign/pull/902.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/panama-foreign/pull/902#issuecomment-1753462865)

### DIFF
--- a/src/java.base/share/classes/jdk/internal/foreign/ArenaImpl.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/ArenaImpl.java
@@ -35,6 +35,8 @@ public final class ArenaImpl implements Arena {
     static {
         try {
             // initialize MemorySegment before touching NativeMemorySegmentImpl to avoid deadlock
+            // if multiple threads try to initialize NativeMemorySegmentImpl and MS through allocateNoInit
+            // and a method on MS at the same time
             MethodHandles.lookup().ensureInitialized(MemorySegment.class);
         } catch (ReflectiveOperationException e) {
             throw new ExceptionInInitializerError(e);
--- a/src/java.base/share/classes/jdk/internal/foreign/ArenaImpl.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/ArenaImpl.java
@@ -28,8 +28,18 @@ package jdk.internal.foreign;
 import java.lang.foreign.Arena;
 import java.lang.foreign.MemorySegment;
 import java.lang.foreign.MemorySegment.Scope;
+import java.lang.invoke.MethodHandles;
 
 public final class ArenaImpl implements Arena {
+
+    static {
+        try {
+            // initialize MemorySegment before touching NativeMemorySegmentImpl to avoid deadlock
+            MethodHandles.lookup().ensureInitialized(MemorySegment.class);
+        } catch (ReflectiveOperationException e) {
+            throw new ExceptionInInitializerError(e);
+        }
+    }
 
     private final MemorySessionImpl session;
     private final boolean shouldReserveMemory;
--- a/src/java.base/share/classes/jdk/internal/foreign/ArenaImpl.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/ArenaImpl.java
@@ -28,19 +28,16 @@ package jdk.internal.foreign;
 import java.lang.foreign.Arena;
 import java.lang.foreign.MemorySegment;
 import java.lang.foreign.MemorySegment.Scope;
-import java.lang.invoke.MethodHandles;
+
+import jdk.internal.misc.Unsafe;
 
 public final class ArenaImpl implements Arena {
 
     static {
-        try {
-            // initialize MemorySegment before touching NativeMemorySegmentImpl to avoid deadlock
-            // if multiple threads try to initialize NativeMemorySegmentImpl and MS through allocateNoInit
-            // and a method on MS at the same time
-            MethodHandles.lookup().ensureInitialized(MemorySegment.class);
-        } catch (ReflectiveOperationException e) {
-            throw new ExceptionInInitializerError(e);
-        }
+        // initialize MemorySegment before touching NativeMemorySegmentImpl to avoid deadlock
+        // if multiple threads try to initialize NativeMemorySegmentImpl and MS through allocateNoInit
+        // and a method on MS at the same time
+        Unsafe.getUnsafe().ensureClassInitialized(MemorySegment.class);
     }
 
     private final MemorySessionImpl session;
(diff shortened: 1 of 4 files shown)


src/java.base/share/classes/jdk/internal/foreign/Utils.java
```diff
@@ -56,6 +56,13 @@ public final class Utils {
 
     public static final boolean IS_WINDOWS = privilegedGetProperty("os.name").startsWith("Windows");
 
+    static {
```
mcimadamore, line 59: Another possibility here is to make sure that the static methods here (and also in `NativeMemorySegmentImpl` go via `reinterpret`, so that there's nothing additional to do.

test/jdk/java/foreign/TestDeadlock.java
```diff
@@ -0,0 +1,70 @@
...
+                latch.await();
+            } catch(InterruptedException e) {
+                throw new RuntimeException(e);
+            }
+
+            // Access ArenaImpl -> NativeMemorySegmentImpl -> MemorySegment
+            arena.allocateFrom(ValueLayout.JAVA_INT, 42);
+        });
+
+        Thread t2 = Thread.ofPlatform().start(() -> {
```
mcimadamore, line 51: We do not seem to test:
* mapped segments
* symbol lookup segments
* linker-generated segments

Is that ok?

JornVernee: I'll try to add a test for mapped segments as well (through FileChannel::map). Though, it might not be possible to reproduce an issue in a test. 

But, the other two do not touch NativeMemorySegmentImpl or MappedMemorySegmentImpl directly, so I'm not sure. Theoretically we can test all the API points (that create MemorySegments), but given the tricky timing (the current test has to carefully pre-initialize several classes to get there), I'm not sure it's useful to spend resources on running those tests.

src/java.base/share/classes/jdk/internal/foreign/Utils.java
JornVernee: Yes, that's also an option, but `reintepret` is CallerSensitive, so this does introduce some overhead in terms of startup perf I think.

At the same time, it seems to require a similar amount of care to be taken, so I'm not sure how much it's an improvement over `ensureClassInitialized`.

test/jdk/java/foreign/TestDeadlock.java
JornVernee: I added a similar test for FileChannel::map, but was not able to reproduce a deadlock. The log shows that the basic conditions are right, but I guess we just miss the timing:

```
[0.182s][info][class,init] 728 Initializing 'java/lang/foreign/MemorySegment' (0x0000021a82050320) by thread "Thread-2"
[0.183s][info][class,init] 729 Initializing 'jdk/internal/foreign/AbstractMemorySegmentImpl' (0x0000021a82050d28) by thread "Thread-2"
[0.183s][info][class,init] 730 Initializing 'jdk/internal/foreign/NativeMemorySegmentImpl' (0x0000021a82051598) by thread "Thread-2"
[0.183s][info][class,init] 731 Initializing 'jdk/internal/foreign/MappedMemorySegmentImpl' (0x0000021a82051e20) by thread "Thread-1"
```

src/java.base/share/classes/jdk/internal/foreign/ArenaImpl.java
```diff
@@ -29,8 +29,17 @@
...
 
+import jdk.internal.misc.Unsafe;
+
 public final class ArenaImpl implements Arena {
 
+    static {
+        // initialize MemorySegment before touching NativeMemorySegmentImpl to avoid deadlock
+        // if multiple threads try to initialize NativeMemorySegmentImpl and MS through allocateNoInit
+        // and a method on MS at the same time
+        Unsafe.getUnsafe().ensureClassInitialized(MemorySegment.class);
```
mcimadamore, line 40: If you want to avoid unsafe, another way is to use:
https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/invoke/MethodHandles.Lookup.html#ensureInitialized(java.lang.Class)

test/jdk/java/foreign/TestDeadlock.java
mcimadamore: That's ok - at least we have that covered, and, hopefully it will start to fail from time to time if we regress

src/java.base/share/classes/jdk/internal/foreign/ArenaImpl.java
JornVernee: Yeah. I figured using Unsafe is a little more lightweight, especially for FileChannel::map. Maybe it doesn't matter that much though.

Would you prefer using Methodhandles::ensureInitialized?

JornVernee:
Superseded by: https://github.com/openjdk/jdk/pull/16187

